PR #701, opened 2014-01-04
Rest mapping fix

cdjackson:
This corrects a bug that was introduced with the sitemap updates. An addition was made to remove quotes around commands, and this actually meant no command was provide in the REST interface if there were not quotes.


DaAwesomeP:
I'm having this issue in OH2 from both BasicUI and HABdroid.
```
Switch item=GarageDoor1 label="Garage Door" mappings=[ON="Toggle"]
```
```
[WARN ] [thome.io.rest.core.item.ItemResource] - Received HTTP POST request at 'items/GarageDoor1' with an invalid status value 'ON'.
```

DaAwesomeP:
Nevermind. The item is a `Number`, so it can only accept numbers.

PR #785, opened 2014-01-24
Chart servlet updates

cdjackson:
This makes a few styling changes to the default chart provider to make it more usable.


kaikreuzer:
@teichsta It seems this PR has never been ported to ESH, is that right?
@cdjackson Could you create a PR for ESH and sync all the changes that you had made at that time?

PR #955, opened 2014-03-31
Daikin binding (for KKRP01A online controller)

sumnerboy12:


teichsta:
Hi Ben (@sumnerboy12), a short question: in another PR-Review of #1997 i complained about a the usage of the require-bundle directive in MANIFEST.MF. @berndpfrommer pointed me to the Daikin and the Sonos Binding which make also use of the HttpClient 4. Would you mind removing the HttpClient from the /libs and use import-package statements instead. I assume at that days we only had HttpClient 3.1 in the target platform. Thanks, Thomas E.-E.

sumnerboy12:
Hi Thomas,

As you may have seen on the forums, I am having real trouble even 
getting openHAB to compile so I won't be able to look at this until that 
is resolved.

Must admit I am getting very frustrated - it seems like a pretty simple 
error but no one seems to have any idea about how to solve it. I presume 
there must be users compiling openHAB on Windows 8.1 x64 but I certainly 
can't get it to work.

Ben

sumnerboy12:
Finally got around to looking at this Thomas - apologies for the delay.

https://github.com/openhab/openhab/pull/2499

Cheers,
Ben

sumnerboy12:
Done - https://github.com/openhab/openhab/pull/2499

PR #1188, opened 2014-06-27
Added functionlity for Eltako dim actuator

philipSerbin-TUDo:
I added some code for the setStateOnDevice function in EnoceanBinding to
be able to send some data to the Actuator. The used Actuator is an
Eltako FSUD-230V and uses a variant of the EEP A5-38-08. There are 3
different parameters that can be used in the item configuration:
TEACH_IN, DIM_VALUE_ACK and DIM_VALUE_NO_ACK. With TEACH_IN a user is
able to send a teach-in telegramm, so that the Actuator knows the used
USB300. With the DIM_VALUE parameters it is possible to send
INCREASE/DECREASE and ON/OFF commands to the Actuator. DIM_VALUE_ACK
uses a simple acknowledgement process to check if the acuator is
reachable. After 3 lost telegramms it send as STOP command to the item,
which i used in the sitemap to change the label colour to red. The used
EnOceanId for the items is the ID of the actuator. To use the EEP
variant the user has to write eep=A5:38:08:eltako.


teichsta:
I'll close this PR due to inactivity. If there is anybody interested in working on the PR in the future, please let me know and i'll reopen this PR …

PR #1436, opened 2014-09-18
added new action: createTimerWithArgument(expiryTime, argument, closure)

berndpfrommer:
to allow creation of timers with an argument passed to the closure.
This allows more complex scripting for e.g. motion sensor timeouts.

This extends the existing createTimer() to allow for scripting like this:

var String cookie = "cookie"
var Timer timer = createTimerWithArgument(now.plusSeconds(10), cookie,
                                     [ x |
                                        logInfo("timer_expired", " with cookie: " + x)
                                     ]
                            )

Having the cookie allows the expiring timer to e.g. look up its timer in hash table.

A combination of Groups, hash maps, and timers with arguments allowed me to write
some powerful rules to avoid cut-and-pasting code.

### DIFF
--- a/bundles/model/org.openhab.model.script/src/org/openhab/model/script/actions/ScriptExecution.java
+++ b/bundles/model/org.openhab.model.script/src/org/openhab/model/script/actions/ScriptExecution.java
@@ -14,6 +14,7 @@ import static org.quartz.TriggerBuilder.newTrigger;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.xtext.xbase.XExpression;
 import org.eclipse.xtext.xbase.lib.Procedures.Procedure0;
+import org.eclipse.xtext.xbase.lib.Procedures.Procedure1;
 import org.joda.time.base.AbstractInstant;
 import org.openhab.core.scriptengine.Script;
 import org.openhab.core.scriptengine.ScriptEngine;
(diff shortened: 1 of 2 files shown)


mavenius:
Any chance of adding a rescheduleWithArgument to supplement this? I would try, but I don't have a Java development environment configured. 

I would like to use this for recursive timers; all I see for something similar is a while() with a sleep in it, and a Timer seems like a better way to go.

berndpfrommer:
How about using cancel + create instead?

mavenius:
I looked into that, but I need to recurse the lambda, and I can't find a
way to access the currently-executing body. I'll dig more into it today.

Thank you,

Mark Avenius
Mark@Aveni.us

mavenius:
Actually, it looks like recursive lambdas do work, but they are not
supposed to:

// val totalSteps, val transitionTimePerStepMs, etc. are defined above

// declare but don't define the temporary timer function, so it can be used
inside its own body
var Functions$Function1 timerFunctionTemp
timerFunctionTemp = [Integer stepNo |
// ...snip... do the actual logic that needs to happen on this step
if (livingRoomColorFaderTimer != null) {
livingRoomColorFaderTimer.cancel
livingRoomColorFaderTimer = null
}
if (stepNo < totalSteps) {
logInfo(logName, "Scheduling timer for " + transitionTimePerStepMs + " ms
from now.")
// This shows as an error in Designer (cannot refer to a non-final
variable,) but log entries for all steps appear
val timerFunction = timerFunctionTemp
livingRoomColorFaderTimer =
createTimerWithArgument(now.plusMillis(transitionTimePerStepMs), stepNo +
1, [arg |
logInfo(logName, "Calling timerFunction with arg " + arg)
timerFunction.apply(arg)
])
}
]
Having a rescheduleTimerWithArgument option would make this a bit cleaner,
but this works fine.

Thanks again!

Mark Avenius

kaikreuzer:
@teichsta: It had been missed to sync this change to ESH as well. Could you please take care of that? Thanks!

teichsta:
yes, i will!

tiktok7:
Assuming the intention was to get the change into eclipse smart home, and by extension OpenHAB2, it [doesn't seem to be present in the smarthome codebase](https://github.com/eclipse/smarthome/search?utf8=%E2%9C%93&q=createTimerWithArguments) and openhab2 rejects calls to CreateTimerWithArgument with `The name 'createTimerWithArgument(<XFeatureCallImplCustom>,<XMemberFeatureCallImplCustom>,<XClosureImplCustom>)' cannot be resolved to an item or type.`

watou:
@tiktok7 `createTimerWithArgument` is indeed not present in [this source file](https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/ScriptExecution.java#L83) in Eclipse SmartHome.  Please open an issue [there](https://github.com/eclipse/smarthome/issues) to request that it is added, assuming it was an oversight that the OH1 runtime change wasn't sync'd with ESH.

tiktok7:
[Done](https://github.com/eclipse/smarthome/issues/2718)

Issue #851, opened 2014-02-12
KNX Reconnection


akileos:
Hi,

 Sometime upon boot the KNX/IP Gateway has not yet freed a socket for client, OpenHAB never retries to reconnect. This could be usefull to implement a reconnect timeout upon failure.

```
15:02:46.234 ERROR o.o.b.k.i.c.KNXConnection[:209]- Error connecting to KNX bus: acknowledge error, failed to connect to control endpoint /192.168.1.252:3671
```

PR #1483, opened 2014-10-01
Fix for #851 KNX Reconnection

Snickermicker:
Added a timer thread, which will try to reconnect to the KNX bus in case
of initial connection failure.
Modified management of DP initialization, such that DPs are not removed
from the queue when connection is not established.
Added debug and trace messages.


teichsta:
@Snickermicker i am not sure anymore … did you send another PR for 1.6.0 these days?

Snickermicker:
Yes, I merged that fix into 1.6.0 with #1517.
But as I wrote before, this is only partly fixing the problem. I saw  a problem when connection is lost in  gateway mode. At first glance it looked as if the binding is stuck sporadically in calimero lib. Contacting the calimero maintainers didn't reveal any insights. So, this is currently unfixed.

PR #1509, opened 2014-10-08
Squeezebox: Add Player State to Speech Commands. Fixes #1474 #1481

### LINKED ISSUE #1474: reset playlist after squeezeboxSpeak
Unfortunately the squeezebox forgets the current playlist after execution of squeezeboxSpeak. Is it possible to add the following feature:

Playlist plays -> fade down current playlist -> squeezeboxSpeak says a text -> remember old playlist and start playing (fading in)

It should fade out and fade in to have smooth transistion between playing music and speaking text.
At the moment I cant use squeezeboxSpeak, because sqeezebox stops playing ...

wezhunter:
Added the current playlist track, timestamp, repeat/shuffle modes for
current player playlist.
Changed the "squeezeboxSpeak" function to an additive mode for speech
playback and add restoration of previous playlist item and time.
Added a new parameter to
"squeezeboxSpeak(playerId,message,volume,resumePlayback)" which allows
for speech commands to override playback resume (item state is still
restored however player will be paused after speech completes). Useful
for Doorbell type workflows.

### DIFF
--- a/bundles/io/org.openhab.io.squeezeserver/src/main/java/org/openhab/io/squeezeserver/SqueezePlayer.java
+++ b/bundles/io/org.openhab.io.squeezeserver/src/main/java/org/openhab/io/squeezeserver/SqueezePlayer.java
@@ -383,7 +383,6 @@ public class SqueezePlayer {
 		}
 	}
 	
-	
 	private synchronized void fireCurrentPlaylistIndexEvent() {
 		PlayerEvent event = new PlayerEvent(this);
 	    Iterator<SqueezePlayerEventListener> itr = squeezeServer.getPlayerEventListeners().iterator();


sumnerboy12:
Hi Wez, I am noticing that when I send my 'good morning' message to the kitchen Squeezebox I get 1 sec of whatever was previously playing on the device, before the 'squeezeboxSpeak' message is played. This used to work just fine prior to your change so I presume it is something to do with your changes to restore state etc.

Are you able to confirm this issue? Sending a squeezeboxSpeak() command to a device with something already in its playlist, but is OFF, exhibits this in my case.

Interested to know your thoughts. Might be something to do with that 1s delay in there which waits for the playlist to update. Is this really necessary? That 1s delay is not ideal for things like doorbell events etc.

PR #1522, opened 2014-10-09
ZWave Heal Modifications

wezhunter:
Do not attempt to heal nodes that are marked as Dead or the Main
Controller
Added debug logging to aid troubleshooting during heal processing

Allow a Dead node to advance back to alive. Fixes a problem with some
devices not init according to sequence.


cdjackson:
I would like to close this PR as it is now very out of date compared to the refactored initialisation and heal code.

PR #1607, opened 2014-10-27
Serial action

joek:
Hi,
I created a serial Action, to push commands to applications which are stateless. Especial for arduino devices it is possible to configure a device in openhab.cfg. This device will be opened once only closed at shutdow. Otherwise arduino devices on linux boxes will be reset at close.


jernejz5:
Hello, since I would really love to have this feature I am wondering if there is anything I can do to help implement this Serial Action or just get it working for the next release? 

Is @joek still actively participating in this issue?

joek:
@teichsta sorry I was on vacation. Yes I implemented a startup rule to check if the feature is working. I will double check the rule.

teichsta:
any news?

joek:
I will not make it within the next year. Maybe somebody like to take it over.

PR #1697, opened 2014-11-20
Update yahoo weather XSLs and add german scale for humidex

cniweb:


distribution/openhabhome/configurations/rules/demo.rules
```diff
@@ -164,13 +164,15 @@ end
 
 // This rule will be used to test Scale transformation service
 rule "Compute humidex"
-when Item Weather_Temperature changed or
-	 Item Weather_Humidity changed 
+when
+        Item Weather_Temperature changed or
+	Item Weather_Humidity changed or
+	System started
```
cniweb: OK i removed System started

PR #1762, opened 2014-11-29
Modify heal so it doesn't abort if GETNEIGHBORS fails

cdjackson:


cdjackson:
Will not merge this to the main branch. This PR is included in the initialisation refactoring so will be merged with that PR.

PR #1786, opened 2014-12-02
Zwave polling - make period configurable

cdjackson:
This allows the user to configure the period that the 'alive check' poll takes place. It also won't perform a poll if all nodes have been contacted within this period, thus eliminating needless traffic.


cdjackson:
Will not merge this to the main branch. This PR is included in the initialisation refactoring so will be merged with that PR.

PR #1802, opened 2014-12-06
Added http status 202 ACCEPTED as valid no-content status

asbkar:
Fibaro Home Center returns 202 ACCEPTED, so I added that as an accepted status code without content.

### DIFF
--- a/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/http/HttpUtil.java
+++ b/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/http/HttpUtil.java
@@ -196,7 +196,7 @@ public class HttpUtil {
 		try {
 			
 			int statusCode = client.executeMethod(method);
-			if (statusCode == HttpStatus.SC_NO_CONTENT) {
+			if (statusCode == HttpStatus.SC_NO_CONTENT || statusCode == HttpStatus.SC_ACCEPTED) {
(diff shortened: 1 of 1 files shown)


watou:
[W3C HTTP Status Code Definitions](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) agrees with @adi32k's comment.  For 202 (SC_ACCEPTED), "The entity returned with this response SHOULD include an indication of the request's current status and either a pointer to a status monitor or some estimate of when the user can expect the request to be fulfilled."  So the method should not yet return when it received 202 ([code](https://github.com/openhab/openhab/blob/a79014b6a3d2f00e81c0ba953aeb0df02ec920ac/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/http/HttpUtil.java#L199)).   Perhaps's @asbkar's original concern could be addressed a different way?

For 204 (SC_NO_CONTENT), returning null is correct as before this commit.

I also think, from experience, that [this logger.warn call](https://github.com/openhab/openhab/blob/a79014b6a3d2f00e81c0ba953aeb0df02ec920ac/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/http/HttpUtil.java#L205) that follows is too heavy handed and should not exist, or be moved to DEBUG level.  For example, every hour the Ecobee binding quite normally returns 500, but there is no reason to log it.

PR #1813, opened 2014-12-08
Network UPS Tools (NUT) binding added

jaroslawmazgaj:
Hi,

I would like to contribute my binding.

Thx


bundles/binding/org.openhab.binding.networkupstools/src/main/java/org/openhab/binding/networkupstools/internal/NetworkUpsToolsBinding.java
```diff
@@ -0,0 +1,209 @@
...
+				Class<? extends Item> itemType = provider.getItemType(itemName);
+				String property = provider.getProperty(itemName);
+				items.put(name, new ItemDefinition(itemName, itemType, property));
+			}
+		}
+		
+		for(String name : items.keySet()) {
+			NutConfig nut = upses.get(name);
+			if (nut == null) {
+				logger.error("No configuration for UPS with name: '{}'", name);
```
teichsta: you should add a "return" otherwise you will get NPEs later on because nut IS null then …

teichsta:
Hi @jaroslawmazgaj,

thanks a lot for this contribution. The Binding looks already good codewise but please improve the documentation. At least each of your classes should provide javadoc on class level (not the generic statements we've prepared for you). Also the inner classes should be documented and at least the public methods.

Looking forward to your Binding (and the according Wiki-Page).

Best, Thomas E.-E.

jaroslawmazgaj:
Thanks for your feedback.

Good catch with the NPE - I have fixed this problem.
I also added comments.

I will create Wiki-Page for the binding soon.

jaroslawmazgaj:
You can find Wiki-Page for this binding here:
https://github.com/openhab/openhab/wiki/Network-UPS-Tools

.idea/.name
```diff
@@ -0,0 +1 @@
+openhab
```
teichsta: please remove this file

teichsta:
> Please also add your Binding as dependency here https://github.com/openhab/openhab/blob/master/distribution/pom.xml.

could you :-) ?

dkorovin:
Where I can download the ready jar-file for openhab, or how to make it?

jaroslawmazgaj:
Thanks for making changes and merging. 
My laptop just returned from a service repair and I wanted to make changes, but they are already done.
Thanks again and sorry for late response.

teichsta:
> Where I can download the ready jar-file for openhab, or how to make it?

@dkorovin you should find the jar bundle within the addons archive available at Cloudbess (see https://openhab.ci.cloudbees.com/job/openHAB/).

Best, Thomas E.-E.

teichsta:
> Thanks again and sorry for late response.

no worries, you're welcome!

Melkor82:
Is it possibile to upgrade the binding to take care of Uom (Unit of Measure)?

9037568:
Really should ask such questions in the community forum.

Issue #1718, opened 2014-11-25
OH1.6 xmpp binding unable to connect


ralf35:
i just upgrade OH from 1.5 to 1.6 and i notice xmpp binding stop working.
Perhaps i missed something but openhab.cfg looks like the same. I check for any typo or encoding issue in my cfg without luck.

Anyway i downgrade xmpp to 1.5 and openhab is online again.

Always the same logs :

 DEBUG o.o.i.h.s.osgi.BundleResource[:156]- org.openhab.action.xmpp_                                                                                                                                                             1.6.0 [246]
12:23:39.245 DEBUG o.o.c.core.ConfigDispatcher[:318]- Parsed configuration line                                                                                                                                                              org.openhab.xmpp:servername=xxxxxx.com
12:23:39.245 DEBUG o.o.c.core.ConfigDispatcher[:318]- Parsed configuration line                                                                                                                                                              org.openhab.xmpp:proxy=xxxxxxx.com
12:23:39.245 DEBUG o.o.c.core.ConfigDispatcher[:318]- Parsed configuration line                                                                                                                                                              org.openhab.xmpp:port=5222
12:23:39.246 DEBUG o.o.c.core.ConfigDispatcher[:318]- Parsed configuration line                                                                                                                                                              org.openhab.xmpp:username=xxxxx
12:23:39.246 DEBUG o.o.c.core.ConfigDispatcher[:318]- Parsed configuration line                                                                                                                                                              org.openhab.xmpp:password=xxxxx
12:23:39.247 DEBUG o.o.c.core.ConfigDispatcher[:318]- Parsed configuration line                                                                                                                                                              org.openhab.xmpp:consoleusers=xxx@xxx.com
        at org.openhab.action.xmpp.internal.XMPPConnect.establishConnection(XMPP                                                                                                                                                             Connect.java:103)
        at org.openhab.action.xmpp.internal.XMPPConnect.updated(XMPPConnect.java                                                                                                                                                             :94)

ralf35:
it is working fine now (openhab 1.8)

PR #1827, opened 2014-12-12
Fix "Could not initialize class SmackConfiguration"

### LINKED ISSUE #1804: Could not initialize class org.jivesoftware.smack.SmackConfiguration
openhab 1.7.0 on raspberry with java 1.8

in a rule , when i try to send a message with xmpp i get this error on log : 

17:14:29.616 ERROR o.o.c.s.ScriptExecutionThread[:50]- Error during the execution of rule 'Attiva Allarme': Could not initialize class org.jivesoftware.smack.SmackConfiguration

Flowdalic:
Fixes #1804 and #1718 (to some amount). The exception ("Could not initialize ...") was thrown because no XmlPullParser was found in the classpath. While developing for openhab, xpp3 slipped into the classpath but wasn't committed. Sorry for that.

Anyway, I'd like to develop further on the xmpp action. Since Smack now enforces SSL/TLS security, I'd like to add support for [java pinning](https://github.com/flowdalic/java-pinning) so that users can enjoy _secure_ TLS connections, even when the server certificate is not signed by a CA cert found in the default SSLContext. Should I base my work on the '1.6.1' or 'master' branch?

### DIFF
--- a/bundles/action/org.openhab.action.xmpp/src/main/java/org/openhab/action/xmpp/internal/XMPPConnect.java
+++ b/bundles/action/org.openhab.action.xmpp/src/main/java/org/openhab/action/xmpp/internal/XMPPConnect.java
@@ -14,6 +14,7 @@ import org.apache.commons.lang.StringUtils;
 import org.jivesoftware.smack.AbstractConnectionListener;
 import org.jivesoftware.smack.ChatManager;
 import org.jivesoftware.smack.ConnectionConfiguration;
+import org.jivesoftware.smack.ConnectionConfiguration.SecurityMode;
 import org.jivesoftware.smack.SmackException;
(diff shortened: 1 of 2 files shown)


itheiss:
PLEASE - We need this applied to master.

teichsta:
at all ... sorry for the late response. The patch has now been applied to master (1.7.0-SNAPSHOT). Thanks for your patience! Best, Thomas E.-E.

teichsta:
> why is that branch called '1.6.1'? Isn't it the '1.6' branch of openhab?

sorry for the confusion! I've renamed the branch to 1.6. Tags for 1.6.0 and 1.6.1 were already available. Thanks for noticing! Best, Thomas E.-E.

PR #1855, opened 2014-12-17
This adds the ability to send any command to a player, usefull for playl...

digitaldan:
...ists and other plugins.  This allows triggering of playlists and other features in a squeeze server.  For example on my system I use both the spotify and pandora plugins, now I can trigger playing a playlist either with a static item bnding like so

```
Switch sq_office_pandora    "pandora" { squeeze="office:command:pandora playlist play item_id:ca30491a.0.14", autoupdate="false"
```

or pass it in via a text item from say a switch (or action or rule)

```
String sq_office_command    { squeeze="office:command", autoupdate="false" }

Switch item=sq_office_command mappings=["pandora playlist play item_id:ca30491a.0.14"="Grand National Radio", "pandora playlist play item_id:de875420.0.14"="Broken Bells Radio"]
```

### DIFF
--- a/bundles/action/org.openhab.action.squeezebox/src/main/java/org/openhab/action/squeezebox/internal/Squeezebox.java
+++ b/bundles/action/org.openhab.action.squeezebox/src/main/java/org/openhab/action/squeezebox/internal/Squeezebox.java
@@ -342,6 +342,18 @@ public class Squeezebox {
 		return true;
 	}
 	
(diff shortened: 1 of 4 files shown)


staehler:
Hi,
I dont understand this command stuff. What can I do with this?
In the wiki I found just commands like power, volume, pause, file:file but nothing, which describes your above mentioned case.
Could you extend the wiki, if this features are not descibed there.
Thanks a lot and kind regards,
Michael

PR #1864, opened 2014-12-20
openhab-designer: set default swt.browser to mozilla

manut:
because using webkit ends with a segfault in libsoup.

This seems to be a common problem with this eclipse version and should
be fixed in newer versions. But according to the replies on the according
bug report: https://bugs.eclipse.org/bugs/show_bug.cgi?id=404776 this should
be a workaround that works for many people.

Signed-off-by: Manuel Traut manut@mecka.net

### DIFF
--- a/products/org.openhab.designer.product/org.openhab.designer.product.product
+++ b/products/org.openhab.designer.product/org.openhab.designer.product.product
@@ -15,7 +15,7 @@ http://p.yusukekamiyamane.com, http://www.icons-land.com
    </configIni>
 
    <launcherArgs>
-      <vmArgs>-Dorg.eclipse.equinox.p2.reconciler.dropins.directory=../../addons -DnoRules=true -Xmx256m -XX:PermSize=128M</vmArgs>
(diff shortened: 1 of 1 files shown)


d0t101101:
Thomas - this is still a problem with OpenHAB Designer v1.8 on Ubuntu 14.04 x64.  I had to add this to openHAB-Designer.ini to get it to start:
    -Dorg.eclipse.swt.browser.DefaultType=mozilla
All the best,
.

colinl:
I am still seeing this issue using openHAB-designer 1.8.2 on 64 bit Ubuntu 16.04 (pre-release).  Adding 
-Dorg.eclipse.swt.browser.DefaultType=mozilla
to openHAB-Designer.ini allowed it to start.

kainhofer:
Crash still happens with openHAB-designer 1.8.3 on 64-bit Ubuntu 16.10 (with the same error message). Manually adding the browser to the .ini file makes the designer start up, but its window size is 1x1 pixel. If you enlarge the window, both Configurations and Items panes show the error messgae "Could not create the view: ord.eclipse.smarthome.designer.u" and the internal browser on the right states "Embedded browser not available"...

PR #1868, opened 2014-12-21
Decrease CPU consumption caused by nrjavaserial library

paulianttila:
Added workaround for both rfxcom and epson projector bindings, which decrease CPU consumption.

Related to issue: https://github.com/NeuronRobotics/nrjavaserial/issues/22

### DIFF
--- a/bundles/binding/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/connector/EpsonProjectorSerialConnector.java
+++ b/bundles/binding/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/connector/EpsonProjectorSerialConnector.java
@@ -11,6 +11,8 @@ package org.openhab.binding.epsonprojector.connector;
 import gnu.io.CommPort;
 import gnu.io.CommPortIdentifier;
 import gnu.io.SerialPort;
+import gnu.io.SerialPortEvent;
+import gnu.io.SerialPortEventListener;
 
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -27,7 +29,7 @@ import org.slf4j.LoggerFactory;
  * @author Pauli Anttila
  * @since 1.3.0
  */
-public class EpsonProjectorSerialConnector implements EpsonProjectorConnector {
+public class EpsonProjectorSerialConnector implements EpsonProjectorConnector, SerialPortEventListener {
 
 	private static final Logger logger = 
 		LoggerFactory.getLogger(EpsonProjectorSerialConnector.class);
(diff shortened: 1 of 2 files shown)


paulianttila:
It seems that nrjavaserial is not stable on my system anymore. It can work weeks, but the sunddly it hangs. Issue might be related to nrjavaserial and 64 bit os x. Someone have introduced patch to nrjavaserial https://code.google.com/p/nrjavaserial/issues/detail?id=32 but patch is not accepted because nrjavaserial is moved to github.

paulianttila:
My stability problems aren't related to this PR, so reopen pull request.

teichsta:
There are even more Bindings incorporating the nrjavaserial Lib. Could you please have a look if these implementation shall also be changed/enhanced? Thanks, Thomas E.-E.

teichsta:
i've added a comment and asked whether we would be allowed to provide the patch at github if original owner does not reply anymore. Would you take over and create a PR for nrjavaserial, @paulianttila ?

paulianttila:
@teichsta: [nrjavaserial 3.9.3](https://github.com/NeuronRobotics/nrjavaserial/releases) contains already new versions of the OS X native libs. Unfortunately v3.9.3 didn't solve my stability problems (see my comments #2312). 

I've already given up the faith to have working cross platform serial port implementation in java :(

I have read lot of good about java-simple-serial-connector, so maybe we should give a one more chance to jSSC.

teichsta:
> I've already given up the faith to have working cross platform serial port implementation in java :(

ok, thanks for this update … how should be proceed with this PR then? Does the given changes affect the stability somehow?

paulianttila:
This PR shouldn't have any influence to rfxcom or epson projector binding stability. See my previous comments. I firstly believed that changes causes the problem, my that was wrong conclusion. Meanwhile when I tried to find my stability problems, @cdjackson have introduced same fixes to [z-wave binding] (https://github.com/openhab/openhab/commit/f094d51bac8a0576e63ef96369bd867df720a640)

teichsta:
hmm … the title of this PR is "decrease CPU usage" which is no longer the purpose if i got you correctly? So remains anything to merge (sorry i am somehow confused :))?

paulianttila:
nrjavaserial have serious CPU consumption problem (see NeuronRobotics/nrjavaserial#22), which can be decreased by this PR. 

When I introduced this PR, I also updated my OS X to Yosemite and java 7. Because I had some stability problems with epson projector binding, I believed that those where caused because of this nrjavaserial workaround. But that was not true, I have similar stability problems with 1.6.2 or any other epson projector binding version as well. So this PR should not cause any problems, but should decrease the CPU consumption caused by nrjavaserial library.

paulianttila:
@teichsta: are you still confused :)

teichsta:
no, thanks for the reminder :-)

PR #1891, opened 2014-12-27
transform: JSONPath support for non-string values.

druciak:
This fix allows to transform other than string values using JSONPath.

### DIFF
--- a/bundles/core/org.openhab.core.transform/src/main/java/org/openhab/core/transform/internal/service/JSonPathTransformationService.java
+++ b/bundles/core/org.openhab.core.transform/src/main/java/org/openhab/core/transform/internal/service/JSonPathTransformationService.java
@@ -41,7 +41,7 @@ public class JSonPathTransformationService implements TransformationService {
 		logger.debug("about to transform '{}' by the function '{}'", source, JSonPathExpression);
 
 		try {
-			String transformationResult = JsonPath.read(source, JSonPathExpression);
+			String transformationResult = JsonPath.read(source, JSonPathExpression).toString();
 			logger.debug("transformation resulted in '{}'", transformationResult);
 			return transformationResult;
 		} catch(InvalidPathException e) {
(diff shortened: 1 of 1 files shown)


teichsta:
Hi @druciak, i referred to https://github.com/jayway/JsonPath#what-is-returned-when. It seems that "JsonPath will automatically try to cast the result to the type expected by the invoker." So i would expect line 44 to work with "String". What exactly is the bug you are going to fix with this PR? Best, Thomas E.-E.

druciak:
Hi @teichsta.
First of all, if there wasn't a bug, I wouldn't try to make a fix. ;) I just was getting an exception when trying to extract a number.

Second thing is that casting does not help in this case. Casting helps when you know what is type of the result, so if you have an integer, you can easily assign it to `int` variable. The same for other type, like `Date` presented in mentioned examples. But in the transformation service we cast the result *always* to `String`, so you can imagine what will happen if you cast a number to string variable. The only way to get this working, is to convert the result to a string. The easiest way is to use `toString()` method.

druciak:
Here is simple rule to test the issue:
```
rule "JSONPath test"
when
	Time cron "0/5 * * * * ?"
then
	var String data = "{\"str_val\": \"a\", \"int_val\": 1, \"float_val\":1.1, \"dict_val\": {\"a\": 1}}"
	logInfo("String", transform("JSONPATH", "$.str_val", data))
	logInfo("int", transform("JSONPATH", "$.int_val", data))
	logInfo("float", transform("JSONPATH", "$.float_val", data))
	logInfo("dict", transform("JSONPATH", "$.dict_val", data))
end
```

The result without patch:
```
2015-01-02 15:48:30.311 [INFO ] [rg.openhab.model.script.String] - a
2015-01-02 15:48:30.340 [ERROR] [.o.m.r.i.engine.ExecuteRuleJob] - Error during the execution of rule JSONPath test
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
```

and with patch:
```
2015-01-02 15:50:45.008 [INFO ] [rg.openhab.model.script.String] - a
2015-01-02 15:50:45.012 [INFO ] [org.openhab.model.script.int  ] - 1
2015-01-02 15:50:45.016 [INFO ] [org.openhab.model.script.float] - 1.1
2015-01-02 15:50:45.019 [INFO ] [org.openhab.model.script.dict ] - {"a":1}
```

Obviously all the values are strings anyway, regardless of the type in JSON, but this can be simply handled on rule side.

teichsta:
> First of all, if there wasn't a bug, I wouldn't try to make a fix. ;)

Sure ;-) Thanks for this explanation and the Fix. Best, Thomas E.-E. (i've also added it to the 1.6 Bugfix Branch)

Issue #1742, opened 2014-11-26
[ERROR] An unexpected exception has occured


kfiasche:
2014-11-26 15:32:56.302 [ERROR] [.o.m.c.i.folder.FolderObserver] - An unexpected exception has occured
java.lang.NullPointerException: null
	at org.openhab.model.rule.internal.engine.RuleEngine.runStartupRules(RuleEngine.java:263) ~[na:na]
	at org.openhab.model.rule.internal.engine.RuleEngine.modelChanged(RuleEngine.java:243) ~[na:na]
	at org.openhab.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:158) ~[na:na]
	at org.openhab.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:99) ~[na:na]
	at org.openhab.model.core.internal.folder.FolderObserver.checkFolder(FolderObserver.java:142) ~[na:na]
	at org.openhab.model.core.internal.folder.FolderObserver.run(FolderObserver.java:99) ~[na:na]

ChrisPlontzke:
Hello,
unfortunately I lately also get this error all 10secs and can´t get rid of it .
Neither disabling Persistence Demo 2 nor disabling any logInfo could help.

Is there another solution ?

thaks,
Chris. Openhab 1.6.1, Rasp Wheezy,

steve-bate:
@dougmeredith, can this issue be closed?

dougmeredith:
I just tried the demo with a fresh install of 1.7.1, and the issue still seems to be there.

2015-12-27 23:44:12.622 [ERROR] [m.r.internal.engine.RuleEngine] - Error during the execution of startup rule 'Initialize light states': Could not invoke method: org.openhab.model.script.actions.BusEvent.postUpdate(org.openhab.core.items.Item,java.lang.Number) on instance: null
2015-12-27 23:44:12.705 [ERROR] [m.r.internal.engine.RuleEngine] - Error during the execution of startup rule 'Initialize heating states': Could not invoke method: org.openhab.model.script.actions.BusEvent.postUpdate(org.openhab.core.items.Item,java.lang.Number) on instance: null
[...]

PR #1913, opened 2014-12-31
Improve logging and handling of null values in RuleEngine

dominicdesu:
This pull request results from the analysis of #1742:
- Improved robustness of RuleEngine against NullPointerExceptions (if ScriptExecutionException contains no cause or no cause message).
- Fix log error caused by demo.rules when no weather history exists yet. Before the fix it could happen that toString was called on a null value.


teichsta:
Thanks @dominicdesu for this contribution, Thomas E.-E.

PR #1915, opened 2015-01-01
SpeechDispatcher:  Reconnect to server if connection is dropped.

engineergreen:
Currently the SpeechDispatcher TTS plugin only connects to the speech dispatcher server once.  If that connection is dropped, it never retries to gain - just logs the following error:
`ERROR o.o.i.m.i.t.SpeechDispatcherConnection[:67]- Error sending text to SpeechDispatcher Host 127.0.0.1: not connected to server`
The only existing recovery was to restart openhab.

This simple Pull Request adds a check to see if the plugin has been disconnected from the server and reconnects automatically.


teichsta:
Thanks @engineergreen for this contribution. Best, Thomas E.-E.

PR #1919, opened 2015-01-01
Fixes comment regarding maxHistory, with weekly rotation it's not keepin...

jensihnow:
...g n days anymore.


teichsta:
Thanks @jensihnow for this contribution. Best, Thomas

PR #1940, opened 2015-01-07
fixed cut&paste mistake in binding archetype

holgero:
nothing big, just a wrong debug message in the binding archetype. I didn't check if it is there in the bindings.


teichsta:
Thanks :-)

Issue #1892, opened 2014-12-27
Feature Request: Netatmo Last Update Time


juelicher:
My Netatmo station occasionally looses the connection to the WiFi Network or the Outdoor and Rain Unit. 

Right now there is no way of checking, if the values the binding retrieves from the Netatmo web site are still valid or if they are to old. So in theses cases openhab might display values which are several hours old.

I would propose to additionally evaluate the time information of the netatmo response and to make it available in an item. So these information could be displayed on the UI and the value could be checked in rules.
##

9037568:
Anyone still working on this? (@TobiasMende, @cdjackson, @clinique)

cdjackson:
Not me - I've never done anything with Netatmo...

TobiasMende:
Me neither.

clinique:
Sorry, thèse informations are handled in v2 of the binding but I don't plan to work back on v1 :(

hmerk:
I am closing this issue as @clinique wrote to not implement this feature, while it is available in oh2.

PR #1947, opened 2015-01-08
Update of the netatmo binding to handle timestamp of values for every Netatmo Module (feature request #1892)

clinique:
...


clinique:
Also updated the wiki page to take this update in account as well as the previous one regarding WifiStatus, BatteryVp, Latitude, Longitude, Altitude ...

PR #1948, opened 2015-01-08
changed log level of fatal errors in charts

holgero:
The log level DEBUG makes it quite hard to find a problem with wrong
parameters in charts. (Particularly as even logback_debug.xml sets the
level of messages from everything below org.openhab.ui to INFO.)

I am not completely sure if ERROR is the right level to use here, perhaps
WARNING (or even INFO) could be enough. But DEBUG is definitely too
low IMHO.


teichsta:
agreed … could you change the level to INFO then? ERROR and WARN is a bit too much since only the Chart is concerned. Thanks, Thomas E.-E.

holgero:
Sure, I'll do that. But it can take a while, I've dismantled my development PC.

PR #1962, opened 2015-01-11
Zwave: update zmnhba.xml

mattnl:
parameter 30 was not saved when setting to 1


cdjackson:
Aren't these two lines exactly the same?  The line feed won't make any difference (it's XML, so whitespace and CR/LF are ignored).
If you had a problem setting a parameter, I don't think this will actually make any difference.
(I'll merge though as it's a neater file :) ).

mattnl:
Well, that's what I though as well, but it seems to do make a difference,
at least on my system. with the CR/LF I got an error when setting this
parameter, now it's working...

Thanks ;)

PR #1966, opened 2015-01-11
Allow a direct response without transformation

domidimi:
There are cases when no transformation of the response is needed.
Therefore put the transformation into a separate function and call it
only if there is a transformation specified in the configuration.

Signed-off-by: domidimi


cribskip:
Thanks for the contribution. I did not see any documentation, is there already any? 

Thanks, Sascha

teichsta:
Hi @cribskip, the Binding itself is documented in Wiki (see https://github.com/openhab/openhab/wiki/Exec-Binding). The "No-Transformation" Feature is not (yet) mentioned, you#re right. @domidimi could you please add some documentation accordingly? Thanks, Thomas E.-E.

domidimi:
Hi,
a note and and example were added to the Wiki.
Best wishes,
Dimi

PR #1968, opened 2015-01-12
Max binding cul config

muesliman:
Current CUL binding only supports only Cul-Devices with 9600 8E1.
At http://busware.de/ there are several more devices basing on the same firmware (culfw) bu using different configurations. e.g. SCC which is used with the Rasberry PI uses different serial settings (36400 8N1).

So I made some changes to be able to setup the serialport in the CUL-Binding configuration.
The properties baudrate and parity are now forwarded from the org.openhab.binding.maxcul config to the org.openhab.io.transport.cul module.

In the org.openhab.binding.maxcul there was a problem converting the parameters "baudrate" to integer.
Added parse method instead of cast operation.

All changes are backward compatible (if there is no property in config-file, the defaults (which are unchanged) are taken)

What is the prefered workflow for updating the documentation?

### DIFF
--- a/bundles/binding/org.openhab.binding.maxcul/src/main/java/org/openhab/binding/maxcul/internal/MaxCulBinding.java
+++ b/bundles/binding/org.openhab.binding.maxcul/src/main/java/org/openhab/binding/maxcul/internal/MaxCulBinding.java
@@ -10,6 +10,7 @@ package org.openhab.binding.maxcul.internal;
 
 import java.util.Collection;
 import java.util.Dictionary;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -242,7 +243,7 @@ public class MaxCulBinding extends AbstractBinding<MaxCulBindingProvider>
 			String deviceString = (String) config.get("device");
 			if (StringUtils.isNotBlank(deviceString)) {
 				logger.debug("Setting up device " + deviceString);
-				setupDevice(deviceString);
+				setupDevice(deviceString, config);
 				if (cul == null)
 					throw new ConfigurationException("device",
 							"Configuration failed. Unable to access CUL device "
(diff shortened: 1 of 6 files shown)


teichsta:
Thanks @muesliman for this Contribution! I'll start the Review of this PR asap (i'll expect it to be ready in the next three weeks). 

> What is the prefered workflow for updating the documentation?

Please change the existing Wiki-Page once this PR has been merged. If possible add an indicator to show that this change is only available in the Nightly Build.

Best, Thomas E.-E.

teichsta:
@dereulenspiegel any chance to review this PR? Thanks, Thomas E.-E.

cyclingengineer:
MaxCUL change looks okay to me as do the changes to the CUL transport -but then I don't 'own' that :). 

I notice you have a check for valid parities - do we need a check for sensible baud rates? There is a limited set of standard baud rates which devices generally adhere to. The usual range is 75, 110, 300, 1200, 2400, 4800, 9600, 19200, 38400, 57600 and 115200

bundles/binding/org.openhab.binding.maxcul/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,8 @@ Bundle-Version: 1.7.0.qualifier
 Bundle-Activator: org.openhab.binding.maxcul.internal.MaxCulActivator
 Bundle-ManifestVersion: 2
 Bundle-Description: This is the MaxCul binding of the open Home Automation Bus (openHAB)
-Import-Package: org.apache.commons.lang,
+Import-Package: com.google.common.collect;version="10.0.1",
```
teichsta: please remove the configured version unless you really need it …

teichsta:
@muesliman please find my review comment inline. What about @cyclingengineer s comment? Validating the port speed as well seems to me a valid point, doesn't it?

muesliman:
@cyclingengineer: I agee, so added baudrate check
@teichsta: removed dependency

sja:
Squashed to one commit.

PR #1982, opened 2015-01-14
Initial version of the ConditionalAction action plugin

juri8:
The ConditionalAction plugin is intended to have items (e.g. switches) that prevent automation commands being sent to actors and still having clean rules (I don't want to have the if(Shutter1_Disabled.state == ON) { sendCommand(Shutter1, DOWN) } on each usage of Shutter1 in my scripts.

Be warned that this pull request contains a fix in the astro binding that I experienced between Dec. 21 (begin of winter) and Jan 1 (a NPE on startup) - I have no idea if this fix is really good but at least the NPE was gone.


teichsta:
Hi @juri8, thanks for your contribution!

> Be warned that this pull request contains a fix in the astro binding that I experienced between Dec. 21 (begin of winter) and Jan 1 (a NPE on startup) - I have no idea if this fix is really good but at least the NPE was gone.

Please don't mix up the intention of PRs. Each PR should have one purpose. Regarding the Astro Binding probably @gerrieg can have a look at your fix once you've created a separate PR for it.

Thanks, Thomas E.-E.

gerrieg:
Hi @juri8 !

The Astro NPE has been fixed with PR https://github.com/openhab/openhab/pull/1875, so please update your local repository.

Regards,
Gerhard

teichsta:
thanks for this Update @gerrieg. @juri8 could you then please remove everything except your new Binding from this PR or create a new one.

Best, Thomas E.-E.

Issue #1980, opened 2015-01-14
Anel Binding consumes 100% CPU


teichsta:
steps to reproduce:

* download fresh runtime 1.6.1
* download fresh add ons 1.6.1
* cp Anel Binding into Add-Ons directory (no configuration in openhab_default.cfg)

CPU usage immediately increased to 100%.  

@paphko could you have a look at this? It would be great to include the fix into the upcoming 1.6.2 Bugfix Release.

Thanks, Thomas E.-E.

paphko:
Sure, I'll have a look at this as soon as possible.

paphko:
Fixed: https://github.com/openhab/openhab/pull/1986
If refresh rate is not set in anel-configuration, the default rate for refresh thread (60s) was not read, so it used 0s instead. This caused the refresh rate to call 'execute()' every 0s which results in 100% cpu usage.

PR #1986, opened 2015-01-15
Bugfix #1980: anel-binding caused 100% cpu usage

paphko:
If refresh rate is not set in anel-configuration, the default rate for refresh thread (60s) was not read properly, so it used 0s instead. This caused the refresh rate to call 'execute()' every 0s which results in 100% cpu usage.


teichsta:
Thanks @paphko for this quick fix!

PR #2013, opened 2015-01-20
Add workaround for SMACK-635 to XMPP action

Flowdalic:
JavaxResolver does not get automatically set. This prevents connection to XMPP services where DNS SRV RR lookups are required in order to connect.

Please also cherry pick into 1.6.


teichsta:
done, thanks!

PR #2023, opened 2015-01-22
Adding COUNT operator to Group item

clinique:


teichsta:
Hi @clinique, this could also be achieved by using the OR operator, right? OR returns the amount of elements being in active state (at least). Could that be useful for your UseCase as well? Best, Thomas E.-E.

clinique:
Hi @teichsta , in fact, I tested without success with the OR operator.
I'm working on a group dinamically filled with NumberItems and the OR operator makes a sum of their states. Then I deduced that the OR operator maybe working this way if inner items are Switch like but not Number.
Reason why I pushed this. Hope I didn't miss any other way.

teichsta:
hmm … as far as i can see there is no code to sum up values in OR :-)

But ok, assuming there are Items with different Types in the same Group it could be hard to define the active and passive State.

For the sake of flexibility: could you nevertheless add the possibility to count only Items with a given State? Kind of Filter …

clinique:
Sure I could but I don't see how I would express a COUNT that would count all items in the group whatever their state ?

Group:Number:Count(???) myGroup

teichsta:
you could use a REGEX which get's applied against the String representation of the State?

clinique:
ok, got it

PR #2024, opened 2015-01-22
Initial version of the ConditionalAction action plugin

juri8:
The ConditionalAction plugin is intended to have items (e.g. switches) that prevent automation commands being sent to actors and still having clean rules (I don't want to have the if(Shutter1_Disabled.state == ON) { sendCommand(Shutter1, DOWN) } on each usage of Shutter1 in my scripts.


teichsta:
Thanks @juri8 for this contribution!

To me the purpose of this ConditionalAction is quite limited:

* the ConditionalAction only works for "Disabled" Items - what if the UseCase requires a positive logic?
* shouldn't it also be possible to work with other "Disabled Items States" than OnOffType? What is about OpenClose or Number?

I would assume that the method API get's quite lengthy when dealing with the different variants that should be included (see). Hence i am not so sure who to proceed with this PR. Have you had a look into the reusable rules using lambda expressions being discussed on the Newsgroup a while ago (see https://github.com/openhab/openhab/wiki/Reusable-Rules-via-Functions)?

Best, Thomas E.-E.

juri8:
Hi,
I totally agree regarding the limited logic. The plugin was something that helped me keeping the rules cleaner and so I thought I share it.
The reusable rules thing seems to be a general purpose solution for the same problem so I'd prefer this one (I didn't know this can be done). Before I will switch to the lambda approach I will need to do some  changes to the RFXCom plugin...

teichsta:
ok, great, then i close this PR for the time being. Anyhow thanks a lot for the contribution and for sharing your efforts with the community! Best, Thomas E.-E.

Issue #2027, opened 2015-01-22
change binding project structure to DS 1.2.0


tdiekmann:
Refactor the sample structure of a binding project to use the DS 1.2.0 schema and use the CAS policy required for the DS component.
Eliminate the implementation of the ManagedService and the updated method. Change the signature of the activate() method to take the BundleContext and a Map of configuration objects.

This refactoring simplifies the work for new bindings. The SCR (Service Component Runtime == DS) will activate a component only if there is a configuration available under the configured configuration-pid. Otherwise, the component will not be activated. This removes the complicated code dealing with timing issues around activation and configuration of a binding.

steve-bate:
@tdiekmann, can this issue be closed?

tdiekmann:
Yes, sure.

PR #2028, opened 2015-01-23
changed the archetype default binding project according to #2720

tdiekmann:
updated the copyright header with the year 2015
added immediate="true" flag to binding.xml, otherwise the binding does
 not get started automatically
removed unnecessary imports
removed the Activator. The bundleContext is available from the call to
 the activate method. The static Activator access is discouraged, since
 it is only valid when the bundle is ACTIVE, however, any code could
 access the static variable at any point in time


tdiekmann:
Thomas, this is the proposed change to the default binding project layout. I hope I followed the process correctly.

kaikreuzer:
Hi Tim,

Thanks, looks very good to me! Just one question: Would it make sense to also add the modified method? Otherwise, if the configuration at startup is invalid, the setProperlyConfigured will never be set to true, even if the configuration is changed. What do you think?

tdiekmann:
Yes, that sounds like a good option. Since you support dynamic updates of
the configuration, it will be worth adding that. I'll update the branch for
#2027. Does that mean I need to send a new pull request?

On Fri, Jan 23, 2015 at 3:05 AM, Kai Kreuzer <notifications@github.com>
wrote:

> Hi Tim,
>
> Thanks, looks very good to me! Just one question: Would it make sense to
> also add the modified method? Otherwise, if the configuration at startup is
> invalid, the setProperlyConfigured will never be set to true, even if the
> configuration is changed. What do you think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openhab/openhab/pull/2028#issuecomment-71178627>.
>



-- 


     Tim.

teichsta:
No, if you update the branch the PR will be updated as well.

- sent from a mobile device -

tdiekmann:
I'll make the change, but I noticed that the openhab.cfg file is not
monitored and, therefore, dynamic configuration changes are not detected
and propagated. This is not a problem, but could be enhanced in the future.

  Tim.

On Fri, Jan 23, 2015 at 8:25 AM, Thomas Eichstädt-Engelen <
notifications@github.com> wrote:

>
> No, if you update the branch the PR will be updated as well.
>
> - sent from a mobile device -
>
>
> > Am 23.01.2015 um 17:17 schrieb Tim Diekmann <notifications@github.com>:
> >
> > Yes, that sounds like a good option. Since you support dynamic updates
> of
> > the configuration, it will be worth adding that. I'll update the branch
> for
> > #2027. Does that mean I need to send a new pull request?
> >
> > On Fri, Jan 23, 2015 at 3:05 AM, Kai Kreuzer <notifications@github.com>
> > wrote:
> >
> > > Hi Tim,
> > >
> > > Thanks, looks very good to me! Just one question: Would it make sense
> to
> > > also add the modified method? Otherwise, if the configuration at
> startup is
> > > invalid, the setProperlyConfigured will never be set to true, even if
> the
> > > configuration is changed. What do you think?
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/openhab/openhab/pull/2028#issuecomment-71178627>.
> > >
> >
> >
> >
> > --
> >
> >
> > Tim.
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openhab/openhab/pull/2028#issuecomment-71219311>.
>



-- 


     Tim.

kaikreuzer:
> This is not a problem, but could be enhanced in the future. 

It actually is already enhanced in future - when you use openHAB 1 bindings on the openHAB 2 runtime, configuration changes are tracked and reloaded :-)

Cheers,
Kai

> Am 23 Jan 2015 um 19:28 schrieb Tim Diekmann <notifications@github.com>:
> 
> I'll make the change, but I noticed that the openhab.cfg file is not 
> monitored and, therefore, dynamic configuration changes are not detected 
> and propagated. This is not a problem, but could be enhanced in the future. 
> 
> Tim. 
> 
> On Fri, Jan 23, 2015 at 8:25 AM, Thomas Eichstädt-Engelen <

kaikreuzer:
@teichsta : Any specific reason why you have put esh and openHAB2 tags on here?
@tdiekmann : Just noticed that we missed to update the wiki accordingly. Would you be able to do that? See https://github.com/openhab/openhab/wiki/How-To-Implement-A-Binding#global-binding-configuration

tdiekmann:
Sure, I’ll take a look at it.


    Tim.

teichsta:
> @teichsta : Any specific reason why you have put esh and openHAB2 tags on here?

as a reminder. Although the Archetypes itself will look differently the changes should be reflected there as well, shouldn't they?

teichsta:
> Sure, I’ll take a look at it.

Thanks!

kaikreuzer:
> Although the Archetypes itself will look differently the changes should be reflected there as well, shouldn't they?

Well, there are no ManagedServices in the ESH archetype, so need for such a refactoring.

teichsta:
> there are no ManagedServices in the ESH archetype

a good argument :-)

watou:
Regarding the recent wiki change (copied from my comment in Google Groups):

The wiki page for How to Implement a Binding instructs binding developers to include `configuration-policy="require"` in their `binding.xml` file if their binding requires configuration properties in `openhab.cfg`.  Mine does, so I added this to my binding, but when I copied the new JAR into a running 1.6.2 system, the binding shutdown but failed to restart.  I backed out `configuration-policy="require"`, rebuilt, and the new JAR is now working properly.

Do the recent edits to the wiki page only refer to some future openHAB release, where following the instructions in the current wiki page will produce bindings that don't work on openHAB 1.6?

Developers are used to being able to have their binding JAR work in different openHAB releases.  What release(s) of openHAB will work with the instructions that are currently in the wiki?

teichsta:
any idea @tdiekmann ?

tdiekmann:
That’s a fair question.

The changes apply to 1.7 onward. I guess I could take the previous version of the documentation and make it available through a link to a different wiki page. Would that work for you?

    Tim.

teichsta:
> The changes apply to 1.7 onward

That is true for the created Bindings … but even the new Bindings should still be (backwards) compatible to a 1.6.2 Runtime since this PR didn't change anything in the Core/Runtime self, shouldn't they?

tdiekmann:
Sure, they are. The archetype is a template for new bindings only. I did not change any of the existing bindings. Their functionality is still valid and supported, nothing has changed for them.


    Tim.

watou:
While old bindings should be forward compatible as you say, the question is more about new bindings being backward compatible with openHAB 1.6.2.  I think we want to prevent a situation where someone follows the new wiki instructions, but produces a binding that will not work on 1.6.2 (or perhaps 1.5.*, 1.4.*, etc.).  As much as I want a nice clean new archetype for bindings, I don't think we want to see chaos with binding compatibility!

tdiekmann:
They are binary compatible. At least, the openhab 1.6.2 version of Equinox DS already supports the 1.2 schema and, therefore, it is fully backward compatible.


    Tim.

watou:
Perhaps my test was flawed in some way, but I think this should be thoroughly tested empirically.

PR #2058, opened 2015-01-30
Fix for scaling decimal values in pilight binding

idserda:
Number values received from pilight would have the decimal point in the wrong position for some devices. This fixes that and adds a testcase. It also makes sure the initial state is fetched.

### DIFF
--- a/bundles/binding/org.openhab.binding.pilight.test/src/test/java/org/openhab/binding/pilight/internal/PilightGenericBindingProviderTest.java
+++ b/bundles/binding/org.openhab.binding.pilight.test/src/test/java/org/openhab/binding/pilight/internal/PilightGenericBindingProviderTest.java
@@ -51,23 +51,15 @@ public class PilightGenericBindingProviderTest {
 		Assert.assertEquals(number0.toBigDecimal().compareTo(new BigDecimal("23")), 0);
 
 		config.setScale(1);
-		DecimalType number1 = (DecimalType) binding.getState("23.3", config);
+		DecimalType number1 = (DecimalType) binding.getState("233", config);
 		Assert.assertEquals(number1.toBigDecimal().compareTo(new BigDecimal("23.3")), 0);
 
-		config.setScale(0);
-		DecimalType number2 = (DecimalType) binding.getState("2.3", config);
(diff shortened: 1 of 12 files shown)


bundles/binding/org.openhab.binding.pilight.test/src/test/java/org/openhab/binding/pilight/internal/PilightGenericBindingProviderTest.java
```diff
@@ -0,0 +1,73 @@
...
+		DecimalType number0 = (DecimalType) binding.getState("23", config);
+		Assert.assertEquals(number0.toBigDecimal().compareTo(new BigDecimal("23")), 0);
+
+		config.setScale(1);
+		DecimalType number1 = (DecimalType) binding.getState("23.3", config);
+		Assert.assertEquals(number1.toBigDecimal().compareTo(new BigDecimal("23.3")), 0);
+
+		config.setScale(0);
+		DecimalType number2 = (DecimalType) binding.getState("2.3", config);
+		Assert.assertEquals(number2.toBigDecimal().compareTo(new BigDecimal("23")), 0);
```
teichsta: i am not sure if you got your test correctly. But 2.3 should return "2" at scale "0", shouldn't it?

bundles/binding/org.openhab.binding.pilight/src/main/java/org/openhab/binding/pilight/internal/PilightBinding.java
```diff
@@ -105,21 +105,26 @@ private void processValueEvent(List<PilightBindingConfig> configs, Status status
 			String property = config.getProperty();
 			if (status.getValues().containsKey(property)) {
 				String value = status.getValues().get(property);
-				State state = getStateFromProperty(value, config);
-				eventPublisher.postUpdate(config.getItemName(), state);
+				State state = getState(value, config);
+				
+				if (state != null)
+					eventPublisher.postUpdate(config.getItemName(), state);
```
teichsta: please add curly braces here (even though it's a One-liner)

idserda:
Thanks @teichsta for the review, I've made a few small changes.

teichsta:
>  I've made a few small changes.

Thanks @idserda!

PR #2063, opened 2015-01-30
crc 16 encapsulation

pluimpje:
added crc 16 encapsulation 0x56 command class

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCRC16EncapsulationCommandClass.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCRC16EncapsulationCommandClass.java
@@ -91,7 +91,7 @@ public class ZWaveCRC16EncapsulationCommandClass extends ZWaveCommandClass {
 		ByteBuffer byteBuffer = ByteBuffer.allocate(2);
 		byteBuffer.putShort(calculatedCrc);
 		if (!Arrays.equals(messageCrc, byteBuffer.array())) {
-			logger.error("NODE %d: CRC check failed message contains %s but should be %s",
+			logger.error("NODE {}: CRC check failed message contains {} but should be {}",
 					this.getNode().getNodeId(), SerialMessage.bb2hex(messageCrc), SerialMessage.bb2hex(byteBuffer.array()));
 			return;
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCRC16EncapsulationCommandClass.java
```diff
@@ -0,0 +1,172 @@
...
+				payload.length);
+		byte[] tocheck = Arrays.copyOfRange(payload, offset - 2,
+				payload.length - 2);
+
+		short calculatedCrc = crc_ccit(tocheck);
+		// check if messageCrc = calculatedCrc
+		ByteBuffer byteBuffer = ByteBuffer.allocate(2);
+		byteBuffer.putShort(calculatedCrc);
+		if (!Arrays.equals(messageCrc, byteBuffer.array())) {
+			logger.error("CRC not equal");
```
cdjackson: I think it would be worth making this error a bit more useful. At least adding the node number so that if errors pop up in the log then someone can see if a particular device is generating lots of errors.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCRC16EncapsulationCommandClass.java
```diff
@@ -0,0 +1,172 @@
...
+				logger.error(String.format(
+						"NODE %d: CommandClass %s (0x%02x) not implemented.",
+						this.getNode().getNodeId(), commandClass.getLabel(),
+						commandClassCode));
+			} else {
+				logger.debug(String.format(
+						"NODE %d: Calling handleApplicationCommandRequest.",
+						this.getNode().getNodeId()));
+				zwaveCommandClass.handleApplicationCommandRequest(
+						serialMessage, offset+1, 0); //TODO get endpoint from somewhere
```
cdjackson: I would probably remove the TODO. As per my email, I don't think it makes any sense to encapsulate a CRC 16 class inside the multi_instance class, since the whole point of CRC16 is to improve reliability, and it doesn't do this if it's not the outer class.

cdjackson:
Thanks Hans.

PR #2092, opened 2015-02-07
hue binding enhancements for hue tap handling

gernoteger:
This pull request contains enhancements to the hue binding in order to support hue tap switches. To control the effect of the tap device a set of actions has been defined to control bridge settings from rules.
The bridge binding has been considerably refactored to support this. A preliminary user-documentation is provided in the readme.md.


gernoteger:
I just figured out that there seems to be a problem with the build in this pull request - I'll doublecheck and let you know the results

gernoteger:
had  a conflict due to the fact that actions need an entry in the designer's feature.xml. This likely to reoccur with anybody adding actions!

gernoteger:
I'll rework this Concept for OpenHab2. The Tap rules implemented here are clumsy to use;  I didn't use it myself, but worked wit my javarunner concept. So I don't intend to get it ready for Openhab1

PR #2109, opened 2015-02-09
Update start & start_debug to use only *.jar files in addons directory

hreintke:
Current implementation includes all files in  addons directory.
This patch limits this to only *.jar files.
Include in start & start_debug for both .sh and .bat versions


teichsta:
Thanks @hreintke will have a look at it, soon! Best, Thomas E.-E.

PR #2114, opened 2015-02-10
fix NPE on very early rebuild polling table

maggu2810:
If allBindingsChanged was called before the initialise function a NPE is
thrown.

Signed-off-by: Markus Rathgeb <maggu2810@gmail.com>


cdjackson:
Hi,
Thanks for this. However I'm expecting to merge the new initialisation changes in the next few days and as this bug is already fixed there it will create a merge conflict. So unless something happens in the next day or two that delays the merge of #2136 I won't merge this.
Cheers
Chris

maggu2810:
Sounds great, all fine.

maggu2810:
Fixed with #2136

Issue #2064, opened 2015-01-31
 KNX-Binding: Clouldn't add to KNX bus reader scheduler


PascalTurbo:
I'm using the KNX read feature. But sometimes after updating an items file I got these messages:

    Clouldn't add to KNX bus reader scheduler

The message appears 4 Times - and I have 4 items using the knx read. 

Maybe this is only a misguided error message because it looks like that the reader works like it should.

Snickermicker:
Hi,
Could you Providence a debug log?

Snickermicker:
Found the issue. It's a case of a "missleading" error message, which can be safely ignored.
I'll propose a pull request to fix this in 1.7

teichsta:
Fix #2128 has been merged into master

PR #2128, opened 2015-02-13
Bug fix for #2064: eliminated false error message

Snickermicker:
Eliminated false error message. Improved log output.


bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBusReaderScheduler.java
```diff
@@ -182,7 +185,7 @@ public synchronized boolean scheduleRead(Datapoint datapoint, int autoRefreshTim
 		if (oldListNumber > 0) { 
 			if (oldListNumber==autoRefreshTimeInSecs) {
 				sLogger.debug("Datapoint '{}' was already in  auto refresh list {}", datapoint.getName(), autoRefreshTimeInSecs);
-				return false;
+				return true;
```
teichsta: why did you change the logic here? There doesn't seem to be complains about an erroneous behaviour?

Snickermicker:
Yes, there was a complaint: #2064. The logic caused a subsequent error message when a record was already in a list and would've been overwritten.
This function shouldn't fail, since the whole purpose for the function is to add a record to the list and if it already existed then it's still OK.

PR #2136, opened 2015-02-15
Zwave initialisation refactoring

cdjackson:
This supersedes https://github.com/openhab/openhab/pull/1678

This PR refactors the node stage advancer that is responsible for initialising the nodes. Previously, this was scattered throughout the command classes, making it difficult to maintain, test, and modify. This has all been brought into a single class. Other changes have been made to improve performance, such as delaying battery node initialisation until mains nodes have completed.

Node initialisation is handled solely within the NodeStageAdvancer. It is not
based on time - it waits for the transactions to complete. Time cannot be
used since with larger networks, it may take a long time for the
initialisation. This is especially true if there are battery nodes since the
PING phase, used to detect if a node is active, will time-out for battery
devices. A timeout takes 5 seconds, and if there are retries active, this may
be extended to 15 seconds. For a network with 8 battery nodes, this could
mean a delay of 2 minutes!

We use the 'listening' flag to prioritise the initialisation of nodes. Rather
than kicking off all nodes at the same time and have battery nodes timing out
and delaying the initialisation of mains nodes, we try and initialise nodes
that are listening first. This is checked after the protocol information
is received, and non-listening nodes are held at a WAIT state until the
transmit queue drops below 2 frames when they are allowed to proceed to PING.

The NodeStageAdvancer registers an event listener during the initialisation
of a node. This allows it to be notified when each transaction is complete,
and we can process this accordingly. The event listener is removed when we
stop initialising to reduce processor loading.

Command classes are responsible for building lists of messages needed to
initialise themselves. The command class also needs to keep track of
responses so it knows if initialisation of this stage is complete. Other than
that, the command class does not have any input into the initialisation
phase, and the sequencing of events - this is all handled here in the node
advancer class.

For each stage, the advancer builds a list of all messages that need to be
sent to the node. Since the initialisation phase is an intense period, with a
lot of messages on the network, we try and ensure that only 1 packet is
outstanding to any node at once to avoid filling up the main transmit queue
which could impact on the performance of other nodes.

Each time we receive an ACK for a message, the node advancer gets called, and
we see if this is an ACK for a message that's part of the initialisation. If
it is, the message gets removed from the list.

Each time we receive a command message, the node advancer gets called. This
is called after the command class has been updated, so at this stage we know
if the stage can be completed.

Two checks are performed to allow a node stage to advance. Firstly, we make
sure we've sent all the messages required for this phase. Sending the
messages however doesn't guarantee that we get a response, so we then run
through the stage again to make sure that the command class really is
initialised. If the second run queues no messages, then we can reliably
assume this stage is completed. If we've missed anything, then we continue
until there are no messages to send.

If a node stalls during initialisation, there is a timer to move things along. This uses a log backoff (ie each time it triggers, the time doubles) and it is reset when the node progresses to the next stage.

Additional stages have been added to automatically configure some features of devices based on information that is in the database. This includes the wakeup and association classes.

The binding also reads association and configuration parameters so that they are available to view in the GUI thus avoiding confusion. This information is only read once and then stored in the persistent XML files for future use.


maggu2810:
https://github.com/cdjackson/openhab/commit/d41abafe6dfd6953e78d03a1e4878038fc8b7fe2#commitcomment-9761470

teichsta:
Hi @cdjackson any chance to squash this PR into a single commit? I am going to write an email about that squashing anyhow in order to slow down the increasing repo size. Thanks, Thomas E.-E.

cdjackson:
Hi Thomas,
Does the repo size matter?  Are Github hassling you to reduce the size?

The problem is loosing the history which I think is important - especially if we want to find out what’s changed, or when it changed, when someone reports a problem. My preference would be to keep the history unless there’s a reason to reduce the size?

Cheers
Chris

teichsta:
hmm … i would think that (in general) PRs should not evolve like this so every PR should reflect a new Feature or a Fix or an Enhancement. For this new Feature the history (all the merge with branch xyz messages) should be quite meaningless. Wouldn't you agree?

For ESH (and thus most likely openHAB2) squashing commits is a rule.

maggu2810:
IMHO:
For PR that adds a new feature this is surely correct, but sometimes a PR is used to add multiple new features and bugfixes.
If all changes are squashed, this could hide some feature / bugfixes.

teichsta:
> but sometimes a PR is used to add multiple new features and bugfixes.

cdjackson:
Another point about this merge is that it's not (quite!) as bad as it looks. As this has been ongoing for the past 4 months or so, there have been a lot of merges of the main codebase into this branch to keep it up to date. Of the 250 commits that GH shows - 94 are showing as merges, and a bunch more are other commits that are already in the main branch (lots of database updates) so at a rough guess, it's only about 1/2 as big as it looks...

cdjackson:
> hmm … i would think that (in general) PRs should not evolve like this so every PR should reflect a new Feature or a Fix or an Enhancement. For this new Feature the history (all the merge with branch xyz messages) should be quite meaningless. Wouldn't you agree?
> 
Absolutely, and this is what I try to do. However, in this case we underwent a major refactor that has taken about 4 or 5 months (I forget when it started). So, some features needed the new code, so they grey in their own branch, and were then merged into the development branch.  This is also quite a normal way of operating, but openhab doesn’t use a development branch, we instead merge into master, but other projects I work on use this as a policy and this is effectively what I’ve tried to do here. I have kept a development branch (in my repo) which a number of people have had input to...

teichsta:
we're getting out of topic again (sorry for that) … i'll soon start a discussion about this issue on the Group.

@cdjackson will you review yourself? Or do you expect me to review/merge?

Thanks, Thomas E.-E.

cdjackson:
> @cdjackson <https://github.com/cdjackson> will you review yourself? Or do you expect me to review/merge?
> 
You’re more than welcome to have a look :) - it’s been a long refactor, so I think it’s ok (I’ve been over it a lot). Most of the changes are in one or two files, but there are small changes to a lot of files to support changes in interfaces. It’s had a LOT of testing (as you’ve probably already gathered from the emails within the group).

So, no, I wasn’t expecting you to review unless you want to - if you want to merge it, feel free - just let me know...

PR #2137, opened 2015-02-15
Change Parameter Settings of Parameter 12

bulliuli:
Change Maximum from 1000 to 100
Change Default from 50 to 20
Change Size from 2 to 1


bulliuli:
In my Operation Manual of FGSS-001 Parameter 12 settings are other then in this xml.

cdjackson:
Agreed - thanks...

PR #2140, opened 2015-02-15
Add Evolve LTM-5 accessory switch to products database.

Jzooor:
Added manufacturer Evolve and LTM-5 Accessory switch to z-wave product database.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/evolve/ltm-5.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/evolve/ltm-5.xml
@@ -1,0 +1,157 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>LTM-5</Model>
+	<Label lang="en">Wall Mount Accessory Switch</Label>
+	<CommandClasses>
+		<Class>
+			<id>0x20</id>
+		</Class>
+		<Class>
+			<id>0x26</id>
+		</Class>
+		<Class>
+			<id>0x27</id>
+		</Class>
+		<Class>
+			<id>0x2b</id>
+		</Class>
+		<Class>
+			<id>0x2c</id>
+		</Class>
+		<Class>
+			<id>0x70</id>
+		</Class>
+		<Class>
+			<id>0x72</id>
+		</Class>
+		<Class>
+			<id>0x73</id>
+		</Class>
+		<Class>
+			<id>0x85</id>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/database/evolve/ltm-5.xml
```diff
@@ -0,0 +1,157 @@
...
+				<Label lang="en">Yes</Label>
+			</Item>			
+		</Parameter>
+	</Configuration>
+
+	<Associations>
+		<Group>
+			<Index>1</Index>
+			<Maximum>5</Maximum>
+			<Label lang="en">Group 1</Label>
```
cdjackson: What does this associatin group do? If it's used to send notifications like updates on a switch state, then I'd suggest adding the following tag -:
```<SetToController>true</SetToController>```
This way the binding will automatically configure the switch.

cdjackson:
Thanks for adding this - just one comment before I merge...

Cheers
Chris

bundles/binding/org.openhab.binding.zwave/database/evolve/ltm-5.xml
Jzooor: I think something in your comment may have been lost in formatting. What is the tag you want me to add?

This group is used to control another switch in a 3-way switching setup.

cdjackson: The tag is ```<SetToController>true</SetToController>```

It's used if you want the switch to automatically notify the binding of a change of state. As it's the only group, I would have thought it would be useful, but I might be wrong as I don't know the device...

Jzooor: I think I could see how this might be useful. For my current setup, OpenHAB won't be handling any signals from the switch; the switch is just associated with my dimmer and it controls the dimmer directly independently of the controller. If I understand correctly, this tag would make it so that the controller is automatically associated into the group so that the switch will send any switching commands to the controller?

cdjackson: Yes - exactly. In this case, the group can have up to 5 nodes, so probably using 1 for the controller is useful... You might not intend to use the controller, but I suspect most people will want the openHAB GUI updated when the switch is changed...

Issue #1429, opened 2014-09-15
wrong file permissions in openhab-runtime from apt-repo


hoedlmoser:
hi!

as I reported already on openhab@googlegroups.com

after successfull fresh (!) install of openhab-runtime 1.5.1 from
apt-repo it didn't start. no error, nothing. then I started it the same
way the init.d script does, but removed the option --background from
start-stop-deamon. and voila ...

```
root@inucha:~# start-stop-daemon --start --quiet --pidfile [...]
<title>Invalid Configuration Location</title>The configuration area at
'/var/lib/openhab/workspace' is not writable.  Please choose a writable
location using the '-configuration' command line option.
```

uh, oh, wrong permissions on /var/lib/openhab/workspace? lets check.

```
root@inucha:~# ls -la /var/lib/openhab/workspace
total 8
drwxr-xr-x 2 root    openhab 4096 Sep 12 22:41 .
drwxr-xr-x 4 openhab openhab 4096 Sep 13 17:30 ..
root@inucha:~#
```

indeed. lets fix it.

```
chown openhab: /var/lib/openhab/workspace
```

oh no, next error.

```
17:33:51,510 |-ERROR in ch.qos.logback.core.FileAppender[FILE] -
openFile(/var/log/openhab/openhab.log,true) call failed.
java.io.FileNotFoundException: /var/log/openhab/openhab.log (Permission
denied)
```

```
root@inucha:~# ls -la /var/log/openhab/
total 8
drwxr-xr-x 2 root openhab 4096 Sep 12 22:41 .
drwxr-xr-x 8 root root    4096 Sep 13 17:30 ..
root@inucha:~#
```

```
chown openhab: /var/log/openhab/
```

next try.

```
17:36:16.596 ERROR o.o.c.internal.CoreActivator[:117] - Couldn't create
file 'webapps/static/uuid'. java.io.FileNotFoundException:
webapps/static/uuid (Permission denied)
17:36:16.599 ERROR o.o.c.internal.CoreActivator[:117] - Couldn't create
file 'webapps/static/version'. java.io.FileNotFoundException:
webapps/static/version (Permission denied)
```

```
root@inucha:~# ls -la /var/lib/openhab/webapps
total 12
drwxr-xr-x 3 root    root    4096 Sep 13 17:30 .
drwxr-xr-x 4 openhab openhab 4096 Sep 13 17:30 ..
drwxr-xr-x 2 root    openhab 4096 Sep 12 22:41 static
root@inucha:~#
[...]

Issue #886, opened 2014-02-23
apt-repo, deb packages 1.4 problems collection


theoweiss:
#### Issue for collecting known bugs and enhancement requests to the 1.4 build:

1. db4o persistence saves data to /etc/openhab/jetty/etc/db4o/store.db4o
2. files in /etc/openhab are owned and writable by the openhab user
3. dependency on java to the openhab-runtime package
4. optional java arguments in /etc/default/openhab like gnu.io.rxtx.SerialPorts
5. dropbox add-on tries to write authfile at runtime into /opt/openhab/authfile.dbx (presumably cwd)

theoweiss:
Recently I stumbled upon https://bintray.com, they offer free for OSS distribution and have maven integration and support deb file and rpm repos? Could be an opportunity, @teichsta what do you think?

teichsta:
> Could be an opportunity

absolutely, thanks for the hint! Have created an openHAB organisation and invited you as a member :-)

theoweiss:
"5. dropbox add-on tries to write authfile at runtime into /opt/openhab/authfile.dbx (presumably cwd)":
After a quick look into the dropbox bundle, there seems to be an undocumented configuration variable "dropbox:contentdir" which could be used configure the output directory. Should we use this variable or switch to the openhab.userdata property?

theoweiss:
Fixed with 1.7.

PR #2141, opened 2015-02-15
Fixes for deb packaging + systemd support

theoweiss:
Fix for #1429

Tested for debian 7 (sysv init) and debian 8 (systemd init).

It would be nice to have some more tests and also upgrade tests!

Changes:
- removed fragile handling of openhab group from deb packaging
- updated db4o persistence service to use the smarthome.userdata system property
- relies on evolved rrd4j persistence service, which uses the smarthome.userdata property
- changed /etc/init.d/openhab to set smarthome.userdata property to /var/lib/openhab
- added systemd service description for openhab
- added script which sets permissions on various runtime directories according to OPENHAB_USER_GROUP variable
- removed symlink configuration from pom.xml for /usr/share/openhab/webapps/static to /var/lib/openhab/webapps,
  because it was broken and may be have unintentional behavior for the greent addon.

How it works
- For security reasons the openHAB runtime runs as unprivileged non-root user
  (openHAB runtime user).
- The default user is named "openhab" and with the primary group "openhab".
- The user and group can be adapted with the variable USER_AND_GROUP in the
  /etc/default/openhab configuration file.
- The openHAB runtime user and group are created using the deb file postinst
  script. It's postinst not preinst because user and group information can be
  and is obtained from /etc/default/openhab.
- Permissions used for the deb file
  - All files and directories contained in the deb file are owned by user root and 
    group root.
  - Most files and directories are readable by others (file mode 644 or 444, directory mode 755).
  - For security reasons some files should only be readable for the openHAB runtime user
    and his primary group (mode 440). In the deb file they are owned by root:root, the
    ownership will be adapted at daemon startup.
    For now these files are /etc/openhab/configurations/user.cfg and /etc/openhab/jetty/etc/keystore.
  - There are files and directories which must be writable by the openHAB runtime user.
    These files / directories like all others are owned by root:root in the deb file.
    The permissions and owner are adapted at daemon startup time. Therefore changes to 
    USER_AND_GROUP in /etc/default/openhab during application lifetime are honored.
- The permissions and owner of files and directories which should be writable by the openHAB 
  runtime user are adopted at startup time.
  - the directory /usr/share/openhab/webapps/static is owned by the openHAB runtime user because
    the "version" and "uuid" files are created at runtime startup.
  - the directory /var/log/openhab is owned by the openHAB runtime user to enable the creation 
    of log files for the openHAB runtime
  - /var/lib/openHAB/workspace is used as osgi workspace and therefore owned by the openHAB 
    runtime user.
  - sub directories of /var/lib/openHAB/ are used for storing user data like persistence 
    database files, therefore /var/lib/openHAB/ is owned by the openHAB runtime user
    - rrd4j persistence stores its files in /var/lib/openhab/rrd4j, this is achieved by setting the 
      the java system property smarthome.userdata=/var/lib/openhab.
    - db4o persistence stores its files in /var/lib/openhab/db4o, this is achieved by setting
      the java system property smarthome.userdata=/var/lib/openhab.
- systemd and sysv init from one deb package
  - is inspired by the deb packaging of elasticsearch:
    https://github.com/elasticsearch/elasticsearch/issues/8493
    and relies on this recommendation from a debian forum:
    http://lists.alioth.debian.org/pipermail/pkg-systemd-maintainers/2014-December/005077.html
  - systemd and sysv init share configuration through /usr/share/openhab/bin/openhab.in.sh and 
    functionality through /usr/share/openhab/bin/setpermissions.sh
- systemd init executes /usr/share/openhab/bin/setpermissions.sh as root, this is granted by
  setting PermissionsStartOnly=true


teichsta:
Hi Theo,

Thanks for this contribution!

A short question: "smarthome.userdata" in openHAB1?

Best, Thomas

- sent from a mobile device -

theoweiss:
Hi Thomas,

yes I found it in the RRD4j service and added it to db4o:
air:openhab theo$ grep -r smarthome.userdata bundles/persistence/
bundles/persistence//org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java:		String progArg = System.getProperty("smarthome.userdata“);

Best,
Theo

> Am 16.02.2015 um 09:17 schrieb Thomas Eichstädt-Engelen <notifications@github.com>:
> 
> Hi Theo,
> 
> Thanks for this contribution!
> 
> A short question: "smarthome.userdata" in openHAB1?
> 
> Best, Thomas

teichsta:
> yes I found it in the RRD4j service and added it to db4o:

i'd prefer to rename this variable to "openhab.userdata" in opposite to the variable used by ESH. Could you please rename both occurrences?

Furthermore i am wondering wich of the commits are the fix for #1429? Assuming there could be an 1.6.3 release i would like add that fix to it but not the while PR.

Thanks, Thomas E.-E.

theoweiss:
> Am 20.02.2015 um 06:44 schrieb Thomas Eichstädt-Engelen <notifications@github.com>:
> 
> yes I found it in the RRD4j service and added it to db4o:
> 
> i'd prefer to rename this variable to "openhab.userdata" in opposite to the variable used by ESH. Could you please rename both occurrences?
yes, no problem.
> 
> Furthermore i am wondering wich of the commits are the fix for #1429? Assuming there could be an 1.6.3 release i would like add that fix to it but not the while PR.
> 
An 1.6.3 release would be great. Cherry picking some changes would be challenging, more or less all changes are required to solve the problems. I  would prefer to find way and a few people beta testing the new packages. There are different Linux OS's and different init systems involved, this made the bug fixing more difficult.

distribution/src/deb/bin/openhab.sh
```diff
@@ -0,0 +1,128 @@
...
+#   USER_AND_GROUP -- The user and group the openHAB is running as. The 
+#                     default is "openhab:openhab". In systemd environments
+#                     openhab.service must be adapted, when changing this.
+#   JAVA_ARGS      -- Additional arguments to the JVM.
+#   OPENHAB_ARGS   -- Additional arguments to the openHAB runtime.
+#   DEBUG          -- Start openHAB in debugging mode. The default is no, set 
+#                     it to yes for debugging purposes.
+#   HTTP_PORT      -- 
+#   HTTPS_PORT     -- 
+#   TELNET_PORT    --
```
sja: The description here could be `OSGi console port`

distribution/src/deb/bin/openhab.sh
```diff
@@ -0,0 +1,128 @@
...
+	HTTPS_PORT=8443
+fi
+if [ -z $TELNET_PORT ]; then
+	TELNET_PORT=5555
+fi
+
+JAVA="/usr/bin/java"
+SETOWNER="no"
+
+LAUNCHER=`ls ${OPENHAB_DIR}/server/plugins/org.eclipse.equinox.launcher_*.jar`
```
sja: startup*.sh locate this jar with a `find` - not optimal, but is this really an reliable way?

theoweiss: As long as there is only one org.eclipse.equinox.launcher jar in ${OPENHAB_DIR} it will work.
The "find  | sort | tail" from start.sh will work even if there are several launcher jars in ${OPENHAB_DIR} and it will traverse directories and find also launchers located in sub dirs. I think this is not needed for starting openHAB, is it?

distribution/src/deb/bin/openhab.sh
theoweiss: Sounds good, I will add it.

theoweiss:
I've prepared a commit with the discussed changes. I will to some testing before committing the changes to this PR.

theoweiss:
Just tested the changes: looks good. 
I've also tested an upgrade from 1.6.2, there is a problem to solve: up to 1.6.2 the rrd4j and db4o files are placed into /var/lib/openhab/persistence/{rrd4j, db4o}. Using the openhab.userdir property they will be created in /var/lib/openhab/{rrd4j, db4o} (without persisentce subdir). On upgrading to (1.6.3/)1.7.x the existing databases will no longer be found by the new rrd4j/db4o service.
I see to options:
1. Try to move existing databases to the new directory when upgrading  the deb package using the postinst script.
2. Change the RRD4J and DB4o persistence service implementations to use openhab.userdir/persistence/{rrd4j, db4o} as directories. This should be handled in a similar way in openHAB2.

Any preferences?

teichsta:
i would go for option 2) (fix the Persistence Services)

theoweiss:
I will give it a try.

theoweiss:
I'm not sure, but I think if the installation is upgraded the contents of workspace directory should be wiped out?

theoweiss:
According to this discussion I think wiping out of the workspace is done on every startup, so there is nothing to be done on upgrading the installation: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/openhab/NzpLBsO9mO0/wEpSfofma_QJ

theoweiss:
There is another open bug concerning the deb packages #886
"5. dropbox add-on tries to write authfile at runtime into /opt/openhab/authfile.dbx (presumably cwd)"
I will make an own PR addressing this issue.

PR #2151, opened 2015-02-17
Greent to support sitemap video encoding property

Jzooor:
Added logic to switch to use image widget instead of video widget when the video item in the sitemap has encoding set to "mjpeg"


teichsta:
Hi @Jzooor thanks for this contribution. I am wondering why the user shouldn't simply configure the an ImageWidget rather the Video when a mjpeg stream is going to be received. I would assume devices (and hence the according items) does not change the content type dynamically. Best, Thomas E.-E.

Jzooor:
A similar change was made to the default web interface. Having been several
weeks since I was tinkering with this I can't remember why just using image
widget wouldn't work. I think this was necessary for it to work in both the
app and web interface.

teichsta:
i see, thanks for clarification!

PR #2160, opened 2015-02-19
Introducing library.gps : CoordinateItem, CoordinateType

clinique:
This proposal of new Item type holds both latitude, longitude and altitude if provided.
Bundle with it's usage implication for the Netatmo Binding.


teichsta:
Thanks @clinique for this contribution. Will have a look at this soon …

kaikreuzer:
Thanks, that's an interesting one.
It makes me think about location support in general again as if we introduce this new item type in openHAB 1, we should do the same for ESH / openHAB 2.
This brings up two questions:

1. Should the extension mechanism for items types be kept or shall the new types be directly added to the core? (This question is also still open for the pending CallItem type.) Although I think it was a nice idea, the extensibility does not really work out and new types have to be added (hard-coded) to the designer, the UIs need to be able to support them etc.

2. How does this new item type align with the bigger picture of https://bugs.eclipse.org/bugs/show_bug.cgi?id=443908? There is no concept yet for locations & persons in general, but some when we will need it. So do we maybe constrain ourselves when adding the CoordinateType now or is it maybe rather a perfect first step to get things moving on this topic?

Looking forward to your thoughts on this!

clinique:
Hello Kai,

1/ It was the first time I used the extension mechanism instead of modifiying xtend definition file. The need to introduce dependencies in script and rules along with .ui bundles breaks the principle to have dynamic extension system as it is currently. Then from my point of view, if let in current state, it shall be removed.

2/ I join you on the fact that it will become a real need in evolutions of the project. From what I read, the CoordinateType I propose is a perfect match of GeoCoordinate previewed in prg.eclipse.smarthome.core.location

kaikreuzer:
1) I am still undecided on this; I would love to have this with the extension mechanism, but then again, every item type requires all other parts of the system to be aware of it (UIs, persistence) - so maybe the simpler solution is to have them all in the core.
2) prg.eclipse.smarthome.core.location is outdated and will not make it (in its current form) into ESH, so this does not have to be any reference. I would like to suggest some change to your CoordinateType/Item though:
I would call the long/lat/alt data type "PointType" - taking the vocabulary from http://www.w3.org/2003/01/geo/#vocabulary.
As a name for the item type I would choose "Location" - which can accept data type "Point" and maybe in future other types (relative coordinates, rooms, whatever) as well. This way we would be flexible for the future and just relying on geo coordinates for positioning might be a bad idea.

What do you think about this proposal?

clinique:
1°) I think so and for sure a dynamic extension mechanism would be great but probably not for OH1.
2°) Ok for your proposal regarding the renaming to data type and item type - it obvioulsy makes it more generic.

kaikreuzer:
Thanks, the changes look good to me. 
@teichsta: No concerns from my side regarding the introduction of the new item type then, feel free to go ahead with the review.

@clinique: Would you also like to introduce this new item type in ESH yourself, so that it becomes available in OH2? Feel free to create a PR at https://github.com/eclipse/smarthome.

One useful addition would probably also be a widget for the Classic UI (or rather in general for the sitemap). Could Google Maps be integrated there easily?

bundles/core/org.openhab.core.library/META-INF/MANIFEST.MF
```diff
@@ -11,7 +11,8 @@ Bundle-License: http://www.eclipse.org/legal/epl-v10.html
 Bundle-SymbolicName: org.openhab.core.library
 Bundle-DocURL: http://www.openhab.org
 Originally-Created-By: Apache Maven Bundle Plugin
-Import-Package: org.openhab.core.events,
+Import-Package: org.apache.commons.lang;version="2.6.0",
```
teichsta: please remove the version unless you really would like to restrict your Bundle to use at least 2.6.0

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/LocationItem.java
```diff
@@ -0,0 +1,73 @@
...
+import org.openhab.core.items.GenericItem;
+import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.PointType;
+import org.openhab.core.types.Command;
+import org.openhab.core.types.State;
+import org.openhab.core.types.UnDefType;
+
+
+/**
+ * This item identifies a telephone call by its origin and destination.
```
teichsta: please fix this copy paste error and add a brief explanation of this ItemType

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/LocationItem.java
```diff
@@ -0,0 +1,73 @@
...
+	public List<Class<? extends State>> getAcceptedDataTypes() {
+		return acceptedDataTypes;
+	}
+
+	public List<Class<? extends Command>> getAcceptedCommandTypes() {
+		return acceptedCommandTypes;
+	}
+	
+	/**
+	 * Compute the distance with another Coordinate type,
```
teichsta: could you add a reference where you got the equation from?

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
```diff
@@ -0,0 +1,158 @@
...
+	public String toString() {
+		return String.format("%1$.2f°N, %2$.2f°W, %2$.2f m", latitude, longitude, altitude);
+	}
+
+	@Override
+	public SortedMap<String, PrimitiveType> getConstituents() {
+		SortedMap<String, PrimitiveType> result = new TreeMap<String, PrimitiveType>();
+		result.put(KEY_LATITUDE, getLatitude());
+		result.put(KEY_LONGITUDE, getLongitude());
+		result.put(KEY_ALTITUDE, getAltitude());
```
teichsta: why did you choose the SortedMap rather than a "normal" Map here?

features/org.openhab.designer.feature/feature.xml
```diff
@@ -263,7 +263,12 @@
          install-size="0"
          version="0.0.0"
          unpack="false"/>
-
+   <plugin
+         id="org.openhab.library.gps"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
```
teichsta: please revert

features/org.openhab.runtime.feature/feature.xml
```diff
@@ -211,7 +211,12 @@
          install-size="0"
          version="0.0.0"
          unpack="false"/>
-
+   <plugin
+         id="org.openhab.library.gps"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
```
teichsta: please revert

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/LocationItem.java
```diff
@@ -0,0 +1,73 @@
...
+			double dLat = Math.pow(Math.sin(Math.toRadians(away.getLatitude().doubleValue() - me.getLatitude().doubleValue()) / 2),2);
+			double dLng = Math.pow(Math.sin(Math.toRadians(away.getLongitude().doubleValue() - me.getLongitude().doubleValue()) / 2),2);
+			double a = dLat + Math.cos(Math.toRadians(me.getLatitude().doubleValue()))  
+							* Math.cos(Math.toRadians(away.getLatitude().doubleValue())) * dLng;
+			
+			dist = PointType.WGS84_a * 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1-a));
+		}
+		
+		return new DecimalType(dist);
+	}
```
teichsta: it would be really great to have a Unit-Test to proof the correctness of this equation

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
```diff
@@ -0,0 +1,158 @@
...
+			latitude = 180 - latitude;
+			longitude += 180;
+		} else if (latitude < -90) {
+			latitude = -180 - latitude;
+			longitude += 180;
+		}
+
+		longitude = ((longitude + 180) % 360);
+		if (longitude <= 0) longitude += 360;
+		longitude -= 180;
```
teichsta: it would be really great to have a Unit-Test to proof the correctness of this calculation

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
```diff
@@ -0,0 +1,158 @@
...
+		this.altitude = altitude.doubleValue();
+	}
+	
+	public DecimalType getGravity() {
+		double latRad = Math.toRadians(latitude);
+	    double deltaG = -2000.0 * (altitude/1000) * EARTH_GRAVITATIONAL_CONSTANT / ( Math.pow(WGS84_a,3.0) );
+        double sin2lat = Math.sin(latRad) * Math.sin(latRad);
+        double sin22lat = Math.sin(2.0*latRad) * Math.sin(2.0*latRad);
+        double result = (9.780327 * (1.0 + 5.3024e-3 * sin2lat - 5.8e-6 * sin22lat) + deltaG); 
+        return new DecimalType(result);
```
teichsta: it would be really great to have a Unit-Test to proof the correctness of this calculation

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
clinique: I used the same principle as in library.tel

teichsta: gotcha ;-)

To answer my question. I have no clue (anymore) …

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/LocationItem.java
clinique: Wondering whether it would be better to include some kind of dedicated library like this one : 
http://www.gavaghan.org/blog/free-source-code/geodesy-library-vincentys-formula-java/

clinique: Code coming from here : http://stackoverflow.com/questions/837872/calculate-distance-in-meters-when-you-know-longitude-and-latitude-in-java
Bug also wondering wether it would be better to use some kind of dedicated library like this one : http://www.gavaghan.org/blog/free-source-code/geodesy-library-vincentys-formula-java/

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/LocationItem.java
clinique: Added a Unit test class

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
clinique: Added a Unit test class

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
clinique: Added a Unit test class

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/LocationItem.java
teichsta: great, thanks!

StefanBertels:
Just reading this after stumbling over this new runtime type in 1.7.0. 

I had a `Group Location` in my items resulting in `NullPointerException`.

I think it is ok to add types directly in the runtime to make developer life easier -- even if this might result in problems like [#2874](https://github.com/openhab/openhab/issues/2874).

I guess it's not an option to allow type names as item names (parsing), to solve any conflicts. But exceptions should contain more information so anyone upgrading to a newer openhab might not run into this.

PR #2167, opened 2015-02-22
Parameter 121 for doorsensor incorrectly defined as byte

wnagele:
Fixing as length 4 int (see: http://www.pepper1.net/zwavedb/device/120)


wnagele:
The problem is also described here: https://groups.google.com/d/msg/openhab/bhnXaX_SIRI/RswXoz8Z5OoJ

cdjackson:
Well spotted! Thanks :)

Issue #2170, opened 2015-02-22
Command 1 not implemented for HRT4-ZW


casperleenheer:
It seems that a SET is not implemented for the HRT4-ZW. The log states:

2015-02-22 12:03:08.781 DEBUG o.o.b.z.i.p.ZWaveController[:191]- Message: class = ApplicationCommandHandler (0x04), type = Request (0x00), payload = 08 03 03 40 01 01 
2015-02-22 12:03:08.781 DEBUG o.o.b.z.i.p.s.ApplicationCommandMessageClass[:38]- NODE 3: Application Command Request (ALIVE:DONE)
2015-02-22 12:03:08.783 DEBUG o.o.b.z.i.p.s.ApplicationCommandMessageClass[:57]- NODE 3: Incoming command class THERMOSTAT_MODE (0x40)
2015-02-22 12:03:08.784 DEBUG o.o.b.z.i.p.c.ZWaveThermostatModeCommandClass[:95]- NODE 3: Received Thermostat Mode Request
2015-02-22 12:03:08.785 WARN  o.o.b.z.i.p.c.ZWaveThermostatModeCommandClass[:101]- NODE 3: Command 1 not implemented.

cdjackson:
This command is the SET command - probably meant to be going from the thermostat to the controller. The binding doesn't implement this because it's normally the unit doing the setting, not being set.

This binding probably should be modified to process this command as if it is a REPORT (we already do this for other classes). Doing this will mean that OH knows that the thermostat has sent a command to the boiler to change the mode.

cdjackson:
@digitaldan - since you wrote most of the thermostat stuff, do you have any issues with this? We do the same in the BASIC command class, and maybe some of the other switches where a light sends a SET in an association, and we interpret this as a report, so I don't think there's any issue here.
If you're happy I'll merge... Casper is hopefully testing this afternoon...

cdjackson:
Casper has confirmed this change is working correctly with the HRT4...

cdjackson:
Closed by #2171

PR #2171, opened 2015-02-22
Treat thermostat mode SET as a REPORT

cdjackson:
This treats a SET as a REPORT. This can happen on the HRT4 systems that have a separate
thermostat controller that controls a boiler controller. The thermostat is sending the SET MODE
command to the boiler, and by handling this command here we find out as well.
https://github.com/openhab/openhab/issues/2170


cdjackson:
Closes #2170

PR #2173, opened 2015-02-22
Mios binding rollershutter support

Jakey69:
In previous versions the mios binding did not allow you to create bindings with "Rollershutter" items. With this update it becomes possible. 

Create the "miosShutterCommand.map" transformation file in "openhab/configurations/transform" with the following contents:

    INCREASE=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=?++)
    DECREASE=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=?--)
    DOWN=urn:upnp-org:serviceId:WindowCovering1/Down
    UP=urn:upnp-org:serviceId:WindowCovering1/Up
    STOP=urn:upnp-org:serviceId:WindowCovering1/Stop
    _defaultCommand=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=??)

Now, in your items file you can create Rollershutter items:

Rollershutter Kitchen 	"Kitchen "	(Shutters)  {mios="unit:micasa,device:13/service/WindowCovering1,command:MAP(miosShutterCommand.map)"}


mrguessed:
Thanks for the contribution Jack!

A few quick comments on the changes:

a) The MiOS Binding syntax for a parameterless UPnP _Action_ call exists, and is:
```
    DOWN=urn:upnp-org:serviceId:WindowCovering1/Down()
```

Basically it has an empty set of brackets "()" at the end, like a regular function/method call.  This is supported by the current code (`DEVICE_PATTERN` has optional elements specified), so there's no need to modify `MiosUnitConnector.java`

b) The example `miosShutterCommand.map` transform file should be part of this checkin.
This would be placed with the other Command map files in the examples/transform/ directory of the MiOS Binding code.  Using (a) above, it'll now look like:
```INCREASE=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=?++)
DECREASE=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=?--)
DOWN=urn:upnp-org:serviceId:WindowCovering1/Down()
UP=urn:upnp-org:serviceId:WindowCovering1/Up()
STOP=urn:upnp-org:serviceId:WindowCovering1/Stop()
_defaultCommand=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=??)
```

c) An example _Rollershutter_ Item binding should be added to the MiOS Binding's `README.md` in Git
Since this is new, it will be handy for other users if the example usage is added to the `README.md` in the MiOS Binding directory in Git. 

I copy/promote these to the Wiki once the release is made, but it'll ensure that others know how to use it correctly.

d) `DeviceDefaults.properties` should be altered to have the relevant default-values for Rollershutters.
It'll be a line like:
```
service/urn\:upnp-org\:serviceId\:WindowCovering1=command:MAP(miosShutterCommand.map)
```
That'll ensure that users can simply bind an Item using just:
```
Rollershutter Kitchen "Kitchen " (Shutters) {mios="unit:micasa,device:13/service/WindowCovering1"}
```

Jakey69:
Hi mrguessed,

Thanks for your feedback! I updated the code according to your remarks:

a) I reverted the MiosUnitConnector.java

b) I committed the miosShutterCommand.map file

c) I added the example to the README.md file

d) I updated the DeviceDefaults.properties file. However, when I tested the Rollershutter with the default mapping it didn't work. To make it work I had to change the DeviceBindingConfig.java file. The SERVICE_IN_PATTERN always expects a serviceVar (or did I misinterpret the pattern?). So I made the serviceVar optional (did I do that the right way?). Now, it works!

Please check my changes.

Kind regards,
Jack

bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/config/DeviceBindingConfig.java
```diff
@@ -193,6 +193,12 @@
 				"urn:upnp-org:serviceId:Dimming1/StepUp()");
 		COMMAND_DEFAULTS.put("DECREASE",
 				"urn:upnp-org:serviceId:Dimming1/StepDown()");
+		COMMAND_DEFAULTS.put("UP",
+				"urn:upnp-org:serviceId:WindowCovering1/UP()");
```
mrguessed: MiOS Action-names are case-sensitive, like their UPnP Counterparts, so these defaults would need to be `Up()`, `Down()` and `Stop()` respectively.

bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/config/DeviceBindingConfig.java
```diff
@@ -241,7 +247,7 @@
 	}
 
 	private static final Pattern SERVICE_IN_PATTERN = Pattern
-			.compile("service/(?<serviceName>.+)/(?<serviceVar>.+)");
+			.compile("service/(?<serviceName>.+)(/(?<serviceVar>.+))*");
```
mrguessed: You'll want to use a REGEX Pattern more like "_X_?" (_X_, once or not at all)  at the end instead of "_X_*" (_X_, zero or more times) so the pattern-match is as tight as possible.

bundles/binding/org.openhab.binding.mios/README.md
```diff
@@ -391,6 +391,26 @@ The `examples/transform/miosDimmerCommand.map` file has a definition that handle
     _defaultCommand=urn:upnp-org:serviceId:Dimming1/SetLoadLevelTarget(newLoadlevelTarget=??)
 
 
+##### A Roller shutter...
```
mrguessed: Looks good!

distribution/openhabhome/configurations/transform/miosShutterCommand.map
```diff
@@ -0,0 +1,4 @@
+DOWN=urn:upnp-org:serviceId:WindowCovering1/Down()
```
mrguessed: The content looks good, but since they're MiOS specific I've put these into the directory:
```
bundles/binding/org.openhab.binding.mios/examples/transform/
```

and let people copy them over (manually) as needed.

mrguessed:
Hey Jack, I added the comments inline to make it easier.  Only a few minor tweaks.

I'm guessing, but I suspect Thomas will also want the Git Commits compressed for the PR as well. I follow this guide that Kai pointed us to a while back, which is quite handy:
* http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

Jakey69:
Made a fix to the SERVICE_IN_PATTERN: variables were parsed incorrectly.

mrguessed:
Looks like the auto-merge kicked in, and failed.  

Two changes of mine were merged this morning, one for the support of Color Item bindings and one for Performance work.

There was also a merge yesterday for the MiOS Action binding work.  This change included reformatting the entire codebase to the [new] 120ch wide comments.

teichsta:
Thanks guys! As Mark pointed out, it would be great if you could squash your commits into a single one in oder to reduce cluttering the history. See Marks comment above for a small How-To. Best, Thomas E.-E.

Jakey69:
Ok, I will give it a try. Never done that before.

Jakey69:
Hi Thomas,

I think I did something wrong... I followed your guide to squash my commits into a single one and committed that one. I then saw that something was wrong in a file and I fixed it and committed. I then did another squash because I would now again have two commits. I pushed that one to my branch. If you now look at my latest commit, you will only see my fix and not the changes to the other files, which were done in the first squash.... Is this a problem for you and if so, how can I help to fix it?

Kind regards,
Jack

mrguessed:
Sometimes it's easier to create a new branch, put your merged changes into that Branch, and then submit a new PR.  I've wrestled with Git and had to do this more than a few time (mostly when I early merged from my origin/master or accidentally from upstream/master)

Over time, it gets better, but it's trivially easy to munge-up multiple works-in-progress ;)

In it's current form, it's a little hard to tell if the changes I'd review are correct or something from the history.

eg. Use of `STOP()` as an Action name vs `Stop()` (UPnP is case sensitive)
eg. I've never had to trap exceptions before when using a `Matcher.group(<namedGroup>)` when that Capture of the pattern is missing, but the Pattern matches overall.

I suspect I'm looking at older stuff, and not the latest.

Jakey69:
I think your suggestion is good. I will create a new branch and put my changes in that one.

teichsta:
ok, give me a sign when you are finished. Thanks, Thomas E.-E.

Jakey69:
I created a new branch with all the changes and created a new pull request #2291

teichsta:
i assume this PR has been superseded by #2309, correct?

mrguessed:
This PR can be nuked, it is no longer needed.

PR #2189, opened 2015-02-24
Change Indication from sec to minutes.

pierok13:
I just saw that the parameter is in minutes, not in seconds...

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/vision/zp3102.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/vision/zp3102.xml
@@ -22,7 +22,7 @@
 			<Maximum>15300</Maximum>
 			<Default>180</Default>
 			<Size>1</Size>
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.zwave/database/vision/zp3102.xml
```diff
@@ -22,7 +22,7 @@
 			<Maximum>15300</Maximum>
 			<Default>180</Default>
```
cdjackson: I think there are some errors here. It's not your fault - clearly they were already in the file, but it would be good if you can fix them...
Default says 180, but I read in the manual that it's 3. Maximum says 15300, but clearly this is not possible for a 1 byte value,so it should be 255. I dont know what minimum is set to but I guess it should be 0?
If you dont mind fixing these it would be great.

pierok13:
updated

PR #2193, opened 2015-02-25
Set lastEventTime to actual time when restarting

JSurf:
Set lastEventTime when restarting, so the binding is only restarted every alive.interval. Without this fix it will be restarted every 60 seconds in configurations with only a few components and low communication.

I currently only have one 4-channel switch and no sensors, so there is not a lot of communication/events. 
Without that fix, communication is restarted every 60 seconds after the first alive interval is over. 

I think it is good to reset the time interval when reconnecting, because we know we just had a successful connection/communication with the CCU even if there were no new events.

also see:
https://groups.google.com/forum/#!searchin/openhab/homematic$2060/openhab/dmAh25FyA98/IcO5jwWbw7MJ


JSurf:
There is an additional commit which should not be here

PR #2195, opened 2015-02-25
Set lastEventTime to actual time when restarting

JSurf:
Set lastEventTime when restarting, so the binding is only restarted every alive.interval. Without this fix it will be restarted every 60 seconds in configurations with only a few components and low communication.

I currently only have one 4-channel switch and no sensors, so there is not a lot of communication/events.
Without that fix, communication is restarted every 60 seconds after the first alive interval is over.

I think it is good to reset the time interval when reconnecting, because we know we just had a successful connection/communication with the CCU even if there were no new events.

also see:
https://groups.google.com/forum/#!searchin/openhab/homematic$2060/openhab/dmAh25FyA98/IcO5jwWbw7MJ

### DIFF
--- a/bundles/binding/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/HomematicCommunicator.java
+++ b/bundles/binding/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/HomematicCommunicator.java
@@ -87,8 +87,10 @@ public class HomematicCommunicator implements HomematicCallbackReceiver {
 
 				homematicCallbackServer.start();
 				homematicClient.registerCallback();
-
+				
 				scheduleFirstRefresh();
+
+				lastEventTime = System.currentTimeMillis();
 			} catch (Exception e) {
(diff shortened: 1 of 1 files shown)


teichsta:
@gerrieg could you please take care of this? Thanks, Thomas E.-E.

gerrieg:
@teichsta 
That's OK, will fix the problem.

@JSurf 
Thanks for fixing this. You still have to increase the alive.interval to a higher value if you have no components which sends data in regular intervals. It's not necessary to reconnect every 300 seconds (default).

BR, Gerhard

gerrieg:
@teichsta 
Don't merge this please! I noticed, that it has an impact of the normal keep alive validation. I will make a PR in the next days which handles both, normal validation and one with low communication (no sensors).

gerrieg:
see https://github.com/openhab/openhab/pull/2332 and homematic:reconnect.interval configuration in openhab.cfg, Wiki updated.

Best, Gerhard

JSurf:
Thanks @gerrieg so I'm closing my pull request

PR #2203, opened 2015-02-27
DSC alarm now sends user code over serial interface

### LINKED ISSUE #2202: DSC alarm can't disarm over serial interface
The DSC alarm binding cannot disarm the alarm when connected over the serial port via an IT-100. The user code from the config file is not being sent to the alarm, and instead a default code of 1234 is supplied, which is rejected by the alarm system.

beowulfe:
When communicating with a DSC alarm system over the IT-100 interface, a user code is required to disarm the system. This user code was being sent when using an envisalink connection to DSC, and we reuse that code for the serial connection with this change. The code is also padded to 6 digits per the IT-100 Developer Guide.

This fixes #2202

### DIFF
--- a/bundles/binding/org.openhab.binding.dscalarm/src/main/java/org/openhab/binding/dscalarm/internal/DSCAlarmActiveBinding.java
+++ b/bundles/binding/org.openhab.binding.dscalarm/src/main/java/org/openhab/binding/dscalarm/internal/DSCAlarmActiveBinding.java
@@ -495,7 +495,7 @@ public class DSCAlarmActiveBinding extends AbstractActiveBinding<DSCAlarmBinding
     	switch (connectorType) {
 	    	case SERIAL:
 	     		if (api == null) {
-	    			api = new API(serialPort, baudRate);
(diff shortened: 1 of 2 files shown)


beowulfe:
@RSStephens - I'd like you to take a look at this PR as well, if you would. I'm not ultra familiar with the DSC ecosystem, so I'm not sure if there's another serial connection besides the IT-100 that would break with this change.

RSStephens:
Hello @beowulfe,

I am glad to see you using the DSC Alarm binding with an IT-100 interface.  There was no testing done with that interface, as I only use the Envisalink interface, so it is good to see you doing this.  I don't see any problems with your added code; that should work fine with the rest of the code.  This binding was only designed to work with the IT-100 or the Envisalink, so there would be no other type of serial interface that would work.  I am also getting ready to submit a PR, for other fixes and features, so I will make sure I incorporate your changes.  Thank you for your contribution.

Regards,

Russell

beowulfe:
Thanks for verifying!

teichsta:
Thanks @beowulfe for your contribution!

PR #2206, opened 2015-02-27
Fix scale for not defined.

benoit934:
No tested but this may work.


teichsta:
I am lacking the context … could you please reference to a Bug or explain what this PR is about?

benoit934:
It's when the value is not defined not a number but, "-" with the map it's working, not with scale.

clinique:
as the scale transform currently falls down to map transform when the value is not a number, it would be better to use the same logic for Undef inseated of extending the current regular expression no ?

benoit934:
I thinks it's a better error handling, maybe faster too, i don't know i'm not a pro of JAVA.

teichsta:
Thanks @clinique and @benoit934 for this discussion. I tend to go with @clinique (use the same logic for Undef). Can you please adjust this PR accordingly?

teichsta:
Any news?

benoit934:
I'm sorry I don't really have the time, maybe the issue can just be closed.

But when I coding in others languages I try to handle everything faster by limiting any sub conditions.

teichsta:
@clinique would you mind adding the proposed behaviour yourself? If no, please reference the new PR here. Thanks, Thomas E.-E.

Issue #2208, opened 2015-02-28
InfluxDB persistence bug?


staehler:
Hi,
I'm quite unsure, if I make a mistake, but all seems to end in a persistence bug.
After restarting OH, the persisted items have wrong values. Let me show this at the debug output:

16:07:42.684 [DEBUG] [.r.i.engine.RuleTriggerManager:438  ] - Scheduled rule Verreist wenn 24 Stunden keine Anwesenheit with cron expression 0 0 0 * * ?
16:10:58.232 [DEBUG] [i.i.InfluxDBPersistenceService:357  ] - query string: select value, time from Verreist limit 1
16:10:58.404 [DEBUG] [.p.internal.PersistenceManager:390  ] - Restored item state from '28.02.2015 16:06:00' for item 'Verreist' -> 'ON'
16:11:06.726 [DEBUG] [i.internal.GenericItemProvider:334  ] - Start processing binding configuration of Item 'Verreist (Type=SwitchItem, State=Uninitialized)' with 'KNXGenericBindingProvider' reader.
16:12:00.452 [TRACE] [o.c.p.internal.PersistItemsJob:56   ] - Storing item 'Verreist' with persistence service 'influxdb' took 6ms
16:13:00.467 [TRACE] [o.c.p.internal.PersistItemsJob:56   ] - Storing item 'Verreist' with persistence service 'influxdb' took 5ms
16:14:00.453 [TRACE] [o.c.p.internal.PersistItemsJob:56   ] - Storing item 'Verreist' with persistence service 'influxdb' took 6ms
16:15:07.805 [TRACE] [o.c.p.internal.PersistItemsJob:56   ] - Storing item 'Verreist' with persistence service 'influxdb' took 6ms

Before stopping OH, I set the item "Verreist" to OFF. The influxDB persisted this value, so I could query  "select value, time from Verreist limit 1" with the result 0.
But have a look at the debug log! The persistenceManager restored the value from '28.02.2015 16:06:00' for item 'Verreist' -> 'ON', which is definately not the value, which I found with my DB query. I've made a snapshot of the influxDB, which proves, that the value was 0 at '28.02.2015 16:06:00'!!!
![influxdb](https://cloud.githubusercontent.com/assets/6258158/6426642/65deeb3c-bf66-11e4-99bf-6a544f6fea8a.PNG)

What is going wrong with this persistence service? Before I used rrd4j and it worked correctly.

Please help!!!

Kind regards, Michael

[...]

PR #2211, opened 2015-02-28
Influxdb bugfix for issue #2208 (restoring switch or contact state)

theoweiss:
Waiting for test results from Michael issue #2208


staehler:
Hi Theo,
my test was successfull (the items wer restored correctly), your fix works. Thank you very much!
Regards,
Michael

PR #2229, opened 2015-03-04
Proposal for an Item created event in rules

clinique:
This PR aims to bring a solution for initializing items in introducing a "Item created" trigger in rules.
The target is to provide an alternate way to go to avoid "System started" rules as discussed in google groups.

Sample usage : 
rule "test"
when Item sys_start created
then
   sys_start.postUpdate(now.toString)
end

### DIFF
--- a/bundles/model/org.openhab.model.rule/src/org/openhab/model/rule/Rules.xtext
+++ b/bundles/model/org.openhab.model.rule/src/org/openhab/model/rule/Rules.xtext
@@ -27,6 +27,7 @@ EventTrigger:
 	UpdateEventTrigger |
 	CommandEventTrigger |
 	ChangedEventTrigger |
+	CreatedEventTrigger |
 	TimerTrigger |
 	SystemTrigger
 ;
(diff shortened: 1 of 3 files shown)


maggu2810:
Sounds like a nice additional rule trigger.
This would work if the "action" block does only access this item and does not rely on other ones.
Or you have to do some more "workarounds".

Ok, here a very theoretical example:
The state of the switch "item1" should be the logical not of the state of "item2".
So, we could write a rule "if item1 is created, set item1.state to !item2.state".
But if item2 is created after item1 this is still not working.

clinique:
Yes, triggering "Item1 created" doesn't mean that item2 is created and the creation order is out of control at this level. I suppose the approach for you case would be to test state of item2 in "Item1 created" and vice versa.

cribskip:
Would be cool if there would also be a "all items loaded" and "persistence loaded" trigger -  then we could test at one stage for asks startup stages

clinique:
Yes, good ideas, will take a look at it - but for the moment don't know if this PR will be accepted

clinique:
From what I see in RuleEngine, "all items loaded" would be equivalent to system started.

PR #2232, opened 2015-03-04
Add distributionSince and varianceSince methods to persistence extensions

edog1973:
Added methods to the persistence extension for distributionSince and varianceSince that allow finding the variance and standard distribution of a value over a given time.

### DIFF
--- a/bundles/core/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/core/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -358,7 +358,7 @@ public class PersistenceExtensions implements ManagedService {
 			return null;
 		}
 	}
-
+	
 	/**
 	 * Gets the average value of the state of a given <code>item</code> since a certain point in time. 
 	 * The {@link PersistenceService} identified by the <code>serviceName</code> is used. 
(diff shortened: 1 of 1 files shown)


bundles/core/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -393,6 +393,90 @@ static public DecimalType averageSince(Item item, AbstractInstant timestamp, Str
...
+	 * Gets the standard deviation value of the state of a given <code>item</code> since a certain point in time. 
+	 * The {@link PersistenceService} identified by the <code>serviceName</code> is used. 
+	 * 
+	 * @param item the item to get the average state value for
+	 * @param the point in time to start the check 
+	 * @param serviceName the name of the {@link PersistenceService} to use
+	 * @return the standard deviation of the value since the given point in time
+	 */
+	static public DecimalType deviationSince(Item item, AbstractInstant timestamp, String serviceName) {
+		DecimalType variance = varianceSince(item, timestamp, defaultService);
```
teichsta: please change "defaultService" into "serviceName" here

edog1973:
After doing more testing I noticed an issue.  When using data that has the current value already persisted to the database the calculations were wrong because the current value was included twice  To mitigate this the current value is now only included in the calculation if it's value is the same as the newest db record.  The issue affected the averageSince so I updated it's logic as well.

teichsta:
looks good to me. Could you please squash your commits into one? Thanks, Thomas E.-E.

kaikreuzer:
Does this apply to ESH as well?

teichsta:
@edog1973 could you finally please squash your commits into one? Thanks, Thomas E.-E.

teichsta:
@edog1973 any chance to squash now? Will reschedule to 1.8 otherwise …

PR #2250, opened 2015-03-07
Added Linear LB60Z-1 to the database

odlund:
I couldn't find this devices in the database in the binding, nor in the http://www.pepper1.net/zwavedb/ database.

Information about the device is gathered from http://www.linearcorp.com/product_detail.php?productId=1756

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/linear/lb60z-1.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/linear/lb60z-1.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>LB60Z-1</Model>
+	<Label lang="en">Dimmable LED Light Bulb</Label>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/database/linear/lb60z-1.xml
```diff
@@ -0,0 +1,68 @@
...
+				<Label lang="en">Disable dim level memory</Label>
+			</Item>
+			<Item>
+				<Value>1</Value>
+				<Label lang="en">Enable dim level memory</Label>
+			</Item>
+		</Parameter>
+	</Configuration>
+		<Associations>
+		<Group>
```
cdjackson: I don't know what this group is used for, but is it needed to send updated to the controller? If so, you need to add the tag ```<SetToController>true</SetToController>```

odlund:
Thanks for feedback, updated the Association group based on the information I could find in the documentation

PR #2259, opened 2015-03-08
Null dereference checks. Indentation fixes.

llamahunter:
Protect against a null dereference when looking up associations for devices that don't support it.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
@@ -539,11 +539,11 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 				if(networkMonitor != null) {
 					record = new OpenHABConfigurationRecord(domain, "LastHeal", "Heal Status", true);
 					if (node.getHealState() == null) {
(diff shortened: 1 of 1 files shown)


cdjackson:
I'm a bit surprised that this has come up now since there are a number of devices that don't have associations... Anyway...

I'm tempted to suggest an alternative approach. Instead of checking for null from the return, the return from ```getProductAssociationGroups``` should return an empty list if there are no associations. This will ensure that other calls to this method don't have the same issue.  This is what we do for the configuration list as well...

So, can I suggest removing the null check in the configuration file here, and add the return of the empty collection in the ```getProductAssociationGroups``` method?

What do you think?

llamahunter:
I didn't change the null check on getting the group list.  I left that more or less as-is.  I changed the code after getting the association command class to check for null.  If the node doesn't support it, it will be null.  Previously the code was written assuming it was never null, and then would throw an NPE when it tried to use it. I copied the pattern more or less exactly from the "parameters/" handler just above, which does deal with the case correctly.

This only seems to happen when trying to expand the association list using HABmin, which is probably why few people have encountered it.  Also, it possibly also requires a zwave node config xml file that says there are association groups, but that the runtime is unable to verify because of a communication issue (in this case, the lack of support for the Security command class to communicate with the schlage lock)

cdjackson:
Ah, ok.  It's difficult to tell what changes with the way GitHub displays the changes.  I'll merge this but might change the way this works with the return from the association. 

Yes, this code is only related to habmin interface. 

I'll merge when I get to the hotel.

PR #2262, opened 2015-03-08
Comfoair: EWT bugfixes / enhancements

joltcoke:
- Fix EWT low/high temperature values
- Add EWT active state
- Minor typos fixed

### DIFF
--- a/bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirCommandType.java
+++ b/bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirCommandType.java
@@ -146,20 +146,20 @@ public enum ComfoAirCommandType {
 		}
 	},
 
-	EWT_TEMPERATUR_HIGH {
+	EWT_TEMPERATUR_LOW {
 		{
-			key = "ewt_temperatur_high";
-			data_type = DataTypeTemperature.class;
+			key = "ewt_temperatur_low";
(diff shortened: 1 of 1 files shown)


teichsta:
@HolgerHees could you please have a look at this fix? Can you confirm this as a Bug/Fix? Thanks, Thomas E.-E.

HolgerHees:
I can't really test it because I does not own a EWT device.

But it is confusing me that the EWT_TEMPERATUR is interpreted as a temperature value and the EWT_TEMPERATUR_LOW and EWT_TEMPERATUR_HIGH as a number value.

based on the protocol description (not a official one)

http://www.see-solutions.de/sonstiges/Protokollbeschreibung_ComfoAir.pdf

all 3 values should be a temperature value.

@joltcoke which zehnder device are you using?

joltcoke:
@teichsta If you're looking at a strictly bug fix only release, then feel free to cherry pick just changeset 37995fe.

@HolgerHees I'm having a CA350 Luxe + Comfofond-L Eco + Ease CC (disconnected). The change fixes two problems, index Low / High swapped and no (temp + 20) * 2 magic applied. For the latter please refer to the spec you've linked. None of the temperatures in reply 0xEC have an asterisk attached - no conversion necessary. The range for low temp is 0 - 15 with +1/-1 increments, thats why they send raw values.

HolgerHees:
@joltcoke you are right :-) thanks for fixing!

@teichsta the other changeset a8ab09c is not a bugfix, but it does not affect any other areas.

i'm fine with the whole commit.

PR #2264, opened 2015-03-09
MapDb persistence

JSurf:
A simple persistence provider that only saves the last value. Intention is to use this for "reloadOnStartup" items because all other presistence options have their drawbacks if values are only needed for reload

They:
a) grow in time
b) require complex installs (mysql, influxdb, ...)
c) rrd4j can't store all item types

See:
https://groups.google.com/forum/#!category-topic/openhab/umfFKg5uAM4


paulianttila:
Great work Jens!

Few additions which would make me happy (at least). Many of us is running OH in flash based systems, so to minimize unnecessary writes to flash based systems.

1) Commit db to file only if item value is changed. Maybe this should be configurable.
2) Introduce commit interval, which define how often db is saved to the file. So updated items are normally just stored to memory map and after commit interval is exceeded, map is saved to file (if any items are updated).

-Pali

JSurf:
1) should be easy, will implement this

i have to think a bit more about 2) Calling commit in an extra thread sounds a little arkward. 

You could workaround this requirement with the persistence configuration and define a time based strategy, so i'm not sure if code in the persistence module is required.

```
// persistence strategies have a name and a definition and are referred to in the "Items" section
Strategies {
	everyFiveMinutes : "*/5 * * * * ?"

	// if no strategy is specified for an item entry below, the default list will be used
	default = everyChange
}

/* 
 * Each line in this section defines for which item(s) which strategy(ies) should be applied.
 * You can list single items, use "*" for all items or "groupitem*" for all members of a group
 * item (excl. the group item itself).
 */
Items {
	* : strategy = everyFiveMinutes, restoreOnStartup	
}
```

paulianttila:
2) Timer should be enough, I think? When item (any one) is updated, start timer if it's not already running. And when persistence service is shutdown, store to file if map is "dirty".

teichsta:
1) please keep in mind that you'd have to a) retain the old state as member variable or b) query the old State from the Service before committing the old c) implement a mixture of both … i would not consider that too easy
2) using a timer seems to me an appropriate approach. See Db4oPersistenceService where i've implemented such a time already.

@JSurf please keep us posted when you'd consider this PR "mergeable".

Thanks, Thomas E.-E.

JSurf:
Thanks for the feedback, will keep you updated about the progress

JSurf:
Current version seems to work fine on fast systems, but hung openhab on my raspberry while starting up. Maybe i have hit another "startup timing"-bug on slow hardware

JSurf:
Wanted to dive deeper in this, and tried again on my raspberry pi. This time it worked flawlessly. Don't know what happened last time. so i would be happy if you could review/comment my latest changes @teichsta @paulianttila

AcidReptile:
Works fine as far as I can judge. Thanks, useful binding.

bundles/persistence/org.openhab.persistence.mapdb/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,25 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: openHAB MapDB Persistence Bundle
+Bundle-SymbolicName: org.openhab.persistence.mapdb
+Bundle-Version: 1.6.2
```
teichsta: please change version to 1.7.0.qualifier

bundles/persistence/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDBItem.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Date;
+
+import org.openhab.core.persistence.HistoricItem;
+import org.openhab.core.types.State;
+
+/**
+ * This is a Java bean used to persist item states with timestamps in the database.
+ * 
+ * @author Jens Viebig
+ * @since 1.0.0
```
teichsta: please change all occurrences of @since to 1.7.0

bundles/persistence/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDBItem.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ */
+public class MapDBItem implements HistoricItem, Serializable {
+
+	/**
+	 * 
+	 */
```
teichsta: remove this empty comment of (even better) briefly explain the purpose of this member

bundles/persistence/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDBPersistenceActivator.java
```diff
@@ -0,0 +1,41 @@
...
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Extension of the default OSGi bundle activator
+ * 
+ * @author Jens Viebig
+ * @since 1.0.0
```
teichsta: change version

bundles/persistence/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDBPersistenceService.java
```diff
@@ -0,0 +1,265 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * This is the implementation of the MapDB {@link PersistenceService}. To learn
+ * more about MapDB please visit their <a href="http://www.mapdb.org/">website</a>.
+ * 
+ * @author Jens Viebig
+ * @since 1.0.0
```
teichsta: change version

JSurf:
Added Wiki Page:
https://github.com/openhab/openhab/wiki/mapdb-Persistence

AcidReptile:
Do not forget that user need to create folder %openhabdir%/etc/mapdb manually. Otherwise service is not working.

theoweiss:
Meanwhile there was a decision to rename openhab.userdata to smarthome.userdata for openHAB2 compatibility reasons. Therefore you should adapt the plugin accordingly.

teichsta:
… and please squash all commits into one. Thanks, Thomas E.-E.

JSurf:
Will have a look soon

JSurf:
Changed openhab.userdata to smarthome.userdata, added code to automatically create mapdb folder when not available and squashed to one commit

PR #2267, opened 2015-03-09
Correction some errors of projects settings

slawekjaranowski:


slawekjaranowski:
Thank you for your approval.

I see that you assign this commit to version 1.6.3 but I don't see this commit in branch  1.6.
Do you need my help to also do it on branch 1.6?

PR #2233, opened 2015-03-05
Initial Commit of the Autelis Pool Control binding for openHAB

digitaldan:
Signed-off-by: Dan Cunningham <dan@digitaldan.com> (github: digitaldan)

PR #2269, opened 2015-03-09
Only listen to changed events, not updates.

digitaldan:
The rest code was broadcasting both update and changed events, we only want to push changed message. Note, i did no cleanup or formatting changes as the already commented out code may prove useful as a guide for later debugging.  This was in reference to PR #2233 and a conversation around caching.

### DIFF
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ResourceStateChangeListener.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ResourceStateChangeListener.java
@@ -130,7 +130,7 @@ abstract public class ResourceStateChangeListener {
 		stateChangeListener = new StateChangeListener() {
 			// don't react on update events
 			public void stateUpdated(Item item, State state) {
-				broadcaster.broadcast(item);
+//				broadcaster.broadcast(item);
 				// if the group has a base item and thus might calculate its state
 				// as a DecimalType or other, we also consider it to be necessary to
 				// send an update to the client as the label of the item might have changed,


digitaldan:
@teichsta, i was looking back through the commit history to see when this code was added, I found https://github.com/openhab/openhab/commit/f025f609a7494f3d19fb8cf3de5373dd5a2462c4  which was titled "merged khome's atmosphere fix"  , do you know what this was fixing?  Looking at the commit, I can't tell if  subscribing to updates was intentional or not.  For reference, the commit subscribed to both update and changed messages which ends up delivering a lot of noise to rest clients that only care about changed data.  My PR is commenting out the subscribing to updates, so that only changed data is delivered.  Thanks!

teichsta:
> do you know what this was fixing

no, admittedly i can't remember anymore! If i got the discussion in #2233 and at  https://groups.google.com/d/msg/openhab/UeexB3MRLBU/T-_NXi5MVrAJ correctly the noise should be handled even by small devices. Do you have any figures?

digitaldan:
In a basic case if I have 5 bindings that poll every 5 seconds, and each binding has 5 items, that means we are sending down 25 items over 5 seconds to REST clients, which is basically all the time, even though nothing has changed.  I agree even small clients can handle this, but those on cellular connections are strained, and devices left on all the time such as touch panels end up wasting a lot of cpu.  The issues we recently addressed in the REST binding had far more volume then this, items were being updated several times per second even thought the values had not changed.  

This PR keeps everything intact with the original  commit, but without subscribing to updates which I now understand was the original intent of the system. From the discussion you mentioned I now realized that the system understands changed vs updated events, which you would think I would know by now ;-)   So I think this is the correct behavior ( and how it use to be)

 Also if I read the comment above the line I disabled it reads "don't react on update events" , so maybe this was just a mistake left in during some testing or debugging? I feel good about this so far in my testing.

teichsta:
ok, let's try this "fix" ;-)

digitaldan:
Fingers crossed!

PR #2276, opened 2015-03-11
some fixes for the CometVisu

peuter:
- skip broadcasting of update events
- changed Interceptor name (due to Atmosphere upgrade)
- fix for broken update mechanisn after item model changes


teichsta:
Thanks @peuter, if there are open issues you are referring to, please link them here. Best, Thomas E.-E.

PR #2277, opened 2015-03-11
ZWave configs for Dragon Tech products PA-100, WD-100, WS-100

llamahunter:
I think the BOM on the products.xml file is actually needed by some tool I use, so I put it back with this change.  It should not be needed (as per the unicode spec), but I was getting very weird conflicts on utf8 characters,  and will probably affect others, so backing out that part of my previous change.


cdjackson:
Thanks for sorting this out :)

PR #2300, opened 2015-03-14
adapted rrd4j chart servlet to be compatible with openHAB 2

kaikreuzer:

### DIFF
--- a/bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java
+++ b/bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java
@@ -58,7 +58,7 @@ public class RRD4jService implements QueryablePersistenceService {
 
 	private static final String DATASOURCE_STATE = "state";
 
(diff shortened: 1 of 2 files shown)


kaikreuzer:
@theoweiss This PR reverts your change of smarthome.userdata to openhab.userdata, because that broke the openHAB 2 compatibility (and such the feature for which this line of code had been added at all). I am not sure, whether you need this for the deb packages in some way, but if so, you would have to come up with some other solution.

theoweiss:
@kaikreuzer I changed this on request of @teichsta https://github.com/openhab/openhab/pull/2141#issuecomment-75192703
No problem to revert this change. But we should also adapt db4o, may be the new mapdb persistence plugin and the startup scripts. I can care for the startup scripts and db4o if you want me to.

kaikreuzer:
> But we should also adapt db4o, may be the new mapdb persistence plugin and the startup scripts. I can care for the startup scripts and db4o if you want me to.

Yes, this would be great if you could test them on openHAB 2 and do the necessary changes! Don't forget to update the [compat matrix](https://github.com/openhab/openhab2/blob/master/docs/sources/installation/compatibility.md) afterwards.

theoweiss:
OK. Face openHAB 2 IDE setup ;-)

teichsta:
> No problem to revert this change.

thanks @theoweiss, sorry for the inconvenience!

PR #2307, opened 2015-03-15
[Z-Wave] Philio PST02 Slim-Multisensor correction.

marcelerkel:
Configuration of the PST02 is similar to, but not compatible with, the PSM02.
* To receive PIR Motion and Contact updates in OpenHAB set the notification type to "Sensor Binary Report" (set bit 4 to 1 of configuration parameter 7 Customer Function)

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
@@ -1,0 +1,235 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>PST02</Model>
+	<Label lang="en">Slim Multi-Sensor</Label>
+	<CommandClasses>
+		<Class><id>0x20</id></Class>	<!-- COMMAND_CLASS_BASIC -->
+		<Class><id>0x30</id></Class>	<!-- COMMAND_CLASS_SENSOR_BINARY_V2 -->
+		<Class><id>0x31</id></Class>	<!-- COMMAND_CLASS_SENSOR_MULTILEVEL_V5 -->
+		<Class><id>0x59</id></Class>	<!-- COMMAND_CLASS_ASSOCIATION_GRP_INFO -->
+		<Class><id>0x5A</id></Class>	<!-- COMMAND_CLASS_DEVICE_RESET_LOCALLY -->
+		<Class><id>0x5E</id></Class>	<!-- COMMAND_CLASS_ZWAVEPLUS_INFO_V2 -->
+		<Class><id>0x70</id></Class>	<!-- COMMAND_CLASS_CONFIGURATION -->
+		<Class><id>0x71</id></Class>	<!-- COMMAND_CLASS_NOTIFICATION_V4 -->
+		<Class><id>0x72</id></Class>	<!-- COMMAND_CLASS_MANUFACTURER_SPECIFIC_V2 --> 
+		<Class><id>0x73</id></Class>	<!-- COMMAND_CLASS_POWERLEVEL -->
+		<Class><id>0x7A</id></Class>	<!-- COMMAND_CLASS_FIRMWARE_UPDATE_MD_V2 -->
+		<Class><id>0x80</id></Class>	<!-- COMMAND_CLASS_BATTERY -->
+		<Class><id>0x84</id></Class>	<!-- COMMAND_CLASS_WAKE_UP_V2 -->
+		<Class><id>0x85</id></Class>	<!-- COMMAND_CLASS_ASSOCIATION_V2-->
+		<Class><id>0x86</id></Class>	<!-- COMMAND_CLASS_VERSION_V2 -->
+		<Class><id>0x8F</id></Class>	<!-- COMMAND_CLASS_MULTI_CMD -->
+		<Class><id>0x98</id></Class>	<!-- COMMAND_CLASS_SECURITY -->
+		<Class><id>0xEF</id></Class>	<!-- COMMAND_CLASS_MARK -->
+	</CommandClasses>
+	<Configuration>
+		<Parameter>
+			<Index>2</Index>
+			<Type>byte</Type>
+			<Default>-1</Default>
+			<Minimum>-1</Minimum>
+			<Maximum>100</Maximum>
+			<Size>1</Size>
+			<Label lang="en">Basic Set Level</Label>
+			<Help lang="en">-1	   ->  	0xFF(-1) turns on the light.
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
```diff
@@ -0,0 +1,235 @@
...
+		<Class><id>0x8F</id></Class>	<!-- COMMAND_CLASS_MULTI_CMD -->
+		<Class><id>0x98</id></Class>	<!-- COMMAND_CLASS_SECURITY -->
+		<Class><id>0xEF</id></Class>	<!-- COMMAND_CLASS_MARK -->
+	</CommandClasses>
+	<Configuration>
+		<Parameter>
+			<Index>2</Index>
+			<Type>byte</Type>
+			<Default>-1</Default>
+			<Minimum>-1</Minimum>
```
cdjackson: Negative numbers are not allowed. ZWave defines this as an unsigned value.

bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
```diff
@@ -0,0 +1,235 @@
...
+		</Parameter>
+		<Parameter>
+			<Index>6</Index>
+			<Type>byte</Type>
+			<Default>4</Default>
+			<Minimum>0</Minimum>
+			<Maximum>127</Maximum>
+			<Size>1</Size>
+			<Label lang="en">Multi-Sensor Function Switch</Label>
+			<Help lang="en">Bit 0: Disable magnetic integrate illumination to turn on the lighting nodes in the association group 2. 1: Disable, 0: Enable
```
cdjackson: This will render badly in the UI. It might be worth encapsulating it in CDATA tag and then adding some HTML to break the lines?

bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
```diff
@@ -0,0 +1,235 @@
...
+		<Parameter>
+			<Index>5</Index>
+			<Type>byte</Type>
+			<Default>0</Default>
+			<Minimum>0</Minimum>
+			<Maximum>127</Maximum>
+			<Size>1</Size>
+			<Label lang="en">Operation Mode</Label>
+			<Help lang="en">Bit 0: Reserved.
+Bit 1: 1 means enable test mode; 0 means normal mode. Notice: Ignored if DIP Switch is not set to Customer Mode.
```
cdjackson: This will render badly in the UI. It might be worth encapsulating it in CDATA tag and then adding some HTML to break the lines?

bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
```diff
@@ -0,0 +1,235 @@
...
+		</Parameter>
+		<Parameter>
+			<Index>7</Index>
+			<Type>byte</Type>
+			<Default>4</Default>
+			<Minimum>0</Minimum>
+			<Maximum>127</Maximum>
+			<Size>1</Size>
+			<Label lang="en">Customer Function</Label>
+			<Help lang="en">Bit 0: Reserved.
```
cdjackson: This will render badly in the UI. It might be worth encapsulating it in CDATA tag and then adding some HTML to break the lines?

cdjackson:
Hi.
Thanks for this. I would like to at least fix the SetToController issue since this is important for the operation of the device.

See -:
https://github.com/cdjackson/HABmin2/wiki/Z-Wave-Product-Database#association-configuration

It would probably be nice if you updated the longer help messages to make them HTML so that they format correctly, but up to you - if you don't want to, then just say.
Cheers
Chris

marcelerkel:
Hi Chris,

Thanks for the feedback. You are right, the formating is totally lost in the rendering. I found an example in the database (qubino/zmnhba.xml) that uses CDATA tags. I also found an example in the database (philio/pan11_1.xml) for the SetToController tag so I'll add that as well.

In the mean time I also found a manual of the Philio PSM02 which seems to have been written by a native English speaker. I will try to find some inspiration from this manual to improve the help messages a bit (as a none native English speaker I have trouble understanding the PST02 manual and I can imagine I'm not alone).
When I'm ready I'll force push the changes to update the PR. This will probably be not be before tomorrow evening (Central European Time) though.

Cheers,
Marcel

bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
marcelerkel: How should we handle these? The manual says 0xFF, but that seems strange for the Minimum (255) value in relation to the Maximum (100) value because then the minimum value is larger than the maximum value. The psm02.xml also uses -1 (that's where I copied it from :) )

marcelerkel:
I probably should have squashed the two into one commit. If necessary I can do it, just let me know.

Cheers,
Marcel

bundles/binding/org.openhab.binding.zwave/database/philio/pst02.xml
cdjackson: I think all we can do is to add min=0 and max=255. I know this isn't really right, but using -1 will definately cause problems (I've had this issue in the past).
FF is normally used in zwave to mean on (but this isn't normally written as -1), which is what you have here, so I think I'd just use 0 to 255 and use the information in the help to let the user know what to do?

marcelerkel: Mmmm.. Even if I try to set it to 255, then habmin still reports is as -1. Also the node.xml file shows -1:
```
      <commandClass>CONFIGURATION</commandClass>
      <configurationCommandClass>
        <version>1</version>
        <instances>1</instances>
        <configParameters>
          <entry>
            <int>2</int>
            <configurationParameter>
              <index>2</index>
              <size>1</size>
              <value>-1</value>
              <readOnly>false</readOnly>
              <writeOnly>false</writeOnly>
            </configurationParameter>
          </entry>
```
This is because Java only uses signed integral types: http://stackoverflow.com/questions/22452609/converting-0xff-gives-me-1-signed-vs-unsigned

To me it would seem confusing to have the help text mention that one needs to enter 255 while habmin actually displays -1.

cdjackson: Ok - let's leave it like it is. I'm just worried this will cause problems - I know it has in the past, but I can't remember what the problem was now.

marcelerkel: ok

PR #2308, opened 2015-03-15
rename openhab.userdata to smarthome.userdata

theoweiss:
rename openhab.userdata to smarthome.userdata. more: #2300


theoweiss:
Tested with openHAB 1; Will do testing with openHAB 2 within the next days.

PR #2332, opened 2015-03-20
Homematic: Added reconnect.interval

gerrieg:


bundles/binding/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/HomematicCommunicator.java
```diff
@@ -109,7 +112,7 @@ public void run() {
 				logger.debug("Initial Homematic datapoints reload");
 				context.getStateHolder().reloadDatapoints();
 			}
-		}, 60000);
+		}, 61000);
```
teichsta: loginfo and actual interval does not match (1m vs. 61sec)

bundles/binding/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/bus/HomematicBinding.java
```diff
@@ -183,15 +183,24 @@ protected void internalReceiveUpdate(String itemName, State newState) {
...
+	 * configured time or the reconnect interval is reached.
 	 */
 	@Override
 	protected void execute() {
-		long timeSinceLastEvent = (System.currentTimeMillis() - communicator.getLastEventTime()) / 1000;
-		if (timeSinceLastEvent > context.getConfig().getAliveInterval()) {
-			logger.info("No event since {} seconds, refreshing Homematic server connections", timeSinceLastEvent);
-			communicator.stop();
-			communicator.start();
+		if (context.getConfig().getReconnectInterval() == null) {
```
teichsta: I would prefer to remove this if-else block and rather initialise reconnectInterval with a reasonable value. What do you think?

teichsta:
Hi @gerrieg thanks for this contribution. Please find my review comments inline. Best, Thomas E.-E.

gerrieg:
Hi @teichsta 

Null is a reasonable value for reconnectInterval. reconnectInterval and aliveInterval are mutually exclusive. If reconnectInterval != null, aliveInterval is ignored.

yes, 61 seconds is not a minute, i know ;-) This is a easy workaround for a rare imperfection when a reconnect and a datapoint reload occurs at the same time. With the extra second that never happens. I've added a comment.

Best, Gerhard

PR #2341, opened 2015-03-21
Update fgsd002.xml

filhome:


cdjackson:
There's a couple of issues with this -:
1) I think you've probably deleted too many classes out of the ```<CommandClasses>``` list, but this doesn't matter too much.
2) You've got ```<SetToController>``` on every association. Normally for Fibaro devices this is only needed on group 3 - for this sensor, it looks like it should be group 1 (Lifeline). I don't think that the others should be needed and I think it would be best to remove them and add them later if you find they are needed.  The problem with adding this to every group is that it might result in multiple notifications, which might (or not) cause some small issues, but it might also reduce the battery life if it is sending notifications to or three times if not needed.

Let me know what you think - otherwise it looks good. Thanks for making the effort to put the file together :)

Chris

filhome:
I made some changes according your comments and my debug logs.

cdjackson:
Thanks for this.

Cheers
Chris

filhome:
It's a pleasure, and thanks for your help and your patience.
I wish I will help for this beautiful project

PR #2342, opened 2015-03-22
Zwave device version comparisson

cdjackson:
This changes the version comparison from a double to a string. The double was previously used when it was thought that the version was a simple number - it's not - the minor and major parts need to be considered separately and here the double causes problems.
This converts all usage of versions to strings. The actual string comparison hasn't changed as that used the OSGi version classes since the minor/major issue was discovered.


cdjackson:
https://github.com/cdjackson/HABmin/issues/226

PR #2345, opened 2015-03-22
Fix: Error on creating tables for DateTimeItems

Dennis650:


teichsta:
Hi @Dennis650, i would tend to accept your proposed change. But honestly i don't know enough about MySQL. Since others use the Service already and didn't complain i am wondering whether is change is correct or not. Didn't the correct piece of information/specification defining the DATETIME correctly. Can you help? Best, Thomas E.-E.

Dennis650:
Hi, 

with DATETIME(3) you can define Fractional Seconds. This is only available on the newest versions of mysql > 5.6.4. The standard mysql-package of many linux distributions (such as debian / raspbian 4 pi) is still mysql 5.5, which doesn't support this optional new feature. For RaspberryPI Users, there is a diskussion here: http://www.raspberrypi.org/forums/viewtopic.php?t=68037&p=496714

So if you want to use openhab on a pi with mysql, you have to build mysql yourself, to use it with the mysql-binding.

Greetings
Dennis

teichsta:
thanks for the clarification …

PR #2355, opened 2015-03-23
MiOS - Improve diagnostics for failed HTTP requests to MiOS Unit.

mrguessed:
When a MiOS Unit is called, as a result of a Command, these calls can fail.  When they do, we need better diagnostics as to when/if it's failing, so we can work out why.

Sometimes, this is just because the configurable Transformations have been incorrectly setup, but we'd fail silently which made it hard to debug in the field.

### DIFF
--- a/bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/MiosUnitConnector.java
+++ b/bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/MiosUnitConnector.java
@@ -35,6 +35,7 @@ import org.openhab.core.types.State;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.ning.http.client.AsyncCompletionHandler;
 import com.ning.http.client.AsyncHttpClient;
 import com.ning.http.client.AsyncHttpClientConfig;
 import com.ning.http.client.AsyncHttpClientConfig.Builder;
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/MiosUnitConnector.java
```diff
@@ -257,7 +258,29 @@ private void callMios(String url) {
 		logger.debug("callMios: Would like to fire off the URL '{}'", url);
 
 		try {
-			getAsyncHttpClient().prepareGet(url).execute();
+			@SuppressWarnings("unused")
+			Future<Integer> f = getAsyncHttpClient().prepareGet(url).execute(new AsyncCompletionHandler<Integer>() {
```
teichsta: does execute() provide a timeout parameter? It would be good to limit the timeout to something reasonable …

mrguessed:
Timeouts are set on the underlying `AsyncHttpClientConfig` object.  In my case, I do that up-front when I'm building the `MiosUnitConnector` object, using user-supplied settings (in the `openhab.cfg`)

This is all setup in the constructor of `MiosUnitConnector`:
      https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/MiosUnitConnector.java#L110

teichsta:
perfect, thanks! Didn't look into the code too deep before asking :-)

Issue #2367, opened 2015-03-26
Ecobee: Bug: Multiple OAuth instances overwrite each other


watou:
If you attempt to connect to two or more Ecobee accounts with settings in openhab.cfg, one sweep across all items bound to Ecobee will be made for each account, generating errors that the execute() method failed to retrieve the specified thermostat identifier.  The real problem is that each distinct account should only attempt to update items that are bound to that distinct account.  (This bug is fixed in an imminent pull request but is being documented here for completeness.)

Issue #2366, opened 2015-03-25
Ecobee: Enh: Add remote sensor support


watou:
Now that the Ecobee API supports access to the ecobee3's [remote sensors](https://www.ecobee.com/home/developer/api/documentation/v1/objects/RemoteSensor.shtml), add support to the Ecobee binding so that these values can be exposed in OpenHAB items.  So, for example, if your ecobee3 with ID 123456789 has 3 remote sensors connected named Kitchen, Basement and Bedroom, allow item binding configurations like

```
Number "Kitchen temperature [%.1f °F]" { ecobee="<[123456789#remoteSensors(Kitchen).capability(temperature).value]" }
Number "Basement humidity [%d %%]" { ecobee="<[123456789#remoteSensors(Basement).capability(humidity).value]" }
Contact "Bedroom occupancy [%s]" { ecobee="<[123456789#remoteSensors(Bedroom).capability(occupancy).value]" }
```

PR #2368, opened 2015-03-26
Add support for ecobee3 remote sensors + bug fixes

watou:
This PR fixes a problem using the binding across two or more accounts #2367, removes interrogation of the item registry to suppress updates (based on recent discussions), and implements access to the wireless remote temperature/occupancy sensors that can be added to the ecobee3.  Issue #2366.

### DIFF
--- a/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
+++ b/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
@@ -51,7 +51,6 @@ import static org.apache.commons.lang.StringUtils.isNotBlank;
 
 import org.openhab.core.binding.AbstractActiveBinding;
 import org.openhab.core.binding.BindingProvider;
-import org.openhab.core.items.ItemRegistry;
 import org.openhab.core.library.types.DateTimeType;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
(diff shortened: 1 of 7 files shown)


bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
```diff
@@ -369,28 +339,19 @@ private void readEcobee(OAuthCredentials oauthCredentials, Selection selection)
 		// Iterate through bindings and update all inbound values.
 		for (final EcobeeBindingProvider provider : this.providers) {
 			for (final String itemName : provider.getItemNames()) {
-				if (provider.isInBound(itemName)) {
+				if (provider.isInBound(itemName)
+						&& oauthCredentials == getOAuthCredentials(provider.getUserid(itemName))) {
```
teichsta: i am pretty sure you didn't want to check for object identity here, right? Better use equals() …

watou: Since the objects come from the same map, I am actually only interested if they are the same object, not equal in content.  If they were equal in content but different references, it would be a bug.  Perhaps there is a better way to do this.  This line is a mirror of line 711.

teichsta: ok! would it then be more readable to extract this line into it's own method with a reasonable comment? I am quite sure even you will ask yourself about the meaning of this statement in several weeks :-)

watou: Added a credentialsMatch method to more clearly document the logic.

watou:
Anything else I need to do with to this pull request, Thomas?

watou:
Hi Thomas, is there anything else you would like me to do to this pull request before you merge it?  Thank you!

teichsta:
Yes, you could please squash the commits into one. Will then merge immediately! Thanks, Thomas E.-E.

teichsta:
thanks, @watou :-)

PR #2375, opened 2015-03-28
Added items to support cooling

frami:


teichsta:
Thanks @frami for this contribution. Could you please:

* change all LogStatements to use the parameterised Logging Feature of SLF4J (see http://logback.qos.ch/manual/architecture.html#parametrized)
* squash the PR into one commit after incorporating the Feedback

Best, Thomas E.-E.

frami:
@teichsta Thanks for the feedback. I'm not so familiar with git. Now everything is in one commit. Do I have to do something with the PR?
Best regards, Michael

PR #2379, opened 2015-03-29
Ihc command specific outbinding

smerschjohann:
This modification allows to specify different resources and the wanted value related to the called command.

# Configuration Syntax
## Before
```
ihc=">123456"
```

Only a single resourceId could be referenced. Values were automatically determined based on the item state.

## After

```
ihc=">[UP:12345:100],>[DOWN:0x9bcd:100],>[STOP:99999]"
```

Syntax:  >[COMMAND:RESOURCE_ID(:VALUE)]

1. Command which is called
2. The resource id to use (hex or int)
3. optional
    * if not specified the value will be determined like before
    * 1: OnOffType.ON is sent to the resource
    * 0: OnOffType.OFF is sent to the resource
    * value > 1: Sequence is sent: OnOfType.ON, sleep(value), OnOffType.OFF

### DIFF
--- a/bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcBinding.java
+++ b/bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcBinding.java
@@ -29,6 +29,7 @@ import org.openhab.core.binding.AbstractActiveBinding;
 import org.openhab.core.binding.BindingChangeListener;
 import org.openhab.core.binding.BindingProvider;
 import org.openhab.core.items.Item;
+import org.openhab.core.library.types.OnOffType;
 import org.openhab.core.types.Command;
 import org.openhab.core.types.State;
 import org.openhab.core.types.Type;
@@ -45,13 +46,14 @@ import org.slf4j.LoggerFactory;
  * Binding also polls resources from controller where interval is configured.
  * 
  * @author Pauli Anttila
+ * @author Simon Merschjohann
  * @since 1.1.0
  */
 public class IhcBinding extends AbstractActiveBinding<IhcBindingProvider>
 		implements ManagedService, IhcEventListener, BindingChangeListener {
 
-	private static final Logger logger = 
-		LoggerFactory.getLogger(IhcBinding.class);
+	private static final Logger logger = LoggerFactory
+			.getLogger(IhcBinding.class);
 
 	private long refreshInterval = 1000;
 
@@ -78,18 +80,19 @@ public class IhcBinding extends AbstractActiveBinding<IhcBindingProvider>
 	 * download from controller
 	 */
 	private static String projectFile = null;
-	
+
 	/** Path for resource's dump. Dump is useful to find out resource id's. */
 	private static String dumpResourceFile = null;
 
 	/** Timeout for controller communication */
 	private static int timeout = 5000;
 
-	/** Store current state of the controller, use to recognize when
-	 * controller state is changed
+	/**
+	 * Store current state of the controller, use to recognize when controller
+	 * state is changed
 	 */
 	private WSControllerState controllerState = null;
-	
+
 	/**
 	 * Reminder to slow down resource value notification ordering from
 	 * controller.
@@ -97,7 +100,7 @@ public class IhcBinding extends AbstractActiveBinding<IhcBindingProvider>
 	private NotificationsRequestReminder reminder = null;
 	private boolean reconnectRequest = false;
 	private boolean valueNotificationRequest = false;
-	
+
 	@Override
 	protected String getName() {
 		return "IHC / ELKO LS refresh and notification listener service";
@@ -112,30 +115,30 @@ public class IhcBinding extends AbstractActiveBinding<IhcBindingProvider>
 	}
 
 	public void deactivate(ComponentContext componentContext) {
-		discnnect();
+		disconnect();
 	}
-	
+
 	protected boolean isReconnectRequestActivated() {
 		synchronized (IhcBinding.class) {
 			return reconnectRequest;
-		}		
+		}
 	}
 
 	protected void setReconnectRequest(boolean reconnect) {
 		synchronized (IhcBinding.class) {
 			this.reconnectRequest = reconnect;
-		}	
+		}
 	}
 
 	protected boolean isValueNotificationRequestActivated() {
 		synchronized (IhcBinding.class) {
-			return valueNotificationRequest;	
+			return valueNotificationRequest;
 		}
 	}
 
 	protected void setValueNotificationRequest(boolean valueNotificationRequest) {
 		synchronized (IhcBinding.class) {
-			this.valueNotificationRequest = valueNotificationRequest;	
+			this.valueNotificationRequest = valueNotificationRequest;
 		}
 	}
 
@@ -144,7 +147,7 @@ public class IhcBinding extends AbstractActiveBinding<IhcBindingProvider>
 	 * 
 	 */
 	public void connect() throws IhcExecption {
-		
+
 		if (StringUtils.isNotBlank(ip) && StringUtils.isNotBlank(username)
 				&& StringUtils.isNotBlank(password)) {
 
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
```diff
@@ -87,50 +94,92 @@ public void processBindingConfiguration(String context, Item item,
...
+		String[] splittedCommands = bindingConfig.split(",");
+		
+		for(String split : splittedCommands) {
+			if (split.startsWith(">")) {
+				try {
+					//out binding
+					String resourceCommand = split.substring(1);
+					
+					final String commandRegex = "\\[([\\w*]+):(0x[0-9a-fA-F]+|\\d+)(?::(\\d+)){0,1}\\]"; 
+					Pattern p = Pattern.compile(commandRegex);
```
teichsta: please extract the Regex String into a precompiled final static Pattern

teichsta:
@paulianttila could you please have a look at these changes? Thanks, Thomas E.-E.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcBinding.java
```diff
@@ -396,28 +399,79 @@ private void updateResource(String itemName, Type type, boolean updateOnlyOutBin
...
+						value = IhcDataConverter.convertCommandToResourceValue(OnOffType.ON, value,
+								enumValues);
+		
+						boolean result = updateResource(value);
+						
+						if (result == true) {
+							logger.debug("Item '{}' trigger started",
+									itemName);
+							
+							Thread.sleep(val);
```
paulianttila: This sleep duration probably should be limited? Long sleep will block another resource id updates.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcBinding.java
```diff
@@ -396,28 +399,79 @@ private void updateResource(String itemName, Type type, boolean updateOnlyOutBin
...
+					
+					ArrayList<IhcEnumValue> enumValues = null;
+	
+					if (value instanceof WSEnumValue) {
+	
+						enumValues = ihc.getEnumValues(((WSEnumValue) value)
+								.getDefinitionTypeID());
+					}
+	
+					Integer val = provider.getValue(itemName, (Command)type);
```
paulianttila: Could you add comment here, how this val and trigger special case works. Without comment, code could be little bit confusing  afterwards.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
```diff
@@ -87,50 +94,92 @@ public void processBindingConfiguration(String context, Item item,
...
+		
+					String resourceId = configParts[0];
+					config.resourceId = getResourceIdFromString(resourceId);
+					config.outBindingOnly = false;
+					config.outCommandMap = null;
+					
+					if (configParts.length == 2)
+						config.refreshInterval = Integer.parseInt(configParts[1]);
+				} else {
+					throw new BindingConfigParseException("Only out binding or normal binding supported at this time");
```
paulianttila: Is this exception message correct?

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
```diff
@@ -87,50 +94,92 @@ public void processBindingConfiguration(String context, Item item,
...
-
-			String resourceId = configParts[0];
-
-			if (resourceId.startsWith("0x")) {
-				config.resourceId = Integer.parseInt(
-						resourceId.replace("0x", ""), 16);
 			} else {
-				config.resourceId = Integer.parseInt(resourceId);
+				if(splittedCommands.length < 2) {
+					String[] configParts = bindingConfig.trim().split(":");
```
paulianttila: Length of the configParts should be checked. If length is bigger than 2, exception should be thrown. In old code this was done in line 93.

smerschjohann:
okay I've changed the code at the given positions. In addition I've added "In-Bindings".

You can now combine In- and Out-Bindings together. Example ihc="<12345,>[*:54321]". The corresponding item will get its updates from the 12345 resourceId, and update 54321 by any given command.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
```diff
@@ -169,10 +276,25 @@ public int getRefreshInterval(String itemName) {
...
+		
+		
+		return isOutBinding;
+	}
+	
+	@Override
+	public boolean hasInBinding(String itemName) {
+		IhcBindingConfig config = (IhcBindingConfig) bindingConfigs.get(itemName);
+		
+		return config != null ? config.resourceId > 0 : null;
```
paulianttila: Does this really work? Resource id is always bigger than 0, so how old "out binding only" situation is now detected?

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcBinding.java
```diff
@@ -552,14 +608,14 @@ public void resourceValueUpdateReceived(EventObject event,
 		for (IhcBindingProvider provider : providers) {
 			for (String itemName : provider.getItemNames()) {
 
-				int resourceId = provider.getResourceId(itemName);
+				int resourceId = provider.getResourceIdForInBinding(itemName);
 
 				if (value.getResourceID() == resourceId) {
 
-					if (provider.isOutBindingOnly(itemName)) {
+					if (!provider.hasInBinding(itemName)) {
```
paulianttila: See my comment in hasInBinding method implementation

paulianttila:
Btw, you should use eclipse code formatter to format you code changes according to openhab code styling.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
smerschjohann: I changed it in that way, that the "In-Binding" will use the resourceId from IhcBindingConfig. Every Out-Going Binding ("1234", ">1234") will use the IhcOutCommandConfig and will get an entry in the outCommandMap.
An out-only binding does not set the resourceId in IhcBindingConfig and therefore it is assigned with zero and hasInBinding would return false.

smerschjohann:
I reformatted the code with the code style. Don't know if I really like that style (not really readable when the function call is in the next line and not indented wide enough), but well it is existing code and I agree it should be one code style in a file.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
paulianttila: Maybe I didn't fully understand your explanation, but for old style out binding, resource id is always set (see IhcGenericBindingProvider.java line 118).

smerschjohann: There exists two classes: IhcOutCommandConfig (Configuration for an out configuration, this is used to configure resources for outgoing "commands").

In addition there is the IhcBindingConfig class. This class groups all configurations available to an item. The resourceId in this class is used for the InBinding and is by default 0.

When you look at the line 118 you will see that the resourceId is set at the IhcOutCommandConfig and not on the IhcBindingConfig and therefore the resourceId responsible for the In-Binding remains zero.

teichsta:
looks good to me altogether now. If you could now squash your commits into one it will be a pleasure to merge into master :-) Thanks, Thomas E.-E.

bundles/binding/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/IhcGenericBindingProvider.java
paulianttila: OK, now I understand :) It's sometimes difficult to follow code just in PR diff, sorry about that.

PR #2381, opened 2015-03-29
Add CommandClasses - Update fgsd002.xml

filhome:


cdjackson:
Just FYI this part of the database is not used at all - it's just there for reference. I'll merge this as it doesn't hurt, but it won't change anything.

PR #2390, opened 2015-04-01
allow negative values for setpoint widget ranges

kaikreuzer:


mepi0011:
Hi, I checked setpoint and negative values with the Build #896 (03.04.2015 03:03:30), but it dose not work or I do something wrong!

My Setpoint:
Setpoint item=HomeTotalPower label="Leistung [%d W]" step=200 minValue=-4000 maxValue=4000

kaikreuzer:
I can confirm that this does not fully seem to solve it.
As a temporary workaround you can have a try with

```
Setpoint item=HomeTotalPower label="Leistung [%d W]" step=200 minValue=-4000.0 maxValue=4000```

PR #2397, opened 2015-04-02
allows negative numbers as commands in widget mappings (see https://code...

kaikreuzer:
....google.com/p/openhab/issues/detail?id=302)


teichsta:
see also https://bugs.eclipse.org/bugs/show_bug.cgi?id=423515 (in case Google-Code will be shutdown completely one day)

Issue #2426, opened 2015-04-09
Enhance SNMP binding to support other protocol versions than v1


J-N-K:
The SNMP binding in its current state uses protocol v1 only. The underlying SNMP4J library supports v2c, too. 

One short-coming of v1 is that 64-bit counters are not supported. Mikrotik Routers (and probably most networking equipment with gigbait Ethernet ports) use these counters for byte-in/out and packet-in/out counters.

It would be better to let the user choose which protocol version should be used, and assume v1 if the corresponding argument is not defined.

The proposed way to do this is to enhance the item's binding configuration to e.g.

<[url:protocol-version:community:OID:refresh]

where possible values for protocol-version are v1, v2c and v3 (v3 not supported by the binding at the moment, but possible in future enhancement).

PR #2427, opened 2015-04-09
Enhance SNMP binding (Issue #2426)

J-N-K:
This PR enhances the SNMP binding as requested in issue #2426.

### DIFF
--- a/bundles/binding/org.openhab.binding.snmp/src/main/java/org/openhab/binding/snmp/internal/SnmpGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.snmp/src/main/java/org/openhab/binding/snmp/internal/SnmpGenericBindingProvider.java
@@ -37,6 +37,7 @@ import org.snmp4j.smi.GenericAddress;
 import org.snmp4j.smi.Integer32;
 import org.snmp4j.smi.OID;
 import org.snmp4j.smi.OctetString;
+import org.snmp4j.mp.SnmpConstants;
 
 /**
  * <p>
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.snmp/src/main/java/org/openhab/binding/snmp/internal/SnmpGenericBindingProvider.java
```diff
@@ -186,45 +196,100 @@ private void parseBindingConfig(SnmpBindingConfig config, Item item, String bind
...
 					newElement.address = GenericAddress.parse("udp:" + outMatcher.group(2).toString() + "/161");
-					newElement.community = new OctetString(outMatcher.group(3).toString());
-					newElement.oid = new OID(outMatcher.group(4).toString());
+					String version = inMatcher.group(3).toString();
+					if (version.equals("v1")) {
+						newElement.snmpVersion = SnmpConstants.version1;
+					} else if (version.equals("v2c")) {
+						newElement.snmpVersion = SnmpConstants.version2c;
+					} else {
+						newElement.snmpVersion = SnmpConstants.version3;
```
teichsta: since v1 was the default before, wouldn't it be better to fall down to v1 in this else case?

teichsta:
Hi @J-N-K, please my review comment inline. Please also squash your commits into one before I'll merge into master. Thanks, Thomas E.-E.

bundles/binding/org.openhab.binding.snmp/src/main/java/org/openhab/binding/snmp/internal/SnmpGenericBindingProvider.java
J-N-K: I can change that, but it doesn't make any difference. If the option is not used, v1 is the default. If there is a protocol version, only v1, v2c and v3 are matched by the Regex, so if it is not v1 and not v2c it sure is v3.

teichsta: sounds reasonable … thanks :-)

J-N-K:
Followed comment and squashed commits, PR is updated.

teichsta:
great, thanks!

Issue #2430, opened 2015-04-10
Absolute path doesn't work in Systeminfo binding on Windows


9037568:
When attempting to get file system information via the Systeminfo binding, if the configuration attempts to specify the directory using an absolute path on a windows system like this:

    Number DirFiles " # of files: [%d]" { systeminfo="DirFiles:60000:f:\stuff" }

the method will fail in the 1.6.2 release, but be flagged as a configuration error in the most recent snapshot.

1.7.0 yields this configuration error: Systeminfo binding must contain 2-3 parts separated by ':'
1.6.2 yields this trace:

    01:39:36.183 [DEBUG] [b.s.internal.SysteminfoBinding:102  ] - item 'DirUsage' is about to be refreshed now
    01:39:36.183 [ERROR] [b.s.internal.SysteminfoBinding:267  ] - Error occured while reading KPI's
    org.hyperic.sigar.SigarException: There are no more files.
    : f
        at org.hyperic.sigar.SigarProxyCache.invoke(SigarProxyCache.java:220) ~[sigar-1.6.4.jar:na]
        at com.sun.proxy.$Proxy74.getDirUsage(Unknown Source) ~[na:na]
        at org.openhab.binding.systeminfo.internal.SysteminfoBinding.getData(SysteminfoBinding.java:222) [bundlefile:na]
        at org.openhab.binding.systeminfo.internal.SysteminfoBinding.execute(SysteminfoBinding.java:108) [bundlefile:na]
        at org.openhab.core.binding.AbstractActiveBinding$BindingActiveService.execute(AbstractActiveBinding.java:156) [org.openhab.core_1.6.2.jar:na]
        at org.openhab.core.service.AbstractActiveService$RefreshThread.run(AbstractActiveService.java:173) [org.openhab.core_1.6.2.jar:na]

I have begun working on a fix for this.  I'm thinking the binding should allow absolute paths to be specified by replacing the colon with a pipe (eg. "f:\stuff" becomes "f|\stuff")

OH: OpenHAB 1.6.2/1.7.0 (SNAPSHOT)
Platform: Windows XP
Related to #2392 / #2393

teichsta:
@paulianttila could you have a look at this, please? Thanks, Thomas E.-E.

paulianttila:
[...]

PR #2443, opened 2015-04-12
Fix absolute path problem in systeminfo Binding and add unit tests

9037568:
This fixes the absolute path problem from #2430 and adds a package of unit tests for the systeminfo binding.

### DIFF
--- a/bundles/binding/org.openhab.binding.systeminfo.test/.classpath
+++ b/bundles/binding/org.openhab.binding.systeminfo.test/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/test/java"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.6"/>
+	<classpathentry kind="output" path="target/test-classes"/>
+</classpath>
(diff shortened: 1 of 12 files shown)


paulianttila:
First of all, thanks for the contribution. It's really nice to have unit tests for system binding as well.

What I'm not fully like is to have special case for DirFiles and DirUsage commands in config parser. I think we should have universal approach here. My proposal for config syntax is following:

DirFiles:60000:'c:\Users\foo'

Where target (path) should also be accepted without ' characters for backward compatibility. 

DirFiles:60000:/Users/foo

Then it should be possible to handle this universally without any special cases. Regex most probably is to easiest way to check the syntax. 

What do you think @9037568?

9037568:
Yes, that could work.  But you're only moving the special case around, not eliminating it.

Another problem is that the apostrophe is rather common in file paths, so you'd have to create special handling for that, too.

paulianttila:
You are right, apostrophe is valid character for file paths, so it can't be used.

Why I still want to have generic solution is that I have introduced FileSystem commands (already merged), which doesn't work windows with absolute paths either. So my opinion is still, that binding provider shouldn't have command specific dependencies.

One possibility is to escape colon character from file path. 

E.g.

DirFiles:60000:c\:\Users\foo

Parser then replace escaped sequence \: by "magic string" before split the config variable and then replace it back to colon after split.

teichsta:
What if you use (escaped) double quotes ' \" '? That should help identifying the different configuration elements better! Any chance to get this fixed until next week? Thanks, Thomas E.-E.

9037568:
Submitted a simplification of the parameter parsing...

teichsta:
Great, if you could now finally sqash your commits into one i am happy to merge this PR. Thanks, Thomas E.-E.

9037568:
Ok. I _think_ that's done. Still don't entirely grok git.

Issue #1845, opened 2014-12-15
Fritzbox Binding: Exception at midnight


juelicher:
Every day at midnight I get the following exception from the Fritzbox Binding:
2014-12-15 00:00:07.708 ERROR o.o.b.f.i.FritzboxBinding[:435]- Lost connection to FritzBox
java.net.SocketException: Socket closed
	at java.net.SocketInputStream.socketRead0(Native Method)
	at java.net.SocketInputStream.read(SocketInputStream.java:152)
	at java.net.SocketInputStream.read(SocketInputStream.java:122)
	at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:283)
	at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:325)
	at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:177)

Probably this is related to reconnecting to the box. The functionality seems not to be affected but I think there should not be any exception if everything is working fine.

hmerk:
@steve-bate can be closed, fix was merged.

steve-bate:
Thanks, @hmerk. Closed.

PR #2445, opened 2015-04-12
Fix issue #1845

J-N-K:
Log.Error if connection closed and no interrupt was received
Log.Info if connection closed by interrupt


teichsta:
@J-N-K this PR seems to be messed up, would you agree? 68 commits and 97 changed files? Could you please fix it? Thanks, Thomas E.-E.

PR #2448, opened 2015-04-12
Bugfix: Handle multiline messages from MAX!Cube

habnefrage:
I added the support for multiline messages, received from MAX!Cube.

When a MAX!Cube has configured many devices with long names and long room names (I have 54 devices with nice german names configured) the configuration data does not fit into one single M_Message. In this case, the cube splits the message into multiple lines. This was not supported by the former implementation.

Now there is a new class "MessageProcessor" that does the job. All messages received from MAX!Cube are passed to an instance of MessageProcessor. The MessageProcessor recignizes when there are additional lines required for a complete message.

### DIFF
--- a/bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
+++ b/bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
@@ -19,6 +19,12 @@ import java.util.Dictionary;
 
 import org.apache.commons.lang.StringUtils;
 import org.openhab.binding.maxcube.MaxCubeBindingProvider;
+import org.openhab.binding.maxcube.internal.exceptions.IncompleteMessageException;
+import org.openhab.binding.maxcube.internal.exceptions.IncorrectMultilineIndexException;
+import org.openhab.binding.maxcube.internal.exceptions.MessageIsWaitingException;
+import org.openhab.binding.maxcube.internal.exceptions.NoMessageAvailableException;
+import org.openhab.binding.maxcube.internal.exceptions.UnprocessableMessageException;
+import org.openhab.binding.maxcube.internal.exceptions.UnsupportedMessageTypeException;
 import org.openhab.binding.maxcube.internal.message.C_Message;
 import org.openhab.binding.maxcube.internal.message.Configuration;
 import org.openhab.binding.maxcube.internal.message.Device;
@@ -28,6 +34,7 @@ import org.openhab.binding.maxcube.internal.message.HeatingThermostat;
 import org.openhab.binding.maxcube.internal.message.L_Message;
 import org.openhab.binding.maxcube.internal.message.M_Message;
 import org.openhab.binding.maxcube.internal.message.Message;
+import org.openhab.binding.maxcube.internal.message.MessageProcessor;
 import org.openhab.binding.maxcube.internal.message.MessageType;
 import org.openhab.binding.maxcube.internal.message.S_Command;
 import org.openhab.binding.maxcube.internal.message.S_Message;
@@ -119,6 +126,12 @@ public class MaxCubeBinding extends AbstractActiveBinding<MaxCubeBindingProvider
 	private BufferedReader reader = null;
 	private OutputStreamWriter writer = null;
 	
+	
+	/**
+	 * Processor that handles lines received from MAX!Cube
+	 */
+	MessageProcessor messageProcessor = new MessageProcessor();
+	
 	/**
 	 * {@inheritDoc}
 	 */
@@ -186,10 +199,15 @@ public class MaxCubeBinding extends AbstractActiveBinding<MaxCubeBindingProvider
 					cont = false;
 					continue;
 				}
-
-				Message message;
+				
+				Message message = null;
 				try {
-					message = processRawMessage(raw);
+					this.messageProcessor.addReceivedLine(raw);
+					if (this.messageProcessor.isMessageAvailable()) {
+						message = this.messageProcessor.pull();
+					} else {
+						continue;
+					}
 
 					message.debug(logger);
 
@@ -244,9 +262,28 @@ public class MaxCubeBinding extends AbstractActiveBinding<MaxCubeBindingProvider
 							cont = false;
 						}
 					}
+				} catch (IncorrectMultilineIndexException ex) {
+					logger.info("Incorrect MAX!Cube multiline message detected. Stopping processing and continue with next Line.");
+					this.messageProcessor.reset();
+				} catch (NoMessageAvailableException ex) {
+					logger.info("Could not process MAX!Cube message. Stopping processing and continue with next Line.");
+					this.messageProcessor.reset();
+				} catch (IncompleteMessageException ex) {
+					logger.info("Error while parsing MAX!Cube multiline message. Stopping processing, and continue with next Line.");
+					this.messageProcessor.reset();
+				} catch (UnprocessableMessageException ex) {
+					logger.info("Error while parsing MAX!Cube message. Stopping processing, and continue with next Line.");
+					this.messageProcessor.reset();
+				} catch (UnsupportedMessageTypeException ex) {
+					logger.info("Unsupported MAX!Cube message detected. Ignoring and continue with next Line.");
+					this.messageProcessor.reset();
+				} catch (MessageIsWaitingException ex) {
+					logger.info("There was and unhandled message waiting. Ignoring and continue with next Line.");
+					this.messageProcessor.reset();
 				} catch (Exception e) {
 					logger.info("Failed to process message received by MAX! protocol.");
 					logger.debug(Utils.getStackTrace(e));
+					this.messageProcessor.reset();
 				}
 			}
 			if(!exclusive) {
@@ -395,8 +432,22 @@ public class MaxCubeBinding extends AbstractActiveBinding<MaxCubeBindingProvider
 					writer.write(commandString);
 					logger.debug(commandString);
 					writer.flush();
+					
+					Message message = null;
 					String raw = reader.readLine();
-					Message message = processRawMessage(raw);
+                    try {
+                        while (!this.messageProcessor.isMessageAvailable()) {
+                        	this.messageProcessor.addReceivedLine(raw);
+                            raw = reader.readLine();
+                        }
+
+                        message = this.messageProcessor.pull();
+                    } catch (Exception e) {
+                        logger.info("Error while handling response from MAX! Cube lan gateway!");
+                        logger.debug(Utils.getStackTrace(e));
+                        this.messageProcessor.reset();
+                    }
+					
 					if (message !=null) {
 						if (message.getType() == MessageType.S) {
 							sMessageProcessing((S_Message)message);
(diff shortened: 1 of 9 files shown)


habnefrage:
I think, BuildHive failed due to some Java 1.7 features I was using in the src code.
I changed it back to 1.6 features. I hope its work now.

teichsta:
@marcelrv could you please have a look at this PR? Thanks, Thomas E.-E.

marcelrv:
@teichsta Hi Thomas, indeed I think we have not seen this many devices and don't handle it correctly. Hence this is a bug.

@habnefrage Hi Christian, great house you must have with 54 devices ;)
I would like to look bit deeper into your code. esp as we will need to integrate it with the OH2 as well, where we improved the binding and allow for multiple cubes to be configured. I'm bit concerned with the current way this is done it conflicts.

habnefrage:
Hi Marcel,
I habe checked out openhab2 repository yesterday and started integrating my
change there aswell.
If you give me two or three days i will push the bugfix in oh2.

Cheers
 Am 14.04.2015 4:17 vorm. schrieb "Marcel" <notifications@github.com>:

habnefrage:
Hi,
while integrating the fix into OH2 repository, I recognized a mistake. I fixed it and extended the JUnitTest to cover this issue.

teichsta:
Please keep the upcoming deadline in mind. Could you both take care to get this Fix done? Thanks a lot, Thomas E.-E.

habnefrage:
@marcelrv let me know if I can support you somhow.

marcelrv:
hi @habnefrage support...unless you are airplane mechanic and fixed my fuel leaking plane yesterday so I would not be stuck in the usa,,, not too much :(

In the meantime i setup the oh1 dev environment on my laptop so I can do some testing on this PR from here

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
```diff
@@ -186,10 +199,14 @@ public synchronized void execute() {
 					cont = false;
 					continue;
 				}
-
-				Message message;
+				
+				Message message = null;
 				try {
-					message = processRawMessage(raw);
```
marcelrv: I think this needs to be consistent. If we replace message = processRawMessage(raw); then replace it in the send message area (L435) as well

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageType.java
```diff
@@ -15,5 +15,19 @@
...
 */
 public enum MessageType {
-	H, M, C, L, S
+	H("H:"),
+	M("M:"),
+	C("C:"), 
+	L("L:"),
+	S("S:");
+	
+	private String messageIndicator;
```
marcelrv: this message indicatior is more or less duplicate of the message.getType() can you see if you can use that instead?

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Processes the raw TCP data read from the MAX protocol, returning the
+	 * corresponding Message.
+	 * 
+	 * @param line
+	 *            the raw data provided read from the MAX protocol
+	 * @return MessageType of the line added
+	 */
+	public static MessageType getMessageType(String line) {
```
marcelrv: this MessageType getMessageType is more or less duplicate of the message.getType() is this really needed?

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+			result = this.currentMessage;
+			reset();
+		}
+
+		return result;
+	}
+
+	/**
+	 * Processes the raw TCP data read from the MAX protocol, returning the
+	 * corresponding Message.
```
marcelrv: I think you mean here returning the message type

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
```diff
@@ -186,10 +199,14 @@ public synchronized void execute() {
...
+					if (this.messageProcessor.addReceivedLine(raw)) {
```
marcelrv: Is it not more elegant to add a messageProcessor.IsMessageAvailable() test instead?

marcelrv:
hi @habnefrage 
Would be great if you could take a look at the comments.
For the rest seems to be running fine

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
habnefrage: Yes, you'r right.
I added it aswell in this block.

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageType.java
habnefrage: I know what mean, but I do not see a suitable way to get rid of this code.
I neet the message type before I compile the actual message object. So I can not use message.getMessage at this point in time.

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
habnefrage: Same as before...
I do not have the message object available at this time. I need to lookup the message indicator from the raw message to handle it correctly and not mix up different messages by accident.

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
habnefrage: Hmmmmm not sure what you mean.
The pull() function returns the fully compoiled message.

Goal of the message processor is, to combine multiple received lines (belonging together) to one single message. Therefore you can put each received line from the MAXCube into the processor . When there is one message ready, you can pull it from the processor to handle it in the normal flow of the binding. 
If you put an other line to the messageProcessor, without pulling the waiting message upfront, you will get an exception thrown indicating, that there is a message waiting.

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
habnefrage: I was thinking about it, too.
In this way it was a bit easyer to implement AND it reduces the code, using the messageProcessor. 
It is a philosophical question...
If you really want me to introduce the messageProcessor.isMsgAvailable() method, I will implement it.

habnefrage: As it was quite fastforward, I implemented the method to check whether there is a message available. Now one can either way.

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
marcelrv: suggest then to make it something like: Note, I think it should be private then as well

```
	public static final String SEPARATOR = ":";

	private static MessageType getMessageType(String line) {

		for (MessageType msgType : MessageType.values()) {
			if (line.startsWith(msgType.name()+SEPARATOR)) {
				return msgType;
			}
		}

		return null;
	}
```

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageType.java
marcelrv: see below suggested solution that would avoid this

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
marcelrv: what I mean to say is that the description does not match what is actually returned
```
* Processes the raw TCP data read from the MAX protocol, returning the 
* corresponding Message 
```
but it is returning a MessageType instead of a Message

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
habnefrage: I do not see the advantage, but you are the boss :-)
Give me one or two hours until the sun goes down...
Am 19.04.2015 5:35 nachm. schrieb "Marcel" <notifications@github.com>:

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
habnefrage: OK, got it...
It is fixed now.

bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/message/MessageProcessor.java
habnefrage: OK, I changed the implementation as recommented by @marcelrv.

marcelrv:
@teichsta LGTM
Tested fine here.

teichsta:
Great :-)

If you could finally squash your commits into one @habnefrage i'll be happy to merge this PR, soon.

Thanks, Thomas E.-E.

habnefrage:
@teichsta, I am not very familiar with GIT. I hope you can use my last commits.

marcelrv:
Hi Christian,
You can squash your commit in eclipse by going to the git screen, then
select the last change before your own commits.
The select in the right click menu reset and select the middle option (not
the hard one). For the first time it I'd maybe best to backup your work
before you do it

This will undo the commits, but keep your chances. Than you can do a single
commit including all of your chances.

If you then push it to github your PR will be updated to the one New commit.

Hope this helps

teichsta:
Thanks @marcelrv for the explanation and @habnefrage for following it ;-)

Issue #2460, opened 2015-04-15
IDE Setup: "3 Generate Script Model.launch" and "4 Generate Script Model.launch"  fail


gernoteger:
When including the current 1.7.0-SNAPSHOT version into Eclipse (I used the yoxoxs profile), I get the Exception described below, no matter if I use a clean checkout or buidling with maven beforehand.

```
4711 [main] INFO  or.validation.JavaValidatorFragment  - generating Java-based EValidator API
8119 [main] INFO  text.generator.junit.Junit4Fragment  - generating Junit4 Test support classes
8177 [main] INFO  text.generator.junit.Junit4Fragment  - generating Compare Framework infrastructure
8286 [main] ERROR g.eclipse.xtext.generator.Generator  - java.io.FileNotFoundException: ..\org.openhab.model.script.tests\META-INF\MANIFEST.MF (Das System kann den angegebenen Pfad nicht finden)
org.eclipse.emf.common.util.WrappedException: java.io.FileNotFoundException: ..\org.openhab.model.script.tests\META-INF\MANIFEST.MF (Das System kann den angegebenen Pfad nicht finden)
	at org.eclipse.xtext.generator.Generator.mergeManifest(Generator.java:418)
	at org.eclipse.xtext.generator.Generator.generateManifestTests(Generator.java:485)
	at org.eclipse.xtext.generator.Generator.invokeInternal(Generator.java:144)
	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent.invoke(AbstractWorkflowComponent.java:126)
	at org.eclipse.emf.mwe.core.lib.Mwe2Bridge.invoke(Mwe2Bridge.java:34)
	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent.invoke(AbstractWorkflowComponent.java:201)
	at org.eclipse.emf.mwe2.runtime.workflow.AbstractCompositeWorkflowComponent.invoke(AbstractCompositeWorkflowComponent.java:35)
	at org.eclipse.emf.mwe2.runtime.workflow.Workflow.run(Workflow.java:19)
	at org.eclipse.emf.mwe2.launch.runtime.Mwe2Runner.run(Mwe2Runner.java:102)
	at org.eclipse.emf.mwe2.launch.runtime.Mwe2Runner.run(Mwe2Runner.java:62)
	at org.eclipse.emf.mwe2.launch.runtime.Mwe2Runner.run(Mwe2Runner.java:52)
[...]

PR #2461, opened 2015-04-15
fix for Issue #2460: can't run "3 Generate Script Model.launch" and "4 G...

gernoteger:
...enerate Rule Model.launch"


gernoteger:
I provided an alternate solution in #2465, which in my opinion is better.

teichsta:
we'll follow #2465 …

PR #2464, opened 2015-04-15
WeatherBinding: Use of API v2 and support of uvIndex and dewpoint

robbyb67:
... for provider WorldWeatherOnline

### DIFF
--- a/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/model/Temperature.java
+++ b/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/model/Temperature.java
@@ -63,7 +63,8 @@ public class Temperature {
 	@ProviderMappings({ 
 			@Provider(name = ProviderName.WUNDERGROUND, property = "current_observation.dewpoint_c"),
 			@Provider(name = ProviderName.FORECASTIO, property = "dewPoint"),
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/model/Temperature.java
```diff
@@ -63,7 +63,8 @@
 	@ProviderMappings({ 
 			@Provider(name = ProviderName.WUNDERGROUND, property = "current_observation.dewpoint_c"),
 			@Provider(name = ProviderName.FORECASTIO, property = "dewPoint"),
-			@Provider(name = ProviderName.HAMWEATHER, property = "dewpointC")
+			@Provider(name = ProviderName.HAMWEATHER, property = "dewpointC"),
+			@Provider(name = ProviderName.WORLDWEATHERONLINE, property = "DewPointC")
```
teichsta: shouldn't the property start with a lowercase "d"?

robbyb67: No - the property is used to literally map to a JSON key from the weather providers response:
... 
                       "cloudcover": "97",
                        "DewPointC": "5",
                        "DewPointF": "41",
                        "FeelsLikeC": "15",
 ...
So for WorldWeatherOnline it's ok. :-)

PR #2465, opened 2015-04-15
fix for Issue #2460: remove references to non-existent tests from Generators

gernoteger:
an alternative to PR #2461: either this one or the other makes sense. I prefer this one


teichsta:
Thanks @gernoteger! I would also follow this approach … could you please remove the commented code and squash into one commit finally? Best, Thomas E.-E.

gernoteger:
I messed up my fork's repo so I created a new PR with clean history

PR #2469, opened 2015-04-17
New features for Freebox binding:

lolodomo:
New features for Freebox binding:
* enable/disable FTP
* enable/disable Air Media
* enable/disable UPnP AV
* enable/disable Samba file/printer shares
* get reachable status for network devices
* optional ability to distinguish accepted/missed/outgoing phone calls


bundles/binding/org.openhab.binding.freebox/src/main/java/org/openhab/binding/freebox/internal/FreeboxBinding.java
```diff
@@ -222,42 +251,162 @@ protected void execute() {
 	@SuppressWarnings("incomplete-switch")
 	@Override
 	protected void internalReceiveCommand(String itemName, Command command) {
+		if ((loginManager == null) || (!loginManager.isConnected())) {
+			try {
+				authorize();
+			} catch (FreeboxException e) {
+				logger.error(e.getMessage());
```
teichsta: please decrease severity to info since this exception does not effect the stability of openHAB itself

bundles/binding/org.openhab.binding.freebox/src/main/java/org/openhab/binding/freebox/internal/FreeboxBinding.java
```diff
@@ -222,42 +251,162 @@ protected void execute() {
...
+				setProperlyConfigured(false);
+				return;
+			}
+		}
```
teichsta: extract whole block into a method connect() (or alike) since this code is (at least) doubled

bundles/binding/org.openhab.binding.freebox/src/main/java/org/openhab/binding/freebox/internal/FreeboxBinding.java
```diff
@@ -222,42 +251,162 @@ protected void execute() {
 	@SuppressWarnings("incomplete-switch")
 	@Override
 	protected void internalReceiveCommand(String itemName, Command command) {
+		if ((loginManager == null) || (!loginManager.isConnected())) {
+			try {
+				authorize();
```
teichsta: since the authorisation is a long running process i would avoid calling it in receiveCommand. Better just check whether the Binding is authorised and in case it isn't return fast and log a reasonable statement explaining the problem.

bundles/binding/org.openhab.binding.freebox/src/main/java/org/openhab/binding/freebox/internal/FreeboxBinding.java
```diff
@@ -222,42 +251,162 @@ protected void execute() {
...
+						logger.error(e.toString());
+					}
+					// Get the current state
+					try {
+						SambaConfig sc = fbClient.getNetShareManager().getSambaConfig();
+						value = sc.getPrintShareEnabled();
+					} catch (Exception e) {
+						logger.error(e.toString());
+					}
+					break;
```
teichsta: decrease severity all log.error statements to INFO level

teichsta:
Hi @lolodomo, thanks a lot for this contribution. Please find my review comments inline. In general two main things should be changed:

* remove authorisation from internalReceiveCommand() and execute() and rather properly log the problem
* decrease the severity of log.error to something like INFO because the given errors doesn't really affect the stability of the platform but "only" of this particular binding

Best, Thomas E.-E.

lolodomo:
Review comments taken into account. Fix just pushed.

If I update my configuration file openHAB.cfg while openHAB is running, I was expecting that method "updated" to be called with my new configuration. That's apparently not the case. Is it normal ?

teichsta:
> That's apparently not the case. Is it normal ?

yes, because the refresh of the openhab.cfg and openhab_default.cfg is not implemented in OH1 (functionality is available in OH2 though).

May i ask you to squash your commits into one, please?

Thanks, Thomas E.-E.

lolodomo:
I tried to squash my 2 previous commits.
I hope that is ok now (I am not yet very familiar with Git).

lolodomo

teichsta:
looks good now!

Mickaelh51:
Hi all,
I have an issue with new Freebox binding (v1.8).
I can see in log: 
"2015-06-07 19:45:08.347 [INFO ] [.o.b.f.internal.FreeboxBinding] - Ce réglage n'est disponible qu'en mode routeur"

Ok it's clear, my freebox is a bridge mode. But before with version 1.6, this binding worked.

With this two versions I have two commands in Items conf:
1.6: OK / 1.8: NOK

String Freebox_Xdsl_Status      "Status XDSL [%s]"                   (G_All_House) {freebox="xdsl_status"}
String Freebox_Cs_State         "Line State [%s]"                    (G_All_House) {freebox="line_status"}

Have you an idea for me ?

thanks in advance

lolodomo:
Hi Mickael.

Maybe I have an idea. It is possible that one of my additional calls fails because your Freebox is in bridge mode and then all the process fails. We have to do a change to be sure that even if a particular feature fails to run, we run the others.
On a more technical side, my idea is that the try/catch surrounding everything in the method execute is too much global.

Mickaelh51:
Hi lolo ;)
thanks for your answer !

I thought  the same thing, an early break. I'm sorry I can't help you in java source code. PERL, C, Bash, etc ... OK , but no Java ;)

Do I check anything other (to try help you)?

clinique:
Yes, I guess that one of the information pull fails because the Freebox server is in bridge mode. Now the thing is witch one ? I saw that current mode is furnished by a call to /api/v3/lan/config/

lolodomo:
There are probably several things to improve. One would be to improve the exception message to know in the logs which call fails. The second would be to replace the unique try/catch in execute method. The third would be to detect the bridge mode and in this case to avoid calling API not available in this mode.
Unfortunately we don't know yet which ones are not available in this mode and we will have to identify them first. My idea is that probably we cannot get the list of network devices when in bridge mode.
I will probably not be able to provide a proper fix very soon, I am sorry Michael.

Mickaelh51:
No problem lolodomo ;)
I'm going to open an "official" issue in github. This binding is "franco-Français", But I expect someone will can help me.
I will try to check myself and help you to find this bug.

bye

lolodomo:
Hi Mickael.

Please go to the openHAB Google group to follow the discussion. I provided a new jar file for you to help understand what's wrong in your case.
https://groups.google.com/forum/?hl=fr#!category-topic/openhab/mJSFoTPZyz8

PR #2472, opened 2015-04-17
Fix: MySQL-binding doesn't reconnect on error because of incorrect default value of errReconnectThreshold 

Dennis650:
If the mySQL connection gets closed because of an error, the binding never reconnects, because the default value of errReconnectThreshold is 0. The default value should be 1, so every error causes a reconnect.


cdjackson:
If you change the default, then it will change everyones configuration. Why not just change your configuration using the config file?
When I added this a couple of years ago, the default was set to not reconnect as some people didn't want it. Personally I don't mind as I use reconnect anyway, but it's not correct that it's an 'incorrect' default - it's just a different default.

teichsta:
i was about to finish the review and write something similar. So i'd tend to close this PR as the proper configuration should be adjusted individually. Nevertheless thanks for this contribution. Best, Thomas E.-E.

Dennis650:
Hi, i changed it and made this pull request, because i think it should reconnect by default. Leaving the connection closed doesn't make sense. Yesterday i updated the mysql version on my server and had to restart it. I didn't recognize, that openhab didn't persist values after the restart of the mysql server. So i think the mysql binding should do a reconnect by default, like the other bindings i am using. If some people dont't want i reconnect (why?), they can change it in the configuration, but that should not be the default.

watou:
FWIW, I agree with Dennis that defaulting to reconnect would seem more sensible, but it's too late to change default behaviour.  I think a prominent update to the wiki page will be as close to a fix that can be done now.  Whatever equivalent exists in OH2 should default to reconnect.

cdjackson:
I don’t disagree - I was pointing out the fact that this isn’t ‘incorrect’ - it’s just a different default. At the time this was added, people didn’t want this function added (I forget why now - too long ago) so I defaulted it to off.

teichsta:
ok, so could we agree to not change the default for OH1 (since the behaviour is well known) already but add some proper notes to the Wiki (and openhab_default.cfg if not already there)? @Dennis650 would you be so kind and take care of this?

PR #2475, opened 2015-04-17
Updates to Mqttitude binding in preparation for new iOS/Android apps

sumnerboy12:
There are some new versions of the Owntracks app in the pipeline which use a slightly different topic for publishing. Along with sending lat/lons as floats (real types) rather than strings as they currently do.

This PR ensures the Mqttitude binding is backward compatible with the current versions of the app but will also work with the new version.

Only change required to binding configuration for the new versions of the Owntracks app, is to add `/event` to the binding topic.

### DIFF
--- a/bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeConsumer.java
+++ b/bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeConsumer.java
@@ -113,9 +113,11 @@ public class MqttitudeConsumer implements MqttMessageConsumer {
 		if (jsonPayload == null)
 			return;
 
-		// only interested in 'location' publishes
+		// only interested in 'location' or 'transition' publishes
 		String type = jsonPayload.get("_type");
-		if (StringUtils.isEmpty(type) || !type.equals("location"))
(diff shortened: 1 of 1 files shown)


sumnerboy12:
I have also updated the Mqttitude WIKI page to reflect these changes.

jpmens:
Thank you very much for doing this, Ben, and I hope it slips into the upcoming openHAB release!

teichsta:
>  I hope it slips into the upcoming openHAB release!

it will :-)

sumnerboy12:
Thanks Thomas!

jpmens:
@teichsta much appreciated, Thomas. Danke! :-)

sumnerboy12:
This will be announced by OwnTracks - I will post something in here when 
that happens tho. The new version should work as long as you are using 
region mode, just ensure your item binding topic is the base topic - 
i.e. no /event.

sumnerboy12:
What is your configuration and OwnTracks version number?

In order to debug subscribe to exactly the same topic as you have 
configured in your Mqttitude openHAB item binding and check what you see 
when your phone moves in/out of that waypoint.

Remember the waypoint/region has to be `shared` in your Owntracks app 
otherwise the `name` is not sent with the event payloads.

sumnerboy12:
Wondered why I couldn't see the question on Github! I have been having 
internet connection problems so apologies for the delay in responding.

Glad it is working for you now.

PR #2482, opened 2015-04-19
Small fix to avoid NullPointerException

marcelrv:
as .equals(null) is not a correct check
Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

### DIFF
--- a/bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
+++ b/bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
@@ -191,9 +191,8 @@ public class MaxCubeBinding extends AbstractActiveBinding<MaxCubeBindingProvider
 				try {
 					message = processRawMessage(raw);
 
-					message.debug(logger);
-
 					if (message != null) {
+						message.debug(logger);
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.maxcube/src/main/java/org/openhab/binding/maxcube/internal/MaxCubeBinding.java
```diff
@@ -334,7 +334,7 @@ public void internalReceiveCommand(String itemName, Command command) {
 		for (MaxCubeBindingProvider provider : providers) {
 			serialNumber = provider.getSerialNumber(itemName);
 
-			if (serialNumber.equals(null)) {
+			if (serialNumber==null || serialNumber.isEmpty()) {
```
teichsta: please use StringUtils.isBlank() instead

teichsta:
Hi @marcelrv thanks for this Fix. Please find my comment and finally please squash your commits into one. Thanks, Thomas E.-E.

marcelrv:
Hi @teichsta ,
Got it.

teichsta:
> Got it.

not quite :-) isBlank() already contains the "null check" …

marcelrv:
ah.. did not know that :-(
Okay.. hope that _now_ I got it :-)

teichsta:
yep, thanks!

PR #2487, opened 2015-04-20
Updated OpenSprinkler plugin to support new API versions

CrackerStealth:
Updated the OpenSprinkler library to support OpenSprinkler API versions 2.1.0 and up.


CrackerStealth:
Put out a call for more testers: https://groups.google.com/forum/#!category-topic/openhab/discussions/opensprinkler/c5N5ajs105I

teichsta:
since only the lib has been updated (and no binding code seem to be affected) why do we need this updated lib? Could you give any hints?

teichsta:
also please squash your commits into one finally …

Thanks, Thomas E.-E.

CrackerStealth:
Squashed commits. You can view the changes that were needed for the library here: https://bitbucket.org/JonathanGiles/opensprinkler/pull-request/2/added-support-for-opensprinkler-firmware/diff#comment-None and https://bitbucket.org/JonathanGiles/opensprinkler/pull-request/3/md5-password-hashing-is-firmware-213-and/diff#comment-6395335

The short story is that the OpenSprinkler developers removed manual mode, removed certain API endpoints and force the password to be used on every API call starting with Firmware version 2.1.0. Starting with Firmware 2.1.3, password needs to passed as an MD5 hashed string instead of clear-text. This completely breaks the library which the OpenSprinkler binding is heavily dependant on to work with the OpenSprinkler API.

Issue #1934, opened 2015-01-05
Mysql persistence don't log anything and no errors


Dannikorsholm:
I got openhab 1.5.1 installed on win 8 and all works great.

Now i got a raspberry pi with openhab 1.6 and can't get mysql to log anything.
I tried a lot of different setup in the persist file. The setup that worked on 1.5.1 don't work.

When i launch openhab start.sh the only thing about mysql i get is:  

2015-01-05 20:21:40.203 [INFO ] [c.internal.ModelRepositoryImpl] - Loading model 'mysql.persist'

When i launch with debug i get this:

21:23:23.029 [INFO ] [c.internal.ModelRepositoryImpl:79   ] - Loading model 'mysql.persist'
21:24:04.321 [DEBUG] [sqlPersistenceServiceActivator:32   ] - mySQL persistence bundle has been started.
21:24:04.861 [DEBUG] [.p.internal.PersistenceManager:146  ] - Initializing mysql persistence service.
21:24:07.304 [DEBUG] [.p.m.i.MysqlPersistenceService:434  ] - mySQL configuration starting
21:24:07.317 [DEBUG] [.p.m.i.MysqlPersistenceService:359  ] - mySQL: Attempting to connect to database jdbc:mysql://127.0.0.1/Openhab

21:24:17.683 [DEBUG] [.p.m.i.MysqlPersistenceService:362  ] - mySQL: Connected to database jdbc:mysql://127.0.0.1/Openhab
21:24:17.800 [DEBUG] [.p.m.i.MysqlPersistenceService:369  ] - mySQL: Setting wait_timeout to 65 seconds.
21:24:17.867 [DEBUG] [.p.m.i.MysqlPersistenceService:495  ] - mySQL configuration complete.

I created a database for openhab and when i launch openhab it create Items/ItemId and ItemName but empty and don't logging anything..

Can someone help ???

Regards Danni.
##

teichsta:
> I created a database for openhab and when i launch openhab it create Items/ItemId and ItemName but empty and don't logging anything..

you  mean the MySQL PersistenceService didn't log anything?

Can you please provide items and persist configuration?

Thanks, Thomas E.-E.

Dannikorsholm:
Yes no log and no errors from the mysql persistence.. Only creates the two data tables..


Items:

Group All
Group Hus	(All)
Group Temp	(All)
Group Lys	(All)

/*Temperature*/

[...]

PR #2503, opened 2015-04-23
MySQL-Persistence: Add errCnt++ to getTable and check connection

Dennis650:
changes discussed in Issue #1934. Adding errCnt to getTable Method
and check connection with JDBC isValid method, before a broken
connection is used to store values, which will result in data loss.


teichsta:
Thanks @Dennis650!

PR #2432, opened 2015-04-10
InsteonPLM binding: introduce "related" devices, feature groups, and bug fixes

berndpfrommer:
- The concept of "related" devices is introduced. When a device changes state, the "related" device is polled. This improves consistency for 3-way switches.
- Feature groups are added to allow multiple features to be queried with one poll message. This cuts down on message load on the Insteon network, and reduces the chances of misinterpreted messages.
- Various potential race conditions now in synchronized blocks.
- Fixes for iMeter calculations.

PR #2513, opened 2015-04-27
Various fixes and improvements to InsteonPLM

berndpfrommer:
- use ConcurrentHashMap to avoid synchronized() sections in InsteonPLMActiveBinding class

    - fixes for iMeter calculations

    - the concept of "related" devices is introduced. When a device
      changes its state, all related devices are polled as well. This is
      useful for 3-way switch configurations

    - a DeviceFeature can now act as a group of other
      DeviceFeatures. These feature groups allow for consolidated polling
      when one message can query multiple features' values at the same
      time.

    - added protecting synchronized statements around m_devices in
      InsteonPLM binding

    - added explicit expiration time to request queue entries to control
      message delay better


berndpfrommer:
This was originally submitted as pull request #2432.

I couldn't figure out how to get this done without deleting the branch, so I did that, which closed #2432.

Please approve.

Issue #2504, opened 2015-04-24
Philips Hue Binding crashes on refresh


mstolt:
Since a recent update of the hue bridge Philips enabled deleting items on the bridge.
This results in non continuous item numbers in the datastore lights section.

The current implementation requires a continuous numbering of the lights. 
The error is the loop in HueBinding.java method "execute" (line 101 and following)

The current code throws an NPE if a light was deleted and e.g. only the lights 1,2 and 4 remain in the bridge.

```
18:41:45.850 [ERROR] [.service.AbstractActiveService:175  ] - Error while executing background thread Hue Refresh Service
java.lang.NullPointerException: null
        at org.openhab.binding.hue.internal.data.HueSettings$SettingsTree.node(HueSettings.java:216) ~[na:na]
        at org.openhab.binding.hue.internal.data.HueSettings.isBulbOn(HueSettings.java:71) ~[na:na]
        at org.openhab.binding.hue.internal.hardware.HueBulb.getStatus(HueBulb.java:74) ~[na:na]
        at org.openhab.binding.hue.internal.hardware.HueBulb.<init>(HueBulb.java:62) ~[na:na]
        at org.openhab.binding.hue.internal.HueBinding.execute(HueBinding.java:104) ~[na:na]
        at org.openhab.core.binding.AbstractActiveBinding$BindingActiveService.execute(AbstractActiveBinding.java:156) ~[na:na]
        at org.openhab.core.service.AbstractActiveService$RefreshThread.run(AbstractActiveService.java:173) ~[na:na]
```

I currently try to setup a development environment and give it a try to fix the issue. Since a am a novice to OpenHAB, Java and GIT it may take a while to fix it.

Greetings
  Matthias

mstolt:
So far i programmed a solution. Now i have to deal with git and create a "push/pull"-Request :)

mstolt:
This issue can be closed.

Issue #2505, opened 2015-04-25
The Philips Hue Bindings takes several minutes to refresh the bulbs


mstolt:
In bigger installations of Philips Hue the refresh cycle take several minutes. 

I trace the problem down to the "multiple" fetch of the HueSettings from the bridge.
Each time a HueBulb is created the bridge gets ask for the settings of all bulbs.
Each of these fetches generates an HTTP Request that takes its time.

I am working on a patch to "reuse" the already fetched settings in the refresh cycle.
So in a 40 lamp installation we will remove 40 calls to the bridge.

mstolt:
It does work with the version on my machine. I now have to find out how to create a correct pull request in GitHub.

mstolt:
This issue can be closed.

PR #2514, opened 2015-04-27
Update the HUE binding to work with non continuous device numbers on the...

mstolt:
... bridge (and speed up the initial loading)

Due the new option to delete bulb on the Hue bridge the old assumption
of continuous device numbers is no longer true. The for loop in
HueBinding.java crashed with a NPE if it hits a deleted bulb.

The second fix in the update is a speed up of updates/initial reads by
removing multiple call to the settings on the bridge. The code now
fetches the settings once per refresh from the bridge and hands the
settings over the the bulb construction.

This fixes the 
issue #2504 and 
issue #2505

### DIFF
--- a/bundles/binding/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/data/HueSettings.java
+++ b/bundles/binding/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/data/HueSettings.java
@@ -9,6 +9,7 @@
 package org.openhab.binding.hue.internal.data;
 
 import java.util.Map;
+import java.util.Set;
 
 import org.codehaus.jackson.map.ObjectMapper;
 import org.slf4j.Logger;
@@ -56,6 +57,19 @@ public class HueSettings {
 	}
 
 	/**
+	 * Return the keys of lights connected to Hue hub
+	 * 
+	 * @return the keys of lights connected to Hue hub
+	 */
+	public Set<String> getKeys() {
+		if (settingsData == null) {
+			logger.error("Hue bridge settings not initialized correctly.");
+			return null;
+		}
+		return settingsData.node("lights").getKeys();
+	}
+	
+	/**
 	 * Determines whether the given bulb is turned on.
 	 * 
 	 * @param deviceNumber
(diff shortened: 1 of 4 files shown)


bundles/binding/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/data/HueSettings.java
```diff
@@ -223,6 +223,10 @@ protected int count() {
 			return dataMap.size();
 		}
 
+		protected Set<String> getKeys(){
+			return dataMap.keySet();
+		}
```
teichsta: are there any other Key types then Integer? If not i would suggest to transform the Strings into Integer once (probably here?)

teichsta:
Hi @mstolt, thanks for this contribution! Please find my review comment inline. Also it would be great if you could have a look at the ESH version of the Hue Binding. It was rewritten from scratch but probably that piece of code has been copied? Could be worth having a look there. Best, Thomas E.-E.

bundles/binding/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/data/HueSettings.java
mstolt: Actually the Philips API is not clear about a type for the ID. Since the usage of the ID in the rest API is always in a string context (and the integer ID is alway converted into string), my next suggestion for a refactoring would have been a complete usage of String for the ID type.
In the ESH the lights ID is also a String.

mstolt:
Hi Thomas, I only looked at the ESH code. It seems to be fine. The light IDs are Strings and the init loop/update loop in the method "run" of HueBridgeHandler (line 91) works with an iterator over the lights key set.

teichsta:
thanks, @mstolt!

mstolt:
I have to thank you (and the other contributors) for this project.

Btw. do you close the associated issues #2504 and 
issue #2505 ? I already commented them as "to be closed"

teichsta:
since you are the creator of both issues you can close them yourself. Thanks, Thomas E.-E.

PR #2521, opened 2015-04-30
Fix so Squeezebox actions automatically reconnects

sumnerboy12:
i.e. if connection to server is lost


teichsta:
Thanks @sumnerboy12 for this contribution but i somehow missed the "reconnect" part of the code :-) Can you point me into the right direction? Best, Thomas E.-E.

sumnerboy12:
Previously the Squeezebox action bundle would stop if the SqueezeServer was connected, before sending the command. Now it will send the command regardless and let the server handle the connection checks and reconnect if possible. This is how the Squeezebox binding works already. 

So basically leaving the connection state checking (and reconnect attempts) to the SqueezeServer IO bundle, rather than in the action bundle. Hope that makes sense!

teichsta:
it does, thanks Ben!

PR #2534, opened 2015-05-02
Improve log messages when attempting a meter reset

sumnerboy12:
Very minor change, just improves the debug messages if a METER_RESET request is made on a device which doesn't support it.


cdjackson:
Thanks Ben.

Issue #2535, opened 2015-05-03
Colorpicker does not work anymore in 1.7 (coming from 1.6.2)


spacemanspiff2007:
Hi,

I installed the snapshot from cloudbees from today (03.05.2015).
The small color-configuration does not work anymore.
When I start an 1.6.2 - Installation everything works fine.

Here are my items:
```
Group	gAll
Group	gUp				(gAll)
Color   Wohnzimmer_Vorne_Oben		"Wohn V O"	(gUp)		{dmx="CHANNEL[17/3], ON[FADE|5000:255,255,255:0], OFF[FADE|3000:0,0,0:0]"}
Color   Wohnzimmer_Rechts_Oben		"Wohn R O"	(gUp)		{dmx="CHANNEL[9/3], ON[FADE|5000:255,255,255:0], OFF[FADE|3000:0,0,0:0]"}
Color   Wohnzimmer_Hinten_Oben		"Wohn H O"	(gUp)		{dmx="CHANNEL[1/3], ON[FADE|5000:255,255,255:0], OFF[FADE|3000:0,0,0:0]"}
```

With the following sitemap:
```
sitemap default label="Main Menu"
{
	Frame label="Licht" {
		Colorpicker item=gAll 	label="Alle"		
		Colorpicker item=gUp	label="Oben"		
	}	
}
```

When I start 1.6.2 everything works fine when I select a color with the colorpicker:
```
2015-05-03 13:49:23 - gUp received command 27.87878787878788,38.8235294117647,100
2015-05-03 13:49:23 - Wohnzimmer_Vorne_Unten received command 27.87878787878788,38.8235294117647,100
2015-05-03 13:49:23 - Wohnzimmer_Rechts_Unten received command 27.87878787878788,38.8235294117647,100
2015-05-03 13:49:23 - Wohnzimmer_Hinten_Unten received command 27.87878787878788,38.8235294117647,100
2015-05-03 13:49:23 - g_Licht_Wohnzimmer_Unten received command 27.87878787878788,38.8235294117647,100
```

Starting the Openhab 1.7.0 snapshot I receive the following error:
```
2015-05-03 13:52:24.511 [WARN ] [.w.internal.servlet.CmdServlet] - Received command '27.768595041322314,47.450980392156865,100' for item 'null', but the item does not exist in the registry
```

The configuration is identical.
What can I provide to help you locate this issue.
Greetings

theoweiss:
Other users reported the same issue on the list:
[...]

PR #2545, opened 2015-05-04
Fix broken color-picker in Classic UI

J-N-K:
Fix issue #2535 (colorItem not properly set after changing of namespace). Formerly colorItem was set from within the snippet, after the change of the namespace the colorpicker-snippert was not updated.


teichsta:
thanks @J-N-K for this quick fix!

PR #2548, opened 2015-05-05
code cleanup, remove skeleton Action from Prowl that does nothing

mstolt:
During looking around who to implement Actions i stumbled across a method remaining from the skeleton build. I think it can be deleted.

@teichsta Is it ok for you to do pull request for such small cleanups or do you prefer to have a bigger pull request with a collection of them? I am just asking, to fit better into your way of development.


mstolt:
I just check the other action addons for exceed skeleton methods and found none.

teichsta:
> @teichsta Is it ok for you to do pull request for such small cleanups or do you prefer to have a bigger pull request with a collection of them? I am just asking, to fit better into your way of development.

i am absolutely fine with this granularity! Thanks @mstolt!

PR #2549, opened 2015-05-05
New Binding for using Onewire Bus on CUNO Devices

querdenker2k:


querdenker2k:
Pull Request closed because, now there is another binding which supports this functionality (not all but i will add missing features there)

PR #2550, opened 2015-05-06
OpenSprinkler Bugfix: Addon Silently Fails to Load When numberOfSprinklers is Defined

CrackerStealth:
- Fixed bug where defining the "numberOfSprinklers" parameter makes the add-on silently fail to load.
- Added number of stations parameter validation.


teichsta:
Thanks a lot @CrackerStealth

PR #2566, opened 2015-05-09
make RFXComMessageFactory message creation type-safe at compile time

juri8:
This makes not-implemented messages much more visible in the code, the existing solution did not provide this.

### DIFF
--- a/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComMessageFactory.java
+++ b/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComMessageFactory.java
@@ -8,11 +8,6 @@
  */
 package org.openhab.binding.rfxcom.internal.messages;
 
-import java.lang.reflect.Constructor;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-
 import org.openhab.binding.rfxcom.internal.RFXComException;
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComMessageFactory.java
```diff
@@ -93,36 +38,12 @@
...
-			return (RFXComMessageInterface) c.newInstance(packet);
-		
-		} catch (ClassNotFoundException e) {
-			throw new RFXComException("Message " + packetType + " not implemented", e);
-			
-		} catch (Exception e) {
-			throw new RFXComException(e);
-		} 
+		final PacketType packetType = getPacketType(packet[1]);
+		return getMessageInterface(packetType);
```
paulianttila: Hi @juri8 

This doesn't work similarly than previous code. Now it return object with default values not object based on data (byte[] packet).

paulianttila:
@juri8,

Before I refactored rfxcom binding, I got some feedback that new message support development should be easier. This was the reason why I used reflection. Idea was just to add new class for unsupported message and no need to touch anything else. Now developers need to touch base message class as well.

-Pali

juri8:
@paulianttila 
Thanks for finding the issue in my pull request. I was about to fix this but I think it makes the PacketType enum hard to read.
I personally prefer code that can be checked at compile time to runtime checks which was the reason for this change. Additionally, I think that if a new message gets implemented it's not the big issue to register it (although I get your point that it's a bit easier without this need)

PR #2571, opened 2015-05-10
Milight Binding: Added Nightmode support for the newer RGBW bulbs, Fixed brightness setting for White.

DJCool1:
Extended night mode support to the newer RGBW bulbs.

The white Milight bulbs have 10 levels of brightness + Off so there are 11 possible settings between 0 and 100%. It was at 10 settings and so a disparity between desired brightness setting vs. actual brightness would accrue during use. Also, to engage the desired bulb, an on signal needed to be sent before the brightness setting. Lastly, white bulbs are powered to lowest setting before off, and when powering on this is taken into account. All these fixes combined finally allows the brightness to be very accurate and reliable for the white bulbs.


DJCool1:
Latest commit: rolls back Jan 21, 2015 change that broke HSB brightness. The 3rd part of HSB color is brightness and it's especially important for the Milight binding because these bulbs store their color brightness separately from their whitemode brightness. So color binding does need to have its own way to set and remember its brightness.

teichsta:
@DJCool1 thanks for this contribution which looks good to me. Could you finally please squash your commits into one? Thanks, Thomas E.-E.

DJCool1:
Done.

PR #2577, opened 2015-05-11
zwave: Invalid CommandClass type causes ClassCastException at startup

johkin:
After adding a remote and configuring SceneActivation in the ZWave binding, OH started to log errors at startup.
It seems that the ZWaveSceneConverter is using the ZWaveBasicCommandClass as a generic type, but since the ZWaveSceneActivationCommandClass doesn't extend ZWaveBasicCommandClass this would cause an exception.

Maybe this change should be handled in another way, by letting the ZWaveSceneActivationCommandClass extend ZWaveBasicCommandClass?


cdjackson:
I'll check this in the next day or so, but it looks ok. I think the issue is that when I originally wrote this I was only looking at the receive side, not the command or refresh, so this wasn't tested :(

Issue #2552, opened 2015-05-06
Rule trigger 'changed from to' does not work for string items


gregeva:
Affected use-case/example:

    rule Alarm_Away
    when
            Item cntAlarm changed from "Disarmed" to "Armed-Away"
    then
            pushover("Alarm system went to away.")
    end

It is unclear if this is normal operation or not, as the documentation references change from and to <STATE>, and I am not sure if a string value is considered a state or not.

The problem has been reported in the group, and the user is using a 1.7 SNAPSHOT build.
https://groups.google.com/forum/#!topic/openhab/nWlt0FT8Vqk

J-N-K:
There is a problem with "changed from ... to ..." rules with String items

Item test changed from state1 to state2

works perfectly fine, if state1 and state2 are strings which only contain alpanumeric characters without any spaces.

It will not work with strings containing a space or hyphen (the string is truncated, i.e. instead of Test-String only Test is stored as the state). If the string is included in quotes, it is not truncated, however the quotes are stored, too. because of that the comparison of test.state with state1 fails.

This is at least not the expected behaviour, if not a bug. I would expect that using quotes is properly supported (i.e. they are either stripped before the compare-state is stored or ignored during the comparison).

The latter could easily be implemented in RuleTriggerManager, but to me the first seems to be the better solution.

(copied from issue #2559 for easier reference)

steve-bate:
Since the PR is merged, can we close this issue?

gregeva:
I would say that you can, as I think this is not only merged, but fixed.




*Greg Eva*
*geva@ge-volution.eu <geva@ge-volution.eu>*
www.gE-volution.eu
France : +33 4 81 68 09 64
Canada : +1 226 476 1192
iNum : +833 51 00 09 90 28 81

PR #2578, opened 2015-05-11
quick-fix for issue #2552, as suggest as option 2

J-N-K:
This enables openHAB to behave as expected for changed-from-to-triggers in rules with string items. If present, double-quotes at the end and beginning of the states are removed.


teichsta:
rescheduled to 1.7.1 - didn't find the time to research in depth. In general Xtext should handle Strings properly (even with '"'). So i'd suggest to better find the root cause for this issue rather than workarounding it (since i don't reviewed possible side effects, too)

teichsta:
seem to be related to #2880 … at least the same symptom both of you fought against the same issue :-)

marcelerkel:
Yes, same symptom. In the next couple of days I'll try to implement the same remedy :)

teichsta:
superseded/replaced by #2933

PR #2590, opened 2015-05-12
InsteonPLM binding: support the standard pre-2014 InsteonHub

berndpfrommer:
The InsteonPLM binding now also supports the "old" InsteonHub via its raw tcp port. (The 2014 Hub is already supported, but only through its http interface). The following line in openhab.cfg will connect to the raw tcp feed of the insteonhub:

insteonplm:port_0=/hub/localhost:9761

### DIFF
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/IOStream.java
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/IOStream.java
@@ -8,14 +8,18 @@
  */
 package org.openhab.binding.insteonplm.internal.driver;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 
 import org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream;
+import org.openhab.binding.insteonplm.internal.driver.hub.OldHubIOStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Abstract class for implementation for I/O stream with anything that looks
- * like a PLM (e.g. the insteon hub)
+ * like a PLM (e.g. the insteon hubs, serial/usb connection etc)
  * 
  * @author Bernd Pfrommer
  * @author Daniel Pfrommer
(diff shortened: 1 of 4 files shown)


bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/hub/OldHubIOStream.java
```diff
@@ -0,0 +1,71 @@
+/**
+ * Copyright (c) 2010-2013, openHAB.org and others.
```
teichsta: please update the license header

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/IOStream.java
```diff
@@ -23,18 +27,50 @@
...
+	public int read(byte [] b, int offset, int readSize) {
+		int len = 0;
+		while (len < 1) {
+			try {
+				len = m_in.read(b, offset, readSize);
+			} catch (IOException e) {
+				while (!reconnect()) {
+					try {
+						Thread.sleep(10000);
+					} catch (InterruptedException ie) {
```
teichsta: please add logging in TRACE level just to don't miss anything

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/IOStream.java
```diff
@@ -23,18 +27,50 @@
...
 	 */
-	public OutputStream out() { return m_out; }
+	public void write(byte [] b) {
+		try {
+			m_out.write(b);
+		} catch (IOException e) {
+			while (!reconnect()) {
+				try {
+					Thread.sleep(10000);
+				} catch (InterruptedException ie) {
```
teichsta: please add logging in TRACE level just to don't miss anything

teichsta:
Thanks @berndpfrommer for this contribution. Regarding log levels: until i didn't consequently follow the same root in all reviews but i think we should start thinking about the log levels. From my point of view we should ERROR only when the stability of the platform is affected. What do you think? Best, Thomas E.-E.

berndpfrommer:
Done all the changes you asked.

You will now ask me to squash into a single commit, I know you :). But I still don't know how to do that cleanly. I can squash the commits in my local workspace, no problem, but when I try to push the squashed commits to github it asks me to pull first, at which point I have to merge, which gives me another commit. Plus it seemed like the un-squashed commits were still on github. Last time I ended up removing the branch and pushing it again, but github didn't like that, and killed the entire pull request. How do other contributors do that?

About logging levels:

Off the top of my head I wouldn't know a single situation where one of the two bindings I've written could bring down the entire platform. At first glance it appears you are suggesting eliminating ERROR logging from most binding code. My guess is that bindings constitute most of openhab code base, so effectively that would compress the range of logging for most of openhab to TRACE, DEBUG,INFO,WARN.

I can't claim consistency, but usually use ERROR when the incident predictably or very likely impacts functionality, i.e. a configuration mistake, a tcp connection going down etc. I use WARN when something happens that is fishy and usually a bad sign, but may not impact operation going forward (e.g. a device did not respond to a single message (could be dropped on the network)).

teichsta:
> so effectively that would compress the range of logging for most of openhab to TRACE, DEBUG,INFO,WARN.

true and if you say like that it somehow doesn't sound like the best idea. So let's follow your suggestion for now :-)

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/IOStream.java
```diff
@@ -23,18 +27,54 @@
...
+		while (len < 1) {
+			try {
+				len = m_in.read(b, offset, readSize);
+			} catch (IOException e) {
+				logger.trace("got exception while reading: {}", e.getMessage());
+				while (!reconnect()) {
+					try {
+						logger.trace("sleeping before reconnecting");
+						Thread.sleep(10000);
+					} catch (InterruptedException ie) {
```
teichsta: the idea was to have a TRACE level statement in case of an error … so one can get an idea where the problem lies.

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/IOStream.java
```diff
@@ -23,18 +27,54 @@
...
+	public void write(byte [] b) {
+		try {
+			m_out.write(b);
+		} catch (IOException e) {
+			logger.trace("got exception while writing: {}", e.getMessage());
+			while (!reconnect()) {
+				try {
+					logger.trace("sleeping before reconnecting");
+					Thread.sleep(10000);
+				} catch (InterruptedException ie) {
```
teichsta: the idea was to have a TRACE level statement in case of an error … so one can get an idea where the problem lies.

teichsta:
> How do other contributors do that?

i am not sure about your specific situation. But i found this https://help.github.com/articles/about-git-rebase/ and this http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html quite helpful.

Hope this helps, Thomas E.-E.

berndpfrommer:
Thomas, thanks for the pointers.

I think this remark is key:
"Single commit - great. All you have to do now is push this branch to your fork. You'll have to use the --force flag to push here, since the rebase messed the ancestry relationship between what's currently in the remote and your local branch."

So is this pull request o.k. otherwise, can I go squash?

teichsta:
yes, please do :-)

- sent from a mobile device -

berndpfrommer:
Done!
Just needed the --force option on the push.

teichsta:
great, thanks @berndpfrommer!

Issue #2322, opened 2015-03-18
ihc binding and java 8


tktreload:
hello all, i am running openhab 1.7.0 snapshot but have seen the same problem in 1.5 and 1.6.

When using java 7 i have no problems connecting to my ihc/elko system after adding the certificate to the store, but repeating the same process for java 8 only yields an authentication error.

paulianttila:
When I implemented IHC binding, I tried to implement certificate manager which trust all certificates, so that certificate isn't need to add certificate store at all. I had many problems, which I could not resolve on that time. But some time ago I got tired to play those certificates, so I tried to solve those problems and in-fact I did.

[Here](https://dl.dropboxusercontent.com/u/1481715/org.openhab.binding.ihc-1.6.0-SNAPSHOT.jar) you can find test version, which does not need any certificates on the certificate store. Could you tried it and give feedback? Maybe it solve your problems on java 8. If not, enable debugs for IHC binding and send full log.

tktreload:
Yes thank you for the quick feedback. I will give it a try while the
girlfriend watches xfactor.

tktreload:
When i start 1.7.0 snapshot with the new binding and java 7 with the certificate added to the keystore everything is fine.
Then after uninstalling java7 and installing java8 I get the following errors:
```
19:53:53.164 INFO  o.o.b.ihc.internal.IhcBinding[:154]- Connecting to IHC / ELKO LS controller [IP='10.0.0.50' Username='admin' Password='******'].
19:53:53.300 WARN  o.o.b.ihc.internal.IhcBinding[:204]- Can't open connection to controller
org.openhab.binding.ihc.ws.IhcExecption: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at org.openhab.binding.ihc.ws.IhcHttpsClient.sendQuery(IhcHttpsClient.java:118)
	at org.openhab.binding.ihc.ws.IhcAuthenticationService.authenticate(IhcAuthenticationService.java:57)
	at org.openhab.binding.ihc.ws.IhcClient.openConnection(IhcClient.java:242)
	at org.openhab.binding.ihc.internal.IhcBinding.connect(IhcBinding.java:162)
	at org.openhab.binding.ihc.internal.IhcBinding.execute(IhcBinding.java:200)
[...]

PR #2598, opened 2015-05-15
IHC binding: Support for java 8 and simplified controller certificate…

paulianttila:
… handling (no need to add certificates to trust store anymore).


paulianttila:
Fix #2322

teichsta:
Thanks for this fix @paulianttila. I am wondering if this code change really require httpcomponents 4.4? Version 4.1 is already included in the target platform. Best, Thomas E.-E.

paulianttila:
Well, IHC controller TLS implementation is tricky (that's the reason why I changed to apache httpclient) and 4.4 was the version which I got it to work. I firstly tried 4.1version, but there was some issues, which I was not able to solve. I haven't used httpclient earlier, so it could also be that I just used it wrongly in "researching" phase. Anyhow API between 4.4 and 4.1 is changed a lot, so it's difficult to test if 4.1 could still work.

Issue #2470, opened 2015-04-17
Fix max value of saturation and brightness of Philips Hue binding


dominicdesu:
According to the latest [Philips Hue API](http://www.developers.meethue.com/documentation/lights-api#16_set_light_state), the maximum possible value for saturation and brightness is 254. In the old version of the Philips Hue API is has appearently been 255, which was an error in the APi description.

The Hue binding in openHAB uses the incorrect 255 as the max value. This seems not to be causing problems at the moment but should be fixed to be in alignment with the official API.

When updating the binding, it should be considered to extract the maximum brightness/saturation into a constant.

PR #2602, opened 2015-05-15
Set max value of saturation and brightness to 254 as defined by Hue API

dominicdesu:
FIx for issue #2470.
Introduced constants for max brightness and saturation and changes both max values from 255 to 254.


teichsta:
Thanks @dominicdesu!

PR #2604, opened 2015-05-16
Sonos: reduced log level for discovered non Sonos devices.

paulianttila:


teichsta:
thanks @paulianttila!

PR #2619, opened 2015-05-19
JPA persistence: Avoid null pointer exception if no user/password defined

cdjackson:
An NPE is thrown if no user/password is supplied. This might be the case if an embedded database is used where the user/pass is not needed as it's linked in with no net access (or I guess if you're silly enough to have a server with no password, but I guess not too many people do that!)...

### DIFF
--- a/bundles/persistence/org.openhab.persistence.jpa/java/org/openhab/persistence/jpa/internal/JpaPersistenceService.java
+++ b/bundles/persistence/org.openhab.persistence.jpa/java/org/openhab/persistence/jpa/internal/JpaPersistenceService.java
@@ -211,8 +211,15 @@ public class JpaPersistenceService implements QueryablePersistenceService {
 		Map<String, String> properties = new HashMap<String, String>();
 		properties.put("javax.persistence.jdbc.url", JpaConfiguration.dbConnectionUrl);
 		properties.put("javax.persistence.jdbc.driver", JpaConfiguration.dbDriverClass);
-		properties.put("javax.persistence.jdbc.user", JpaConfiguration.dbUserName);
-		properties.put("javax.persistence.jdbc.password", JpaConfiguration.dbPassword);
+		if(JpaConfiguration.dbUserName != null) {
+		    properties.put("javax.persistence.jdbc.user", JpaConfiguration.dbUserName);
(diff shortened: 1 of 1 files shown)


bundles/persistence/org.openhab.persistence.jpa/java/org/openhab/persistence/jpa/internal/JpaPersistenceService.java
```diff
@@ -211,8 +211,12 @@ protected EntityManagerFactory newEntityManagerFactory() {
...
 		properties.put("javax.persistence.jdbc.url", JpaConfiguration.dbConnectionUrl);
 		properties.put("javax.persistence.jdbc.driver", JpaConfiguration.dbDriverClass);
-		properties.put("javax.persistence.jdbc.user", JpaConfiguration.dbUserName);
-		properties.put("javax.persistence.jdbc.password", JpaConfiguration.dbPassword);
+		if(JpaConfiguration.dbUserName != null) {
+		    properties.put("javax.persistence.jdbc.user", JpaConfiguration.dbUserName);
+		}
+		if(JpaConfiguration.dbPassword != null) {
+		    properties.put("javax.persistence.jdbc.password", JpaConfiguration.dbPassword);
+		}
```
teichsta: probably a good idea to INFO/WARN log a message stating that we strongly advice to set a username + password?

cdjackson: Ok - although this might also confuse since my application is where no user/pass is required since the database is actually embedded - it’s not a server…

How about if user is not null and password is null, then add a warning? I’m guessing that any server based system will at least require a login name.

teichsta: good catch … would you change this PR accordingly?

cdjackson: > good catch … would you change this PR accordingly?
> 
Will do...

PR #2623, opened 2015-05-20
Add anchor to proxy URL

LukeOwncloud:
URLs for WebView using /proxy are missing the anchor/fragment attribute.

Check e.g. rest/sitemaps/ours/0102 (which must contain a webview with anchor (alias fragement) specified using #

In my case:

    Webview url="http://www.sample.com/?test&sp=IGERS72#forecast"

becomes

    http://192.168.0.78:93/proxy?sitemap=ours.sitemap&widgetId=01020300

however, it should be:

    http://192.168.0.78:93/proxy?sitemap=ours.sitemap&widgetId=01020300#forecast

This is fixed by this tiny PR.

Please test first! I was not able to compile openHAB due to some weird mvn errors...

Without this, the Android client cannot handle the anchors. (Browser view is not affected.)

### DIFF
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/resources/SitemapResource.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/resources/SitemapResource.java
@@ -423,6 +423,9 @@ public class SitemapResource {
 					throw new RuntimeException(ex.getMessage(), ex);
 				}
 			}
+			if(uri.getFragment() != null) {
+				sb.append("#" + uri.getFragment());
+			}
 			sbBaseUrl.append(sb.toString());
 			bean.url = sbBaseUrl.toString();
     	}


teichsta:
Thanks @LukeOwncloud! This might also apply to ESH … would you mind creating a PR there as well?

LukeOwncloud:
What is ESH?

teichsta:
Eclipse Smarthome :-) (see https://github.com/eclipse/smarthome)

LukeOwncloud:
Okay, now I understand. I saw the issue tracker for ESH on some Eclipse site and did not find the source code. So there it is and no issue tracker. That makes sense.

I am afraid I cannot help out there. Actually I hardly understand the openHAB source:
First for each `widget instanceof Image, Video, WebView` `bean.url` is set (lines 388, 395, 400) And then it is overwritten without first reading it -> https://github.com/openhab/openhab/blob/master/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/resources/SitemapResource.java#L430
Some re-factoring or cleaning up seems to be necessary.

As for 
https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
`bean.url` is set to some "proxy/..." value for `widget instanceof Image and Video` separately (duplicated code!) but for ``widget instanceof Webview` the widget.URL is taken as it.

I'd love to dig in there and do some cleaning up, I there was just more time...

Issue #2561, opened 2015-05-07
Freebox binding: outgoing phone calls are bypassed


lolodomo:
As discussed on the group board, there is a bug in the Freebox binding relative to the handling of outgoing phone calls.

I will provide a fix.

teichsta:
merged fix #2632 into master …

PR #2632, opened 2015-05-23
Freebox binding: fix phone calls management (issue 2561)

lolodomo:
Please find the correction for issue 2561


lolodomo:
I am sorry but I am not sure if this is normal that I have in my PR the first item (merge pull request #1 from openhab/master). My real intention was to have only the second item. Maybe it is due to changes in openHAB since my last fork ?

teichsta:
Fixes #2561

PR #2633, opened 2015-05-23
fix OSGi injection for persistence service due to bind/unbind ordering

hoegertn:
Under some circumstances dynamic references will call bind before unbind to ensure the presence of the reference. With the old implementation this led to the field set to null.
Problem might be throughout the code.

Signed-off-by: Thorsten Hoeger <thorsten.hoeger@hoegernet.com> (github: hoegertn)

/cc @teichsta @theoweiss

### DIFF
--- a/bundles/persistence/org.openhab.persistence.influxdb/java/org/openhab/persistence/influxdb/internal/InfluxDBPersistenceService.java
+++ b/bundles/persistence/org.openhab.persistence.influxdb/java/org/openhab/persistence/influxdb/internal/InfluxDBPersistenceService.java
@@ -16,6 +16,7 @@ import java.util.Dictionary;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.commons.lang.StringUtils;
 import org.influxdb.InfluxDB;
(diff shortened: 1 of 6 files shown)


kaikreuzer:
> Under some circumstances dynamic references will call bind before unbind

Could you elaborate on what these circumstances are? The current code is imho the usual pattern when using OSGi DS. Do you have a reference that this is actually the wrong way to do things?

teichsta:
any logs to prove your observation?

hoegertn:
If a service reference is mandatory and the policy is set to dynamic, the bind method is called before the unbind method to ensure the presence of the reference.

This possibility is specified in OSGi compendium, section Declarative Services, References to Services, Event Methods:

```
If the policy is dynamic then the bind method of a replacement service can be called before its corresponding unbind method.
```

kaikreuzer:
Interestingly, this sentence has only been introduced in 4.3, it wasn't there in 4.2.
Am I right that it only applies for services that have policy "dynamic" and are 0..1 or 1..1?
And then again, for services where it is clear that there is only one available in the system (like the itemRegistry), there should be no need to change the code, would you agree?

hoegertn:
I think that it only applies to dynamic services that are required. But for the second part I am not sure as the spec is very unclear about it. I changed it to be safe but if you think it is not needed just reject the PR.

kaikreuzer:
I definitely learned something from your PR, so many thanks for that :+1: 
But as your fix makes the code imho more difficult to read, I would suggest to reject the PR and only go for this approach in situations where we really have a problem, because there can be alternative unary services being swapped at runtime (which should rather be a corner case).

PR #2636, opened 2015-05-23
Update of de.map

spacemanspiff2007:
added "-=unbekannt"


teichsta:
using the "-" as default was the intended translation. If you personally like the "unbekannt" to be your favourite translation please feel free to change your local de.map as you want. However thanks a lot for your contribution. Best, Thomas E.-E.

Issue #2210, opened 2015-02-28
Dimmer item keeps sending INCREASING/DECREASING command


Eggn1n3:
Hi,

I am having issues with the DIMMER item. When I click in my sitemap on button up to increase the volume of my Sonos system, I get a lot of INCREASE commands that do not stop. The same for DECREASE when clicking button down. This espically happens when I click rappidly on up/down button. I already changed the frequence setting but that doesn help.

I am using OpenHAB 1.7.0

.item:
Dimmer       woonkamerVolume " [%d %%]"      { sonos="[OFF:woonkamer:volume], [ON:woonkamer:volume], [DECREASE:woonkamer:volume], [INCREASE:woonkamer:volume]", autoupdate="false" }

.sitemap:
Slider item=woonkamerVolume label="Volume [%d %%]" icon="settings" sendFrequency=1000	

Nothing in rule file

events.log:
2015-02-28 20:41:42 - woonkamerVolume received command INCREASE
2015-02-28 20:41:42 - woonkamerVolume received command INCREASE
2015-02-28 20:41:42 - woonkamerVolume state updated to 98
2015-02-28 20:41:42 - woonkamerVolume received command INCREASE
2015-02-28 20:41:42 - woonkamerVolume state updated to 99
2015-02-28 20:41:42 - woonkamerVolume received command INCREASE
2015-02-28 20:41:42 - woonkamerVolume state updated to 100
2015-02-28 20:41:42 - woonkamerVolume received command INCREASE
2015-02-28 20:41:42 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:43 - woonkamerVolume received command INCREASE
2015-02-28 20:41:44 - woonkamerVolume received command INCREASE
2015-02-28 20:41:44 - woonkamerVolume received command INCREASE
[...]

PR #2644, opened 2015-05-25
ClassicUI: Fix invalid empty reponse

sja:
Firefox and other Browsers crash in JS if server sends empty reponse. WebApp.net does not handle this correctly. It would be much more work to fix this, so we just fill the response body with an empty root node to make it valid xml.


sja:
This will fix nearly all of the `no element found` errors mentioned in #2210

teichsta:
Many thanks, @sja!

PR #2663, opened 2015-05-28
Small tweak for the Denon binding

idserda:
Some devices do not support the device info URL, this change makes the binding continue nevertheless.


teichsta:
Thanks @idserda!

PR #2670, opened 2015-05-31
H2 SQL Persistence

cdjackson:
This is a simple, high performance embedded SQL persistence service. The only installation required is to put the JAR into the addons folder and add a h2sql.persist file (see wiki on the .persist files).

This requires further testing - I've tested it on a Linux box, and on a BeagleBone Black and performance looks good (better than mySQL server running on the same machine). I've not tested all item types, so any feedback is welcome - the schema might require some adjustment for some items...

@teichsta - please don't merge this just yet.


teichsta:
ok, just let me know when this PR is ready for review …

mdbergmann:
Why is this add-on required?
H2 would also work with the JPA add-on. The only thing required would be the expansion of the classpath in the start.sh script.

cdjackson:
I agree, and I my first attempt was to use the JPA, but there are two problems. The first one is I wanted something that was simple to install and had a low footprint - the current JPA binding is large as it has a lot of drivers. Ok - we can solve this and I started to move them into separate bundles.

However, the killer was performance. JPA performance was really quite slow - I tried Derby and H2.

The other issue was the JPA binding stores everything as text (I think - I didn't look at this in great detail) - I guess this provides better compatability, but again, performance takes a hit, and this will also hit disk space.

My use for the H2 binding is an embedded system - so it needs to be small, fast, and use minimal storage. I didn't find JPA fitted the need.

mdbergmann:
OK. Operating this on an embedded system might be another matter.

However, JPA or other persistence layers, are used in extremely performance intensive systems, like Banks and others where it performs extremely well.
Compared with the plugins InfluxDB or DB4O, JPA seems to performs a lot better
(when querying large date ranges using HABmin).

Storing as string representation is not uncommon. SQLite does store everything as string.

cdjackson:
I don't disagree - and as I said, it was my original thought to use the JPA binding - I just didn't find that it was optimum for the need I had which is for embedded purposes. I don't criticise the JPA binding at all, and if I could have got the performance I would have been happy to use it - maybe if I'd spent more time on it I could have improved performance, but I also read some information on the web that said JPA was not as fast as other native interfaces. For running on an embedded machine, speed was my main criteria.

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+ * 
+ * In the store method, type conversion is performed where the default type for
+ * an item is not as above For example, DimmerType can return OnOffType, so to
+ * keep the best resolution, we store as a number in SQL and convert to
+ * DecimalType before persisting to H2SQL.
+ * 
+ * @author Chris Jackson - Initial contribution
+ * @since 1.8.0
+ */
+public class H2SqlPersistenceService implements QueryablePersistenceService, ManagedService {
```
teichsta: since update() is empty there is no more need to implement ManagedService, please also remove update() and the reference in OSGI-INF

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+        this.itemRegistry = null;
+    }
+
+    // public void setPersistentStateRestorer(PersistentStateRestorer persistentStateRestorer) {
+    // this.persistentStateRestorer = persistentStateRestorer;
+    // }
+
+    // public void unsetPersistentStateRestorer(PersistentStateRestorer persistentStateRestorer) {
+    // this.persistentStateRestorer = null;
+    // }
```
teichsta: remove commented code

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+            logger.warn("H2SQL: No connection to database. Can not persist item '{}'", item);
+            return;
+        }
+
+        // A bit of profiling!
+        long timerStart = System.currentTimeMillis();
+
+        // Default the type to double
+        String sqlType = new String("DOUBLE");
+        String itemType = item.getClass().toString().toUpperCase();
```
teichsta: rather use getSimpleName() which only contains the class itself

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * Connects to the database
+     */
+    private void connectToDatabase() {
+        try {
+            logger.debug("H2SQL: Connecting to database");
+            Class.forName(driverClass).newInstance();
+
+            final String USERDATA_DIR_PROG_ARGUMENT = "smarthome.userdata";
+            final String eshUserDataFolder = System.getProperty(USERDATA_DIR_PROG_ARGUMENT);
```
teichsta: move these declaration to the top of this class

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlItem.java
```diff
@@ -0,0 +1,53 @@
...
+	public Date getTimestamp() {
+		return timestamp;
+	}
+
+	@Override
+	public String toString() {
+		return DateFormat.getDateTimeInstance().format(timestamp) + ": " + name + " -> "+ state.toString();
+	}
+
+}
```
teichsta: add newline

bundles/persistence/org.openhab.persistence.h2sql/build.properties
```diff
@@ -0,0 +1,6 @@
+output.. = target/classes/
+bin.includes = META-INF/,\
+               OSGI-INF/,\
+               .,\
+               lib/h2-1.3.176.jar
+source.. = java/
```
teichsta: add newline

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+        if (item.getState() == null || item.getState() instanceof UnDefType) {
+            logger.debug("H2SQL: State of {} [{}] is {}. Store aborted", item.getName(), item.getClass()
+                    .getSimpleName(), item.getState());
+            return;
+        }
+
+        // Connect to H2SQL server if we're not already connected
+        if (!isConnected()) {
+            connectToDatabase();
+        }
```
teichsta: isn't there probably a leaner method to check the connection than isConnected(). Is connected() seem to take a timeout of 5000ms which will potentially slows down the whole persistenceService implementation (not only this Service)

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+            logger.debug("Got as HSB: {}", value);
+        } else {
+            // All other items should return the best format by default
+            value = item.getState().toString();
+        }
+        logger.debug("H2SQL: State is {}::{}", item.getState(), value);
+
+        try {
+            statement = connection.createStatement();
+            sqlCmd = new String("INSERT INTO openhab." + item.getName() + " (TIME, VALUE) VALUES(NOW(),'" + value + "');");
```
teichsta: no need for "new String()"

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+
+            final String USERDATA_DIR_PROG_ARGUMENT = "smarthome.userdata";
+            final String eshUserDataFolder = System.getProperty(USERDATA_DIR_PROG_ARGUMENT);
+            String databaseFileName = "etc/";
+            if (eshUserDataFolder != null) {
+                databaseFileName = eshUserDataFolder + "/";
+            }
+            databaseFileName += "h2sql/openhab";
+
+            String url = "jdbc:h2:file:" + databaseFileName;
```
teichsta: all these declarations are static and should be declared at the top of this file

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+                filterString += " AND";
+            }
+            filterString += " TIME<'" + sqlDateFormat.format(filter.getEndDate().getTime()) + "'";
+        }
+
+        if (filter.getOrdering() == Ordering.ASCENDING) {
+            filterString += " ORDER BY Time ASC";
+        } else {
+            filterString += " ORDER BY Time DESC";
+        }
```
teichsta: for the sake of readability you could use the tertiary operator ("?")

bundles/persistence/org.openhab.persistence.h2sql/java/org/openhab/persistence/h2sql/internal/H2SqlPersistenceService.java
```diff
@@ -0,0 +1,464 @@
...
+            // Retrieve the table array
+            Statement st = connection.createStatement();
+
+            String queryString = new String();
+            queryString = "SELECT Time, Value FROM openhab." + filter.getItemName();
+            if (!filterString.isEmpty()) {
+                queryString += filterString;
+            }
+
+            logger.debug("H2SQL: " + queryString);
```
teichsta: use parameterised logging

teichsta:
please find my review comments inline. Please also squash your commits into one. Thanks, Thomas E.-E.

cdjackson:
Thanks Thomas - I'll try and take a look at this next week.

lewie:
Hello @cdjackson, @teichsta ,
*sqlTypes are missing "CALLITEM" and "LOCATIONITEM".
*"SWITCHITEM" should be VARCHAR(3) because you get "ON" and "OFF" as String back. CHAR(3) we will get "ON " with an additional space.
(Thought I had changed this in Mysql-Service, but now it is back to CHAR(3))

It seems H2 Binding is bases on an an older version of [Mysql Binding](https://github.com/openhab/openhab/blob/master/bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java), there are some more changes i think.
OpenHub gets some more redundancy in Persistence services with a new "single-Database-Binding" based on Mysql-Binding. @teichsta convinced me to take back my PostgreSQL Binding (also based on Mysql). 

You and me talked about a [generic SQL Binding](https://github.com/openhab/openhab/issues/2804), which can be extended.

So I have prepared a generic DBMS Binding (including H2), which is already working on my PC. 
But it takes a few days to ripen, then I will introduce it ;-)

Rgds
Helmut

cdjackson:
Hi Helmut,
Ok - I’ll take a look when you’re done :)

The idea for H2 was to have something simple, small, fast, and robust for simple devices like RPi/BBB etc. You’re right - I did base it on an older MySQL - the PR is many months old now :)

Cheers
Chris

lewie:
Hi @cdjackson,
absolutely, H2 (derby, sqlite) embedded would be fine for Raspis etc.

Rgds
Helmut

cdjackson:
I tried derby first and found it very slow compared to H2 - I also found a comparison somewhere on the web that showed similar issues…

My motive was simplicity, and speed - I have NO configuration in the H2binding (at least none is actually required)… I was supporting someone who couldn’t get to grips with configuring things, so that was one of the ideas...

mdbergmann:
Hi Guys.

A generic DBMS binding based on JDBC?
I assume this requires quite a bit of work for customizations for each DBMS.
Isn’t this what JPA or Hibernate already have done?


Manfred

watou:
The database layout for JPA persistence is different from the requirement (described elsewhere) for a Postgres persistence mechanism, where the desire is to have one table per item using SQL types like the MySQL bundle.  I would not characterize it as a lot of work to accommodate different databases under JDBC, but definitely some work, and exactly how much remains to be seen.  A careful implementation can add support for new JDBC drivers without breaking existing implementations, eliminate duplicate code, and shrink the codebase by supporting MySQL.

mdbergmann:
With one table per item I assume you mean per item type?

watou:
No, per item.

mdbergmann:
Hmm, what would be the benefit of having 10’s or 100’s of tables?

lewie:
@cdjackson,

cdjackson:
> @cdjackson <https://github.com/cdjackson>,
> 
???

lewie:
Sorry, pushed the button before copy and paste, once again:

Hi @cdjackson, take a look now I've done now. :)
New generic JDBC Persistence Service

Wiki page is ready too.

Rgds
Helmut

lewie:
Ahrg... :-(

Sorry, pushed the button before copy and paste, once again:

Hi @cdjackson, take a look now I've done now. :)
[New generic JDBC Persistence Service](https://github.com/openhab/openhab/pull/3248)

 [Wiki page](https://github.com/openhab/openhab/wiki/JDBC-Persistence) is ready too.

Rgds
Helmut

teichsta:
ok, so how are we going to proceed with this PR? From my point of view it can be merged, right?

teichsta:
@cdjackson any update on this? Would be great to have this PR done …

teichsta:
@cdjackson how are we going to proceed with this contribution? Move over to 1.9 or do you have the changes in your local branch available already?

cdjackson:
Hi Thomas.
Sorry - I’ve been umming a bit this since the JDBC binding now also supports H2. I’ve not tried it myself, but assuming it’s a good-un, it probably makes sense to drop this PR and stick with the single service.  For now though, let’s look at this for 1.9 and I’ll try and try the JDBC binding…

Chris

(and happy new year :) )

kaikreuzer:
> For now though, let’s look at this for 1.9

Afair, this was also something pending for 2.0 - so maybe after the 1.8 release it would be enough to go for 2.0 then.

Happy New Year as well :-)

teichsta:
fine for me!

Happy New Year :-)

Issue #1403, opened 2014-09-07
Several Actions aren't properly resolved in Designer


udo1toni:
Although I copied the action-file to Designer/addons-folder, the actions aren't recognized in the rules-files. So the designer shows errors, where aren't errors at all, also the auto completion does not work (evident)
This Problem exist at least with `Win7-64, Oracle Java7u67-32` and with `openHAB nightly build 716` or older. With the same Installation, `xmpp-action` works as expected, so I think, the problem could be in squeezebox-action-file.

See also http://knx-user-forum.de/openhab/35875-designer-meckert-squeezebox-actions.html

udo1toni:
Question: Do I have to import something to rulefile, to make correct use of squeezebox-actions? (Maybe something like **import org.openhab.action.squeezebox** ?)

staehler:
I've the same problem unfortunately. Is there no hint, how to get rid of this anoying errors?
Meanwhile this false errors lead to overlook real faults.

teichsta:
so you are working with 1.7 SNAPSHOT, right?

staehler:
Thats correct. I have the build number not at my hand, but it's quite 
recent (this week)

Gesendet mit AquaMail für Android
http://www.aqua-mail.com


Am 17. April 2015 15:24:36 schrieb Thomas Eichstädt-Engelen 
<notifications@github.com>:

teichsta:
strange … when running the Designer from IDE the Squeezebox and xPL Action appear without any problems. So it seems to be an assembly problem …

staehler:
Hi Thomas,
hmm, do I need to include a special lib into my rules file? 
I get following error messages:
Couldn't resolve reference to JvmIdentifiableElement 'squeezeboxPlayUrl'.
or
Couldn't resolve reference to JvmIdentifiableElement 'squeezeboxSpeak'.
but the actions work well in openhab.

BTW I get such a error message for my.openhab actions ax´s well:
Couldn't resolve reference to JvmIdentifiableElement 'sendNotification'.

teichsta:
as i said it only works when starting the Designer out of the IDE. When running the binary i can reproduce the given problem

udo1toni:
:) happy that you can reproduce it :)

lolodomo:
Please note that we have the same issue with the MiOS actions too, at least with Designer1.6.2. I have not tested with Designer 1.7.0. So please check the MiOS actions too.

watou:
[...]

PR #2672, opened 2015-05-31
Re-enable CORS for REST API

idserda:
Support for Cross-Origin Resource Sharing (CORS) in the REST API was lost in 1.7.0 because of the update of Atmosphere. This again adds the necessary interceptor.

### DIFF
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/RESTApplication.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/RESTApplication.java
@@ -197,6 +197,7 @@ public class RESTApplication extends Application {
         // use the default interceptors without PaddingAtmosphereInterceptor
         // see: https://groups.google.com/forum/#!topic/openhab/Z-DVBXdNiYE
         final String[] interceptors = {
+    			"org.atmosphere.interceptor.CorsInterceptor",
 			"org.atmosphere.interceptor.CacheHeadersInterceptor",
 			"org.atmosphere.interceptor.AndroidAtmosphereInterceptor",
 			"org.atmosphere.interceptor.SSEAtmosphereInterceptor",


teichsta:
thanks @idserda - is this to considered as a bug?

idserda:
Without CORS enabled you can't access the REST API in a webpage hosted on a different host/port. Since this was working in 1.6.x I would say this is indeed a bug.

kaikreuzer:
Is CORS working for openHAB 2? If not, the solution might be different as there is no atmosphere used anymore.

idserda:
No it's not working there. Seems like OH2 uses JAX-RS so might be possible to add an additional ContainerResponseFilter to enable this.

Testato:
i tried to add a filter to the end of /opt/openhab/etc/webdefault.xml 
but not only the CORS do not work but all icons in the ClassicUI disappear

Witch is the way for enable CORS ?
I'm on 1.7.0 but if it is necessary i can update
Thanks

<filter>
<filter-name>cross-origin</filter-name>
<filter-class>org.eclipse.jetty.servlets.CrossOriginFilter</filter-class>
</filter>
<filter-mapping>
<filter-name>cross-origin</filter-name>
<url-pattern>/*</url-pattern>
</filter-mapping>

</web-app>

idserda:
This will be fixed in version 1.7.1, which should be released soon.

Testato:
on 1.7.1 the CORS is enabled by default ?
Is there a nightly available so i can made a test ?

idserda:
Yes, it's fixed in the latest nightly builds.

Testato:
sorry for the question but i do not found where i may download the nightly 1.7.1 

and is there a tutorial for apply the update ? i will loose some configuration after update ?

Thanks

Testato:
sorry for the question but where i can download the 1.7.1 nightly ?

kaikreuzer:
According to https://groups.google.com/d/msg/openhab/pGrmyVDHWxQ/M3GlgDsDhPsJ <https://groups.google.com/d/msg/openhab/pGrmyVDHWxQ/M3GlgDsDhPsJ>, it will be published yesterday, so it probably won’t take too long before it is available!

Testato:
thanks Kai,
I understand that in this case it is better await for the stable because it will arrive soon, but if an user want install a nightly:
1) there is a direct link for the download (for manual installation) ?
2) there is a nightly repository for apt-get ?

Thanks a lot for your work, in this moment i'm working on a client App so also one or two day it is important for me, and this CORS bug stop me completely.

pasqualevitiello:
+1

teichsta:
sorry folks (also see #1403) unfortunately i am still struggling with making the my.openHAB Action available in the Deigner. All the other Actions are available already. Hope to finalize this issue soon!

kaikreuzer:
Nightly builds of 1.8.0-SNAPSHOT are available at https://openhab.ci.cloudbees.com/job/openHAB/ <https://openhab.ci.cloudbees.com/job/openHAB/>, I would assume that the CORS bugfix is there as well.
Afaik, there are no nightly builds for patch branches like 1.7.1.

Regards,
Kai

Testato:
Hi,
today I installed the 1.7.1 and the CORS work, Thanks a lot for it.

but I discover a little bug, in the Allowed Method there is no PUT, so by ClassicUI is possible do a PUT, by an external website is not possible, this is my situation:
```
Item state URI: http://192.168.100.254:8080/rest/items/Light_FF_Bath_Ceiling/state
Request Header: Access-Control-Request-Method:PUT
Response Header: Access-Control-Allow-Methods:OPTIONS, GET, POST

PR #2678, opened 2015-06-02
Fix for NullPointerException in jsr223 add-on. This would happen if a…

mdbergmann:
…n exception was throws during evaluation of the script.

If an exception was throws during evaluation, the script wasn't added to a map.
The removeScript() method however was removing the script from the map nevertheless.
The getRules() method was then called on a null reference.


teichsta:
also merged into 1.8.0-SNAPSHOT

PR #2682, opened 2015-06-03
Fix to YamahaReceiver binding MANIFEST.MF bundle-version

sumnerboy12:
This somehow got missed and in the current 1.7 release the Yamaha binding is stamped as v1.5! Whoops!


teichsta:
thanks @sumnerboy12!

PR #2699, opened 2015-06-07
Add Vision Security ZM 1602 (Type 2009, ID:908)

CondormanFr:


cdjackson:
This device is already in the products.xml file, so you can't add it again in this way.  Instead of adding the whole section for this device, you should just add the new reference section... Take a look a few lines further down at the Fibaro devices to see what I mean.

CondormanFr:
Sorry for that. I made the change to add only a new reference instead of creating a new product entry.

cdjackson:
Thanks.

PR #2713, opened 2015-06-10
[Ecobee] Changed default polling to 3 minutes; added binding example.

watou:
Lengthened default polling to three minutes to conform to API documentation.


teichsta:
thanks, @watou

PR #2715, opened 2015-06-10
Open http or https port for jetty only when configured

slawekjaranowski:
There is a situation that we want to have only one connector http or https.

When we use only one kind of connector only proper connector should be registered in discovery service.

### DIFF
--- a/bundles/io/org.openhab.io.cv/src/main/java/org/openhab/io/cv/CVApplication.java
+++ b/bundles/io/org.openhab.io.cv/src/main/java/org/openhab/io/cv/CVApplication.java
@@ -156,9 +156,25 @@ public class CVApplication extends PackagesResourceConfig  {
 			} catch (Exception e) {
 				logger.error("Could not start Jersey framework", e);
 			}
-        	
-    		httpPort = Integer.parseInt(bundleContext.getProperty("jetty.port"));
-    		httpSSLPort = Integer.parseInt(bundleContext.getProperty("jetty.port.ssl"));
+
+			String bundleProperty =  bundleContext.getProperty("jetty.port");
+			if (bundleProperty != null) {
+				httpPort = Integer.parseInt(bundleProperty);
+			} else {
+				httpPort = -1;
+			}
+
+			bundleProperty = bundleContext.getProperty("jetty.port.ssl");
+			if (bundleProperty != null) {
+				httpSSLPort = Integer.parseInt(bundleProperty);
+			} else {
+				httpSSLPort = -1;
+			}
+
+			if (httpPort < 0 && httpSSLPort < 0) {
+				logger.error("Please set at least one of jetty.port, jetty.port.ssl property");
+				throw new RuntimeException("There are no jetty ports settings");
+			}
 
 			httpService.registerServlet(CV_SERVLET_ALIAS,
 				new AtmosphereServlet(), getJerseyServletParams(), createHttpContext());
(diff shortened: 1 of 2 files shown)


teichsta:
@peuter could you please have a look at this? Thanks, Thomas E.-E.

teichsta:
@peuter any chance to have a look into this one? Thanks, Thomas E.-E.

teichsta:
@peuter probably any news?

hakan42:
@slawekjaranowski , could you add a warning if both http and https ports are configured? If I read the first comment correctly, this was the case which you wanted to avoid with your patch.

slawekjaranowski:
Hi, I think if both ports are configured is nothing bad.
Of course if you wish I can add warning.

My case is that I want to use only one port http or https, and when I don't use one of them it is not necessary to define system property and also is not necessary to register port in discover service.

hakan42:
Looks good to me with the use case mentioned by @slawekjaranowski

kaikreuzer:
Guys, please note that this PR does changes on the core, which is not allowed unless critical. If it is merged, it must be guaranteed that the code is synched to Eclipse SmartHome before merging.

Is any openHAB 1 maintainer taking care of that?

hakan42:
Oops, sorry, was not aware of that. Do you want me to revert the change or should we create a PR against ESH?

As this seems not to be a critical change, I would opt for reverting.

teichsta:
> As this seems not to be a critical change, I would opt for reverting.

+1 would you please revert this PR? @slawekjaranowski please provide your changes again for ESH. Once it is merged there we can "downport" this changes to OH1 as well, ok?!

hakan42:
o.k., reverting...

hakan42:
For the record, the merge is reverted in the master branch.

kaikreuzer:
Sorry, couldn't respond any earlier. I don't think that it was necessary to physically revert it. The problem with merging PRs here is that they are then out of sight and with a 99% chance nobody will ever remember to port it to ESH and we will only notice it once people migrate to OH2 and notice a regression. That's why I asked @teichsta to always first make sure that things are contributed to ESH before merging (=closing) the PR here.

teichsta:
>with a 99% chance nobody will ever remember to port it to ESH and we will only notice it once people migrate to OH2 and notice a regression. 

and that's why i suggested to revert the PR. See https://community.openhab.org/t/some-rules-for-reviewing-prs/4724 where we could add this as rule to first add and merge at ESH and "down port" to OH1 afterwards.

kaikreuzer:
Well, reverting does not reopen this issue, right? That's why "getting out of sight" is not solved, unless you have opened an new issue for it.

slawekjaranowski:
As I see that ESH has different architecture in this point.
One place which I found, to provide my change is:

https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/MDNSAnnouncer.java

@kaikreuzer , @teichsta  please confirm me it is good place.
I  will prepare correct change to ESH.

kaikreuzer:
What is this issue exactly about? The title does not seem to match the description and the code change.
From the title, I assumed it is some feature to allow configuring Jetty for http and/or https - but the issue is really about mDNS service registration instead, right?
If so, than this is indeed something hat I have implemented a long while ago in ESH already - the class that you reference is doing that. So no need to create any PR there, sorry for the hassle.

But please update the title of this issue to make clear that it is about mDNS announcements and not http ports.

slawekjaranowski:
In this PR I resolved 2 issues for OH1: 
- NullPointerException if one property for jetty port is not defined 
- not necessary mDNS announcements

In ESH MDNSAnnouncer  line 49, 54 is still affected by NullPointerException if property is missing. "mdnsService.registerService" is also called everytime.

kaikreuzer:
> In ESH MDNSAnnouncer line 49, 54 is still affected by NullPointerException if property is missing. 
> "mdnsService.registerService" is also called everytime.

This is not correct. the Integer.parseInt() throws a NumberFormatException, if a null parameter is passed. Hence there is no NPE and `mdnsService.registerService` isn't called either.

slawekjaranowski:
You are right. So in ESH we have this feature. 
Up to you - what next in this case with this PR.

PR #2718, opened 2015-06-11
Webui CSS fixes for non-webkit browsers

resetnow:
As of 2015, the majority of browsers supports `border-radius` CSS property [without any vendor-specific prefixes](http://caniuse.com/#feat=border-radius). Moreover, modern Firefox versions do not support this property with `-moz-` prefix. Therefore, non-prefixed versions were added. I think it's probably safe to remove all prefixed values (thus getting rid of firefox.css), but as for now this works too.

Also, there is a fix for “back” button display in Firefox.

### DIFF
--- a/bundles/ui/org.openhab.ui.webapp/web/WebApp/Design/Firefox.css
+++ b/bundles/ui/org.openhab.ui.webapp/web/WebApp/Design/Firefox.css
@@ -51,6 +51,11 @@
 	-moz-border-radius:4px;
 }
 
+#waBackButton, #waHomeButton {
+	border-style: solid;
+	border-color: transparent;
+}
+
 #waBackButton {
 	-moz-border-radius-topleft:22px;
 	-moz-border-radius-bottomleft:22px;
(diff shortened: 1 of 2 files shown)


resetnow:
>ERROR: Failed to parse POMs
>Test Result (no failures)

teichsta:
Thanks for providing this fix. Will have a look soon …

teichsta:
I've tested this PR which looks quite good already. However the back button is still broken in my browser (38.0.5 OS X)

resetnow:
@teichsta I don't have a working OSX instance around, I'll try to find something and test it this weekend. In the meanwhile, could you please confirm the back button is not displayed correctly on [this page](http://codepen.io/vladislav/full/Nqabvp/) too? Also, did you try reloading the page without using the cache (`ctrl`-`F5`) ?

resetnow:
@teichsta I have checked with OS X and Firefox 38, and it looks as it should: http://i.imgur.com/RZ9Vzqi.png

resetnow:
Apparently “Home” button needed this fix too, I have added a commit.

teichsta:
great thanks, if you could finally squash your commits into one i would be happy to merge …

resetnow:
Done.

resetnow:
@teichsta <i>ping</i>

teichsta:
Thanks @resetnow!

kaikreuzer:
@teichsta Doesn't this also apply to ESH?

teichsta:
yes, sure, sorry!

@resetnow would you mind sending this fix also to (Eclipse Smarthome)[https://github.com/eclipse/smarthome]?

resetnow:
Thank you for merging! ESH changes are a bit different because here at OpenHAB some things were already fixed in `Firefox.css`. PR: https://github.com/eclipse/smarthome/pull/360

Issue #2679, opened 2015-06-02
PiFace binding reads wrong state for output pins upon startup


sbjdk:
The binding confuses input pins and output pins. Follow this to reproduce:

Simple items

```
Switch     PifaceSwitch0       "Switch 0" (all,piface) { piface="piface1:OUT:0" }
Switch     PifaceSwitch1       "Switch 1" (all,piface) { piface="piface1:OUT:1" }
Switch     PifaceSwitch2       "Switch 2" (all,piface) { piface="piface1:OUT:2" }
Switch     PifaceSwitch3       "Switch 3" (all,piface) { piface="piface1:OUT:3" }
Switch     PifaceSwitch4       "Switch 4" (all,piface) { piface="piface1:OUT:4" }
Switch     PifaceSwitch5       "Switch 5" (all,piface) { piface="piface1:OUT:5" }
Switch     PifaceSwitch6       "Switch 6" (all,piface) { piface="piface1:OUT:6" }
Switch     PifaceSwitch7       "Switch 7" (all,piface) { piface="piface1:OUT:7" }

Contact    PifaceContact0      "Contact 0" (all,piface){ piface="piface1:IN:0" }
Contact    PifaceContact1      "Contact 1" (all,piface){ piface="piface1:IN:1" }
Contact    PifaceContact2      "Contact 2" (all,piface){ piface="piface1:IN:2" }
Contact    PifaceContact3      "Contact 3" (all,piface){ piface="piface1:IN:3" }
Contact    PifaceContact4      "Contact 4" (all,piface){ piface="piface1:IN:4" }
Contact    PifaceContact5      "Contact 5" (all,piface){ piface="piface1:IN:5" }
Contact    PifaceContact6      "Contact 6" (all,piface){ piface="piface1:IN:6" }
Contact    PifaceContact7      "Contact 7" (all,piface){ piface="piface1:IN:7" }
```

1. Switch on all PiFace outputs (e.g. using openHAB)
2. Shutdown openHAB
3. While holding down two of the physical input buttons on the PiFace, start up openHAB and let it initialize completely 

Messages like these should appear in the log (INFO)
```
PifaceSwitch1 state updated to ON
PifaceContact1 state updated to CLOSED
```

When looking at the log and the UI, it becomes evident that the state of the output pins are not read correctly. All Switches are OFF - unless (as above) the input pin with the same number is CLOSED.

I suspect that the 'digital read' method should have BindingType (IN, OUT) included since both have pin numbers 0-7.

[...]

PR #2733, opened 2015-06-13
PiFace binding reads wrong state for output pins upon startup (#2679)

sbjdk:
Here is proposed fix for #2679. Has been discussed with Ben Jones (@sumnerboy12).

Some of the changes involves migrating from the use of an -1 error return code to the use of an Exception in the sendCommand method. The method only returns a byte and we need the whole byte for the output pin state.


teichsta:
This PR fixes #2679. Thanks @sbjdk!

PR #2738, opened 2015-06-15
Increase compatibility with some thermostats

khendricks:
Some thermostats (Trane) do not support setpoint values other than integers.  This change will strip any trailing zeros on BigDecimal numbers to ensure a minimum scale.


khendricks:
What do you guys think about doing this all in ZWaveCommandClass.encodeValue()?

cdjackson:
I've not looked at this in detail, but it might make more sense if it's in the encodeValue method.

I'll take a better look tonight.

PR #2749, opened 2015-06-17
Update zme_05461.xml

mstormi:
XML syntax error fixed


cdjackson:
Just to note that the translations aren't used, so you won't see any difference...

PR #2759, opened 2015-06-18
This fixes #2747 - decreases severity of lib version logging and decr…

### LINKED ISSUE #2747: KNX binding floods log
With the 1.7.0 version of the KNX binding, I have my log full of entries like 
```
2015-06-16 19:42:30 [INFO ] [tuwien.auto.calimero          :51   ] - [KNXnet/IP receiver] link 127.0.0.1:3671: indication from 1.1.7
```
when starting in debug mode. I would expect that those are suppressed as it was in the past and to only have the openhab KNX log messages instead.

Furthermore, when starting in normal mode, I have a log entry 
```
2015-06-16 19:45:17.259 [INFO ] [.b.knx.internal.bus.KNXBinding] - Calimero library version 2.2.0-alpha
```
which imho also should not be there as it is irrelevant for a normal user.

teichsta:
…eased calimero logger to ERROR

### DIFF
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
@@ -68,7 +68,7 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider> implements
 	private boolean mKNXConnectionEstablished;
 
 	public void activate(ComponentContext componentContext) {
-		logger.info("Calimero library version {}", Settings.getLibraryVersion());
+		logger.debug("Calimero library version {}", Settings.getLibraryVersion());
 		logger.trace("KNXBinding: activating");
 		KNXConnection.addConnectionListener(this);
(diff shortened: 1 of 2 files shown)


teichsta:
please review @kaikreuzer

kaikreuzer:
Thanks! But shouldn't logback_debug.xml be changed instead of logback.xml?
I will have to add the same to openHAB 2 then (which is a bit nasty having to add binding specific logger settings in the global logback configuration, but what to do...)

teichsta:
> But shouldn't logback_debug.xml be changed instead of logback.xml?

true i would rather suggest to change both …

kaikreuzer:
> true i would rather suggest to change both …
Why? The root level is already WARN, so there should not be any logging visible from calimero.

teichsta:
> but what to do...

we could also change LogAdapter and map calimeros INFO log level onto openHAB DEBUG?

kaikreuzer:
> we could also change LogAdapter and map calimeros INFO log level onto openHAB DEBUG?

Sounds rather complex to do within a certain binding, so I'd suggest to leave it as it is, thanks!

Issue #2642, opened 2015-05-24
Insteon binding conflicts with zwave binding


xsnrg:
I have an Insteon PLM modem in use with the insteonplm binding, and a Z-stick in use with the zwave binding.  These are 1.7 bindings from the current build of OH 1.7.  I am running them under OH2.

What I am seeing is documented in this ticket:  https://github.com/cdjackson/HABmin2/issues/41

The insteon binding always loads before the zwave binding.  The insteon binding always come up and works.  The zwave binding works perhaps 25% of the time and OH2 needs to be restarted until the zwave binding works correctly.

If I remove the insteon binding the zwave binding starts every time I have tried, on the first try.

berndpfrommer:
At some point I had introduced what can only be called a bug in the insteonplm binding:

try {
			/* by default, RXTX searches only devices /dev/ttyS* and
			 * /dev/ttyUSB*, and will so not find symlinks. The
			 *  setProperty() call below helps 
			 */
			System.setProperty("gnu.io.rxtx.SerialPorts", m_devName);
			CommPortIdentifier ci =
					CommPortIdentifier.getPortIdentifier(m_devName);
			CommPort cp = ci.open(m_appName, 1000);
}

By calling System.setProperty() I was overwriting the serial ports GLOBALLY for all bindings, since "System" is address-space wide.

But this was fixed (or at least so I thought) on March 26th with this commit:

https://github.com/berndpfrommer/openhab/commit/cfa0e00565476258ddb1225447d1c1bb270a18d2

Now the relevant section of code now looks like this:

try {
			/* by default, RXTX searches only devices /dev/ttyS* and
			 * /dev/ttyUSB*, and will so not find symlinks. The
			 *  setProperty() call below helps 
			 */
			String ports = System.getProperty("gnu.io.rxtx.SerialPorts");
			String sp = ((ports == null) ? "" : (ports + ":")) + m_devName;
			// note: calling setProperty() is not threadsafe, in particular if
			// multiple bindings use the serial port
			System.setProperty("gnu.io.rxtx.SerialPorts", sp);
			CommPortIdentifier ci =
					CommPortIdentifier.getPortIdentifier(m_devName);


Do you have that commit? There is still the issue of non-thread safety, but that race condition should be so hard to hit that I doubt that is the problem you are seeing.

Bernd

xsnrg:
Hi Bernd,

[...]

PR #2760, opened 2015-06-18
InsteonPLM: attempted fix for incompatibility with Z-Wave binding

berndpfrommer:
Attempted fix for issue #2642

### DIFF
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/SerialIOStream.java
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/driver/SerialIOStream.java
@@ -16,7 +16,11 @@ import gnu.io.SerialPort;
 import gnu.io.UnsupportedCommOperationException;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Enumeration;
 
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


teichsta:
Thanks for this fix, @berndpfrommer.

With regards to your comments about thread safety of setting the System property. Wouldn't it be better if we would implement something generally available (in io.transport.serial) that deals with this issue? Otherwise we would have check this again and again in all new Addons using serial.

Might also be interesting for OH2

berndpfrommer:
I'm all for it.

Since I added the same setProperties() call to the URTSI binding, it also suffered from the same incompatibility issues: https://github.com/openhab/openhab/issues/1351

The updateSerialProperties() method I added for the fix purposefully does not reference any local variables, so it can be readily moved elsewhere. Adding a "synchronized" statement to it would also take care of the race condition mentioned in the comments.

I'm really clumsy with the build environment though. Would you mind adding a utils class to io.transport.serial?

teichsta:
> I'm really clumsy with the build environment though. Would you mind adding a utils class to io.transport.serial?

i could try … please ping me if nothing happened in the next three weeks.

Thanks, Thomas E.-E.

xsnrg:
As this was merged to master, should this ticket be closed and a new one opened for the broader issue?

It has been working for me since merged by the way, but I only have the two serial interfaces.

wishmoooop:
While developing a binding between OH2 and MySensors, where I'm using the io.transport.serial library, I'm too getting in conflict with the zwave binding from 1.X. 
If I add the code for the fix from @berndpfrommer (updateSerialProperties()) in: https://github.com/cdjackson/openhab/blob/master/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java

Both bindings work together. MySensors most of the time is initialized before zwave at startup. A general solution for all bindings that use io.transport.serial would be nice.

plajjan:
I too am seeing conflicts but between benq serial and zwave. I'm having difficulty following all the issues used here to track various things. Do we have one to track the more generic problem? Given that I wrote the benq serial transport, am I expected to update that?

aroop:
I am running into this issue with openhab 2.3.0. I only use two serial ports one for zwave and another for insteon. 

If I remove the insteon bindings zwave binding will be able to find its COM port or else zwave does not work

Issue #2732, opened 2015-06-13
[MySQL.persist] Saving group as string? But datatype is double!?


Bubelbub:
2015-06-13 23:31:01.220 [ERROR] [.p.m.i.MysqlPersistenceService] - mySQL: Could not store item 'Floor' in database with statement 'INSERT INTO Item50 (TIME, VALUE) VALUES(NOW(),'OFF') ON DUPLICATE KEY UPDATE VALUE='OFF';': Data truncated for column 'Value' at row 1
	2015-06-13 23:31:01.336 [ERROR] [.p.m.i.MysqlPersistenceService] - mySQL: Could not store item 'Kitchen' in database with statement 'INSERT INTO Item55 (TIME, VALUE) VALUES(NOW(),'OFF') ON DUPLICATE KEY UPDATE VALUE='OFF';': Data truncated for column 'Value' at row 1
	2015-06-13 23:31:01.372 [ERROR] [.p.m.i.MysqlPersistenceService] - mySQL: Could not store item 'Bath' in database with statement 'INSERT INTO Item57 (TIME, VALUE) VALUES(NOW(),'OFF') ON DUPLICATE KEY UPDATE VALUE='OFF';': Data truncated for column 'Value' at row 1
	2015-06-13 23:31:01.506 [ERROR] [.p.m.i.MysqlPersistenceService] - mySQL: Could not store item 'Livingroom' in database with statement 'INSERT INTO Item63 (TIME, VALUE) VALUES(NOW(),'OFF') ON DUPLICATE KEY UPDATE VALUE='OFF';': Data truncated for column 'Value' at row 1
	2015-06-13 23:31:01.569 [ERROR] [.p.m.i.MysqlPersistenceService] - mySQL: Could not store item 'Office' in database with statement 'INSERT INTO Item66 (TIME, VALUE) VALUES(NOW(),'OFF') ON DUPLICATE KEY UPDATE VALUE='OFF';': Data truncated for column 'Value' at row 1

They are groups!

    // default.items
    Group	Floor	"Floor"
	Group	Kitchen	"Kitchen"
	Group	Bath	"Bathroom"
	Group	Livingroom	"Livingroom"
	Group	Office	"Office"

But they exists in database table `items`.

![](http://i.imgur.com/O56qDzw.png)

Thank's in Advance for fixing.

teichsta:
@jo-we @Dennis650 @cdjackson could anybody of you take care of this one? Thanks in advance, Thomas E.-E.

[...]

PR #2764, opened 2015-06-19
[MySQL.persist] Saving group as string? But datatype is double!? #2732 

lewie:
default/fallback ItemType is StringType ("STRINGITEM") now.
if GroupItem:<ItemType> is not defined in *.items then using StringType
If higher Types like DOUBLE or INT needed for serialisation, it can be set in openhab.cfg


teichsta:
Thanks @lewie i'll have a look at this Fix soon! Can others confirm this fix is working in general?

lewie:
Hi @Thomas, please ignore this pull request! 
I must again make adjustments. (see https://github.com/openhab/openhab/issues/2732)

Thanks
Helmut

teichsta:
> Hi @Thomas, please ignore this pull request! 

will do :-)

PR #2765, opened 2015-06-20
Ecobee bug: Only update item if we know we retrieved the thermostat it refers to.

watou:
For users of the Ecobee binding who have more than one thermostat that are managed under the same account at ecobee.com, this is an important bugfix that ought to be in 1.7.1.  The `readEcobee` method was iterating through all in-bound items from that account and setting their states, even if the thermostat that that item refers to had not changed and was therefore not in the payload returned from ecobee, resulting in sometimes Uninitialized items.


teichsta:
Thanks @watou!

PR #2777, opened 2015-06-22
Add Linear FS20Z fixture module to database

khendricks:


cdjackson:
How come this is listed in two manufacturers? Do you have examples of both manufacturers?
I think I would prefer to split this out to two separate files for the two manufacturers if that is needed.

khendricks:
I believe they are identical hardware.  See this link:
http://smile.amazon.com/Linear-FS20Z-1-Z-Wave-Fixture-Module/dp/B00ER6MH22

Thats the Linear FS20Z.  If you look at the image, it says LFM-20 right on the front.

cdjackson:
Do you have both of these though? If not, how do you know the type/id codes? Clearly they can't be 'identical' if they have different manufacturers (as you've added in the database) - are you sure that the type/id codes are correct for all versions?

If you don't have both versions, then you should just add the information for the one you have - otherwise it's guessing and the problem is that when someone wants to add a code that might conflict, we have to work out what's wrong...

Also, I've seen in the past that sites like amazon use a photo for a similar unit - it doesn't mean that's necessarily exactly the same.

I would prefer to only add codes for devices that we know to avoid future confusion.

khendricks:
Should be all set.

cdjackson:
Thanks.

PR #2778, opened 2015-06-22
Added Leviton VRPD3-1LW plug-in dimmer

kabili207:
There doesn't seem to be any information about the command classes this module supports anywhere, so I just copied it from one of the existing leviton dimmers.


cdjackson:
Command classes aren't used in the XML anyway (unless they have additional information included), so it doesn't matter if it's correct...

PR #2780, opened 2015-06-22
onewire binding: fixed bug to make onewire:tempscale work

berndpfrommer:
fixed a bug in the onewire binding so the onewire:tempscale=FAHRENHEIT is parsed correctly.


berndpfrommer:
Another pull request fixing the same issue has been submitted with pull request #2787.
#2787 is a bit fancier since it throws a config exception as well so I recommend closing this PR in favor of #2787.

teichsta:
ok, thanks for the update - will follow #2787

PR #2782, opened 2015-06-22
bug fixes to make insteonplm binding more robust

berndpfrommer:
- retry modem database download if it fails
- much improved handling of corrupt data packets as received from the
  InsteonHub raw port
- discard spurious messages for the insteon thermostat to avoid
  outlier temperature and humidity values


teichsta:
Thanks @berndpfrommer!

PR #2784, opened 2015-06-22
Fatek PLC binding

slawekjaranowski:
Fatek PLC is one kind of programmable logic controller.

This binding support native Fatek protocol to communicate with PLC.


teichsta:
Hi @slawekjaranowski,

thanks for this contribution. Hope to find the time to review soon. Could you meanwhile check if this binding also works with OH2? You could also start creating the according Wiki page.

Thanks in advance, Thomas E.-E.

slawekjaranowski:
I have already prepared Readme file in binding directory. 
I hope it will be also good for Wiki page.

slawekjaranowski:
I checked binding with OH2.

It seems to work correctly, I didn't see any strange event i log file even with DEBUG level. All communications occurs in proper way.

I can read items data by rest.

teichsta:
great, thanks @slawekjaranowski for this update. Hope to start the review soon. Best, Thomas E.-E.

slawekjaranowski:
Should I do something more, to help you?
I rebased my PR to latest master version.

teichsta:
no, i am "just a bit" busy currently … thanks for your patience, Thomas E.-E.

zacofany:
Hi, is there any progress with this binding? I would very appreciate if you release it ;-) Cheers

slawekjaranowski:
I see that I should rebese my PR again. I will do it in few days.

@zacofany  if you would test this binding let me know, I can share binary addons from my build.

slawekjaranowski:
I have rebased my PR with the latest master branch.
I have also tested it with OpenHab version 1.8.1 and 2.0.2.b2 - looks ok.

I'm waiting for your review.

slawekjaranowski:
If you have question - you can ask on forum.
In post I put link to compiled binding - so you can ease start to testing.
https://community.openhab.org/t/fatek-binding/5645

zacofany:
Hi All,
I was testing this binding for 2 weeks on my system configuration:
BananaPi with OpenHAB 1.8.2 running on latest bananian (Debian port) with oracle java 8 (java version "1.8.0_77")
+Fatek FBs-20MC
+FBs-CM25E
+FBs-CB55
+8xFBs-24EY
+4xFBs-24EX
OpenHAB connected to Fatek via ethernet (TCP) as described in manual with configuration:
fatekplc:refresh=100 # (milliseconds)

It works fine with full provided functionality. I haven't found any issues in my log connected to this binding.
The only change is that I get is much more frequent warnings:
[WARN ] [osphere.cpr.DefaultBroadcaster] - Duplicate resource (...) Could be caused by a dead connection not detected by your server. Replacing the old one with the fresh one
However I believe this warning does not have anything to do with the binding itself.

Also I have to admit that performance of this binding is a lot better than modbus binding, so if anybody has Fatek PLC connected to OpenHAB then this is the best choice.
Next month, once I get my Pine64 i will test it on this platform with OpenHAB 2.
Thanks @slawekjaranowski for this binding. Cheers.

watou:
Still in my review queue, I haven't forgotten. 😄

slawekjaranowski:
Ok. I only rebase with the latest master version to avoid merge conflict.

I do it from time to time, because when I didn't do it for long time I can get many conflicts to resolve.

bundles/binding/org.openhab.binding.fatekplc/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.openhab.binding.fatekplc</name>
+	<comment>This is the plcfatek binding of the open Home Automation Bus (openHAB)</comment>
```
watou: fatekplc ?

bundles/binding/org.openhab.binding.fatekplc/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+If we start register name with + (plus sign) values read from plc will be recognized as unsigned number.
+
+If we end register name with "F" values read from plc will be recognized as float number.
+For float number 32 bits data register must be used.
+
+We can also specific factor value, all read data will be multiplied by this value,
+before write value from OpenHab will be divided by factor.
+
+For Number item we can provide transformation function, eg: JS('file.js')
```
watou: The use of single quotes is a different format than most other bindings seem to use.  Almost universally (it seems to me) the de facto notation is `JS(file.js)`.  Would it be possible to change this in code and doc?

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/FatekPLCBindingProvider.java
```diff
@@ -0,0 +1,28 @@
...
+ * Binding provider interface for Fatek PLC
+ * @author Slawomir Jaranowski
+ * @since 1.9.0
+ */
+public interface FatekPLCBindingProvider extends BindingProvider {
+
+	/**
+	 * Get Fatek PLC item by item name
+	 * @param name item name
+	 * @return Item
```
watou: `FatekPLCItem`

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
```diff
@@ -0,0 +1,278 @@
...
+	 *            <li>3 A configuration was changed
+	 *            <li>4 A configuration was deleted
+	 *            <li>5 The component was disposed
+	 *            <li>6 The bundle was stopped
+	 *            </ul>
+	 */
+	public void deactivate(final int reason) {
+
+		setProperlyConfigured(false);
+		logger.info("deactivate executed reason={}", reason);
```
watou: should be a debug or trace

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	@Override
+	public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+
+		logger.debug("update start");
+		try {
+			setProperlyConfigured(false);
```
watou: side effects are possible when toggling properlyConfigured; more happens than changing a boolean value (AbstractActiveService class).  Might be better to set it at the end of method with final answer?

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
```diff
@@ -0,0 +1,278 @@
...
+		} catch (NumberFormatException e) {
+			ret = defaultVale;
+			logger.warn("Property {} parse exception {}", key, e.getMessage());
+		}
+
+		return ret;
+	}
+
+	/**
+	 * @{inheritDoc}
```
watou: `{@inheritDoc}`

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
```diff
@@ -0,0 +1,278 @@
...
+					logger.error("update items error", e);
+				}
+			} else {
+				logger.warn("Unknown slave: {} to process update: {}", slaveName, items);
+			}
+		}
+	}
+
+	/**
+	 * @{inheritDoc}
```
watou: `{@inheritDoc}`

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
```diff
@@ -0,0 +1,278 @@
...
+					} catch (Exception e) {
+						logger.error(String.format("internalReceiveCommand(%s, %s)", itemName, command), e);
+					}
+				}
+			}
+		}
+	}
+
+	/**
+	 * @{inheritDoc}
```
watou: `{@inheritDoc}`

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
```diff
@@ -0,0 +1,278 @@
...
+	 */
+	@Override
+	protected void internalReceiveUpdate(String itemName, State newState) {
+		logger.debug("internalReceiveUpdate({},{}) is called!", itemName, newState);
+	}
+}
```
watou: Since the binding does not handle item state updates, this override can be removed.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCGenericBindingProvider.java
```diff
@@ -0,0 +1,64 @@
...
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public String getBindingType() {
+		return "fatekplc";
+	}
+
+	/**
+	 * @{inheritDoc}
```
watou: `{@inheritDoc}`

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
watou: In order for the binding to run properly under openHAB 2, please add these method signatures:
```
    protected void addBindingProvider(FatekPLCBindingProvider bindingProvider) {
        super.addBindingProvider(bindingProvider);
    }

    protected void removeBindingProvider(FatekPLCBindingProvider bindingProvider) {
        super.removeBindingProvider(bindingProvider);
    }
```

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCGenericBindingProvider.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void processBindingConfiguration(String context, Item item,
+			String bindingConfig) throws BindingConfigParseException {
+
+		super.processBindingConfiguration(context, item, bindingConfig);
```
watou: Move to end of method, in case your parsing throws bcpe.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
```diff
@@ -0,0 +1,162 @@
...
+					logger.info("New Fatek PLC connectionUri={}", value);
+				} catch (Exception e) {
+					fatekPLC = null;
+					throw new ConfigurationException(name, e.getMessage());
+				}
+			} else {
+				throw new ConfigurationException(name, "Unknow property");
+			}
+		} catch (ConfigurationException e) {
+			logger.error("", e);
```
watou: Would some message be a little helpful here? 😄

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
```diff
@@ -0,0 +1,162 @@
...
+				item = itemRegistry.getItem(fatekItem.getItemName());
+				currentState = item.getState();
+			} catch (ItemNotFoundException e) {
+				logger.warn("", e);
+				currentState = null;
+				item = null;
+			}
+
+			logger.debug("new state: {}, current state: {}, item: {}", newState, currentState, item);
+			if (!newState.equals(currentState)) {
```
watou: I believe it is unwise to suppress item state updates by inspecting the item's current state—bindings should be as stateless as possible.  Do you have extraordinary circumstances with Fatek PLC that justifies suppressing repeated updates of the same state?

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
```diff
@@ -0,0 +1,162 @@
...
+	public void command(FatekPLCItem config, Command command) {
+
+		logger.debug("Command config={}, command={} {}", config, command, command.getClass());
+
+		try {
+			synchronized(fatekPLC) {
+				config.command(fatekPLC, command);
+			}
+		} catch (CommandException e) {
+			logger.error("", e);
```
watou: Would some message be a little helpful here? 😄

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekColorItem.java
```diff
@@ -0,0 +1,280 @@
...
+	public FatekColorItem(Item item, List<String> confItems)
+			throws BindingConfigParseException {
+
+		super(item, confItems);
+
+		step = getParamsFromConfAsInt(confItems, "step", 5);
+
+		if (confItems.size() < 3 || confItems.size() > 5) {
+			throw new BindingConfigParseException(
+					"Incorect binding for color item- should be: plcName:R1:R2:R3[:R4][:RGB|HSB]");
```
watou: "Incorrect"

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekContacSwitchtItem.java
```diff
@@ -0,0 +1,60 @@
...
+import com.github.s4u.jfatek.registers.UnknownRegNameException;
+
+/**
+ * Common implementation for Contact and Switch items.
+ *
+ * @author Slawomir Jaranowski
+ * @since 1.9.0
+ */
+public abstract class FatekContacSwitchtItem extends FatekPLCItem {
+
```
watou: `FatekContactSwitchtItem`, with a "t"

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekContacSwitchtItem.java
```diff
@@ -0,0 +1,60 @@
...
+	protected boolean negate1;
+
+	FatekContacSwitchtItem(Item item, List<String> confItems) throws BindingConfigParseException {
+
+		super(item, confItems);
+
+		if (confItems.size() < 1 || confItems.size() > 2) {
+			throw new BindingConfigParseException("Incorect binding for contact, switch item");
```
watou: "Incorrect"

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekContactItem.java
```diff
@@ -0,0 +1,54 @@
...
+import com.github.s4u.jfatek.registers.Reg;
+import com.github.s4u.jfatek.registers.RegValue;
+
+/**
+ * Contact item implementation.
+ *
+ * @author Slawomir Jaranowski
+ * @since 1.9.0
+ */
+public class FatekContactItem extends FatekContacSwitchtItem {
```
watou: They inheritance hierarchy seems reversed, based on the class names.  The subclass usually has the more specific-sounding name.  Could the classes be renamed in such a way as to reflect this, in addition to their true roles?

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekContactItem.java
```diff
@@ -0,0 +1,54 @@
...
+
+	FatekContactItem(Item item, List<String> confItems) throws BindingConfigParseException {
+		super(item, confItems);
+
+		if (confItems.size() != 1) {
+			throw new BindingConfigParseException("Incorect number of register for contact binding");
```
watou: "Incorrect"

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekDimmerItem.java
```diff
@@ -0,0 +1,128 @@
...
+	private int step;
+
+	public FatekDimmerItem(Item item, List<String> confItems) throws BindingConfigParseException {
+
+		super(item, confItems);
+
+		step = getParamsFromConfAsInt(confItems, "step", 5);
+
+		if (confItems.size() != 1) {
+			throw new BindingConfigParseException("Incorect binding for dimmer item");
```
watou: "Incorrect" (please search/replace any others)

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekRollershutterItem.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Rollershutter item implementation.
+ *
+ * @author Slawomir Jaranowski
+ * @since 1.9.0
+ */
+public class FatekRollershutterItem extends FatekPercentItem {
+
+	// private static final Logger logger = LoggerFactory.getLogger(RollershutterItemConfig.class);
```
watou: Remove commented code.

bundles/binding/org.openhab.binding.fatekplc/src/main/resources/jfatek/SERIALFatekConnectionFactory
```diff
@@ -0,0 +1 @@
+org.openhab.binding.fatekplc.serial.SerialFatekConnectionFactory
```
watou: How does this work?  Should the file name have some helpful extension?

distribution/openhabhome/configurations/openhab_default.cfg
```diff
@@ -2450,3 +2450,12 @@ tcp:refreshinterval=250
...
 #powermax:autoSyncTime=false
+
+################################### Fatek PLC Binding #######################################
+#
+# the refresh interval in milliseconds which is used to poll values from the Fatek PLC
+# (optional, defaults to 60000ms / 60 seconds)
+# fatekplc:refresh=60000
+#
+# fatek connection uri, support tcp and udp
+# fatekplc:plc1.connectionUri=udp://192.168.1.100?plcId=1
```
watou: Please add newline at end of file

features/openhab-addons-external/src/main/resources/conf/fatekplc.cfg
```diff
@@ -0,0 +1,8 @@
+################################### Fatek PLC Binding #######################################
+#
+# the refresh interval in milliseconds which is used to poll values from the Fatek PLC
+# (optional, defaults to 60000ms / 60 seconds)
+# fatekplc:refresh=60000
+#
+# fatek connection uri, support tcp and udp
+# fatekplc:plc1.connectionUri=udp://192.168.1.100?plcId=1
```
watou: Please add newline at end of file

features/openhab-addons/src/main/feature/feature.xml
```diff
@@ -175,6 +175,13 @@
         <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.exec/${project.version}</bundle>
     </feature>
 
+    <feature name="openhab-bindign-fatekplc" description="Fatek PLC Binding" version="${project.version}">
+        <feature>openhab-runtime-base</feature>
+        <feature>openhab-runtime-compat1x</feature>
```
watou: Does this/will this binding depend on `openhab-transport-serial`?

watou:
Very nice work, Slawomir @slawekjaranowski.  Please see my comments inline.  Thanks for your extreme patience!

bundles/binding/org.openhab.binding.fatekplc/.project
slawekjaranowski: should be fatekplc

bundles/binding/org.openhab.binding.fatekplc/README.md
slawekjaranowski: Wrong example in doc only

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/FatekPLCBindingProvider.java
slawekjaranowski: should be `FatekPLCItem`

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
slawekjaranowski: changed to debug

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
slawekjaranowski: Ok, I remove this.
At end of methods if no exceptions occurs `setProperlyConfigured(true);` is called.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
slawekjaranowski: Binding skeleton was generated by maven archetype as I remember.
So this was wrong in code of archetype, we can find many such misspell in other bindings.
Of course I will correct all occurrence in my code.

If you want I can prepare new PR which correct this in other bindings.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
slawekjaranowski: Why protected? 
In super class `AbstractActiveBinding` this methods are public.

I added according to your comment.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCGenericBindingProvider.java
slawekjaranowski: ok.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
slawekjaranowski: This catch is unnecessary, I removed it.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
slawekjaranowski: I tried to minimize traffic on openhub bus. I removed this checking. 
All states will be published to bus.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
slawekjaranowski: ok.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/items/FatekContactItem.java
slawekjaranowski: Class `FateContactSwitchItem` was implemented as common code for Contact and Switch item.
Now I see that maybe this class is not necessary.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCBinding.java
watou: FTR: https://github.com/openhab/openhab/pull/3861 is about these method signatures.

bundles/binding/org.openhab.binding.fatekplc/src/main/resources/jfatek/SERIALFatekConnectionFactory
slawekjaranowski: This resource contains name of class which is responsible for implementing serial transport for JFatek library.

As you see Serial transort for this binding is not implemeted yet.
I only add skeleton class which must be completed to support this trasport.

features/openhab-addons/src/main/feature/feature.xml
slawekjaranowski: Not yet.

`SerialFatekConnectionFactory` is only skeleton, which must be implemented and then depend on `openhab-transport-serial` should be add.

slawekjaranowski:
Thanks @watou , I changed my code according to your comments.

Now I will test if everything is ok and then I push new version in a few days.

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
slawekjaranowski: Maybe to detect current state of item and publish to bus only changed items is not bad idea.

https://community.openhab.org/t/modbus-tcp-polling-generates-events-even-if-there-is-no-state-change/6708

ssalonen: Please note the discussion about this here https://github.com/openhab/openhab/pull/4291#issuecomment-210759837. 

There is motivation also to publish unchanged values in order to detect issues with the connection in the rules.

Also : @watou's suggestion of changing rule trigger might do the trick also.

In any case, would be nice that general consensus is reached on the topic, and decision would be documented in the developer guides.

watou: Modbus is one of the rare cases where the binding does have to poll the hardware at such a high frequency that posting so many updates to the event bus would lead to a performance problem.  If a binding can post repeated updates without unacceptably hurting system performance, it's preferable to do so, so the responsibility for listening for changed states falls on the listeners.  I have not come to an opinion if I think the FatekPLC binding falls into the same category as the Modbus binding, but the default situation for bindings is to be stateless with respect to items, post all updates, and leave it to the listeners to listen only for changes.  Do you think FatekPLC meets these special circumstances like Modbus does, @slawekjaranowski @ssalonen, and if so, how?

watou:
Thanks @slawekjaranowski, comment here when it's ready for a final pass!

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
slawekjaranowski: I think, that Fatek PLC is also case where users want to poll items many times per second. Please look at @zacofany coment about testing, he sets refresh rete to 100 ms.

@zacofany  how many items do you poll in your configurations.

Now I changed code to publish all events, I have last version, so I can compare performance and system utilization in both case.

watou: If it would be common to post an update for items ten times per second unless "no change" updates were filtered out, then yes, you would probably want to avoid that!  But don't let me discourage you from pursuing real data on which approach is best! 😄

ssalonen: Excellent to see the performance comparison! Would really appreciate if results are commented in the modbus PR as well.

slawekjaranowski:
Below there is table where I compare system utilization according to publish all or only changed events.

My system:
- OH 1.8.0
- Raspberry PI model B - 512 MB
- Refresh = 100
- Number of updated items from PLC = 110

Events | Number of clients | cpu usage | loadavg
---------| --------------------- | ------------- | ----------
all       |                      1     |          40% |   0,9
all       |                      0     |          30% |   0,7
changed       |            1     |          20% |   0,5
changed       |            0     |          16% |   0,4

When we publish all change I can see that client refresh your page very often, which also make more network traffic and more system utilization.
Refreshing page without change is not comfortable especially when we have charts on page. 

But when we publish only changed items - client waits to real change.

So I want to keep code responsible for testing if item state was changed.

slawekjaranowski:
@watou new version of code is pushed and ready for a final pass

ssalonen:
Excellent performance comparison @slawekjaranowski! I think that indeed proves that the change detection in the binding makes sense.

Only one comment that I [posted to modbus binding PR #4291](https://github.com/openhab/openhab/pull/4291#issuecomment-210759837) on this same topic, quoting here for your convenience :

> In my use case I'm actually interested having explicit events whenever read is successfull. This allows rules to react to broken connections (e.g. "if no modbus value received in 5 minutes -> send sms") and communication faults. I'm using this for z-wave wireless sensors but why not have the same functionality here also?

So coming from this perspective would it make sense to make the binding change detection configurable? 


Thanks
Sami

bundles/binding/org.openhab.binding.fatekplc/src/main/java/org/openhab/binding/fatekplc/internal/FatekPLCSlave.java
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		// distribute read data
+		for (FatekPLCItem fatekItem : items) {
+
+			State newState = fatekItem.getState(response);
+			State currentState;
+			Item item;
+			try {
+				item = itemRegistry.getItem(fatekItem.getItemName());
```
ssalonen: I believe that ItemRegistry can be null, at least in theory. Perhaps that should be checked separately to avoid null pointer exception?

watou: Looking at the 1..1 cardinality of the service [here](https://github.com/slawekjaranowski/openhab/blob/fatekplc/bundles/binding/org.openhab.binding.fatekplc/OSGI-INF/binding.xml#L31), it looks to me like itemRegistry cannot be `null` since the dependent service is mandatory (cardinality starts with 1, not 0).  Did I miss something, @ssalonen ?

ssalonen: my bad, thank you for correcting me!

watou:
I will wait for @slawekjaranowski to respond to @ssalonen's [comment](https://github.com/openhab/openhab/pull/2784#issuecomment-214965701) about the potential need to mitigate the loss of "received update" triggers that is caused by suppressing duplicate updates, as the code currently stands.  Please ping me after this issue is considered and any further commits are made and tested, and I will finish the review, towards merging this PR.

watou:
I will also add [this quote from Kai](https://community.openhab.org/t/openhab2-new-issue-with-the-startup-process-mios-binding/8320/18?u=watou) regarding the item registry:

> It was never (officially) allowed to use the ItemRegistry within a binding.

Please read the whole story about that.  In other words, there may be a problem using the item registry, but also wanting the binding to work on OH2, at least right now.

slawekjaranowski:
@ssalonen, @watou The first - the loss of "received update" triggers can be easy mitigate by adding some of special register to items which change state evry time like RTC (seconds, minutes). Fatek PLC also has special register (R4151) which is incremented every 1ms when PLC is working.

slawekjaranowski:
As we see prevent to publish the same value of items has a big impact to system utilization and other hand as you say binding should be stateless and using ItemRegistry is not allowed (officially)  in binding.

@watou, @kaikreuzer What do you suggest in this situation?

watou:
@slawekjaranowski if the only way to be stateless w.r.t. items is to post an unacceptably high volume of updates onto the event bus, then I suppose the binding can't be stateless.  And if working correctly under OH2 is in doubt because of the binding's use of the item registry isn't officially supported and shows issues in other bindings, then using the item registry to track previously known states is also to be avoided, which would leave you with managing your own item state cache (or similar) in the binding code.

However, I would wait for @kaikreuzer's recommendation before heading down any new path.

slawekjaranowski:
@kaikreuzer how is your recommendation?

kaikreuzer:
Sorry for the very late reply!

> which would leave you with managing your own item state cache (or similar) in the binding code.

Yes, this is also my suggestion. Use of item registry is indeed something that shouldn't be done, while the "stateless" behavior is rather a recommendation, but not a must. There are often situations, where bindings need to track what they were sending before to make sure that they know when and what to send next.

slawekjaranowski:
In last commit I removed dependency to item registry. I repleced it by own cache.

zacofany:
@kaikreuzer & @watou is it ready to be added as officiall binding to OH? Can I help with anything as binding user?

watou:
I gave it another quick review and it looks good.  Consider submitting a new PR to add a few lines here:

https://github.com/openhab/openhab-docs/blob/gh-pages/_data/oh1addons_infos.yml#L92

Thanks!

slawekjaranowski:
I am very pleasure that you merge my PR.
In next few days I will add wiki page and put link in documentation.

Thanks.

watou:
> I am very pleasure that you merge my PR.

You're very welcome, and thank you for your contribution and patience.  Sorry for the delay.

> In next few days I will add wiki page and put link in documentation.

Excellent!  Thanks.

PR #2788, opened 2015-06-24
Improve BenQ projector binding connection handling and exception handling

cyclingengineer:
These commits contain 3 improvements:

1) Tidy up of string handling so that non-numeric strings in the case where a numeric string is expected is handled properly

2) Improve the connection and reconnection process in the case where the host is not available or goes down. This also has a fix to avoid null pointer exceptions in the closing of reader/writer/sockets.

3) Delay reading the power status after a power command is sent to the projector. This stops you getting an item toggle when you send a power off request and then read the status when the projector will return 'ON' as it is cooling down. This means that rules tied to the power status will behave correctly (before my projector screen would start going up, then go back down and then finally go up again - now it will just go up).

I am testing locally. Would be good to get feedback from users who have had similar problems that have been reported in the forum. See https://groups.google.com/d/msg/openhab/KboLzufnsnc/NGBE3SGaNPgJ

### DIFF
--- a/bundles/binding/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/BenqProjectorBinding.java
+++ b/bundles/binding/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/BenqProjectorBinding.java
@@ -161,7 +161,11 @@ public class BenqProjectorBinding extends
 
 			String deviceIdString = (String) config.get("deviceId");
 			if (StringUtils.isNotBlank(deviceIdString)) {
-				setProperlyConfigured(transport.setupConnection(deviceIdString));
+				transport.setupConnection(deviceIdString);
+				/* we always want the refresh service to start, it will attempt
+				 * to reconnect next time if necessary
+				 */				
+				setProperlyConfigured(true); 
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


teichsta:
I've also commented on the group. it would be great if you could find testers and let the comment here. Please also test the binding against openHAB2 and note your findings here (see https://github.com/openhab/openhab2/blob/master/docs/sources/addons.md#compatible-1x-add-ons).

Thanks in advance, Thomas E.-E.

teichsta:
btw: please format the code using the standard Eclipse Code Formatter.

cyclingengineer:
I'll follow up on those actions as soon as I can.

Thanks for the feedback.
Paul

On 29 June 2015 at 21:22, Thomas Eichstädt-Engelen <notifications@github.com

teichsta:
any news @cyclingengineer?

teichsta:
ping @cyclingengineer :-) ?

plajjan:
I could potentially have a look at this as I have a BenQ projector, although I'm not using the TCP/IP to RS232 converter but I'll have a read through the code to see if it's applicable to direct serial as well.

Are there binaries built for pull requests, like on CloudBees or do I need to compile this myself?

teichsta:
:+1:  that would be great! No binaries available, so you would have to compile them yourself …

teichsta:
move this PR to 1.9

9037568:
Is @cyclingengineer still planning to finish this PR?

PR #2790, opened 2015-06-24
added outputmode 51 for led dimmers like GE-UMV200

georgwiltschek:
Small change to support LED dimmers which have another output mode on Digitalstrom than "normal" dimmers.


teichsta:
Thanks @georgwiltschek!

Issue #2474, opened 2015-04-17
Slider arrows do not react on short/normal clicks in Classic UI


dominicdesu:
The slider is represented by an up and a down arrow in the Classic UI.
When one of these arrows is pressed and the mouse button is hold, the value of the slider changes until the mouse button is released again. This works fine.
However, when one of these buttons is clicked for a short time only, nothing happens. I would expect that the value of the slider would change little by little for each click on one of the arrows.

Note that I am talking about the slider element with switchSupport deactivated (default value).

Is the slider on purpose designed to react only on long clicks or is this a bug?

kaikreuzer:
> Is the slider on purpose designed to react only on long clicks or is this a bug?

Well, I wouldn't call it a bug, but I agree that it would be more user friendly if the repeating sending of INCREASE/DECREASE would start right on the first moment of pressing the button.

Nonetheless, I would suggest to close this issue as https://github.com/openhab/openhab/pull/2792 will bring a proper slider.

PR #2792, opened 2015-06-24
Implement slider widget for WebUI

resetnow:
This commit implements slider widget in Classic UI. 

Changes overview:
* `slider.html`: the snippet was modified and now contains slider elements along with the initialization parameters
* `Logic.js`: contains slider code. Added `sliderchange` callback.
* `main.html`: processes value change events and sends requests to OpenHAB.
* `Render.css`: slider stylesheet

Tested with recent Firefox and Chrome versions, Android 5.1 stock browser and Chrome for Android. I don't have any Apple devices, so it would be great if someone could test it with iphone/ipad.

Here is how it looks like: [Desktop](http://i.imgur.com/Lu9Li0u.png), [Android](http://i.imgur.com/dVNEdQe.png).

References:  #2474


resetnow:
Regarding `Logic.min.js

teichsta:
Thanks @resetnow, this looks pretty nice already! @sja could you please have a look at this one?

udo1toni:
still work in progress?

teichsta:
good question … @sja?

sja:
@resetnow You're right, its not so nice to check .min.js in, but here we don't have any infrastructure to generate this on build time. So I chose the simple way.
For OH2, we use the frontend-maven plugin to dynamically set up the nodejs/grunt/bower stack.

@udo1toni Sorry for the delay, thanks for the reminder. I was getting confused by the other issues around ClassicUI. Will answer this PR today.

bundles/ui/org.openhab.ui.webapp/web/WebApp/Action/Logic.js
```diff
@@ -1713,7 +1731,151 @@ var WebApp = (function() {
...
+		_t.opt = item.parentNode;
+		_t.progress = item.querySelector(".iSliderProgress");
+		_t.handle = item.querySelector(".iSlider");
+		_t.valueEl = _t.opt.parentNode.querySelectorAll("span")[4];
+		_t.max = 100;
+		_t.min = 0;
+		_t.dragState = undefined;
+		_t.interval = undefined;
+		_t.computedOffset = 
+			parseInt(_style.getPropertyValue("border-top-left-radius")) * 2 +
```
sja: Style: Missing radix

bundles/ui/org.openhab.ui.webapp/web/WebApp/Action/Logic.js
```diff
@@ -1713,7 +1731,151 @@ var WebApp = (function() {
 			}
 		}
 	}
-
+	
+	function Slider(item) {
+		var _t = this;
```
sja: Style: Well, yes, this scripts var names look like so, but wouldn't it be nice to start new functions with more expressive var names? Especially when it's getting minified, it wouldn't make a difference.

sja:
Tested and looks good to me. Thanks to @resetnow for this contribution. I think that will help new users a lot to understand that there is now a real slider.

bundles/ui/org.openhab.ui.webapp/web/WebApp/Action/Logic.js
resetnow: I just wanted something to refer to instance of Slider in nested scopes. Sometimes `that` is used as a variable name. Should I change it to `that` or something other?

sja: Maybe `that`. The underscore confused me at first sight. `(function(a, t)...` is another example. And some comments for functions would be nice. But you're right, the rest of `Slider` is much more readable that the original Logic.js.

resetnow:
The problems mentioned here were fixed.

resetnow:
If there aren't any other problems, I suppose it can be merged then?

resetnow:
Shame on me! I must have been testing a cached version...

resetnow:
Just a note: previously mentioned issues are fixed and it works...

teichsta:
Thanks for the update, guys! Since you have implemented the Basic-UI for ESH meanwhile i assume this PR won't make it into ESH anymore, right?

teichsta:
Thanks @resetnow!

resetnow:
>Since you have implemented the Basic-UI for ESH meanwhile i assume this PR won't make it into ESH anymore, right?

Yes, unless ESH maintainers would like to see this ported. Personally I think it's good as it is, there is a new ui anyway — although some fixes still need to be ported to ESH (like colorpicker constant update).

kaikreuzer:
> Yes, unless ESH maintainers would like to see this ported. Personally I think it's good as it is, there is a new ui anyway

No! This would mean a regression once openHAB 1 users switch to openHAB 2. The Classic UI will continue to exist for the moment, so all fixes must be ported as well!

resetnow:
@kaikreuzer ok, I'll port it then, probably this weekend.

resetnow:
@kaikreuzer, a question: ClassicUI in Eclipse Smarthome contains a minified Logic.js version. Openhab, however, has unminified original version introduced in https://github.com/openhab/openhab/commit/a5a77baf358a01459668af9e69cf3a4394d12b55. Should the uncompressed version (and minify script) be ported by @sja or can I just copy Logic.js from Openhab?

kaikreuzer:
Probably a question that @sja should answer - I wasn't actually aware that we have minified in ESH and non-minified in openHAB. I would prefer if it is done in the same way...

resetnow:
@sja so can you please copy this commit to ESH or do you mind if I copy it?

sja:
The latest and greatest was done for OH1, so yes, there is a gap.

On adding the minifiy script, I also did a small change in Logic.js. The minified version Logic.min.js was there from the beginning. I added the unminified version to be able to make changes and minify it again. Because I don't know the current state of the OH2 Logic(.min).js. I will sync them.

PR #2755, opened 2015-06-18
Added Ecobee action bundle.

watou:
The new bundle adds ten new actions for use with the Ecobee binding.

Parameters in _italics_ are optional, in which case each unused parameter must be replaced with `null`.  Read the [API documentation](https://www.ecobee.com/home/developer/api/documentation/v1/functions/using-functions.shtml) to be sure you know the rules for calling these functions.

* ecobeeAcknowledge(Item item, String thermostatIdentifier, String ackRef, String ackType, _Boolean remindMeLater_) - Acknowledge an alert.
* ecobeeControlPlug(Item item, String plugName, String plugState, _Date startDateTime_, _Date endDateTime_, _String holdType_, _Integer holdHours_) - Control the on/off state of a plug by setting a hold on the plug.
* ecobeeCreateVacation(Item item, String name, DecimalType coolHoldTemp, DecimalType heatHoldTemp, _Date startDateTime_, _Date endDateTime_, _String fan_, _Integer fanMinOnTime_) - Create a vacation event on the thermostat.
* ecobeeDeleteVacation(Item item, String name) - Delete a vacation event from a thermostat.
* ecobeeResetPreferences(Item item) - Set all user configurable settings back to the factory default values.
* ecobeeResumeProgram(Item item, _Boolean resumeAll_) - Remove the currently running event providing the event is not a mandatory demand response event.
* ecobeeSendMessage(Item item, String text) - Send an alert message to the thermostat.
* ecobeeSetHold(Item item, _DecimalType coolHoldTemp_, _DecimalType heatHoldTemp_, _String holdClimateRef_, _Date startDateTime_, _Date endDateTime_, _String holdType_, _Integer holdHours_) - Set the thermostat into a hold with the specified temperature.
* ecobeeSetOccupied(Item item, Boolean occupied, _Date startDateTime_, _Date endDateTime_, _String holdType_, _Integer holdHours_) - Switches a (EMS model only) thermostat from occupied mode to unoccupied, or vice versa.
* ecobeeUpdateSensor(Item item, String name, String deviceId, String sensorId) - Update the name of an ecobee3 remote sensor.

PR #2793, opened 2015-06-24
Ecobee binding bug: BigDecimal needs precision/rounding when converting to/from Celsius

watou:
Avert ArithmeticExceptions for non-terminating decimal expansion when converting to/from Celsius.  I imagine very few Ecobee binding users are using Celsius, but for those that are, this fix would be required.

### DIFF
--- a/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/messages/Temperature.java
+++ b/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/messages/Temperature.java
@@ -9,6 +9,7 @@
 package org.openhab.binding.ecobee.internal.messages;
 
 import java.math.BigDecimal;
+import java.math.MathContext;
 
 import org.codehaus.jackson.annotate.JsonCreator;
 import org.codehaus.jackson.annotate.JsonValue;
@@ -132,7 +133,7 @@ public class Temperature {
 	 *            the Celsius temperature
 	 */
 	public static Temperature fromCelsius(BigDecimal celsius) {
-		return new Temperature(celsius.multiply(NINE).divide(FIVE).add(THIRTY_TWO).movePointRight(1));
+		return new Temperature(celsius.multiply(NINE).divide(FIVE, MathContext.DECIMAL32).add(THIRTY_TWO).movePointRight(1));
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


watou:
[Initial bug report and report of fix working.](http://developer.ecobee.com/api/topics/any-chance-of-seeing-room-temperature-from-smartsi-via-mqtt#reply_15826858)  Let's wait to hear one final report in a day or two before you would consider merging, Thomas.  And thanks.

watou:
User reports that this fix is solid.  If you don't disagree, can this be included in 1.7.1, @teichsta ?  (This same change is included in the [PR that adds the action bundle](https://github.com/openhab/openhab/pull/2755).)  Thank you, Thomas!

bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/messages/Temperature.java
```diff
@@ -132,7 +133,7 @@ public static Temperature fromFahrenheit(BigDecimal fahrenheit) {
 	 *            the Celsius temperature
 	 */
 	public static Temperature fromCelsius(BigDecimal celsius) {
-		return new Temperature(celsius.multiply(NINE).divide(FIVE).add(THIRTY_TWO).movePointRight(1));
+		return new Temperature(celsius.multiply(NINE).divide(FIVE, MathContext.DECIMAL32).add(THIRTY_TWO).movePointRight(1));
```
robnielsen: Why not just multiply by 1.8 instead of multiplying by 9 then dividing by 5

watou: That would make more sense.  I used some BigDecimal F-to-C example from somewhere that does it this way, and most conversion examples in a Google search seem to prefer 9/5 (probably because they are explaining how to convert in your head).  A single multiply would virtually certainly be faster, since there is one fewer intermediate objects created.  But since I have a user who has already signed off on the current code, I will leave it unless Thomas would like me to do it differently.  Thanks, Rob, for the comment!

teichsta: i am fine with this one for 1.7.1. You could probably change the code to 1.8 in #2755?

teichsta:
Thanks @watou!

watou:
Will simply conversion in #2755 as suggested.  Thanks @teichsta.

Issue #2789, opened 2015-06-24
Freebox binding not working well when in bridge mode


lolodomo:
When the Freebox Revolution is in bridge mode, the binding does not retrieve information and does not update items.
The reason is that UPnP AV and AirMedia settings are not available when the Freebox is in bridge mode.

lolodomo:
Fixed by PR #2796.

PR #2796, opened 2015-06-24
Fix for Freebox in bridge mode (issue 2789)

lolodomo:
Please find a fix for the issue #2789.


teichsta:
Thanks for this fix @lolodomo. Are there any users that can confirm your fix is working?

lolodomo:
There was exchanges with Mickael on the "forum", the guy who discovered the problem. I provided an intermediate version to clearly identify where was the problem and this version already fixed his problem.
This PR is not exactly this intermediate version but I am confident (100%) that this PR fix the issue.

teichsta:
Thanks for the update, @lolodomo!

> There was exchanges with Mickael on the "forum", the guy who discovered the problem.
ok, could you please add the link to this discussion here?

Since this fix is quite huge i tend to not cherry-pick it into 1.7. What do you think?

lolodomo:
Thomas, here is the link to the discussion: https://groups.google.com/d/msg/openhab/mJSFoTPZyz8/uWp1IyC6uWgJ

The change is in fact not huge at all even if the diff looks big. What I have changed is mainly cutting a big part of code surrounding by a unique try/catch into several small parts each one surrounding by its try/catch.
In my opinion, there is no risk to simply merge.

teichsta:
Thanks @lolodomo!

PR #2800, opened 2015-06-26
allow using PersistenceExtensions by using object "pe".

smerschjohann:
I've added the ability to use the PersistenceExtensions in jsr223.

In addition, this pull request adds jsr223engine.xml to build.properties and hopefully fixes the automatically built jar file.

### DIFF
--- a/bundles/core/org.openhab.core.jsr223/build.properties
+++ b/bundles/core/org.openhab.core.jsr223/build.properties
@@ -2,3 +2,4 @@ source.. = src/,
 bin.includes = META-INF/,\
                .,\
                OSGI-INF,\
+               OSGI-INF/jsr223engine.xml,\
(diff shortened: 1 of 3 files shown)


lewie:
Oh, same Feature and more I extended here: https://github.com/openhab/openhab/pull/2802
Regards
Helmut

teichsta:
I tend to merge this PR (once my little feedback has been fixed) since it is smaller and hence gives more transparency. Would you agree, too?

bundles/core/org.openhab.core.jsr223/build.properties
```diff
@@ -2,3 +2,4 @@ source.. = src/,
 bin.includes = META-INF/,\
                .,\
                OSGI-INF,\
+               OSGI-INF/jsr223engine.xml,\
```
teichsta: please add '/' to OSGI-INF (line before) rather than referencing each file separately

smerschjohann:
Well I think I would prefer the other merge as it allows another script engine (nashorn) as well. Nashorn is directly integrated in Java 8, so it has some benefit.

But what we could do: split the merges into fix and enhancement, otherwise simply merge lewies enhancements.

Suggestion: close, in favor for #2802

teichsta:
superseded by #2802

Issue #2804, opened 2015-06-26
PostgreSQL: Initial new PostgreSQL Persistence Service.


lewie:
I am using it since OpenHab 1.4, this Version I have extended now for 1.8.
Its rewritten and extended originally from the MySQL-Service.
Main Differences:
- Tablenames are more readable
- Tablenames can be reconfigured
- Can work with original Databasemodel of MySQL- Service

Advantage over jba-Persistence-Binding:
Every Item gets its own Table with matching data type.

```
############################ PostgreSQL Persistence Service ##################################
#
# the database url like 'jdbc:postgresql://<host>:<port>/<user>'
# postgresql:url=jdbc:postgresql://192.168.0.1:5432/ohDatabase

# the database user
#postgresql:user=

# the database password
#postgresql:password=

# I T E M   O P E R A T I O N S
# optional tweaking postgresql datatypes
# see: http://www.postgresql.org/docs/9.3/static/datatype.html
# defaults:
#postgresql:sqltype.CALL=			VARCHAR(200)
#postgresql:sqltype.COLOR=			VARCHAR(70)
#postgresql:sqltype.CONTACT=		VARCHAR(6)
#postgresql:sqltype.DATETIME=		TIMESTAMP
#postgresql:sqltype.DIMMER=			SMALLINT
#postgresql:sqltype.LOCATION=		VARCHAR(30)
#postgresql:sqltype.NUMBER=			DOUBLE PRECISION
#postgresql:sqltype.ROLLERSHUTTER=	SMALLINT
#postgresql:sqltype.STRING=			VARCHAR(65500)
#postgresql:sqltype.SWITCH=			VARCHAR(3)

# For Itemtype "Number" default decimal digit count (optional, default: 3) 
#postgresql:numberDecimalcount=

# T A B L E   O P E R A T I O N S
# Tablename Prefix generation, using Item real names or "item" (optional, default: false -> "item") 
#postgresql:tableUseRealItemNames=

# Tablename Suffix length (optional, default: 4 -> 0001-9999) 
#postgresql:tableIdDigitCount=

# Rename existing Tables using tableUseRealItemNames and tableIdDigitCount (optional, default: false) 
# USE WITH CARE! Deactivate after Renaming is done!
#postgresql:rebuildTableNames=
```

Regards 
Helmut

hakan42:
[...]

PR #2805, opened 2015-06-26
 PostgreSQL: Initial new PostgreSQL Persistence Service. #2804 

lewie:


bundles/persistence/org.openhab.persistence.postgresql/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,23 @@
...
+Import-Package: org.apache.commons.collections.iterators,
+ org.apache.commons.lang,
+ org.openhab.core.autoupdate,
+ org.openhab.core.binding,
+ org.openhab.core.items,
+ org.openhab.core.library.items,
+ org.openhab.core.library.types,
+ org.openhab.core.persistence,
+ org.openhab.core.types,
+ org.openhab.io.net.http,
```
teichsta: is that import really used in your code?

bundles/persistence/org.openhab.persistence.postgresql/java/org/openhab/persistence/postgresql/internal/PostgresqlItem.java
```diff
@@ -0,0 +1,49 @@
...
+	public Date getTimestamp() {
+		return timestamp;
+	}
+
+	@Override
+	public String toString() {
+		return DateFormat.getDateTimeInstance().format(timestamp) + ": " + name + " -> "+ state.toString();
+	}
+
+}
```
teichsta: add new line

bundles/persistence/org.openhab.persistence.postgresql/java/org/openhab/persistence/postgresql/internal/PostgresqlPersistenceService.java
```diff
@@ -0,0 +1,976 @@
...
+
+			if (rowId == 0) {
+				throw new SQLException(getName()+": Creating table for item '{}' failed.", itemName);
+			}
+
+			// Create the table name with real Item Names
+			//String pre = itemName.replaceAll(ITEM_NAME_PATTERN, ""); 
+			//tableName = pre + rowId;
+			// Create the table name
+			//tableName = getTableNamePrefix(itemName) + formatRight(rowId, tableIdDigitCount);
```
teichsta: remove commented code

teichsta:
Hi @lewie,

i had a quick look through the code. Please find my comments inline and some general comments below:

* move code to src/main/java
* add license header to each file (or update to 2015) - use ‘''mvn -Dyear=2015 license:format’'' to accomplish that easily
* finally squash your commits into one

Since this binding seem to be a merely a copy of the mysql binding this leads to tremendous code duplication which i am not very keen on. Have you had a look into the JPA binding? Would that work for you as well?

Best, Thomas E.-E.

lewie:
Absolutely, I agree! But JPA is saving Item values all in one Table as Strings, for me this is not a good solution (limited scalability).

As you can see, I have build in a Function for Table renaming (Prefix as Itemnames or simple "item", Suffix as ID with zeros filled).
Thought about to extend this idea for JPA, so that it can work with to Database models:
**Model1:** one Table with Strings 
OR 
**Model2:** every Item a Table with typical datatypes for each Item (only for SQL-Drivers).

In short, this would be a Merge of Mysql-, PostgreSql-, JPA-Service to one Extended-JPA-Service.

But it takes Time, we all don't have, isn't it.
I would like to implement, but rather facing to OpenHab2.

Rgds
Helmut

lewie:
OK, STOP! 
I hate code duplication too much.
Will do the light solution:
**Model3:** Merging Mysql- and PostgreSql-Service for now and OpenHab1.

What do you think of this solution?

What would be a good name:
 "MyPostreSQL"?

Rgds
Helmut

watou:
Just to add one voice as a user of MySQL persistence: I would _not_ like to see the existing MySQL persistence code changed to accommodate an additional database engine, since this will cause issues.  Perhaps either evolve the JPA persistence bundle, or make a new, generic JDBC persistence bundle that works with PostgresSQL and possibly other configurable JDBC drivers (but knowing that driver-swappable JDBC under OSGi is a configuration headache, along the lines of [this](http://freemanfang.blogspot.ie/2012/03/how-to-use-jdbc-driver-in-osgi.html), for example).

lewie:
@watou: Then a generic JDBC persistence bundle would be the best Choice, isn't it.
Externalize the Driver (jar) and SQL Part from Service Logic.
I will do some Tests... stay tuned.

Either way, it would, however, be an additional Bundle.
The Naming would be more generally. (not PostgreSql as this PR) 

Rgds
Helmut

watou:
Based on a quick read through the MySQL and JPA code, I agree that a more generic JDBC persistence bundle would make sense, with the goal that it would support as many JDBC drivers as users expressed interest in using.  Eventually, it could obsolete the MySQL persistence bundle and yet be fully compatible with existing MySQL databases that were created by the MySQL persistence bundle.  That way, you get PostgresSQL support but without creating "yet another" database-specific persistence bundle, and without disturbing the existing MySQL bundle. I'm sure Kai, Thomas and others have thoughts and know the history better, but this is my take for now.

lewie:
@watou: That sounds great for me.

The only Thing I want to say:
I'm not a SQL Guru for several Databases and its peculiarities.
I would begin to program it for MySql and PSql at first, based on the ideas in JPA, MySql and PostgreSql services.
True specialists could then do improvements for SQL and the program structure. But a start to bring together all JDBC persistence services would be made. 

My name proposal for the new Service would be: "JDBC" ;-)

@teichsta: In the meantime, I withdraw my PR

Rgds
Helmut

Issue #2814, opened 2015-06-28
http binding ignores timeout parameter in config


spacemanspiff2007:
Hi,

I want to set the timeout for the http-config higher and changed the config to the following:
```
# timeout in milliseconds for the http requests (optional, defaults to 5000)
http:timeout=20000
```
Unfortunately this has now effect as I get the following error:
```
2015-06-28 10:29:46.394 [ERROR] [.o.b.http.internal.HttpBinding] - given config key 'timeout' does not follow the expected pattern '<id>.<url|updateInterval>'
```

watou:
Reading the code, it looks like time the `http:timeout` value is being correctly parsed and set, but is generating the error message regardless.  If exceptions were added for `http:timeout` and `http:granularity` config options to [this line](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/HttpBinding.java#L409), you should not see the error message further.

I also note that the timeout and granularity config options are not documented in the [wiki page](https://github.com/openhab/openhab/wiki/Http-Binding).

 I can submit a PR if you would test the change?

spacemanspiff2007:
I can test the change, this should be no problem.
Maybe you can include a message output on startup which shows the values.

watou:
I don't think extra INFO logging would be well received, and it shouldn't be necessary in my opinion, when the binding is well documented and doesn't accidentally issue misleading log entries.  I will update the wiki page with descriptions of the default values for `http:timeout` and `http:granularity` so it's clearer.

spacemanspiff2007:
It doesn't show the warning anymore.
Thank you for your help.

PR #2815, opened 2015-06-28
[http] False error logged when timeout or granularity given in openhab.cfg

watou:
Do not mistakenly log errors when `http:timeout` or `http:granularity` are specified in openhab.cfg.  Addresses #2814 .


teichsta:
Thanks @watou!

Issue #2599, opened 2015-05-15
averageSince/deltaSince not working with RRD4J COUNTER


J-N-K:
Latest rrd4j-persistence added the ability to use COUNTER as type. This works well for graphing.However, there is a problem when using "averageSince" or "deltaSince".

The reason is that RRD4J already returns deltas and not the absolute values (which is perfectly fine for graphing).

averageSince is adding all these values and divides by the number of values. The problem is that the latest value (read from the item's state) is also added. Because this is the absolute value (which can be very large for counters), the average is wrong. A solution would be to leave the last data point out.

A similar problem is deltaSince. deltaSince uses the item's historic state (which is in our case a "delta" at that time and substracts that from the item's current state (which is the absolute value). The solution is to add all historical data.

However, both functions are used for all persistence services, not specialized to RRD4J. Any suggestions?

J-N-K:
@teichsta: Could you please comment om this issue?

I have implemented a sumSince in my development branch, this fixes the deltaSince issue. However, averageSince is still affected.

J-N-K:
After thinking a while about the averageSince problem: IMHO it is wrong to use the current value in calculations of averageSince or similar functions, at least the current implementation is bad. 

We check whether the last persisted value is equal to the current value. If so, we use only persisted data, if not, we add another datapoint with the curent value. This assumes that the current value was persisted if the values are the same. This may be right or wrong, depending on how fast the value changes. The values could also be equal by chance.

To my mind the persistence extension should only work on persisted data, as this is the only way to get consistent results.  The only exception is deltaSince, this should be the difference of the persisted value and the actual value. For RRD4J counter one should use sumSince to get this result.

watou:
Are there other relevant discussion threads besides [this one](https://groups.google.com/forum/#!category-topic/openhab/owlt-euiGqw)?

J-N-K:
Unfortunately not, TTBOMK.

hmerk:
closed as it seems to be fixed with #2854

J-N-K:
Since #2854 was not merged, this was not fixed.

PR #2818, opened 2015-06-28
add new feature sumSince to persistence extensions

J-N-K:
adds solution for second part of #2599 (deltaSince with rrd4j counter db)


teichsta:
Thanks @J-N-K! Would you mind adding this to ESH as well?

J-N-K:
@teichsta: I probaly would, but couldn't find the right place in the openhab2 repo.

kaikreuzer:
@J-N-K The code has moved here: https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.persistence/src/org/eclipse/smarthome/model/persistence/extensions/PersistenceExtensions.java

J-N-K:
done.

PR #2820, opened 2015-06-28
Added support to Netatmo for unit system and pressure unit

robnielsen:
Implemented both unit system and pressure unit in the same manner as implemented by the Netatmo apps (web, Android). Unit systems supported are Metric and US. Pressure units supported are mbar, inHg, and mmHg.

### DIFF
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
@@ -49,6 +49,7 @@ import org.slf4j.LoggerFactory;
  * @author Andreas Brenk
  * @author Thomas.Eichstaedt-Engelen
  * @author Gaël L'hopital
+ * @author Rob Nielsen
  * @since 1.4.0
  */
 public class NetatmoBinding extends
@@ -63,6 +64,8 @@ public class NetatmoBinding extends
 	protected static final String CONFIG_CLIENT_SECRET = "clientsecret";
 	protected static final String CONFIG_REFRESH = "refresh";
 	protected static final String CONFIG_REFRESH_TOKEN = "refreshtoken";
+	protected static final String CONFIG_PRESSURE_UNIT = "pressureunit";
+	protected static final String CONFIG_UNIT_SYSTEM = "unitsystem";
 
 	/**
 	 * The refresh interval which is used to poll values from the Netatmo server
@@ -73,6 +76,10 @@ public class NetatmoBinding extends
 	private PointType stationPosition = null;
 
 	private Map<String, OAuthCredentials> credentialsCache = new HashMap<String, OAuthCredentials>();
+
+	private NetatmoPressureUnit pressureUnit = NetatmoPressureUnit.DEFAULT_PRESSURE_UNIT;
+
+	private NetatmoUnitSystem unitSystem = NetatmoUnitSystem.DEFAULT_UNIT_SYSTEM;
 
 	/**
 	 * {@inheritDoc}
@@ -152,11 +159,19 @@ public class NetatmoBinding extends
 						case RAIN:
 							final String requestKey = createKey(deviceId,
 									moduleId);
-							final BigDecimal value = deviceMeasureValueMap.get(
+							BigDecimal value = deviceMeasureValueMap.get(
 									requestKey).get(measureType.getMeasure());
 							// Protect that sometimes Netatmo returns null where
 							// numeric value is awaited (issue #1848)
 							if (value != null) {
+								if (measureType == NetatmoMeasureType.TEMPERATURE) {
+									value = unitSystem.convertTemp(value);
+								} else if (measureType == NetatmoMeasureType.RAIN) {
+									value = unitSystem.convertRain(value);
+								} else if (measureType == NetatmoMeasureType.PRESSURE) {
+									value = pressureUnit.convertPressure(value);
+								}
+
 								state = new DecimalType(value);
 							}
 							break;
@@ -196,8 +211,8 @@ public class NetatmoBinding extends
 													device.getLatitude()),
 											new DecimalType(device
 													.getLongitude()),
-											new DecimalType(device
-													.getAltitude()));
+											new DecimalType(Math.round(unitSystem.
+													convertAltitude(device.getAltitude()))));
 								}
 								if (device.getId().equals(deviceId)) {
 									switch (measureType) {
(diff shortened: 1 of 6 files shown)


bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
```diff
@@ -0,0 +1,59 @@
...
+		}
+		throw new IllegalArgumentException("Invalid pressureUnit: "
+				+ pressureUnit);
+	}
+	
+	public BigDecimal convertPressure(BigDecimal value) {
+		if (this == DEFAULT_PRESSURE_UNIT) {
+			return value;
+		} else if (this == INHG) {
+			return value.multiply(new BigDecimal(0.02952998751));
```
watou: If the BigDecimal multiplier values were static final class variables, you wouldn't have to "new" them with each conversion, which might be more efficient.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
```diff
@@ -0,0 +1,71 @@
...
+		}
+		throw new IllegalArgumentException("Invalid unitSystem: " + unitSystem);
+	}
+	
+	public double convertAltitude(double value) {
+		if (this == DEFAULT_UNIT_SYSTEM) {
+			return value;
+		}
+		
+		return value * 3.2808399;
```
watou: Why double and not BigDecimal?

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
```diff
@@ -0,0 +1,71 @@
...
+		
+		return value.multiply(new BigDecimal(0.0393700787));
+	}
+	
+	public BigDecimal convertTemp(BigDecimal value) {
+		if (this == DEFAULT_UNIT_SYSTEM) {
+			return value;
+		}
+		
+		return value.multiply(new BigDecimal(9)).divide(new BigDecimal(5)).add(new BigDecimal(32));
```
watou: Better to have private static final BigDecimal values NINE, FIVE, THIRTY_TWO to avoid creating new objects on each conversion.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
```diff
@@ -0,0 +1,71 @@
...
+		
+		return value * 3.2808399;
+	}
+	
+	public BigDecimal convertRain(BigDecimal value) {
+		if (this == DEFAULT_UNIT_SYSTEM) {
+			return value;
+		}
+		
+		return value.multiply(new BigDecimal(0.0393700787));
```
watou: Better to have private static final BigDecimal constant instead of new on each conversion.

distribution/openhabhome/configurations/openhab_default.cfg
```diff
@@ -1323,6 +1323,12 @@ tcp:refreshinterval=250
 # The Netatmo refresh token (see http://dev.netatmo.com/doc/authentication/usercred)
 #netatmo:refreshtoken=
 
+# The Netatmo unit system where M=Metric system, US=US System (optional, defaults to M)
```
watou: Would be good to explicitly document that US gives Fahrenheit, inches; M gives Celsius, mm, or whatever it gives, so the user doesn't have to guess.  Better would be to give the user control over each kind of unit, since they may want inches and Celsius, which is not an option.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
watou: This code may throw ArithmeticException due to non-terminating decimal expansion, unless you add precision and rounding like MathContext.DECIMAL32.  (The Ecobee binding did with very similar code until I added precision and rounding.)

watou: See #2793 .

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -196,8 +211,8 @@ protected void execute() {
 													device.getLatitude()),
 											new DecimalType(device
 													.getLongitude()),
-											new DecimalType(device
-													.getAltitude()));
+											new DecimalType(Math.round(unitSystem.
+													convertAltitude(device.getAltitude()))));
```
watou: Should convertAltitude work in BigDecimal instead?

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
robnielsen: Agreed, changes made.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
robnielsen: Switch to private statics and used 1.8 instead of 9/5 to avoid division and make more efficient

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
robnielsen: Agreed, changes made.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
robnielsen: See above

distribution/openhabhome/configurations/openhab_default.cfg
robnielsen: Added description for US and Metric. If user needs more control, they can always create a rule to do this conversion.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
robnielsen: Trying to keep the code as similar as before

teichsta:
Hi @robnielsen thanks for this enhancement which looks good to me (and @watou as well). Could you please squash your commits into one finally? Thanks, Thomas E.-E.

robnielsen:
Hi @teichsta, I've squished the commits, and thanks @watou for your help. Rob

robnielsen:
Wiki has been updated

watou:
Hey Rob, could you spell out in the [wiki page](https://github.com/openhab/openhab/wiki/Netatmo-Binding) how to specify metric as an openhab.cfg choice?  I know Metric is the default, but if someone gives `netatmo:unitsystem=Metric` instead of `netatmo:unitsystem=metric`, is it case insensitive?  Just trying to avoid future confusion!

robnielsen:
Hi John,  I updated the Wiki with your suggestions. Thanks!

watou:
Thank you, Rob, for making the Netatmo binding better!

teichsta:
Thanks @robnielsen!

PR #2823, opened 2015-06-30
Linear Window\Door Sensor Update

SkynetHome:
Removed non applicable configuration parameters and association groups.


cdjackson:
I'm a bit confused - the switch file seems to turn up in multiple commits (now 3 commits I think)? 

Can you just commit the changes you make, otherwise we'll end up in a bit of a mess with merge errors (actually, it will not be possible to merge after the first one).

SkynetHome:
Every time I commit a new file, even if I do it under a new branch, it pulls that damn zwave  comit in with it. Same thing happened on the HeatMiser edit. I have been creating new branches for each comit and only comitting the files needed.

Issue #2705, opened 2015-06-08
MySQL persistence service is vulnerable to SQL injection


kabili207:
The MySQL persistence service is using plain string concatenation in order to create SQL commands. A PreparedStatement object should be used instead when inserting new records and whatnot ([example here](http://www.vogella.com/tutorials/MySQLJava/article.html#javaconnection)).

Below is an example from my system. I have an Obi100 VOIP adapter that I'm checking with the http binding. In this case, it's just an inconvenience because the value contains single quotes, but this could just as easily be exploited by an attacker for more nefarious purposes.
```
2015-06-08 11:42:00.480 [ERROR] [.p.m.i.MysqlPersistenceService] - mySQL: Could not store item 'Phone_Last_Call' in database with statement 'INSERT INTO Item28 (TIME, VALUE) VALUES(NOW(),''Buzzer ' 1234567890') ON DUPLICATE KEY UPDATE VALUE=''Buzzer ' 14127871737';': You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Buzzer ' 1234567890') ON DUPLICATE KEY UPDATE VALUE=''Buzzer ' 1234567890'' at line 1
```

teichsta:
@kabili207 thanks for letting us know! Since you seem to have knowledge in this area would you mind creating an appropriate fix for the MySQL persistence service? Thanks, Thomas E.-E.

kabili207:
I'll see if I can come up with a fix.I don't have too much free time on my hands, so it might take a while.

teichsta:
> I don't have too much free time on my hands, so it might take a while.

yes … a well know problem ;-)

lewie:
Changed Statements to preparedStatement
Updates coming with: https://github.com/openhab/openhab/issues/2732#issuecomment-113710184
Rgds Helmut

watou:
Would it be possible to have separate pull requests for these two separate issues?

teichsta:
> separate pull requests for these two separate issues

which two issues do you see?

PR #2824, opened 2015-06-30
Mysql fix prepared statement #2705 squashed

lewie:


lewie:
Did not squash! Close it!

PR #2828, opened 2015-06-30
Add GE Device ZW4005

SkynetHome:
Added  GE ZW4005 to database and updated products.xml


cdjackson:
Thanks for persevering :)

SkynetHome:
Thank you for sticking with me. I promise if at all possible I will make it worth your while!

Issue #2177, opened 2015-02-23
Heatmiser binding bug for PRTE stats with remote sensor


NigelCoxon:
I have 13 PRT stats and 2 PRTE stats, using the v3 heatmiser protocol. The PRTE stats are setup to use the remote sensor air temp (in bathrooms). The binding works fine for the PRT stats, but always returns 6553.5 for the Roomtemp from the PRTE stats.
I have looked at the binding source code in the file HeatmiserPRT.java and there is no code to handle the remote sensor air temp.

I think the code needs enhancing to look at the Sensor Selection byte and extract either the built-in airtemp, or the remote air temp

The code  
               dcbRoomTemperature = getTemp(41)
needs to be replaced with something like
               if data [22] ==1 or data [22] == 4 {
                       getTemp (37)
                       }
               else {
                       getTemp (41)
               }

I'd be happy to test

teichsta:
@sprehn could you probably take care of this? Appreciate your support! Best, Thomas E.-E.

SkynetHome:
NigelCoxon: I added suggested code for testing, if the merge is approved please let me know how this works out!

SkynetHome:
Now NigelCoxon: Hopefully this one is done right and will be merged for you to test! Sorry about the confusion earlier, I am learning!

steve-bate:
Since the related PR is merged, can we close this issue?

cdjackson:
No. We need to resolve this yet.

SkynetHome:
Has anyone ever tested this? Does the change simply move problem from remote senor to main sensor or does this work?

9037568:
@cdjackson any update on this one?

cdjackson:
Sorry - not had the chance to look at it.

PR #2829, opened 2015-06-30
Heatmiser update per Issue #2177

SkynetHome:
Added suggested code referenced in Issue #2177  to handle remote sensor air temp.

### DIFF
--- a/bundles/binding/org.openhab.binding.heatmiser/src/main/java/org/openhab/binding/heatmiser/internal/thermostat/HeatmiserPRT.java
+++ b/bundles/binding/org.openhab.binding.heatmiser/src/main/java/org/openhab/binding/heatmiser/internal/thermostat/HeatmiserPRT.java
@@ -28,11 +28,15 @@ public class HeatmiserPRT extends HeatmiserThermostat {
 	public boolean setData(byte in[]) {
 		if (super.setData(in) == false)
 			return false;
-
+		
+		if ((data[22]==1)||(data[22]==4)){
+			dcbRoomTemperature = getTemp(37);
+		}else{	
+			dcbRoomTemperature = getTemp(41);
+		}
 		dcbState = data[30];
 		dcbHeatState = data[44];
 		dcbFrostTemperature = data[26];
-		dcbRoomTemperature = getTemp(41);
 		dcbSetTemperature = data[27];
 		dcbFloorTemperature = getTemp(39);
 		dcbHolidayTime = (data[34] & 0xFF) + ((data[33] & 0xFF) * 256);


cdjackson:
I'm interested in why this change was made like this? This is labeled as a bug, but it wasn't really a bug, but it's a change in 'concept' on how the binding works...

With this change the binding now selects different temperature sensors - why not just make both available - it doesn't have to select different sensors like this?

SkynetHome:
Just  trying to help out. A user indicated that they couldn't get a proper sensor reading because there was no code to handle the PRTE sensor and suggested a code update. I took the suggested update and wrote it into the binding in a manner that I thought would achieve the desired result without comprising the rest of the binding and uploaded for approval and testing. It was merged so I assumed that my additions were reviewed and deemed useful. I have requested a report back from the OP in his issue to find out if the code addition is working properly. As of yet, no response!

cdjackson:
I just don’t think it’s the best way forward as it means that it’s no longer possible to read the internal sensor if you have the external sensor as well. Both sensors should be separately readable, and we’ve now solved one ‘bug’ with another…

ie - someone couldn’t read the external sensor as only the internal was available, but now you can’t read the internal sensor if you also have the external, as now only the external is available...

SkynetHome:
Could the problem be solved by simply adding the internal sensor inside the if statement before the else? So if you have external sensor read both internal and external sensor else read on without setting external sensor?

cdjackson:
I don’t think so - or at least I don’t understand how? If I understand what you’re suggesting, then this would effectively reverse your change anyway as only the second = statement will do anything?

The only way I think this can be solved is to handle the sensors separately so they can be assigned to separate items. At the end of the day, there’s only a single value being read - it can only be the internal OR the external sensor that resides in the dcbRoomTemperature variable - never both. Before it was the internal, and now it’s the external...

For sensors that only support one or the other, it doesn’t matter, but I have both connected...

SkynetHome:
Chris, 

Was this issue ever addressed? Did you revert the changes or is it working as I submitted? I was wondering if it would be possible to add another variable for the remote sensor or if the variables that are included in the binding are all that are supported. ie. add dcbRemoteRoomTemperature so that both sensors could be read and rectify the problem.

cdjackson:
I’ve not had the chance to do anything yet, and as my heating is off, it’s not killing me right now :)

> I was wondering if it would be possible to add another variable for the remote sensor or if the variables that are included in the binding are all that are supported. ie. add dcbRemoteRoomTemperature so that both sensors could be read and rectify the problem.

Yes, this is exactly what I meant previously - this is how I think it should have been handled so that the user can display any or both sensors...

Issue #2745, opened 2015-06-16
Please add tz78 from TKB to the database


solucionesmipc:
Hi:

this switch is recognized by habmin as tz68 but those are very different, please add tz 78 to database

http://www.tkbhome.com/?cn-p-d-269.html

thanks a lot

Diego Ruales

cdjackson:
Please provide a link to the manual. Also, please provide a link to the tz68 manual so I can see the difference and Ill try and work out why the configuration is showing up as a tz68...

solucionesmipc:
Hi:
link for manual tz68 http://manuals.zwaveeurope.com/make.php?lang=en&type=&sku=TKB_TZ68-E&pdf=1

link for manual for tz78 http://portonlatino.com/tz78manual.pdf

i think tz78 is similar to tz06 in configuration edge, pulse, toggle pulse 

Diego

SkynetHome:
I have been unable to find any specific data on the TZ78. The manual provided overs next to nothing in the way of explaining what the device is capable of. I found one forum where a user suggested that the TZ78 functions like a TZ06. I can copy that configuration and you can test and see how it works out. Though without more specifics there is no way for me to know that your device has all of the same capabilities!

SkynetHome:
Correction the Z06 is a dual relay where is the TZ8 is a single relay. I will try to edit for differences and issue a PR for you to test!

solucionesmipc:
Great, i am ready to test it.

thanks a lot

SkynetHome:
My pleasure. I had to hack that together based on best guesses. So be
patient it gets merged and there is a problem let me know and I'll get it
worked out!

SkynetHome:
solucionesmipc: Please let us know if the device is working properly as soon as possible as this was merged despite some concerns about the parameters with so little information to work from!

solucionesmipc:
Hi, my tz78 is broken now, i just ordered a new one. i will update as soon as i receive it.

thanks a lot

Diego

SkynetHome:
@solucionesmipc Have you had the opportunity to test your new device yet.  Feedback would be much appreciated so this can be closed!

solucionesmipc:
i need one more week, the item is coming from china. cheap mail.

solucionesmipc:
hi;
 i got new device now;
i changed the new zwave binding jar in addons, 

how do i force to recognize the item as tz 78?  it is showing the item as tz68 

thanks a  lot


diego

[...]

PR #2830, opened 2015-06-30
Added Zwave Device TKB T78

SkynetHome:
Added per request reference Issue #2745 
There is little to no information available on this device. This is a
best guess configuration based on what is available. Please test and
report back!

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -1668,6 +1668,15 @@
 		</Product>
 		<Product>
 			<Reference>
+				<Type>0121</Type>
+				<Id>0133</Id>
+			</Reference>
+			<Model>TZ68</Model>
+			<Label lang="en">In Wall Micro Switch</Label>
+			<ConfigFile>tkb/tz78.xml</ConfigFile>
+		</Product>
+		<Product>	
+			<Reference>
 				<Type>0002</Type>
 				<Id>0002</Id>
 			</Reference>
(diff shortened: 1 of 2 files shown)


cdjackson:
Looking at the original request, we can't solve this like this. If the device is already recognised by the binding, then we can't just add another device with the same IDs. We would need to understand why it is wrong rather than change it, otherwise the next request will be from the people who were using the original definition asking for it to be changed back.

SkynetHome:
If i understand this situation correctly it is as the op says a totally different device than what it is being recognized as. The documentation he provided showed that he has an in wall microswitch that is being recognized as a receptacle. He compared it to another in wall micro switch in his OP but that was a dual relay not single so I tried to match his device as closely as I could based on limited documentation.  It looks like a knock off of an aeotec or other similar micro switch.

cdjackson:
I'm slightly concerned that this is being added without any information, so we need to be careful. If there are parameters added that don't exist, it could cause the initialisation to slow down or stall, so please keep an eye out for any timeouts during the initialisation procedure.

Ideally, I would have preferred this to be tested before merging, but since the auto build is down and @solucionesmipc is ready to test(and there are only 2 parameters anyway!), I'll merge now. Please report any findings as soon as possible.

solucionesmipc:
hi:
i am trying to test it, but openhab still recognizes the device as tz68 (tz68 and tz 78 xml are included in database)

do you how to force it?

diego

cdjackson:
You can't force it - it is detected from the device, so either you have a device that has been badged incorrectly (this does happen), or the database is wrong.

PR #2832, opened 2015-07-01
Added telnet SerialPort implementation

msteigenberger:
SerialPort can be accessed remotely via telnet. Changing serial port
parameters is allowed via RFC2217. Tested with ser2net on the server
side. Compatible with nrjavaserial (few methods are not supported as
they mainly makes no sense on a tcp session).
The work is based on (or more or less a copy with some changes) of https://github.com/archiecobbs/jvser/.


teichsta:
Thanks @msteigenberger will have a look soon. Best, Thomas E.-E.

9037568:
Not clear to me whether OH can use this code, as it's copied from copyrighted code. Any @openhab/maintainers able to comment?

msteigenberger:
In the meantime this code was included into nrjavaserial itself. We can close this PR.

PR #2735, opened 2015-06-14
Add Log Info for GPIO binding

Speederc:
Add Log Info for GPIO send & receive commands

PR #2834, opened 2015-07-01
Add Log Info for GPIO binding

Speederc:
New PR to remove the problem with #2735
Sorry for these trials and errors, i'm starting with github...

### DIFF
--- a/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOBinding.java
+++ b/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOBinding.java
@@ -86,8 +86,10 @@ public class GPIOBinding extends AbstractBinding<GPIOBindingProvider> implements
 						if (gpioPin != null) {
 							if (command == OnOffType.ON) {
 								gpioPin.setValue(GPIOPin.VALUE_HIGH);
+								logger.debug("Send command 1 to GPIO {}", gpioPin.getPinNumber());
 							} else {
 								gpioPin.setValue(GPIOPin.VALUE_LOW);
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOBinding.java
```diff
@@ -86,8 +86,10 @@ protected void internalReceiveCommand(String itemName, Command command) {
 						if (gpioPin != null) {
 							if (command == OnOffType.ON) {
 								gpioPin.setValue(GPIOPin.VALUE_HIGH);
+								logger.debug("Send command 1 to GPIO " + gpioPin.getPinNumber().toString());
```
watou: Doesn't `GPIOPin.getPinNumber()` return an `int`, which is a primitive type and therefore has no methods like `toString()`?  Also, if you were to use parameterized arguments to logger.debug(), you would avoid creating string objects when DEBUG logging is not enabled, like `logger.debug("Send command 1 to GPIO {}",  gpioPin.getPinNumber());` (this would work because the `int` primitive will get autoboxed to `java.lang.Integer`)

Speederc:
Hi Watou, thanks for your comments. I finally found time to install the Openhab environment and compile this modified version. You were right, the getPinNumber didn't return an int. I correct the file. I also had to add a "catch (IOException e)" part in the onEvent function to be able to use getPinNumber. Is it a correct implementation ? Regards

watou:
Your `logger.debug` changes look correct to my eye, but I don't know more about the surrounding code.  Your testing will be most important in ensuring correct behaviour! :smile:

teichsta:
Thanks @Speederc

Issue #2772, opened 2015-06-21
RFXCOM Add binding support for OWL CM113, Electrisave, cent-a-meter


BoydHitchcock:
Is there a possibility that OWL - CM113 could be added to the RFXCOM binding.

openhab error = 
08:46:07.931 [ERROR] [.rfxcom.internal.RFXComBinding:279  ] - Error occured during packet receiving, data: 0D59010C8900070000001A000079
org.openhab.binding.rfxcom.internal.RFXComException: Message CURRENT not implemented
        at org.openhab.binding.rfxcom.internal.messages.RFXComMessageFactory.getMessageInterface(RFXComMessageFactory.java:121) ~[bundlefile:na]

from rfxmngr the packet is =
03/05/2015 08:49:43
Packettype    = CURRENT
subtype       = ELEC1 - OWL CM113, Electrisave, cent-a-meter
Sequence nbr  = 0
ID            = 35072
Count         = 0
Channel 1     = 0 ampere
Channel 2     = 2.6 ampere
Channel 3     = 0 ampere
Signal level  = 5
Battery       = OK

Speederc:
Hi BoydHitchcock,
I prepared 2 files to add this CM113 support but unfortunately I'm not able to compile them before next week. Are you able to do it ? If yes the files are there : 
https://github.com/Speederc/openhab/blob/Speederc-RfxComCM113/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComCurrentMessage.java
https://github.com/Speederc/openhab/blob/Speederc-RfxComCM113/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/RFXComValueSelector.java
I hope that it works ! The value can be retrieve by the item using "Channel1Amps", "Channel2Amps", "Channel3Amps"

BoydHitchcock:
Hi Speederc,
Thank you for producing the files, I will give it a go compiling and let you know the result.

BoydHitchcock:
Hi Speederc,

I have a compilation error for the following lines:-
		data[7] = (byte) (((channel1Amps * 10.0) >> 8) & 0xFF);
		data[8] = (byte) ((channel1Amps * 10.0) & 0xFF);
		data[9] = (byte) (((channel2Amps * 10.0) >> 8) & 0xFF);
		data[10] = (byte) ((channel2Amps * 10.0) & 0xFF);
[...]

PR #2835, opened 2015-07-01
Add RFXCOM binding support for OWL CM113, Electrisave, cent-a-meter

Speederc:
Add RFXCOM binding support for OWL CM113, Electrisave, cent-a-meter.
It was validated in issue  #2772


teichsta:
Thanks @Speederc

PR #2840, opened 2015-07-02
[Energenie] State response changed? 

tinnerm:
Maybe through a new Firmware version on the device the state response changed. (a wide space was removed) so the search string has to be changed as well. 
Also the if condition could be improved. lastIndexOf() returns -1 if it fails so the if condition was executed anyway because there was tested on !=0 and later on a null pointer exception happened. Now this is solved with the new condition.


teichsta:
Thanks @tinnerm! Could that cause any backwards compat issues? Does there have to be a switch rather than removing the old code?

tinnerm:
Dear Thomas

I think that it should not cause any issues but i can't test it because i
only have the actual hardware here :(

Greetings Marco

2015-07-06 22:58 GMT+02:00 Thomas Eichstädt-Engelen <
notifications@github.com>:

teichsta:
@hmerk @mattwire could you have a look, please?

hmerk:
@teichsta
shure, will have a look later today.

teichsta:
Thanks @tinnerm and @hmerk!

Issue #2553, opened 2015-05-06
Qubino PWM module is unsupported


easypl:
Hi!
Qubino has recently released PWM thermostat module:

http://qubino.com/products-2/flush-pwm-thermostat/

Is it possible to add support for this module in openhub?

Technical docs are available here:

http://www.bluetek.pl/index.php/fileuploader/download/download/?d=0&file=custom%2Fupload%2FFile-1430901177.pdf

cdjackson:
It would be great if you could generate the XML - the information is here (https://github.com/cdjackson/HABmin2/wiki/Z-Wave-Product-Database) or you can probably copy one of the original Qubino files as a starter).

SkynetHome:
I'll run with this Chris and see if I can put it together.

cdjackson:
Please confirm this is ok.

FSSDawid:
Hi @SkynetHome 
It's not working for me looks like it's has the wrong mapping:
https://www.dropbox.com/s/kbl1zdapprlh0lw/Screenshot%202015-07-17%2014.20.41.png?dl=0
type should be 5 not 1 @ least in the EU version. 

Kind regards
Dawid

SkynetHome:
Thanks for the update will take a look right away.

FSSDawid:
If you need someone to test this let me know. I don't have the build environment for openhab so you would need to send me the z-wave jar with the fix.

SkynetHome:
Sorry I just woke up and didnt fully look it this. It appears you updated this on your own and created a PR for it but your edits dont match the screen shot you linked.

SkynetHome:
I have edited the file and committed but not pulled yet as i dont want to have multiple PR's for the same product change. Do you want to fix yours and squash or close it and i'll issue my PR then.

FSSDawid:
That's how you end up when you edit something in the browser  ;) 
I have closed off my PR. 
You can push yours. 
Thanks

SkynetHome:
Thanks PR issued. If I get time today I will compile the binding for you.

FSSDawid:
Grate appreciate it.
Cheers

FSSDawid:
Just to confirm it's working like a charm.
Thanks for the effort!

cdjackson:
Thanks for the feedback - @SkynetHome thanks for the addition :)

PR #2845, opened 2015-07-04
Qubino zmnhla binding in process.

SkynetHome:
Chris, I am committing my work so far for your perusal. Its about 1/3rd
the way done but I thought I would commit in case you want to review the
progress thus far. This is in reference to Issue #2553

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -2123,6 +2123,15 @@
 			<Label lang="en">Flush on/off thermostat</Label>
 			<ConfigFile>qubino/zmnhia.xml</ConfigFile>
 		</Product>
+		<Product>
+			<Reference>
+				<Type>3</Type>
+				<Id>1</Id>
+			</Reference>
+			<Model>ZMNHLA</Model>
+			<Label lang="en">Flush PWM thermostat</Label>
+			<ConfigFile>qubino/zmnhla.xml</ConfigFile>
+		</Product>
 	</Manufacturer>
 	<Manufacturer>
 		<Id>97</Id>
--- a/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhla.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhla.xml
@@ -1,0 +1,605 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>ZMNHLA</Model>
+	<Label lang="en">Flush PWM thermostat</Label>
+	<CommandClasses>
+		<Class>
+			<id>0x20</id>
+		</Class>  <!-- COMMAND_CLASS_BASIC -->
+		<Class>
+			<id>0x25</id>
+		</Class>  <!-- COMMAND_CLASS_SWITCH_BINARY -->
+		<Class>
+			<id>0x26</id>
+		</Class>  <!-- COMMAND_CLASS_SWITCH_MULTILEVEL -->
+		<Class>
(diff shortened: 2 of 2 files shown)


cdjackson:
It looks fine from a quick look, however for parameters that are numeric (ie have a range such as 11, 12, 13, 14) please add the minimum and maximum value elements.

Also, please can you merge the future changes on this PR together so that we don't have lots of commits for a single PR? If you're still using SourceTree, there's an option to do this in the "Commit Options" box when you go to commit changes.

SkynetHome:
Got it, will make corrections, finish this and try and merge it.

SkynetHome:
Made the corrections and then completed the file and added updated products xml. If edits are required I will figure out how to merge that commit or commits with source tree.

cdjackson:
Thanks - looks good. I've one comment. For the associations, the labels are quite long and I don' think will work well in the GUI. Maybe it would be good to changethese to ```<Help>``` and use a short version for the label?

eg for ```basic on/off (triggered by Too low temperature limit, it sends FF )``` maybe just make it ```Low Temperature```?

cdjackson:
Oh - and thanks for doing these - I really do appreciate it. It gives me a bit more time to work on other things (most notably the OH2 zwave binding at the moment)...

SkynetHome:
Np and I see your point. Will make change. Any hints for merging with Source Tree? I have not been using it for anything other than pulling openhab master into local branch. That I have got down thanks to your explanation, if you have any tips for merging the commits they would be welcome. I will play around with it in the meantime!

SkynetHome:
Chris, I cannot rebase children in Source Tree because I cannot set these files to ignore. Untracking them only makes them turn red. These files have been hanging around since I first setup the Eclipse IDE should they have been committed?
problems

cdjackson:
These are auto generated files I think (other than gitignore) - although I’m not 100% sure. It’s probably easiest to stash the changes, or to simply reset them. gitignore probably changes each time you ignore a file.  I wouldn’t commit them though…

SkynetHome:
rebase children allowed me to combined the last two commits in one, which left me with a file to push. When I tried to rebase the new combined commit with the parent commit nothing happens. So I pushed the change and that just set everything back to where I started with three individual commits. I guess I should have been doing this as I committed each file and not after the fact.

cdjackson:
If using SourceTree, you need to click on the one before the one you want to combine (if that makes sense!) and then click on the “interactive rebase” option. To me this seems a little strange - I’d before to click on the one I want to combine - not it’s parent, but that’s the way ST works…

It is however easier to do this as you go and just combine with the previous commit, but you ca’t go back and do this now… There are also other tricks that I’m not that familiar with (I’m far from a Git expert unfortunately!).

If you can’t merge them, then let’s not bother too much now and keep it in mind for future commits...

SkynetHome:
I figured out the ST process with the parent child relationship being sort of inverse to what you would expect and was able to rebase 3 into 2 but not the new 3-2 combination into 1 for some reason. As you said I would much prefer to leave this one as is and get it merged.  I will make sure and proceed with future commits through Source Tree merging each with the next.

SkynetHome:
So is this good to go or do I need to make additional changes?

cdjackson:
Sorry - I missed this one... I'll take a look tomorrow - gotta think about some sleep (it's 1:30)

SkynetHome:
No problem, and I appreciate the continued support. Goodnight.

SkynetHome:
OMG...I think I finally got it. I dont have a clue how to make this work with Source Tree but git terminal appears to have done the trick. Please let me know if this has panned out but it certainly seems to have from my end!!

cdjackson:
Looks good - thanks.

PR #2848, opened 2015-07-04
MiOS Binding: Add Service Aliases for UPnP and Micasaverde Services shipped with UI5

mrguessed:
Add a more complete set of ServiceId Aliases for the UPnP Services that are shipped with Vera/MiOS UI5.

These are provided to avoid people coding MiOS Actions/openHAB Rules with `"...:serviceId:..."` in them, and to keep the Binding declarations simpler when converting over (these aliases are used in both MiOS Action calls, and in the Binding declarations)


teichsta:
Thanks @mrguessed!

PR #2852, opened 2015-07-05
Optimized CCU metadata extraction and support for PARTY_* datapoint for device HM-CC-RT-DN

gerrieg:


teichsta:
Thanks @gerrieg! I am wondering if we should add this to 1.7.1 as well?

gerrieg:
Sure, if you like you can add it to 1.7.1 too. But i would label it as enhancement and not as a bug ;-) 
Beside the optimisation it's a workaround for a CCU problem, it does not send the metadata for this device correctly

teichsta:
Thanks @gerrieg!

Issue #2855, opened 2015-07-06
Popp Z-Weather -> Product Database


ZeroNine00:
Hi ,

can you please add the Z-Weather Device from Popp ?

http://www.pepper1.net/zwavedb/device/717

Did you think this device work with the z-wave binding ?

SkynetHome:
I'll get to work on it.

ZeroNine00:
One question how can interact with the Z-Weather ? Is it SENSOR_MULTILEVEL Sensor ? 
if yes how can I access the wind speed ? sensor_type ?

Thx

SkynetHome:
I am guessing based on this from the manual, 

Product Description
Z-Weather is a cup anemometer for Z-Wave installations. It continually measures
the wind speed and can, for example, trigger closing of the awning and blinds if it
is too windy. It also measures the amount of rotation, enabling calculation of the
wind yield per day or per year. In addition, Z-Weather includes sensors for humidity,
temperature, air pressure and dew point, essentially providing the core sensors of
a weather station.

That you could use  for example:
Number zweather  "Temperature [%.1f °C]" {zwave="32:command=sensor_multilevel,sensor_type=1"}

Number zweather  "Humidity   [%.0f %%]"  { zwave="32:command=sensor_multilevel,sensor_type=5" }

etc. 
More examples for the multilevel sensor class can be found in Z Wave Binding Wiki

ZeroNine00:
Hi , 
thx and what is the sensor_type for wind_speed ? I don't find any information about it ?
regards

cdjackson:
> what is the sensor_type for wind_speed ? I don't find any information about it ?

Can I suggest that when you get the device you can provide the XML that the binding will (hopefully!) generate following initialisation and we can update the code then since I'm not aware of this information being available...

ZeroNine00:
ok great I will post the xml as soon I have device

ZeroNine00:
Here is the xml
<node>
  <deviceClass>
    <basicDeviceClass>ROUTING_SLAVE</basicDeviceClass>
    <genericDeviceClass>MULTILEVEL_SENSOR</genericDeviceClass>
    <specificDeviceClass>ROUTING_SENSOR_MULTILEVEL</specificDeviceClass>
  </deviceClass>
  <homeId>0xeb5ec87f</homeId>
  <nodeId>5</nodeId>
  <version>4</version>
  <manufacturer>0x169</manufacturer>
  <deviceId>0x1</deviceId>
[...]

PR #2856, opened 2015-07-06
Added Popp Z Weather

SkynetHome:
Added Pop Z Wind and Weather device to database per Issue #2855


bundles/binding/org.openhab.binding.zwave/database/popp/zweather.xml
```diff
@@ -0,0 +1,98 @@
...
+		<Class>
+			<id>0x5a</id>
+		</Class>
+	</CommandClasses>
+	<Associations>
+		<Group>
+			<Index>1</Index>
+			<Maximum>1</Maximum>
+			<Label lang="en">Life Line</Label>
+			<Help>Press the button 3 times within 1.5 seconds to add Z-Weather to
```
cdjackson: Why did you put this in here? To me, this looks like either inclusion information, or wakeup info - but I'm not sure it has anything to do with association groups? (or does it?)

SkynetHome: Lol I guess they wouldn't be looking at the habmin help for a device if it
wasn't already included would they. Yep, sorry that was a bone head move on
my part. Will remove it.

SkynetHome:
LOL...I tried to use this as a test to learn how to merge commits and managed to create a bunch of problems and end up with 3 new commits instead of one!

PR #2860, opened 2015-07-06
POPP Z Weather Added

SkynetHome:
Added support for POPP Z Weather Device in response to Issue #2855


SkynetHome:
Try this one instead Chris. Had to abandon ship on that last PR before I got any deeper into dicey water!

cdjackson:
Looks good - thanks.

PR #2866, opened 2015-07-07
MiOS Action: Modify the README.md to reference to the MiOS Doc for UPnP Calls.

mrguessed:
Sync documentation with Wiki version, so we always have the latest with the codebase.

Include a hyperlink reference for the MiOS Documentation for UPnP Calls,
since the Actions / Parameters aren't always obvious.


teichsta:
Thanks @mrguessed!

PR #2870, opened 2015-07-07
Fix state properties Plex binding latest PMS

idserda:
The latest release of the Plex Media Server (PMS, version 0.9.12.4) slightly changed the format of the machine ID in the session information. Because of this properties like state, title, etc. were no longer working in the binding. This PR fixes this. 

Could this be included in 1.7.1? Bug is also referenced here: https://groups.google.com/d/msg/openhab/-xkZGLY_zOs/Ekd-t11c4TYJ


teichsta:
Thanks @idserda!

Issue #2817, opened 2015-06-28
Z-Wave: Help with vision security In-Wall 2 relays switch & In-Wall relay switch xml


Mr-Token:
-
Dual Relay Switch 
-
Its a no logo switch from monoprice, but comes up as Vision Security 

http://downloads.monoprice.com/files/manuals/11990_Manual_141204.pdf
http://www.pepper1.net/zwavedb/device/462 

My items looks like this.
Switch Z_NodeTwoZero  "Node 2 binding of 0" (Lights) {zwave="2:0:command=SWITCH_BINARY"}
*I added the above otherwise on start up i get an warnings about no binding on  endpoint 0.*

Switch Z_lightEntryWay 	"Entry Way" (Lights) {zwave="2:1:command=SWITCH_BINARY"}
Switch Z_lightFrontDoor "Front Door" (Lights) {zwave="2:2:command=SWITCH_BINARY"}

From this informations the xml is suppose to look something like this :

zl7432.xml file I created and I associated it to node 1 for updates and I only get status reports for endpoint 0 (Z_NodeTwoZero).  
```
<?xml version="1.0" encoding="utf-8"?>
<Product>
	<Model>ZL7432</Model>
	<Label lang="en">In-Wall Dual Relay Switch</Label>
	<CommandClasses>
		<Class><id>0x60</id></Class>
		<Class><id>0x25</id></Class>
		<Class><id>0x27</id></Class>
		<Class><id>0x85</id></Class>
		<Class><id>0x72</id></Class>
		<Class><id>0x86</id></Class>
		<Class><id>0x20</id></Class>
	</CommandClasses>

	<Associations>
		<Group>
			<Index>1</Index>
			<Maximum>5</Maximum>
			<Label lang="en">Control Command</Label>
			<Help lang="en"><![CDATA[Control Command when trip]]></Help>
		</Group>
	</Associations>
</Product>

```

I add endpoints <Endpoints>2</Endpoints> xml  under the model number, with no change. 
```
<Product>
	<Model>ZL7432</Model>
	<Endpoints>2</Endpoints>
```

I added endpoints to class xml and still only updated to Z_NodeTwoZero when switch 1 or switch 2 is toggled. 
```
	<CommandClasses>
		<Class><id>0x60</id><endpoint>1</endpoint></Class>
		<Class><id>0x60</id><endpoint>2</endpoint></Class>
		<Class><id>0x25</id><endpoint>1</endpoint></Class>
[...]

PR #2873, opened 2015-07-08
Added Vision zl7431

SkynetHome:
Added Vision zl7431 per Issue #2817


SkynetHome:
I am about to have a fit with Source Tree and squashing commits. I think I am going to try using Git Terminal instead. Here is the completed file in one commit!

cdjackson:
Thanks

PR #2878, opened 2015-07-08
Error handling for cul network devices

querdenker2k:
Without this handling my CUNO lost the connection every 24h. This bug is tested for about a half year and no more connection failures occures.


teichsta:
Thanks @querdenker2k!

PR #2879, opened 2015-07-08
new persistence functions

querdenker2k:


watou:
Just today I was looking for a changesSince persistence method, in order to determine how often an item's state had changed since some time. There are a few potential use cases and it's straightforward to implement. Thoughts?

teichsta:
Thanks @querdenker2k, hope to find the time for reviewing soon … Best, Thomas E.-E.

querdenker2k:
will this pull request merged to openhab 2 as well???

teichsta:
@querdenker2k i am very sorry for not coming back to you earlier! As i remember some discussions about missing functions in PersistenceExtensions, is this related to that discussion? If not i'd suggest to first add these changes to ESH core and merge into OH1 once this is done. Move to 1.9 hence …

querdenker2k:
Which discussion do you mean?

kaikreuzer:
@teichsta What was the status here, could you clarify?

9037568:
@kaikreuzer should we merge or drop this PR?

kaikreuzer:
Well, the PR made sense in 2015 when it was created, but I am afraid that by now, we have to close it here as there won't be any further OH1 runtime release.
I'd therefore like to suggest to @querdenker2k to open a PR at https://github.com/eclipse/smarthome/ instead - I'd guess most of the code can be kept and this should mainly require a namespace refactoring. Very sorry that this PR has never been process, I hope @querdenker2k isn't too frustrated and will give us a new chance at ESH :-)

Issue #1837, opened 2014-12-14
OpenHAB Web: mapped strings cannot be quoted


gersilex:
Hey there, OpenHAB webapp receives an uncaught exception when trying to use quotes ```"``` inside a command.

This is my simplified command from the sitemap:
```
Selection item=AACC mappings=["color=daywhite"="Tageslicht", color="Eigene Farbe", sunrise="Sonnenaufgang (30m)"]
```
The ```AACC``` Item is a String or a Switch with a serial binding, working well.

If I click one of the selections with the ```color``` or ```sunrise``` command, everything works well.
If I click the selection with command ```color=daywhite``` the webapp returns to the home screen of openhab and outputs the following on the console:
```
Uncaught SyntaxError: Unexpected token ILLEGAL @ openhab.app:1
```
The mentioned line only contains ```<!DOCTYPE html>``` which is not very helpfull indeed.
The command is _not_ being sent to the OpenHAB server.

You can see that the quotes are really needed. Works fine on openhab.android.

I know about the way with aliasing this with, let's say, numbers. That's how it was before, but I wanted to remove the unnecessary step and send the commands immediately.

Thanks, folks!

steve-bate:
@marcelerkel, can this issue be closed?

marcelerkel:
PR2880 should indeed have solved this as well. @gersilex can you confirm that this is no longer an issue?

gersilex:
22:32:09 INFO o.o.m.c.i.ModelRepositoryImpl | Refreshing model 'default.sitemap'
    22:32:24 INFO r.busevents | scene received command Essen=toll

fix confirmed. Thank you very much!

Issue #1414, opened 2014-09-12
Double Quotes in Classic UI Switch Dont Send Command


jdrake4:
Using the classic UI and trying to do mappings with a switch where the command contains a space requires. That command is never sent.  This appears to work in the greent gui, (although a slightly different problem in that gui)

Sitemap file Does not work:
Switch item=Lights_Kitchen_String label="Scene" icon="dd_light" mappings=["Scene A"=A, "Scene B"=B, "Scene C"=C, "Scene D"=D]

and in my items file:
String Lights_Kitchen_String	"Kitchen [%s]"	(Lights_Kitchen)	{omnilink="unit:1"}

Sitemap file that does work:
Switch item=Lights_Kitchen_String label="Scene" icon="dd_light" mappings=[A=A, B=B, C=C, D=D].

(The does work sitemap shows in the log that the command is sent, however the omni binding requires me to send a "Scene A" command. So actually the UI send the command but the omni link replies it is a bad command)

teichsta:
fixed through #2880

PR #2880, opened 2015-07-08
Strip off the double quotes from the mapping command in case they exist.

marcelerkel:
This fixes the root cause of the Classic UI not supporting double quotes in Mapping commands.

When enclosing the command in double quotes, the double quotes become part of command. This causes two problems:
* HTML gets formatted as ""the command"" instead of "the command"
* "the command" != the command so there can never be a match and as a result the switch and selection item do not work properly when the command is wrapped in double quotes

Reported problems/issues:
* This fixes the problem reported [here](https://groups.google.com/forum/#!category-topic/openhab/sitemaps/wFbXRT1LWxg) with the Harmony Hub Binding.
* #1414
* Probably #1837 as well.

Discussion topic can be found [here](https://groups.google.com/forum/#!category-topic/openhab/sitemaps/D-PpPY-ixns).

### DIFF
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/resources/SitemapResource.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/resources/SitemapResource.java
@@ -330,18 +330,7 @@ public class SitemapResource {
     		Switch switchWidget = (Switch) widget;
     		for(Mapping mapping : switchWidget.getMappings()) {
     			MappingBean mappingBean = new MappingBean();
-				// Remove quotes - if they exist
-				if(mapping.getCmd() != null) {
-					if(mapping.getCmd().startsWith("\"") && mapping.getCmd().endsWith("\"")) {
-						mappingBean.command = mapping.getCmd().substring(1, mapping.getCmd().length()-1);
-					}
-					else {
-						mappingBean.command = mapping.getCmd();
-					}
-				}
-				else {
-					mappingBean.command = mapping.getCmd();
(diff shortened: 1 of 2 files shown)


marcelerkel:
I can squash them into one commit if you like, but for now I chose a separate commit so that it is clear that I added an additional change to the PR.

teichsta:
seem to be related to #2880 … at least the same symptom both of you fought against the same issue :-)

marcelerkel:
Implementation updated based on lessons learned while fixing PR #2933.

teichsta:
@J-N-K did you have a chance to look into this PR until tonight? It would be great if you could verify this in your environment before releasing it in 1.7.1. Thanks, Thomas E.-E.

marcelerkel:
@teichsta Are you sure your not mixing up this PR #2880 with PR #2933? They are very similar, but not the same.
* PR #2880 (this PR) is about stripping of double quotes from mapping commands.
* PR #2933 is about stripping off double quotes around state/commands in rules. PR #2933 is the PR that replaces PR #2578 by @J-N-K.

teichsta:
ah, you're right … sorry for the confusion!

teichsta:
Thanks @marcelerkel! Would you mind creating an according PR also for ESH?

kaikreuzer:
This is a good example why PRs should not be merged before a porting PR is created for ESH...
Million thanks to @digitaldan for finding it and for creating https://github.com/eclipse/smarthome/pull/809!

Will anyone also take care of https://github.com/openhab/openhab/pull/2933, which seem to be in the same situation?

PR #2883, opened 2015-07-08
Fix bug in config parameter to allow size=3

cdjackson:
https://github.com/cdjackson/HABmin2/issues/51


xsnrg:
verified working with parameter size=3.  Thanks!

cdjackson:
Great - thanks for the quick feedback.

PR #2884, opened 2015-07-09
z-wave binding zw096 - added additional international Type ID codes for US, EU, AU, CN

xsnrg:
removed maximum settings where they did not make sense
added additional international Type ID codes for US, EU, AU, CN

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
@@ -189,8 +189,9 @@
         <Parameter>
             <Index>83</Index>
             <Type>byte</Type>
-            <Default>DDA0DD</Default>
+            <Default>14524637</Default>
             <Minimum>0</Minimum>
+            <Maximum>16777215</Maximum>
             <Size>3</Size>
             <Label lang="en">RGB value when it is in Night light mode</Label>
             <Help lang="en">Configure the RGB value when it is in Night light mode. 
@@ -202,8 +203,9 @@
         <Parameter>
             <Index>84</Index>
             <Type>byte</Type>
-            <Default>505050</Default>
-            <Minimum>010101</Minimum>
+            <Default>3289650</Default>
+            <Minimum>0</Minimum>
+            <Maximum>6513507</Maximum>
             <Size>3</Size>
             <Label lang="en">LED brightness in Energy Mode</Label>
             <Help lang="en">Configure the brightness level of RGB LED (0%‐100%) when 
--- a/bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
@@ -191,7 +191,6 @@
             <Type>byte</Type>
             <Default>DDA0DD</Default>
             <Minimum>0</Minimum>
-            <Maximum>FFFFFF</Maximum>
             <Size>3</Size>
             <Label lang="en">RGB value when it is in Night light mode</Label>
             <Help lang="en">Configure the RGB value when it is in Night light mode. 
@@ -205,7 +204,6 @@
             <Type>byte</Type>
             <Default>505050</Default>
             <Minimum>010101</Minimum>
-            <Maximum>999999</Maximum>
             <Size>3</Size>
             <Label lang="en">LED brightness in Energy Mode</Label>
             <Help lang="en">Configure the brightness level of RGB LED (0%‐100%) when 
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -673,6 +673,18 @@
 				<Type>0003</Type>
 				<Id>0060</Id>
 			</Reference>
+			<Reference>
+                <Type>0103</Type>
+                <Id>0060</Id>
+            </Reference>
+            <Reference>
+                <Type>0203</Type>
+                <Id>0060</Id>
+            </Reference>
+            <Reference>
+                <Type>1D03</Type>
+                <Id>0060</Id>
+            </Reference>
 			<Model>ZW096</Model>
 			<Label lang="en">Smart Switch 6</Label>
 			<ConfigFile>aeon/zw096.xml</ConfigFile>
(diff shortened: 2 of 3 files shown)


bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
```diff
@@ -191,7 +191,6 @@
             <Type>byte</Type>
             <Default>DDA0DD</Default>
             <Minimum>0</Minimum>
-            <Maximum>FFFFFF</Maximum>
```
cdjackson: We probably shouldn't remove this - however it should be in decimal rather than hex

bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
```diff
@@ -191,7 +191,6 @@
             <Type>byte</Type>
             <Default>DDA0DD</Default>
```
cdjackson: This value should be in decimal. I know you didn't add it, but do you mind fixing it?

bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
```diff
@@ -205,7 +204,6 @@
             <Type>byte</Type>
             <Default>505050</Default>
```
cdjackson: This value should be in decimal. I know you didn't add it, but do you mind fixing it?

xsnrg:
I didn't look at the code for the binding close enough to see how it handles these, but the values are not simply a conversion to decimal. Does the binding know that when you set size=3 that it is 3 sets of values, and not a single large one for the MAX?  That is FFFFFF is 255255255 not 16777215.  This question would answer all 3 of your comments I think.  I will understand the format for MAX values.

cdjackson:
> Does the binding know that when you set size=3 that it is 3 sets of values, and not a single large one for the MAX?

No - unfortunately not.  The max can't be 255255255 - it needs to be the decimal of FFFFFF when converted from hex (ie 16777215).


> This would seem to make configuration more difficult for an end user that then has to convert their hex color values into a single decimal representation.

Unfortunately this is the case. The binding doesn't have the ability to split separate parts of the configuration into its components. Really, I think I will blame the manufacturer for this - they should (in my opinion :) ) keep different configs separate...

I also don't think this will be possible in OH2 - certainly not as it stands, but I'll have a think about how/if this could be better achieved, but it's likely to be difficult as the reality is there are 3 parameters here that really need to be defined separately and then combined somehow.

xsnrg:
Thanks Chris, that is quite helpful.  Perhaps sometime in the future the size attribute can be used to ask for  the number of inputs to match the size, and then the binding just combines them.  Not a big deal for now.  I will convert to decimal max.  
What you are saying about vendors is true as well.  In this binding alone they mix and match decimal with hex.  One parameter is actually an alarm time setting where they use a hex field (because they are really all hex) as a time HH : MM.  Their example says set a time such as 1800 to have a mode turn off at 18:00.  0x18 = 24, so nicely done there folks.

I will make the changes.

cdjackson:
> 
> Perhaps sometime in the future the size attribute can be used to ask for the number of inputs to match the size, and then the binding just combines them. Not a big deal for now. I will convert to decimal max.
> 
Unfortunately I fear it’s not that simple. In this case, where we have 3 parameters that are the same (except for the name of the colour) you might be able to do something. But in other instances where the parameters are very different (e.g. one byte might be a colour, another the hue and another something else) it doesn’t work. Really, they need to be treated as a number of separate parameters, with separate definitions in the config file, and then reconstituted after the data has been entered...

bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
xsnrg: Wow, so this is another instance that just leaves me scratching my head.  The value is a size 3, and is the brightness in 1-100% for each color.  Here is what I will do, and you tell me if it makes sense.  The default is 50% brightness across each, so that equates to 505050, or 323232 if hex.  What they don't say is if 0x32, which is 50 decimal is the 50% number of if 256/2 = 128 = 0x80 is the 50% number.  Let's assume the lower to be safe, and see if we can find out from Aeon what the real scoop is.

So...  323232 equates to decimal 3289650 and I will make that the default value.
and .. 636363 equates to decimal 6513507 and that will be the max.

As noted previously, this smells bad.

xsnrg:
>Unfortunately I fear it’s not that simple. In this case, where we have 3 parameters that are the same >(except for the name of the colour) you might be able to do something. But in other instances where >the parameters are very different (e.g. one byte might be a colour, another the hue and another >something else) it doesn’t work. Really, they need to be treated as a number of separate parameters, >with separate definitions in the config file, and then reconstituted after the data has been entered...

I think we are saying the same thing.  The user would be prompted for 3 different inputs and the config file would store them that way as well.  The binding is where the size=3 would come in and know to combine them to talk to the device.

bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
xsnrg: modified:

```
            <Default>14524637</Default>
            <Minimum>0</Minimum>
            <Maximum>16777215</Maximum>
```

bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
xsnrg: modified

xsnrg:
commit 438cdee should roll everything up.

cdjackson:
> On 11 Jul 2015, at 15:11, Jim <notifications@github.com> wrote:
> 
> The binding is where the size=3 would come in and know to combine them to talk to the device.

No - that’s what I meant earlier - you can’t assume that there are 3 things that are the same (or even 3 things that are different) from the size. There might be a size = 4 - with the bottom 2 bytes meaning one thing and the top byte being something else, with 8 single bit values in the middle… It all needs to be separated out completely and described independently in the configuration… anyway, that’s not one to worry about for now :)

bundles/binding/org.openhab.binding.zwave/database/aeon/zw096.xml
cdjackson: > On 11 Jul 2015, at 15:05, Jim <notifications@github.com> wrote:
> 
> So... 323232 equates to decimal 3289650 and I will make that the default value.
> and .. 636363 equates to decimal 6513507 and that will be the max.
> 
> As noted previously, this smells bad.
> 
I agree - it’s not the best, but there’s not much we can do about this, so I think it’s ok...

xsnrg:
got it.  Thanks for the commit.

PR #2887, opened 2015-07-09
fix common handling for double broadcasts within rest (websocket)

querdenker2k:
Not sending double broadcasts is good, but if registering for state changes on multiple items just the first change to e. g. ON is sended, all other following ON commands on other items will not be broadcasted.
This pull excludes double broadcast handling for */state registrations.

### DIFF
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/filter/DuplicateBroadcastProtectionFilter.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/filter/DuplicateBroadcastProtectionFilter.java
@@ -64,6 +64,13 @@ public class DuplicateBroadcastProtectionFilter implements PerRequestBroadcastFi
 	private boolean isDoubleBroadcast(HttpServletRequest request,
 			Object responseEntity) throws JsonGenerationException,
 			JsonMappingException, IOException {
+		
+		if (request.getPathInfo().endsWith("state")) {
+			// extra handling for state requests.
+			// if an item and another item is switched to the same state
+			// the message must be broadcasted as well
+			return false;
+		}
 	
 		String clientId = request.getHeader(HeaderConfig.X_ATMOSPHERE_TRACKING_ID);
 
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/filter/DuplicateBroadcastProtectionFilter.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/filter/DuplicateBroadcastProtectionFilter.java
@@ -64,14 +64,6 @@ public class DuplicateBroadcastProtectionFilter implements PerRequestBroadcastFi
 	private boolean isDoubleBroadcast(HttpServletRequest request,
 			Object responseEntity) throws JsonGenerationException,
 			JsonMappingException, IOException {
-		
-		if (request.getPathInfo().endsWith("state")) {
-			// extra handling for state requests.
-			// if an item and another item is switched to the same state
-			// the message must be broadcasted as well
-			return false;
-		}
-	
 		String clientId = request.getHeader(HeaderConfig.X_ATMOSPHERE_TRACKING_ID);
 
 		// return false if the X-Atmosphere-tracking-id is not set
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ItemsStateChangeListener.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ItemsStateChangeListener.java
@@ -1,0 +1,68 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.io.rest.internal.listeners;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.openhab.core.items.Item;
+import org.openhab.io.rest.RESTApplication;
+import org.openhab.ui.items.ItemUIRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is the {@link ResourceStateChangeListener} implementation for item REST requests
+ * 
+ * @author Robert Delbrück
+ * @since 1.8.0
+ */
+public class ItemsStateChangeListener extends ResourceStateChangeListener {
+
+	static final Logger logger = LoggerFactory.getLogger(ItemsStateChangeListener.class);
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Object getResponseObject(HttpServletRequest request) {	
+		return getItemAndState(this.lastChange);
+	}
+	
+	public static String getItemAndState(final Item item) {
+		if (item != null) {
+			return item.getName() + ":" + item.getState().toString();
+		}
+		return null;
+	}
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	protected Object getSingleResponseObject(Item item, HttpServletRequest request) {
+		return getResponseObject(request);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	protected Set<String> getRelevantItemNames(String pathInfo) {       
+		ItemUIRegistry registry = RESTApplication.getItemUIRegistry();
+        final Set<String> result = new HashSet<String>();
+        for (Item item : registry.getItems()) {
+        	result.add(item.getName());
+        }
+		return result;
+	}
+
+}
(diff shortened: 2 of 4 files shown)


teichsta:
@digitaldan @petrklus could you please have a look at this one? Thanks, Thomas E.-E.

petrklus:
@querdenker2k Thank you very much for your contribution. I understand what issue this solves, however, I am not 100% familiar with the duplicate filtering functionality and I am wary of any blanket exceptions.

Is there any case the exceptions would be too much and would lead to double broadcasts of the state updates? @digitaldan @teichsta , any thoughts?

teichsta:
@digitaldan it would be great if you could have a look at this …

digitaldan:
Sorry!  Missed this in all my email noise,  Yes, I will take a look, thanks.

teichsta:
great, thanks!

digitaldan:
@querdenker2k , I'm not familiar with atmosphere and using websockets, how are you subscribing to multiple items?  Can you give me an example?

querdenker2k:
Yes, my current code is this and with the fix its working great.
```java
final String itemStateUrl = restUrl + ITEMS + PATH_SEP + itemStateUpdater.item.getName() + PATH_SEP + STATE;

        LOG.debug("creating websocket for item '{}'", itemStateUpdater.item);

        final AtmosphereClient client = ClientFactory.getDefault().newClient(AtmosphereClient.class);
        final AtmosphereRequest request = client.newRequestBuilder()
                .method(Request.METHOD.GET)
                .uri(itemStateUrl)
                .header("X-Atmosphere-tracking-id", itemStateUpdater.item.getName())
                .header("Accept", MediaType.TEXT_PLAIN)
                .transport(Request.TRANSPORT.WEBSOCKET)
                .build();
        final Realm.RealmBuilder realmBuilder = new Realm.RealmBuilder();
        if (this.username != null && this.password != null) {
            realmBuilder.setPrincipal(this.username)
                    .setPassword(this.password)
                    .setUsePreemptiveAuth(true)
                    .setScheme(Realm.AuthScheme.BASIC);
        }
        final AsyncHttpClientConfig httpClientConfig = new AsyncHttpClientConfig.Builder()
                .setRealm(realmBuilder.build())
                .build();
        final DefaultOptions options = client.newOptionsBuilder()
                .runtime(new AsyncHttpClient(httpClientConfig))
                .reconnect(true)
                .build();
        final Socket socket = client.create(options);
        try {
            socket.on(Event.MESSAGE, new Function<Object>() {
                @Override
                public void on(Object o) {
                    listener.changed(o.toString());
                }
            }).on(Event.ERROR, new Function<String>() {
                @Override
                public void on(String e) {
                    LOG.error("error reading websocket", e);
                    // reconnecting
                    String value = null;
                    do {
                        try {
                            Thread.sleep(5000);
                        } catch (InterruptedException ex) { }
                        value = getValue(itemStateUrl, MediaType.TEXT_PLAIN);
                    } while (value == null);
                    listener.changed(value);
                    // connection established
                    registerForItemStateChanges(itemStateUpdater, listener);
                }
            }).on(Event.REOPENED, new Function<Object>() {
                @Override
                public void on(Object o) {
                    itemStateUpdater.item.setOnline(true);
                    LOG.info("reconnected: " + o);

                }
            }).on(Event.CLOSE, new Function<Object>() {
                @Override
                public void on(Object o) {
                    itemStateUpdater.item.setOnline(false);
                    LOG.info("socket closed: {}", o);
                }
            }).on(Event.OPEN, new Function<Object>() {
                @Override
                public void on(Object o) {
                    itemStateUpdater.item.setOnline(true);
                    LOG.info("socket opened: {}", o);
                }
            }).on(Event.STATUS, new Function<Object>() {
                @Override
                public void on(Object o) {
                    LOG.info("status: {}", o);
                }
            }).open(request);
        } catch (IOException e) {
            e.printStackTrace();
        }
```

digitaldan:
Thanks for the details. So this is listening for one state change correct?  Are you running it concurrently for other states, so you have multiple web sockets connected?  Also this line:
`.header("X-Atmosphere-tracking-id", itemStateUpdater.item.getName())`  seems to be unusual, i was under the impression you needed to set this to 0 for the first request and let the server generate a tracking id for you (https://github.com/Atmosphere/atmosphere/wiki/Understanding-the-Atmosphere-Protocol ) .  

The Only reason I'm asking about this is the change seems simple enough, I'm just not sure what else might break (say long polling requests).  The original code seems to just compare the last cached entry state with the current state which would seem to indicate this really was designed to only have a single state being monitored by any give tracking id.   It's all very fragile and I want to make sure we don't make this less stable for others ;-)

querdenker2k:
Yes its listening for other state changes, otherwise i didn't have a problem. I can test it without this header parameter and for long-polling and streaming as well.
The change is really simple, i can't image that something breaks. Actually the server push is never send no matter which client or which item is registered if the response would be the same.
But isn't there an error as well if somebody sends in openHAB Core just e. g. ONs. The "change" will never send to clients. Of course if the client is just showing the state of openHAB it doesn't matter but if something should be triggered it will not work.
Why is the double broadcast filter implemented? Maybe it can/should be completly removed?

digitaldan:
A while ago I tried removing this class, unfortunately this resulted in many duplicate messages in the broadcaster cache for a given client.  This stopped the IOS client from working as it could not easily parse more then one response object on any given poll request. 

Instead of your fix, can you try passing in 0 for the tracking id and see if that solves it?  I'm hoping the atmosphere  framework will then generate you a unique UUID for each request which is what this cache is keyed on.

querdenker2k:
Ok i understand. No passing 0 does not work. If i send no extra header args, 0 is send automatically (seen in log message as well). I didn't really know if atmosphere generates a new client id, of course it should and i think it does it, but it did not work. It seems to be that the double broadcast filter is not just for a single client.
I investigated a lot of hours to get it working and tried a lot of clients for android, but no works. The fix i pushed is the only working version for me.
I did not really understand why messages are duplicated if they just change once in the openHAB bus. But in this case this implementation is really a little bit fragile. Currently i can they the fix works with the iOS client, the android client and my client implementation. The most/all clients are registering for page updates so nearly no client is affected by this change. Aren't they?

querdenker2k:
I have another suggest. This pull requests can be regarded. I implement a new websocket method where i can subscribe on every item changes. URL would be like this: /rest/items/state. The callback would be like "Livingroom_Light:ON" or "Heater_Kitchen:17". With this i do not have a lot of open socket Connections. In the Receiver logic a can decide to handle this or blow it away.
My special handling of double broadcast wouldn't be neccessary because the Response contains Itemname with State and this is ever unique.
What do you think?

digitaldan:
I guess I would need to see it ;-)  But I appreciate you working around changing this class, I only know from very painful experiences that making small changes in this binding can lead to bad things once a lot people start using it!

querdenker2k:
@digitaldan can you please take a look at the new commit

digitaldan:
Yes, I'll try and get to it soon, thanks.

teichsta:
Hi, any news regarding the latest changes of @querdenker2k? Would be great to have this fix added to 1.7.1 which i will release tonight. Thanks a lot, Thomas E.-E.

digitaldan:
I have not reviewed it yet, the latest pr has a much larger set of changes then the first one, I might be able to finish today, I'm not sure that would be in time.

querdenker2k:
But the new pr does not affect the old behaviour. I added just a new 
rest/websocket method. (If you can spent some time that would be nice)

digitaldan:
Could you explain what the new functionality is and how this is different then what is already exposed?  I see there's a new state endpoint, how is this different then whats exposed with /item/state ?   Thanks!

querdenker2k:
The new endpoint is for registering on all changes for all items therefore the new endpoint /rest/items/state (states for all items) instead of /rest/items/<item>/state (changes for one item).
In android resources are small. I am getting into trouble when i open 30 or more sockets to openhab to be notified for changes.
The websocket response for the new api is <item>:<state>. The rest GET call will response with a list of  all items state in the format: <item>:<state>\n<item>:<state>\n...

digitaldan:
Ok, I think I see, this returns a plain text, "name=state" style, new line delimited output at the endpoint /rest/items/state for all items , instead of using "/rest/items/itemname/state"  for each item and geting xml or json back?

querdenker2k:
Yes that is correct.

querdenker2k:
Github has removed my words in arrow brackets, Sorry. I cannot edit the comment in the mobile Version...

digitaldan:
So right now the endpoint /rest/items returns back I think what you want (all items and state), it just does not support websockets or listening for changes, what if we added this functionality to it? I'm not sure if the duplicate message class would still be a problem.   I need to run for a little bit, I'll check back in a few hours.

querdenker2k:
Yes of course it does support websockets (includes listening for changes). I have already written a client which uses this functionality and it works for over 24 hours. The double broadcast filter is not a problem because the last message was e. g. Livingroom_Light:ON and the current is Bathroom_Light:ON, so they differ and the double broadcast filter does not reject the current message.

digitaldan:
You are talking about your changes correct?  I am talking about existing functionality, I don't think the endpoint /rest/items supports the atmosphere update functionality, but I am proposing we add it which could satisfy your requirement and keep the existing API as is.

querdenker2k:
Ok i got it. No the current api call /rest/items does not support websockets. But there would be a lot of overhead, because i just need the state and within this call i will get the full item. No problem for me and its not bad to add this functionality to /rest/items.
So i create a new pull request and just extending the interessting api call. If i get into network troubles i still can add the functionality from this pull request once again.
What do you think?

querdenker2k:
This is the new pull request, which implements the discussed changed:
https://github.com/openhab/openhab/pull/2973

Therefore this pull request can be closed.

PR #1851, opened 2014-12-16
GPIO binding enhancements and fix for Debian package issue

dpslavov:
New functionality for GPIO binding: pinmaps, console commands and possibility to forcibly take control of already exported pins.
Fix for Debian package for runtime, extract "webapps/static/index.html" into "/var/lib/openhab".

PR #2889, opened 2015-07-09
gpio force

JSurf:
As discussed in #1851 i tried to extract the gpio "force" option into a separate pull request. I think the code looks good but it does not work it seems the "force" option does not make the way from openhab.cfg to the io bundle. I'm a bit lost and hope for some insight from the experts to get this working.

Maybe you can review the pull request and tell me whats wrong

### DIFF
--- a/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/GPIOBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/GPIOBindingProvider.java
@@ -30,6 +30,14 @@ public interface GPIOBindingProvider extends BindingProvider {
 	public int getPinNumber(String itemName);
 
 	/**
+	 * Query for configured pin force.
+	 * 
+	 * @param itemName the name of the item for which we make query
+	 * @return the configured force option
+	 */
+	public boolean getPinForce(String itemName);
+
+	/**
 	 * Query for configured debounce interval.
 	 * 
 	 * @param itemName the name of the item for which we make query
--- a/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOBinding.java
+++ b/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOBinding.java
@@ -300,7 +300,7 @@ public class GPIOBinding extends AbstractBinding<GPIOBindingProvider> implements
 		try {
 			int direction;
 
-			GPIOPin gpioPin = gpio.reservePin(provider.getPinNumber(itemName));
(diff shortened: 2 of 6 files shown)


teichsta:
> but it does not work it seems the "force" option does not make the way from openhab.cfg to the io bundle

how do you know? The code LGTM but note, updated is only called if there is any given property starting with ```gpio:```.

JSurf:
I think it does not work, because i have gpio:force set to true in openhab.cfg and still get a "device busy " error. 

Seems i need to do some more debugging. Problem is, i only have my "live" system to test

teichsta:
ok, please give me a sign when this PR is ready to merge … Thanks, Thomas E.-E.

JSurf:
After more testing, it seems to work...kind of...sometimes...
Seems i have hit another startup bug on the raspberry pi. 

So if the option is set it is successfully applied through the "updated" method. But it seems that sometimes the "reservePin" method is called before the "updated" method. Seems to be timing related. 
I don't know if the binding has any influence on that timinig or if this is all done via core and osgi.

I have upgraded my system from oh 1.6 to 1.7 which made overall startup very unreliable on the pi. I plan to upgrade my system to a PI 2 in the future in hopes that the startup gets more reliable with more CPU power

So overall this PR could be merged if you think the timing is not an issue on PI2 and other hardware that provides GPIOs

teichsta:
in order "stabilise" the startup behaviour you could try to use the new configuration option which does not use ManagedServices anymore. Please have a look into the current maven archetype for bindings which makes use of the new version.

JSurf:
Thanks, sounds good, will have a look when i got the time

teichsta:
any update @JSurf ?

JSurf:
Started to look into it

JSurf:
@teichsta Rebased the PR to be compatible with 1.8 but did not change the code until now

If there are any bindings which are already changed from the "old" to the "new" behaviour this would be helpful, because from looking at the new binding archetype i'm not exactly sure what to do next

JSurf:
One more question. Would the new binding style work on 1.7.1 or is it only working with 1.8 ? Don't want to upgrade my system to a dev version

JSurf:
So the new configuration option in the binding.xml  is configuration-policy="optional". Does "optional" help to stabilize startup or would only "require" help. 

In the gpio binding, all options cfg options are optional, so 
"require" would not work for all cases

JSurf:
Went another way to make this useful and reliable. "force" option can now be configured on an per-item base, so timing is not an issue. 

The global option is still there and might work on systems that are not that timing sensitive.

I could look at refactoring this to the new binding style once there are any examples on how to use it correctly. 

What do you think @teichsta ?

teichsta:
manually squash and merged (see https://github.com/openhab/openhab/commit/8cdf87adc794edfef64eef31d3a019f4dcc2678a)

> I could look at refactoring this to the new binding style once there are any examples on how to use it correctly. What do you think @teichsta ?

would be great if you would send this change in a second PR, thanks!

Issue #1820, opened 2014-12-10
Netatmo Binding: Nullpointer Exception


juelicher:
I found the following exception in my log file
2014-12-10 18:25:46.646 ERROR o.o.c.s.AbstractActiveService[:175]- Error while executing background thread Netatmo Refresh Service
java.lang.NullPointerException: null
	at org.openhab.binding.netatmo.internal.NetatmoBinding.execute(NetatmoBinding.java:120)
	at org.openhab.core.binding.AbstractActiveBinding$BindingActiveService.execute(AbstractActiveBinding.java:156)
	at org.openhab.core.service.AbstractActiveService$RefreshThread.run(AbstractActiveService.java:173)

Immediately before that was an connection error, but I am not sure if these errors are related:
2014-12-10 18:25:46.097 WARN  o.openhab.io.net.http.HttpUtil[:205]- Method failed: HTTP/1.1 403 Forbidden

juelicher:
Sorry, I forgot to post an update. I tried to double the update frequency to 600000 but I still get this exception.

robnielsen:
I am also receiving an intermittent 403 after a few hours. I'm running in eclipse with debugging turned on, and the error message returned from Netatmo is:
```
2015-07-13 06:33:32.862 [DEBUG] [.o.b.n.internal.NetatmoBinding] - Querying Netatmo API
2015-07-13 06:33:32.862 [DEBUG] [g.openhab.io.net.http.HttpUtil] - About to execute 'https://api.netatmo.net/api/getmeasure?access_token={removed}'
2015-07-13 06:33:35.693 [WARN ] [g.openhab.io.net.http.HttpUtil] - Method failed: HTTP/1.1 403 Forbidden
2015-07-13 06:33:35.693 [DEBUG] [g.openhab.io.net.http.HttpUtil] - {"error":{"code":2,"message":"Invalid access token"}}
2015-07-13 06:33:35.693 [DEBUG] [.o.b.n.internal.NetatmoBinding] - Request: MeasurementRequest[accessToken={removed}]
2015-07-13 06:33:35.693 [DEBUG] [.o.b.n.internal.NetatmoBinding] - Response: MeasurementResponse[error=NetatmoError[code=2,message=Invalid access token],status=<null>,body=<null>]
2015-07-13 06:33:35.694 [ERROR] [.o.b.n.internal.NetatmoBinding] - Invalid access token
```
I'll try and see if I can come up with a fix for this.

[...]

PR #2891, opened 2015-07-10
Netatmo: Added additional types, fix for #1820, #3242, and other improvements

robnielsen:
Added support to Netamato binding for *min_*, *max_*, sum_rain for the intervals 30 minutes, 1 hour, 3 hours, 1 day, 1 week and 1 month. The interval is configurable and defaults to 1 day.

The pull request https://github.com/openhab/openhab/pull/2820 that I created didn't have any scaling of converted values, this was giving the false impression of higher precision than is actually available (many digits of precision). The values are treated as intermediate vales scaled according to the Significant Figures and Significant Arithmetic rules which is widely used with scientific data.

Fix for issue https://github.com/openhab/openhab/issues/1820, if the token was expired, the token was refreshed and a recursive call was made to re-execute the binding, this would cause null pointer exceptions in the original execute call to the binding.

Fix for issue https://github.com/openhab/openhab/issues/3242, if the Netatmo api returns a null altituded, the binding was throwing a null pointer exception.

Added support to also refresh the token if the call to netamoto indicated the token was invalid.

Added better error message if the StartCom CA certificate is not installed

### DIFF
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoMeasureType.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoMeasureType.java
@@ -12,22 +12,40 @@ import org.apache.commons.lang.StringUtils;
 
 /**
  * @author Gaël L'hopital
+ * @author Rob Nielsen
  * @since 1.6.0
  *
  *        This enum holds all the different measures and states available to be
  *        retrieved by the Netatmo binding
  */
 public enum NetatmoMeasureType {
-	CO2("CO2"), TEMPERATURE("Temperature"), HUMIDITY("Humidity"), NOISE("Noise"), PRESSURE(
-			"Pressure"), RAIN("Rain"), WIFISTATUS("WifiStatus"), ALTITUDE(
-			"Altitude"), LATITUDE("Latitude"), LONGITUDE("Longitude"), RFSTATUS(
-			"RfStatus"), BATTERYVP("BatteryVp"), TIMESTAMP("TimeStamp"), MODULENAME(
-			"ModuleName"), STATIONNAME("StationName"), COORDINATE("Coordinate");
+	CO2("CO2", NetatmoScale.MAX), TEMPERATURE("Temperature", NetatmoScale.MAX),
+	HUMIDITY("Humidity", NetatmoScale.MAX), NOISE("Noise", NetatmoScale.MAX),
+	PRESSURE("Pressure", NetatmoScale.MAX), RAIN("Rain", NetatmoScale.MAX),
+	WIFISTATUS("WifiStatus", NetatmoScale.MAX), ALTITUDE("Altitude",NetatmoScale.MAX),
+	LATITUDE("Latitude", NetatmoScale.MAX), LONGITUDE("Longitude", NetatmoScale.MAX),
+	RFSTATUS("RfStatus", NetatmoScale.MAX), BATTERYVP("BatteryVp", NetatmoScale.MAX),
+	TIMESTAMP("TimeStamp", NetatmoScale.MAX), MODULENAME("ModuleName",NetatmoScale.MAX),
+	STATIONNAME("StationName", NetatmoScale.MAX), COORDINATE("Coordinate", NetatmoScale.MAX),
+	MIN_TEMP("min_temp", NetatmoScale.ONE_DAY), MAX_TEMP("max_temp", NetatmoScale.ONE_DAY),
+	MIN_HUM("min_hum", NetatmoScale.ONE_DAY), MAX_HUM("max_hum", NetatmoScale.ONE_DAY),
+	MIN_PRESSURE("min_pressure", NetatmoScale.ONE_DAY), MAX_PRESSURE("max_pressure", NetatmoScale.ONE_DAY),
+	MIN_NOISE("min_noise", NetatmoScale.ONE_DAY), MAX_NOISE("max_noise", NetatmoScale.ONE_DAY),
+	MIN_CO2("min_co2", NetatmoScale.ONE_DAY), MAX_CO2("max_co2", NetatmoScale.ONE_DAY),
+	SUM_RAIN("sum_rain", NetatmoScale.ONE_DAY), DATE_MIN_TEMP("date_min_temp", NetatmoScale.ONE_DAY),
+	DATE_MAX_TEMP("date_max_temp", NetatmoScale.ONE_DAY), DATE_MIN_HUM("date_min_hum", NetatmoScale.ONE_DAY),
+	DATE_MAX_HUM("date_max_hum", NetatmoScale.ONE_DAY), DATE_MIN_PRESSURE("date_min_pressure", NetatmoScale.ONE_DAY),
+	DATE_MAX_PRESSURE("date_max_pressure", NetatmoScale.ONE_DAY), DATE_MIN_NOISE("date_min_noise", NetatmoScale.ONE_DAY),
+	DATE_MAX_NOISE("date_max_noise", NetatmoScale.ONE_DAY), DATE_MIN_CO2("date_min_co2", NetatmoScale.ONE_DAY),
+	DATE_MAX_CO2("date_max_co2", NetatmoScale.ONE_DAY);
 
-	String measure;
+	final String measure;
 
-	private NetatmoMeasureType(String measure) {
+	final NetatmoScale defaultScale;
+
+	private NetatmoMeasureType(String measure, NetatmoScale defaultScale) {
 		this.measure = measure;
+		this.defaultScale = defaultScale;
 	}
 
 	public String getMeasure() {
@@ -43,5 +61,26 @@ public enum NetatmoMeasureType {
 			}
 		}
 		throw new IllegalArgumentException("Invalid measure: " + measure);
+	}
+
+	public NetatmoScale getDefaultScale() {
+		return defaultScale;
+	}
+
+	public static boolean isPressure(NetatmoMeasureType measureType) {
+		return measureType == NetatmoMeasureType.PRESSURE
+				|| measureType == NetatmoMeasureType.MIN_PRESSURE
+				|| measureType == NetatmoMeasureType.MAX_PRESSURE;
+	}
+
+	public static boolean isRain(NetatmoMeasureType measureType) {
+		return measureType == NetatmoMeasureType.RAIN
+				|| measureType == NetatmoMeasureType.SUM_RAIN;
+	}
+
+	public static boolean isTemperature(NetatmoMeasureType measureType) {
+		return measureType == NetatmoMeasureType.TEMPERATURE
+				|| measureType == NetatmoMeasureType.MIN_TEMP
+				|| measureType == NetatmoMeasureType.MAX_TEMP;
 	}
 }
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
@@ -23,9 +23,10 @@ public enum NetatmoPressureUnit {
 	MBAR("mbar"), INHG("inHg"), MMHG("mmHg");
 
 	public static final NetatmoPressureUnit DEFAULT_PRESSURE_UNIT = NetatmoPressureUnit.MBAR;
-	private static final BigDecimal MBAR_TO_INHG = new BigDecimal(0.02952998751);
 
-	private static final BigDecimal MBAR_TO_MMHG = new BigDecimal(0.750061683);
+	private static final BigDecimal MBAR_TO_INHG = new BigDecimal("0.0295");
+
+	private static final BigDecimal MBAR_TO_MMHG = new BigDecimal("0.7500");
 
 	String pressureUnit;
 
@@ -50,6 +51,16 @@ public enum NetatmoPressureUnit {
 				+ pressureUnit);
 	}
 	
+	/**
+	 * Convert to appropriate measurement.
+	 *
+	 * The Barometer is accurate to +-1 mbar or +- 0.03 inHg
+	 *
+	 * @param value
+	 *            pressure in mbars
+	 *
+	 * @return value in proper measurement
+	 */
 	public BigDecimal convertPressure(BigDecimal value) {
 		if (this == DEFAULT_PRESSURE_UNIT) {
 			return value;
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
@@ -26,11 +26,11 @@ public enum NetatmoUnitSystem {
 
 	private static final double METERS_TO_FEET = 3.2808399;
 
-	private static final BigDecimal MM_TO_INCHES = new BigDecimal(0.0393700787);
+	private static final BigDecimal MM_TO_INCHES = new BigDecimal("0.0394");
 
-	private static final BigDecimal ONE_POINT_EIGHT = new BigDecimal(1.8);
+	private static final BigDecimal ONE_POINT_EIGHT = new BigDecimal("1.8");
 
-	private static final BigDecimal THIRTY_TWO = new BigDecimal(32);
+	private static final BigDecimal THIRTY_TWO = new BigDecimal("32");
 
 	String unitSystem;
 
@@ -53,6 +53,14 @@ public enum NetatmoUnitSystem {
 		throw new IllegalArgumentException("Invalid unitSystem: " + unitSystem);
 	}
 	
+	/**
+	 * Convert to appropriate measurement.
+	 *
+	 * @param value
+	 *            altitude in Meters
+	 *
+	 * @return value in the proper measurement
+	 */
 	public double convertAltitude(double value) {
 		if (this == DEFAULT_UNIT_SYSTEM) {
 			return value;
@@ -61,6 +69,17 @@ public enum NetatmoUnitSystem {
 		return value * METERS_TO_FEET;
 	}
 	
+	/**
+	 * Convert to appropriate measurement.
+	 *
+	 * The Rain gauge is accurate to 1 mm/h or 0.04 in/h, and the range starts
+	 * at 0.2 mm/h or 0.01 in/h.
+	 *
+	 * @param value
+	 *            rain in Millimeters
+	 *
+	 * @return value in the proper measurement
+	 */
 	public BigDecimal convertRain(BigDecimal value) {
 		if (this == DEFAULT_UNIT_SYSTEM) {
 			return value;
@@ -68,7 +87,17 @@ public enum NetatmoUnitSystem {
 		
 		return value.multiply(MM_TO_INCHES);
 	}
-	
+
+	/**
+	 * Convert to appropriate measurement.
+	 *
+	 * The Thermometer is accurate to +- 0.3°C or +- 0.54°F
+	 *
+	 * @param value
+	 *            temperature in Celsius
+	 *
+	 * @return value the in proper measurement
+	 */
 	public BigDecimal convertTemp(BigDecimal value) {
 		if (this == DEFAULT_UNIT_SYSTEM) {
 			return value;
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/NetatmoBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/NetatmoBindingProvider.java
@@ -10,6 +10,7 @@ package org.openhab.binding.netatmo;
 
 import org.openhab.core.binding.BindingProvider;
 import org.openhab.binding.netatmo.internal.NetatmoMeasureType;
+import org.openhab.binding.netatmo.internal.NetatmoScale;
 
 /**
  * This interface is implemented by classes that can provide mapping information
@@ -19,6 +20,7 @@ import org.openhab.binding.netatmo.internal.NetatmoMeasureType;
  * taken into account.
  * 
  * @author Andreas Brenk
+ * @author Rob Nielsen
  * @since 1.4.0
  */
 public interface NetatmoBindingProvider extends BindingProvider {
@@ -62,4 +64,13 @@ public interface NetatmoBindingProvider extends BindingProvider {
 	 */
 	String getModuleId(String itemName);
 
+	/**
+	 * Returns the scale to use when querying the Netatmo measure of the given
+	 * {@code itemName}.
+	 *
+	 * @param itemName
+	 * @return the Netatmo scale of the Item identified by {@code itemName} if
+	 *         it has a Netatmo binding, <code>null</code> otherwise
+	 */
+	NetatmoScale getNetatmoScale(String itemName);
 }
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
@@ -12,9 +12,11 @@ import static org.apache.commons.lang.StringUtils.isNotBlank;
 import static org.openhab.binding.netatmo.internal.messages.MeasurementRequest.createKey;
 
 import java.math.BigDecimal;
+import java.util.Calendar;
 import java.util.Collection;
 import java.util.Dictionary;
 import java.util.Enumeration;
+import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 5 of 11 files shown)


bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoMeasureType.java
```diff
@@ -44,4 +62,25 @@ public static NetatmoMeasureType fromString(String measure) {
 		}
 		throw new IllegalArgumentException("Invalid measure: " + measure);
 	}
+
+	public NetatmoScale getDefaultScale() {
+		return defaultScale;
+	}
+
+	public static boolean isPressue(NetatmoMeasureType measureType) {
```
watou: isPressure

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
```diff
@@ -52,11 +53,11 @@ public static NetatmoPressureUnit fromString(String pressureUnit) {
...
 	public BigDecimal convertPressure(BigDecimal value) {
 		if (this == DEFAULT_PRESSURE_UNIT) {
-			return value;
+			return value.setScale(0, RoundingMode.HALF_UP);
 		} else if (this == INHG) {
-			return value.multiply(MBAR_TO_INHG);
+			return value.multiply(MBAR_TO_INHG).setScale(2, RoundingMode.HALF_UP);
 		} else { // MMGH
-			return value.multiply(MBAR_TO_MMHG);
+			return value.multiply(MBAR_TO_MMHG).setScale(1, RoundingMode.HALF_UP);
```
watou: Should so much precision be lost at formatting time instead of during conversion?

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
```diff
@@ -52,11 +53,11 @@ public static NetatmoPressureUnit fromString(String pressureUnit) {
 	
 	public BigDecimal convertPressure(BigDecimal value) {
 		if (this == DEFAULT_PRESSURE_UNIT) {
-			return value;
+			return value.setScale(0, RoundingMode.HALF_UP);
 		} else if (this == INHG) {
-			return value.multiply(MBAR_TO_INHG);
+			return value.multiply(MBAR_TO_INHG).setScale(2, RoundingMode.HALF_UP);
```
watou: Should so much precision be lost at formatting time instead of during conversion?

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
```diff
@@ -63,17 +64,17 @@ public double convertAltitude(double value) {
 	
 	public BigDecimal convertRain(BigDecimal value) {
 		if (this == DEFAULT_UNIT_SYSTEM) {
-			return value;
+			return value.setScale(1, RoundingMode.HALF_UP);
```
watou: Same question about losing all precision here instead of during formatting %.1f

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoMeasureType.java
robnielsen: Fixed spelling mistake, thanks @watou

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
robnielsen: Replicated the same functionality as displayed in Netatmo apps, this is .1 of a millimeter

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
robnielsen: Replicated the same functionality as displayed in Netatmo apps, this is .01 of a inch

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoUnitSystem.java
robnielsen: Replicated the same functionality as displayed in Netatmo apps, this is .1 of a millimeter

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoPressureUnit.java
watou: My only concern was the risk of possibly accumulating rounding errors, if for example, persistence functions are used on sets of these values.  You may know this to not be a concern; just pointing it out!

robnielsen: This is well withing the range of accuracy of the device:  Accuracy: +-1 mbar / +- 0.03 inHg. https://www.netatmo.com/en-US/product/specifications/station

 I'll go back and review the ranges to make sure that I don't unnecessary remove accuracy.

teichsta:
i would go with @watou precision should be held as long as possible (within a reasonable range). Only the UI should decide how many digits to show. Think about rules calculating things with the various values …

teichsta:
@watou do you have Netatmo devices at hand?

watou:
@teichsta no, I've used the binding against the demo system in the past located in the makers' Paris office, but not recently.

robnielsen:
@watou, @teichsta, The scales are at or greater than the precision of the devices. My reasoning for changing is values were getting reported like:

Temp: 79.16000000000000116351372980716405436396598815917968750 vs 79.16 inches
Rain: 2.5393700761499998878878425756511205690912902355194091796875 vs 2.54 inches
Pressure: 29.9788433201520011783802743821070180274546146392822265625000 inHg
Pressure: 762.812731610999978704157342690450605005025863647460937 mmHg

I realized that I don't need to scale the default values so I removed scaling from these, but it is needed on the ones for the conversions I added in a prior pull request.

watou:
@robnielsen, I think the issue with precision is that, even though Netatmo's native interface shows converted values rounded to tenths or hundredths, your Metric to "U.S." conversion multipliers in NetatmoUnitSystem.java are precise to 7 and 10 digits after the decimal.  When you multiply by such precise numbers, as you definitely should, but then round the result to 1 or 2 digits, you are introducing rounding errors that can end up noticeably skewing any computations that are later performed on sets of those rounded numbers.  So even though the input Metric values are only accurate to one or two digits of precision, when you convert them to "U.S." values, the U.S. equivalents need to be as accurate as the input values, which means preserving many more digits of precision in the output U.S. values.  Only then will any functions like averageSince, sumSince, etc., remain as accurate as the original Metric values. 

For example, if you pass in [MathContext.DECIMAL64](http://docs.oracle.com/javase/7/docs/api/java/math/MathContext.html) to the BigDecimal.multiply function, it will properly constrain the amount of precision to something "reasonable," and yet allow nearly exact results between averaging your U.S. values, vs. averaging your metric values and then converting that to U.S. units.  With the current rounding, those results will likely be very different.

robnielsen:
@teichsta, what do you consider a reasonable range? for example, if the device is accurate to 0.04 of an inch, is 0.01 a reasonable value? For example, the code is currently scaling 2.5393700761499998878878425756511205690912902355194091796875  inches of rain to 2.54 inches of rain, this seems reasonable to me, anything beyond the 2nd digit is well beyond the accuracy of the device. If we had further precision, doesn't this give indication to somebody that the values are indeed accurate to the level of precision provided?

BTW, I've also found out that switching the constructor from:
```java
private static final BigDecimal MM_TO_INCHES = new BigDecimal(0.0393700787);
```    
to
```java
private static final BigDecimal MM_TO_INCHES = new BigDecimal("0.0393700787");
```

Helps considerably with precision, the same calculation, it produces the value 2.53937007615. Below is a note from the http://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#BigDecimal(double) constructor:

>The results of this constructor can be somewhat unpredictable. One might assume that writing new BigDecimal(0.1) in Java creates a BigDecimal which is exactly equal to 0.1 (an unscaled value of 1, with a scale of 1), but it is actually equal to 0.1000000000000000055511151231257827021181583404541015625. This is because 0.1 cannot be represented exactly as a double (or, for that matter, as a binary fraction of any finite length). Thus, the value that is being passed in to the constructor is not exactly equal to 0.1, appearances notwithstanding.

watou:
Since the BigDecimal class gives you the ability to achieve exact results with only the needed degree of precision when multiplying your conversion factors constructed from strings instead  of doubles, I recommend you use those values without further rounding.  Since you expressed your conversion factors with 7 and 10 digits of precision, BigDecimal will keep all further computations accurate w.r.t. those conversion factors.  Introducing further rounding will lead to odd forms of inaccuracy when users try to compute totals, averages, etc. from those values.

robnielsen:
@watou, For C to F you get a reasonable result if you use String constructor since multiple will use the scales of the numbers of both multiplicands. But this still gives odd results, for example the web services returns values such as 0 or 0.1, and the conversions are 32.0 and 32.18 . If it was additionally scaled, by 2 you'd get 32.00 and 32.18 

The problem comes with the other conversions. You want to use as precise of a number as possible when multiplying, but then scale to a reasonable value, which is what I asked @teichsta in my previous comment. If a device is only accurate to 0.04" or (0.1 cm), it's just silly to report accuracy higher than 0.01" for rain or barometric pressure.  For example say we use 1 cm which is  0.0393700787" unscaled or 0.04" scaled, and we had 1000 readings of that value we get the following:

0.0393700787 * 1000 = 39.3700787000
0.04 * 1000 = 40.00

Maybe I'm way off base here, but 40.00 is what I'd expect in this case, not  39.3700787000

watou:
In your example, I would expect the value to be recorded as 39.3700787000", but I would probably format it for display as 39.4". Or if I were interested in only inches, it would be formatted as 39".  A result of 40" is less accurate than 39", and it would bother me as a weather station operator to know that my values' accuracy was being rounded away for each recorded value, causing me to doubt accuracy over past ranges of values. 

When Netatmo says their rain gauge is only accurate to 0.1 cm, I don't interpret that to mean that it makes sense to further sacrifice accuracy by rounding it away when that value is recorded in some other unit system.  When you do, you actually worsen accuracy.  The only place where I'm aware that the extra digits are unwanted is in formatting for display, which is well handled with formatting strings in the label text.  Is there some other problem created by allowing the full precision of your conversions into the item state and persistence?  Am I incorrect that computations (like average and sum) will not begin to veer away from the values reported by Netatmo due to rounding each sample?

If Netatmo's native reporting units of measure were reversed (Fahrenheit, feet and inches) and were converting to Metric units, I would make the same points.

robnielsen:
@watou, check out https://en.wikipedia.org/wiki/Significance_arithmetic it has the following:

> Significance arithmetic is a set of rules (sometimes called significant figure rules) for approximating the propagation of uncertainty in scientific or statistical calculations. These rules can be used to find the appropriate number of significant figures to use to represent the result of a calculation. If a calculation is done without analysis of the uncertainty involved, a result that is written with too many significant figures can be taken to imply a higher precision than is known, and a result that is written with too few significant figures results in an avoidable loss of precision. Understanding these rules requires a good understanding of the concept of significant and insignificant figures.

Which supports my point of view by reporting values with  precision of 10 digits we are falsely given the indication of considerably higher precision than the device supports (which is 2 digits).

The article also refers to the https://en.wikipedia.org/wiki/Significant_figures which has the following under https://en.wikipedia.org/wiki/Significant_figures#Arithmetic:

>When performing a calculation, do not follow these guidelines for intermediate results; keep as many digits as is practical (at least 1 more than implied by the precision of the final result) until the end of calculation to avoid cumulative rounding errors.

@teichsta, this basically answers my question on what's reasonable (one additional digit). Do you agree? BTW .001" is 0.0254mm, and he device reports values in 0.1mm

watou:
I read those quoted passages differently.  

If you formatted for display all digits of your unit conversion, then you would indeed be presenting misleading precision.  But that is easily avoided by giving examples and documentation that clearly states that you are merely faithfully converting unit systems (which I believe users would expect of the binding implementation), not implying that the original measurements were accurate to that degree of precision.  ([The current wiki page for the Netatmo binding](https://github.com/openhab/openhab/wiki/Netatmo-Binding#example-items), as does all the binding documentation I've read, applies formatting strings to DecimalType values, like %.1f, so the number of decimal places are limited for display, which is sensible.)  Think of it as merely projecting the Metric values through a lens as a convenience to the user who wants non-Metric units.  The user would expect the lens to project a clear picture, not one that was altered before they even had access to the actual data.

Also, the second passage supports my position, that it is necessary to maintain "as many digits as is practical" in intervening calculations, such as when your goal is to compute sums, averages or other calculations that depend on the most accurate sample values.  To discard whatever accuracy you have through rounding each sample, you will further lose accuracy when adding up rainfall, computing average temperatures, etc.  

Users would see diminished value in the binding if they had lower faith in the accuracy of the reported values, and perhaps even open a new issue against it so they could restore the available accuracy.  In the meantime, the values already recorded in their persistence stores would be dubious.

robnielsen:
I actually discussed this with my son who is a Junior in Electrical Engineering at a university that is ranked as one of the top 20 in the US. He's the one who provided with the info above. Also, I wrote some code to show that 3 digits (possibly 4 is enough):

When converting 0.0, 0.1, ... 99.9 from mm to Inches and back to mm, you need a precision of 3:

```java
import java.math.BigDecimal;
import java.util.Random;

public class Conversion {
	private static final BigDecimal MM_TO_INCHES = new BigDecimal("0.0393700787");
	private static final BigDecimal INCHES_TO_MM = new BigDecimal("25.4");

	public static void main(String[] args) {
		mmToInToMm(2);
		mmToInToMm(3);
	}
	
	public static void mmToInToMm(int precision) {
		int num = 0;
		int same = 0;
		for (int i = 0; i < 100; i++) {
			for (int j = 0; j < 10; j++) {
				num++;
				
				BigDecimal mm = new BigDecimal("" + i + "." + j);
				BigDecimal in = mm.multiply(MM_TO_INCHES).setScale(precision, BigDecimal.ROUND_HALF_UP);
				BigDecimal mm2 = in.multiply(INCHES_TO_MM).setScale(1,  BigDecimal.ROUND_HALF_UP);
				
				if (mm.equals(mm2)) {
					same++;
				}
				//System.out.println(mm + ":" + mm2 + ":" + (mm.equals(mm2) ? "same" : "different"));
			}
		}
		
		System.out.println("num: " + num + " same: " + same);
	}
}
```
produces the result:
```
num: 1000 same: 394
num: 1000 same: 1000
```
So for straight conversion to/from mm/inches 3 digits of precision is sufficient, any more precision provides no value.

I also wrote the following program to calculate sums using no scaling, 3 and 4 digits of scaling when summing:

```java
import java.math.BigDecimal;
import java.util.Random;

public class Conversion {
	private static final BigDecimal MM_TO_INCHES = new BigDecimal("0.0393700787");
	private static final BigDecimal INCHES_TO_MM = new BigDecimal("25.4");

	public static void main(String[] args) {
		sum(1000000);
	}

	public static void sum(int iterations) {
		System.out.println("summing " + iterations + " values");
		BigDecimal sum1 = BigDecimal.ZERO;
		BigDecimal sum2 = BigDecimal.ZERO;
		BigDecimal sum3 = BigDecimal.ZERO;
		
		Random r = new Random();
		int i = 0;
		while (i++ < iterations) {
			// random number between 0.3 and 99.9
			BigDecimal mm = new BigDecimal("" + r.nextInt(1000)).movePointLeft(1);
//			System.out.println(mm);
			
			BigDecimal in = mm.multiply(MM_TO_INCHES);
			sum1 = sum1.add(in);
			
			BigDecimal in2 = in.setScale(3, BigDecimal.ROUND_HALF_UP);
			sum2 = sum2.add(in2);
			
			BigDecimal in3 = in.setScale(4, BigDecimal.ROUND_HALF_UP);
			sum3 = sum3.add(in3);
		}
		
		// scale back to 2 digits
		BigDecimal sum1Scaled = sum1.setScale(2, BigDecimal.ROUND_HALF_UP);
		BigDecimal sum2Scaled = sum2.setScale(2, BigDecimal.ROUND_HALF_UP);
		BigDecimal sum3Scaled = sum3.setScale(2, BigDecimal.ROUND_HALF_UP);
		
		System.out.println("sum1 scaled: " + sum1Scaled + " unscaled: " + sum1);
		System.out.println("sum2 scaled: " + sum2Scaled + " unscaled: " + sum2);
		System.out.println("sum3 scaled: " + sum3Scaled + " unscaled: " + sum3);
		System.out.println("sum1 sum2 scaled diff: " + sum1.subtract(sum2));
		System.out.println("sum1 sum3 scaled diff: " + sum1.subtract(sum3));
	}
}
```
Here's the results of a few different runs:
```
summing 1000000 values
sum1 scaled: 1966739.42 unscaled: 1966739.41925376831
sum2 scaled: 1966739.00 unscaled: 1966739.004
sum3 scaled: 1966739.39 unscaled: 1966739.3948
sum1 sum2 scaled diff: 0.41525376831 accuracy: 0.00002111381732855886%
sum1 sum3 scaled diff: 0.02445376831 accuracy: 0.000001243365952327248%

summing 1000000 values
sum1 scaled: 1964500.88 unscaled: 1964500.87988597288
sum2 scaled: 1964500.37 unscaled: 1964500.373
sum3 scaled: 1964500.86 unscaled: 1964500.8575
sum1 sum2 scaled diff: 0.50688597288 accuracy: 0.00002580227772203500%
sum1 sum3 scaled diff: 0.02238597288 accuracy: 0.000001139524706209313%

summing 1000000 values
sum1 scaled: 1966658.47 unscaled: 1966658.47437196111
sum2 scaled: 1966658.29 unscaled: 1966658.294
sum3 scaled: 1966658.49 unscaled: 1966658.4900
sum1 sum2 scaled diff: 0.18037196111 accuracy: 0.000009171493854193497%
sum1 sum3 scaled diff: -0.01562803889 accuracy: -7.946493554245969E-7%

summing 1000000 values
sum1 scaled: 1965534.50 unscaled: 1965534.49799515481
sum2 scaled: 1965534.59 unscaled: 1965534.586
sum3 scaled: 1965534.44 unscaled: 1965534.4441
sum1 sum2 scaled diff: -0.08800484519 accuracy: -0.000004477400181974162%
sum1 sum3 scaled diff: 0.05389515481 accuracy: 0.000002742010118111540%
```
Over a million iterations, this isn't much of a difference. I don't see how somebody can have lower faith in the accuracy of the binding. Again 3 should be sufficient.

watou:
It's no doubt true that the amount of accuracy lost by rounding at hundredths or thousandths of an inch is small, but I still don't understand what problem is solved by knowingly losing accuracy by rounding values you convert from Metric units.  I asked earlier but didn't see your answer.  I would like to know, as a user, that any values I was handed by the binding were equivalent in all possible ways for the purposes of any calculations I wanted to perform against them, regardless of the unit system they were presented in.  If one is rounded off before I get it and the other isn't, then they are not equivalent, and I will know that any averages, sums, etc. using them will be needlessly inaccurate.  As a user, this would bother me.

I feel I've run out of ways to present my points to you.  If I can be of any further assistance, please reference @watou in your comment and I will chime in.  All the best! - John

robnielsen:
@watou, haven't you done essentially did the same thing at https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/messages/Temperature.java#L162? It has:
```java
public final BigDecimal toCelsius() {
		return temp.movePointLeft(1).subtract(THIRTY_TWO).divide(NINE.divide(FIVE), MathContext.DECIMAL32);
	}
```

which is essentially no different than the code:

```java
public final BigDecimal toCelsius() {
		return temp.movePointLeft(1).subtract(THIRTY_TWO).divide(NINE.divide(FIVE), MathContext.DECIMAL128).setScale(5, BigDecimal.ROUND_HALF_UP);
	}
```

In this case, a decision was made that 5 digits of precision was reasonable. Just to show that this works out the same way, the code:
```java
		BigDecimal nine = new BigDecimal("9");
		BigDecimal five = new BigDecimal("5");
		BigDecimal thirtyTwo = new BigDecimal("32");

		BigDecimal temp = new BigDecimal("789");
		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).divide(nine.divide(five), MathContext.DECIMAL32));
		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).divide(nine.divide(five), MathContext.DECIMAL128).setScale(5, BigDecimal.ROUND_HALF_UP));
		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).divide(nine.divide(five), MathContext.DECIMAL64));
		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).divide(nine.divide(five), MathContext.DECIMAL128));
```
produces the result:
```
26.05556
26.05556
26.05555555555556
26.05555555555555555555555555555556
```
Using your argument, the code at https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/messages/Temperature.java#L162 should be using MathContext.DECIMAL128 instead of MathContext.DECIMAL32 shouldn't it?

Also, once the code is simplified to use a single multiplication instead of two divisions, you'll run into the same problem since  5/9 is the number is 0.555555555..., when do you stop? For example, the code:
```java
		BigDecimal thirtyTwo = new BigDecimal("32");
		BigDecimal zeroPointFives = new BigDecimal("0.5555555556");
		BigDecimal mmToInches     = new BigDecimal("0.0393700787"); // using same number of digits for both

		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).multiply(zeroPointFives, MathContext.DECIMAL32));
		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).multiply(zeroPointFives).setScale(5, BigDecimal.ROUND_HALF_UP));
		System.out.println(temp.movePointLeft(1).subtract(thirtyTwo).multiply(zeroPointFives));
```
produces the result:
```
26.05556
26.05556
26.05555555764
```
So again, I'm back to the same question, what is a reasonable level of precision? I don't think we can solve it here. I'll post a question to the group.

watou:
@robnielsen, you chose 7- and 10-digit multipliers for your unit conversions.  The Ecobee binding code you referenced has to use an infinite-digit divisor, so in order to avoid an ArithmeticException (#2793), I chose 7 digits of precision.  I still don't understand what problem you are solving by rounding your conversions to so many fewer digits of precision than are available.  I hope the wider audience of Netatmo users can help you resolve this.  Regards, John

robnielsen:
Interesting @watou, so you are basically doing the same thing I am. I could use your same argument against you that you should be using a MathContext.DECIMAL128.

robnielsen:
@teichsta, I've implemented the scaling following https://en.wikipedia.org/wiki/Significant_figures#Arithmetic, since the value is consumed downstream, I've considered it a intermediate result since it's published and don't want to give downstream consumers the indication of greater precision. If this is acceptable, I'll also update the wiki to indicate this.

The PR also includes a fix for defect #1820 which I was able to reproduce.

nesys:
fantastic piece of code, Rob :) it works like a charm. please add in the final release as soon as possible!!!

teichsta:
@robnielsen @watou thanks a lot arguing all pros and cons better than i would be able to do! But i wouldn't take it too philosophically :-)

Since you never know when to treat a value as intermediate or as final result (States can be stored using the PersistenceService and can be recalculated in other context) i would simply use what BigDecimal provides without any further scaling (except in the UI).

The cited False Precision definition talks about the ```the number of significant figures used in the presentation of data``` which definitely shouldn't imply a better precision as the real sensors can provide.

Although i agree that from the mathematicians point of view it might not be 100% correct to leave the precision as is (False Precision) following this general implementation and review rule might be simpler in more cases as it leads to a false impression of accuracy.

Can we agree on that?

watou:
@teichsta I agree.

robnielsen:
@teichsta, @watou is essentially scaling to IEEE 754R Decimal32 format at https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/messages/Temperature.java:
```java
/**
	 * Convert this Temperature to Celsius.
	 * 
	 * @return temperature in Celsius
	 */
	public final BigDecimal toCelsius() {
		return temp.movePointLeft(1).subtract(THIRTY_TWO).divide(ONE_POINT_EIGHT, MathContext.DECIMAL32);
	}
```
I assume if the same scaling is used with this pull request, it would be acceptable?

watou:
As [explained in July](https://github.com/openhab/openhab/pull/2891#issuecomment-120811407), this is the only way I am aware of to avoid the `ArithmeticException` created by an infinite-digit divisor.  Since your unit conversions only involve `BigDecimal.multiply`, they are not at risk of throwing `ArithmeticException`s, and so no rounding or truncation would be needed.

robnielsen:
Yes @watou , but you choose to use MathContext.DECIMAL32, why not MathContext.DECIMAL64 or MathContext.DECIMAL128.  I'll remove the scaling, but I'm also going to use less precise values in the calculation since 9-10 digits precision make no sense when we are talking measurements in degrees or inches.

robnielsen:
@teichsta, I removed the scaling, and used a multiplier with a precision of 4 decimal places for conversion calculations. I hope this is acceptable. 

The commits have also been squashed and are ready to get merged if you approve.

teichsta:
yes, thanks a lot @robnielsen and @watou!

robnielsen:
Thanks @teichsta, I'll get the wiki updated this weekend with the new functionality.

PR #2898, opened 2015-07-11
Possible fix for HeatMiser Binding

SkynetHome:
An attempt at solving the problem mentioned in Issue #2177  and the
corresponding problem caused by my PR #2829


cdjackson:
Thanks - I appreciate you taking a look at this, but I don't think this will work. It is however along the lines of what I was thinking but the issue is that there's a missing link with the items themselves, so there's no way to get the new variable out to the real world...

When I get some time I'll take a closer look...

SkynetHome:
Thanks Chris, I know this is not a pressing issue now but I just thought I would take a stab at trying to solve it before it is. Let me know if there is anything else I can do!

steve-bate:
@teichsta @cdjackson Should we close this pull request?

teichsta:
to me it looks like the issue isn't really fixed by this PR. So could you @cdjackson and @SkynetHome please advise how to proceed with this PR? Thanks, Thomas E.-E.

PR #2901, opened 2015-07-12
[XBMC] Fixes properties like Player.Artist always being empty.

marcelerkel:
Properties returned as lists were only processed if there also was an indexer property available. However, some properties come back as a list without an indexer. For those, instead of not processing them at all, return the first in the list.
This fixes the problem where the artist is never displayed, while the album name and song title are displayed.

### DIFF
--- a/bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetItem.java
+++ b/bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetItem.java
@@ -21,7 +21,7 @@ import com.ning.http.client.AsyncHttpClient;
 /**
  * Player.GetItem RPC
  * 
- * @author tlan, Ben Jones
+ * @author tlan, Ben Jones, Marcel Erkel
  * @since 1.5.0
  */
 public class PlayerGetItem extends RpcCall {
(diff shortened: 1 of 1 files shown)


teichsta:
can anybody confirm this fix is working?

Fjuxx:
i would like to test, but still pretty new to openhab,
Though i also have this issue with the XBMC binding with KODI.

is there a snapshot available so i can test it?

i'm still struggling getting the dev environment running ;)

marcelerkel:
@Fjuxx I have attached the snapshots of the Action and the Binding in [this](https://groups.google.com/forum/#!category-topic/openhab/xbmc/7UHWt9XWXC8) topic.
Don't worry that the version indicates 1.8.0-SNAPSHOT, they work fine with openHAB 1.7.0. Make sure you remove the 1.7.0 version of the Action and Binding from the addons directory.

teichsta:
any news?

marcelerkel:
I have opened a [topic](https://groups.google.com/forum/#!category-topic/openhab/xbmc/7UHWt9XWXC8) on the group and got confirmation of one tester stating that it works.

marcelerkel:
@sumnerboy12 Ben, your names pops up in the source code of the XBMC binding quite a lot and you're an active member of the openHAB community. Could you have a look whether the change I'm proposing is a sensible one?

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetItem.java
```diff
@@ -80,23 +80,29 @@ public String getPropertyValue(String property){
...
-			// check if the index is valid
-			List<?> values = (List<?>)value;
-			if (propertyIndex < 0 || propertyIndex > values.size())
-				return null;
+			// some properties come back as a list with an indexer
+			String paramPropertyIndex = getPropertyValue(paramProperty + "id");
+			// other properties come back as a list without an indexer,
+			// e.g. artist, for these we return the first in the list
+			int propertyIndex = 0;
+			if (!StringUtils.isEmpty(paramPropertyIndex)) {
```
sumnerboy12: It looks fine Marcel - the only thing I would do differently is have an explicit check for a missing `paramPropertyIndex` and return the first element in the array - just so it is obvious to anyone reading the code that in this case if there is a missing indexer we just return the first item. Minor refactoring tho, it looks fine to me otherwise.

marcelerkel:
Thanks for your feedback Ben, I have refactored the code accordingly.

openhab-kodi-artist-fix

teichsta:
Thanks guys!

PR #2865, opened 2015-07-07
[satel] Connection status implemented

druciak:
This PR allows to define items for connection status to the module. Currently there are 3 new binding configurations added:
* module:connected - simple binary flag active when there is connection established,
* module:connected_since - a timestamp describing when connection has been established,
* module:connection_errors - a number describing number of connection errors in a row; it is cleared after connection has been established.

PR #2913, opened 2015-07-13
[satel] Connection status implemented

druciak:
This PR replaces #2865 

copied description: This PR allows to define items for connection status to the module. Currently there are 3 new binding configurations added:

module:connected - simple binary flag active when there is connection established,
module:connected_since - a timestamp describing when connection has been established,
module:connection_errors - a number describing number of connection errors in a row; it is cleared after connection has been established.


teichsta:
Thanks @druciak!

PR #2915, opened 2015-07-13
Edit Popp Z - Weather

SkynetHome:
Added missing configuration parameters not mentioned in Pepper Database. Ref PR #2855

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/popp/zweather.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/popp/zweather.xml
@@ -58,6 +58,29 @@
 			<id>0x5a</id>
 		</Class>
 	</CommandClasses>
+	<Configuration>
+		<Parameter>
+			<Index>1</Index>
+			<Type>byte</Type>
+			<Default>6</Default>
+			<Minimum>0</Minimum>
+			<Maximum>30</Maximum>
+			<Size>1</Size>
+			<Label lang="en">Wind Speed Action Threshold</Label>
+			<Help lang="en">Available Parameters: 0  - Off
+			1 - 30 in milliseconds Default = 6 m/s</Help>
+		</Parameter>
+		<Parameter>
+			<Index>2</Index>
+			<Type>byte</Type>
+			<Default>37</Default>
+			<Minimum>0</Minimum>
+			<Maximum>100</Maximum>
+			<Size>1</Size>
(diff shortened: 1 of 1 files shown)


ZeroNine00:
Hi Sorry but the update of product.xml is missing. As I wrote in issue for z-weather device popp have a manufactor id !

SkynetHome:
@cdjackson What do you think about this? I cannot find a reference to a new Mfg ID in any of the literature. Pepper has ID for this device as 0154 as well. Although as I have seen Pepper is far from perfect. Should I make changes based on what habmin is showing this users device to be or based on the literature?

cdjackson:
Update it with the value that HABmin is showing. I’ve seen multiple manufacturer IDs pop up in the past - we just need to be mindful that it might be wrong and we might need to change the name later. For example, sometimes a manufacturer gets another company to develop a unit, and we might fond other random devices use the same ID, but supposedly come from different suppliers…

Anyway, for now, use the value HABmin provides - it’s gotta be right and we’ll worry about the name if something conflicts later…

SkynetHome:
Thanks will update now!

cdjackson:
Thanks - hopefully this will sort it :)

SkynetHome:
Agreed. I just downloaded and am running openhab2. I saw your latest post on the openhab2 format but havent had a chance to look at it as ive run into some problems with OH2 I'm trying to sort out. Will take a look in a bit and start making any changes to database in accordance with your new format!

cdjackson:
Don’t get too far ahead with the OH2 files yet - things may still change around for a day or three. I’ve currently got my dimmer running under the new OH2 binding, but as I start to incorporate a few more devices it might change a little. I’ve also got to sort out some of the configuration issues yet, so there might be some changes…

I have put a converted version of the database into my repo though - it’s linked near the top of the wiki page so it’s a good starter :)

SkynetHome:
So far when I run OH2 the only binding that is recognized is Astro binding.
Should I be breaking down my openhab.cfg file that I copied over into
individual cfg files or do I just need to edit out the name spaces in the
openhab.cfg

cdjackson:
You need to move (for example) the zwave:port= config from the openhab.cfg file into a zwave.cfg file, and remove the zwave part (so just leaving port=… in the zwave.cfg file).

SkynetHome:
Thanks will try that tommorrow it's 3am I've got to get to bed.

PR #2918, opened 2015-07-15
add support for the Aeon MultiSensor 6

paixaop:
Add support for the Aeon MultiSensor 6 ZW100-A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,11 +15,3 @@ distribution/openhabhome/etc/rrd4j/*.rrd
 *.xml_gen
 *.orig
 *~
-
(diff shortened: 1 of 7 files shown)


cdjackson:
Thanks - this all looks good, but I have 2 requests....
1) Can you please check that there aren't 3 association groups for this device? There are config parameters associated with groups 2 and 3, but these association groups are included.
2) Can you please use the formatter in Eclipse to reformat the XML just to make it easier to read (ie in the source menu).

Thanks
Chris

paixaop:
Chris 

I think it is fixed now.

cdjackson:
Unfortunately you've made a number of other changes to the repository that need to be reverted - we just need the zwave files only...

paixaop:
Chris rolled back the repo and re-added the XML files.

PR #2924, opened 2015-07-15
Netatmo: 1.7.1 fix for issue #1820

robnielsen:
1.7.1 fix for issue https://github.com/openhab/openhab/issues/1820, if the token was expired, the token was refreshed and a recursive call was made to re-execute the binding, this would cause null pointer exceptions in the original execute call to the binding.


teichsta:
Thanks @robnielsen!

PR #2927, opened 2015-07-16
Fix pollution of / URL space

### LINKED ISSUE #2742: Move Classic UI jquery .js files
Somewhere in 1.7 jquery .js files were added right to the / of openHAB web server which is a very bad idea. The files are: jquery.miniColors.css, jquery.min.js, jquery.miniColors.js.
There should be some normal directory structure for such files, I would suggest to keep all .js in /static/js/ folder.

sja:
This fixes #2742.

* All assets are moved from URL Path `/` to `/classicui`
* Moved all classicui related assets below `/classicui` (BaseServlet.java)
* Moved CMD Servlet to `/classicui/CMD`
* Fixed all Widgets to use relative URLs
* HTML-Variable `%servletname&` is now replaced by full servlet URL.
* Fixed all image URLs to always use `/images/...`
* Fixed refresh handling to use relative URL

Testen on Chrom, IE 11 and Firefox.

Signed-off-by: Sebastian Janzen <sebastian@janzen.it>

### DIFF
--- a/bundles/ui/org.openhab.ui.webapp/snippets/button.html
+++ b/bundles/ui/org.openhab.ui.webapp/snippets/button.html
@@ -1,1 +1,1 @@
-<a href="javascript:void(0)" onclick="OH.changeState('../CMD?%item%=%cmd%')" class="iButton iB%type%" style="%labelstyle%">%label%</a>
+<a href="javascript:void(0)" onclick="OH.changeState('/CMD?%item%=%cmd%')" class="iButton iB%type%" style="%labelstyle%">%label%</a>
--- a/bundles/ui/org.openhab.ui.webapp/snippets/colorpicker.html
+++ b/bundles/ui/org.openhab.ui.webapp/snippets/colorpicker.html
@@ -1,17 +1,17 @@
 <li>
 	<img src="/images/%icon%.png" width="29" height="29" class="iFull" />
 	<span class="options">
-		<img src="images/_down.png" 
-			ontouchstart="OH.repeatedRequest('../CMD?%item%=DECREASE', %frequency%, 1, event)" 
-			ontouchend="OH.stopRepeatedRequest('../CMD?%item%=OFF', event)" 
-			onmousedown="OH.repeatedRequest('../CMD?%item%=DECREASE', %frequency%, 1)" 
-			onmouseup="OH.stopRepeatedRequest('../CMD?%item%=OFF')" />
+		<img src="/images/_down.png" 
+			ontouchstart="OH.repeatedRequest('/CMD?%item%=DECREASE', %frequency%, 1, event)" 
+			ontouchend="OH.stopRepeatedRequest('/CMD?%item%=OFF', event)" 
+			onmousedown="OH.repeatedRequest('/CMD?%item%=DECREASE', %frequency%, 1)" 
+			onmouseup="OH.stopRepeatedRequest('/CMD?%item%=OFF')" />
 		<img src="/images/colorwheel.png" width="29" height="29" border="0" 
 			onclick="OH.colorpicker.setup('%item%', '%purelabel%', '%state%')" />
-		<img src="images/_up.png" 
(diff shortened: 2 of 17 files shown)


belovictor:
So should I add /classicui to my.openhab.org list of proxied URLs?

kaikreuzer:
Is this a clean backport of https://github.com/eclipse/smarthome/pull/358 (i.e. are the urls identical)?

sja:
:-/ I had not in mind that there already has to be a solution on oh2.

My URLs are:
 - http://localhost:8080/classicui/openhab.app?sitemap=demo
 - http://localhost:8080/classicui/CMD?DemoSwitch=TOGGLE&...

I just shifted classicui in there to leave the rest of the URL the same. I can sync that tomorrow if you want.

kaikreuzer:
I am actually not sure if this whole thing is a good idea for openHAB 1. #2742 is about moving .js files, which is ok because they are only used internally within the Classic UI.
But changing the url of the servlet (openhab.app as well as CMD) is a breaking change - I would not want to imagine how many people out there rely on these urls for some custom integration they did.
For openHAB 2, it is clearly communicated that things will break and will require a manual migration. But for an openHAB 1.8 this should imho be avoided.

sja:
Ok, breaking URLs is not nice for a Minor-Release.

So should I change to this?:
 - http://localhost:8080/openhab.app?sitemap=demo  <-- As before
 - http://localhost:8080/classicui/jquery.min.js <-- new location, referred in HTML from URL above

Alternatively, this should be possible, too:
 - http://localhost:8080/openhab.app/jquery.min.js <-- new location, referred in HTML from URL above

The first variant is like in OH2, the second does not split to two base uris.

watou:
I believe these kinds of changes affect my.openhab.org, so best to consider that possibility as well (not advocating for or against!).

kaikreuzer:
@watou: Note that the issue https://github.com/openhab/openhab/issues/2742 had been entered by @belovictor exactly because of my.openHAB :-)

> Alternatively, this should be possible, too
@sja I don't think this is possible - once you have registered openhab.app as a servlet, you cannot register any sources as a sub-url anymore. So I would rather go for the first option.

watou:
Aha!  Thanks @kaikreuzer.  Just nervous about a possible new regression; glad it's well in hand. :-)

teichsta:
let me know, when this PR is ready to merge …

sja:
I updated the PR. Please review it.

I keep the move of CMD servlet, because it's only used internally from classicui. And I think this clears out, that its part of classicui because `openhab:8080/CMD...` looks like a global API.

kaikreuzer:
> I keep the move of CMD servlet

Not a good idea, see my https://github.com/openhab/openhab/pull/2927#issuecomment-122197031:
 > But changing the url of the servlet (openhab.app as well as CMD) is a breaking change - I would not want to imagine how many people out there rely on these urls for some custom integration they did.

This comment was especially about the CMD servlet (actually much more about it than about openhab.app).

sja:
Forgot to mention that I updated this PR.

belovictor:
Guys, I need a clear statement which URL's should I add to my.openHAB ;-)

kaikreuzer:
Afaik, you will need "classicui" as a new url and you can remove all the *.js files that were directly in the root url. The rest should stay unchanged.

sja:
@kaikreuzer is right, we now have:

 * /CMD
 * openhab.app
 * classicui/

The first two with all parameters passed through and the last one is a complete path.

teichsta:
any updates @sja?

sja:
@teichsta As I wrote 23 and 20 Days ago: It's done.

teichsta:
so as all changes seem to be done to the OH1 code, do we have to synchronise this merge with another merge in my.openHAB @belovictor? If not, i would go ahead and merge this PR finally …

hakan42:
@teichsta , is this still valid for 1.8.0 ? It could very well be a breaking change...

PR #2929, opened 2015-07-17
Product Qubino Flush PWM thermostat fix id in db

FSSDawid:
Product Qubino Flush PWM thermostat fix id in products.xml
#2553


FSSDawid:
Product Qubino Flush PWM thermostat fix id in products.xml
#2553

SkynetHome:
Thanks for the update. Will fix right away.

PR #2930, opened 2015-07-17
Nest-Binding: Suppress occasionally failing update echo cancellations.

watou:
This code change attempts to suppress rare cases where an attempt to cancel an update echo fails.  An update echo is when the `execute()` method polls the Nest API and sends updates on the event bus for bound items.  The binding's `internalReceiveUpdate` method, which is intended to handle updates from all other sources, consults a dynamic list of events to ignore and upon finding and removing a matching event, does not forward the update back to the Nest API.  In rare cases and as yet unexplained, the echo cancellation fails and the binding sends the update back to the Nest API.

This code change employs two approaches to suppress this behaviour:

1. Instead of using a string concantenation of the `itemName` and string version of the new `State` as previously, it places `Update` objects on the list, which are comprised of the `itemName` and the `state`.  The `Update` class's `equals` method is used by the list's `remove` method to find and remove the first matching update.  Previously `String.equals` was used, but appeared to be occasionally failing for as yet unknown reasons.
2. When updates arrive for non-outbound items, a warning is logged and sending the update to Nest is skipped.

Note: This PR only attempts to work around the issue of unsuppressed echoes.  My working theory, lacking other insight, is some kind of JVM bug.  Please test this PR extensively before I merge it (and I will continue to test it), but hopefully some kind of fix or workaround can make it into openHAB 1.7.1.

### DIFF
--- a/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/NestBinding.java
+++ b/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/NestBinding.java
@@ -77,7 +77,32 @@ public class NestBinding extends AbstractActiveBinding<NestBindingProvider> impl
 	/**
 	 * used to store events that we have sent ourselves; we need to remember them for not reacting to them
 	 */
-	private List<String> ignoreEventList = Collections.synchronizedList(new ArrayList<String>());
+	private static class Update {
+		private String itemName;
+		private State state;
+
+		Update(final String itemName, final State state) {
+			this.itemName = itemName;
+			this.state = state;
+		}
+
+		@Override
+		public boolean equals(Object o) {
+			if (o == null || !(o instanceof Update)) {
+				return false;
+			}
+			return (this.itemName == null ? ((Update) o).itemName == null : this.itemName.equals(((Update) o).itemName))
+					&& (this.state == null ? ((Update) o).state == null : this.state.equals(((Update) o).state));
+		}
+
+		@Override
+		public int hashCode() {
+			return (this.itemName == null ? 0 : this.itemName.hashCode())
+					^ (this.state == null ? 0 : this.state.hashCode());
+		}
+	}
+
+	private List<Update> ignoreEventList = Collections.synchronizedList(new ArrayList<Update>());
 
 	/**
 	 * The most recently received data model, or <code>null</code> if none have been retrieved yet.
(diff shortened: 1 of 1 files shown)


mrguessed:
I've been running the patch for ~4hs now, and there's no trace of issue in the logs.  Prior to that, I'd get:
```
07:23:46.312 WARN  o.o.b.n.i.m.AbstractRequest[:147]- Method failed: HTTP/1.1 400 Bad Request
07:23:46.316 ERROR o.o.b.n.internal.NestBinding[:352]- Error updating data model: DataModelResponse[devices=<null>,structures=<null>]
```
every ~10-15 minutes.

In my environment, I only ever use _sendCommand_ in my rules/scripts, and these are infrequently done (during Alarm arming/disarming events for the house).  As a result, I'm not really testing the _update_ codepath for anything other than suppressing echos.

watou:
Thanks @mrguessed! Please check again tomorrow if you can, just to make sure this resolves it for good.

Google Group discussion [here](https://groups.google.com/forum/#!category-topic/openhab/zHEMrJkm_yg).

mrguessed:
Your patch is correct, and avoids the thread-safety issue... in openHAB Core.

As it turns it out, the issue is in [DateTimeType.java](https://github.com/openhab/openhab/blob/master/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/DateTimeType.java#L25), specifically the use of a shared/static `SimpleDateFormat` object, which is specifically listed as an issue in it's JavaDoc:
    http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

It's an oldie, but a goodie, I remember running into that problem in the late 90's:
> Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.


Your new changes avoid the issue since you're no longer using `toString()` on your `DateTimeType` objects, so they're not handed a corrupted value string... which later fail comparison.

watou:
Criminy, what a nuisance to have a non-threadsafe toString() method.  My patch should also run faster and create less garbage, so I'll merge it.  Thanks very much, Mark.

mrguessed:
Yeah, it's not limited to it's `DateTimeType.toString()` method either, since the `SimpleDateFormat` objects are shared and used in `parse()` also.

I suspect if there's enough "DateTime" concurrency going on in an openHAB deployment that there will be little corruptions going on all over.

eg. multiple Bindings parsing, or logging, `DateTimeType` data at the same time.

I _think_ it can be fixed by changing the const's into methods, of the same name, and having them return a new object for each call.  Unfortunately, the const declarations in this class were made `public` so it makes the fix a little more sticky to fix since there may be other callers (including Rule code, potentially)

PR #2933, opened 2015-07-17
Fix for #2552. Allow the use of double quotes around state/commands in rules

marcelerkel:
In rules, double quotes should not be part of the state/command string. This fix removes them.

    Item <item> received command "Armed Away"
    Item <item> received update "Armed-Away"
    Item <item> changed from Disarmed to "Armed-Away"

Without removing the double quotes, the rule will never trigger because Armed-Away != "Armed-Away".

Fixes issue #2552 Rule trigger 'changed from to' does not work for string items
Replaces PR #2578 quick-fix for issue #2552, as suggest as option 2

I tested this using a String item which I changed to "Armed Away", "Armed-Away", "Disarmed" and Disarmed using both sendCommand() and postUpdate(). It would be good if the reporter of issue #2552 could verify the fix.

### DIFF
--- a/bundles/model/org.openhab.model.rule/src/org/openhab/model/rule/RulesRuntimeModule.java
+++ b/bundles/model/org.openhab.model.rule/src/org/openhab/model/rule/RulesRuntimeModule.java
@@ -11,6 +11,7 @@
  */
 package org.openhab.model.rule;
 
+import org.eclipse.xtext.conversion.IValueConverterService;
 import org.eclipse.xtext.scoping.IScopeProvider;
 import org.eclipse.xtext.xbase.featurecalls.IdentifiableSimpleNameProvider;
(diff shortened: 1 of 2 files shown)


marcelerkel:
On hold. I get an error when using String::format() in a rule.

teichsta:
thanks for the update @marcelerkel. Please give me a sign when you've fixed the issue. Thanks, Thomas E.-E.

marcelerkel:
@teichsta I fixed the bug (I extended the RuleConverters class from the wrong base class). I updated the code based on [this](https://github.com/eclipse/xtext/blob/master/plugins/org.eclipse.xtext/src/org/eclipse/xtext/common/services/Ecore2XtextTerminalConverters.java) implementation.

marcelerkel:
@teichsta The on-hold label can be removed.

teichsta:
Thanks @marcelerkel! Would you mind creating an according PR also for ESH?

PR #2939, opened 2015-07-19
Modbus Status update fix

DCandE:
Fix modbus status scrolling to only show status on change

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
@@ -130,6 +130,8 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider> 
 			}
 
 			State currentState = config.getItemState();
+			logger.debug("internalUpdateItem: currentState '" + currentState + "' newState : '" + newState + "'");
+
 			if (! newState.equals(currentState)) {
 				eventPublisher.postUpdate(itemName, newState);
 			}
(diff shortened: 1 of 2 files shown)


teichsta:
Hi @DCandE could you elaborate a bit on the purpose of this PR? To me it looks like your a now caching the state (which is no good idea generally since the binding should be stateless). Best, Thomas E.-E.

sipvoip:
The problem is that without it you get an update every default 250 ms even when the state does not change.

teichsta:
any chance to have a quick look into this @ssalonen? Thanks, Thomas E.-E.

ssalonen:
This might be a little far fetched but I think this might be related to the fact that we have unsymmetrical implementation when it comes to reading/writing values with the modbus binding. To put it more simply, the `valuetype` is not considered when writing data to modbus (see the [discussion in the community forums](https://community.openhab.org/t/writing-values-in-modbus-binding-not-considering-valuetype-and-binding-state/5247)).

For example, item with float32 type would go something like this
1. item value is changed to `x` (e.g. `x=3.14`) -> modbus binding receives command with the changed value
2. binding truncates the value to integer (e.g. `3.14 -> 3`)
3. binding writes the truncated integer as 16bit integer (as two's complement. Note *not* in IEEE 32bit float format)
On next poll, the modbus binding reads the value as IEEE 32bit float format -- which does not make much sense since the value was written as integer in the writing phase.

Even if the above would be the issue here, I don't see how the value would change all the time.

I'm just wondering if the additional state cache introduced in this PR might just hide some real issue? 

@sipvoip @DCandE do you have specific openhab configuration that produced the unnecessary updates? That would help a lot

teichsta:
there seem to be some discussion necessary … moved to 1.9 hence.

PR #2942, opened 2015-07-19
Ecobee-Binding: Suppress occasionally failing update echo cancellations.

watou:
Identical problem as in #2930.  I had implemented a lesser-quality workaround already in the PR for the ecobee action bundle (#2755) but I will merge this one into that PR instead.


watou:
Hi @teichsta, it looks like this PR didn't end up in the 1.7.1 release after all.  Did I do something wrong on my end?  What is the workflow for merging and releases?  Sorry if I screwed up something.

PR #2949, opened 2015-07-22
Add Panasonic TV to default configuration file

dominicdesu:
Panasonic TV was missing in default config file.


teichsta:
Thanks @dominicdesu!

PR #2966, opened 2015-07-26
bintray-upload-debs: add automatic publishing

theoweiss:
@teichsta auto publishing should work but I could not test it.


teichsta:
thanks @theoweiss!

Issue #2407, opened 2015-04-03
Weather binding -- error with sitemapFactory


xsnrg:
Hello,

I have been trying the 1.7 weather binding with OpenHAB2.  Unfortunately it is not working correctly.  I set up the IDE to see what was going on.

The area of concern seems to be around SitemapFactory not being resolved.

private String replaceItem(Token token) throws ItemNotFoundException {
		Item item = itemUIRegistry.getItem(token.itemName);
		if (SUFFIX_ITEM_VALUE.equals(token.itemSuffix)) {
			return item.getState().toString();
		}

		Widget w = SitemapFactory.eINSTANCE.createText();
		w.setLabel(itemUIRegistry.getLabel(token.itemName));
		w.setItem(token.itemName);
		String label = itemUIRegistry.getLabel(w);

and in the manifest are marked as missing:

 org.openhab.model.sitemap,
 org.openhab.model.sitemap.impl,  

I brought this up on the OH2 mail list, and was asked to post here to see if any of the authors of the binding are able to look at this and refactor the sitemapFactory dependency?

Thank you
Jim

xsnrg:
@gerrieg @teichsta anyone?

teichsta:
could you please have a look at this @gerrieg? Thanks, Thomas E.-E.

kaikreuzer:
Would be nice to have that in 1.7 as it has been reported early April already...

teichsta:
@gerrieg any chance to have a look into this? See also discussion at https://groups.google.com/forum/?hl=en#!topic/openhab2/9j3tvS_OYuQ. I've definitely overlooked this UI dependency in WetherTokenResolver during my review. Could you help removing this dependency in order to make the binding OH2 compatible through the Compatlayer?

Let me know if i can be of any help! Thanks, Thomas E.-E.

gerrieg:
removed dependency with https://github.com/openhab/openhab/pull/2607

If the solution is OK for you (method public String getFormattedLabel(Item item) in ItemUIRegistry), i can add it to OH2.

kaikreuzer:
@gerrieg, thanks that should work the the compat layer now. Although I still do not like a dependency from a binding the the ItemUIRegistry much either. I haven't looked into the weather binding deep enough yet to understand if and why this is really necessary. I'll leave the judgement to @teichsta, when he reviews the PR :-)

teichsta:
[...]

PR #2967, opened 2015-07-26
fix for #1403 - makes Actions visible in Designer

teichsta:
especially added bundle activationpolicy to dependencies in order to activate them as soon as classes are needed


teichsta:
unfortunately my.openHAB Action is still not working properly

bundles/action/org.openhab.action.astro/META-INF/MANIFEST.MF
```diff
@@ -11,6 +11,8 @@ Bundle-ManifestVersion: 2
 Bundle-Description: This is the Astro action of the open Home Aut
  omation Bus (openHAB)
 Import-Package: org.apache.commons.lang.builder,
+ org.openhab.binding.astro.internal.calc,
```
kaikreuzer: A bundle must not import internal packages from other bundles.

bundles/action/org.openhab.action.homematic/META-INF/MANIFEST.MF
```diff
@@ -10,7 +10,10 @@ Bundle-Activator: org.openhab.action.homematic.internal.HomematicActivator
 Bundle-ManifestVersion: 2
 Bundle-Description: This is the Homematic action of the open Home Aut
  omation Bus (openHAB)
-Import-Package: org.openhab.core.items,
+Import-Package: org.openhab.binding.homematic.internal.common,
+ org.openhab.binding.homematic.internal.communicator.client,
```
kaikreuzer: A bundle must not import internal packages from other bundles!

bundles/action/org.openhab.action.tinkerforge/META-INF/MANIFEST.MF
```diff
@@ -11,10 +11,12 @@ Bundle-ManifestVersion: 2
 Bundle-Description: This is the TinkerForge action of the open Home Aut
  omation Bus (openHAB)
 Import-Package: org.apache.commons.lang,
- org.openhab.core.scriptengine.action,
+ org.openhab.binding.tinkerforge.ecosystem,
+ org.openhab.binding.tinkerforge.internal.model,
```
kaikreuzer: A bundle must not import internal packages from other bundles!

features/org.openhab.designer.feature/feature.xml
```diff
@@ -376,4 +383,88 @@
...
+
+   <plugin
+         id="org.openhab.binding.weather"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.openhab.ui"
```
kaikreuzer: no binding should ever have a dependency to the ui bundle - this should not be added to the designer

teichsta: absolutely agreed and @gerrieg knows about that already … but for the time being we can decide to not add the Action to the Designer or (meanwhile) life with this flaw.

kaikreuzer: > and @gerrieg knows about that already

Do you have an issue reference?

bundles/action/org.openhab.action.homematic/META-INF/MANIFEST.MF
teichsta: Absolutely right and we should change this (out of the scope of this Bugfix). What i changed was simply removing the require-bundle deep and add the import-package instead. @gerrieg could you please take care of this and create another PR moving the externally required classes in the appropriate package?

bundles/action/org.openhab.action.astro/META-INF/MANIFEST.MF
teichsta: Absolutely right and we should change this (out of the scope of this Bugfix). What i changed was simply removing the require-bundle deep and add the import-package instead. @theoweiss could you please take care of this and create another PR moving the externally required classes in the appropriate package?

bundles/action/org.openhab.action.tinkerforge/META-INF/MANIFEST.MF
teichsta: Absolutely right and we should change this (out of the scope of this Bugfix). What i changed was simply removing the require-bundle deep and add the import-package instead. @theoweiss could you please take care of this and create another PR moving the externally required classes in the appropriate package?

bundles/action/org.openhab.action.homematic/META-INF/MANIFEST.MF
kaikreuzer: Actually, there shouldn't even a dependency from an action bundle to a binding bundle at all. The shared code should rather be externalized into an io bundle as it is e.g. done for squeezebox.

bundles/action/org.openhab.action.tinkerforge/META-INF/MANIFEST.MF
kaikreuzer: Actually, there shouldn't even a dependency from an action bundle to a binding bundle at all. The shared code should rather be externalized into an io bundle as it is e.g. done for squeezebox.

teichsta: yes, true … there are two options: a) remove the Action from the Designer due to this flaw or b) accept it for now, fix this issue and work on it afterwards

bundles/action/org.openhab.action.homematic/META-INF/MANIFEST.MF
watou: @kaikreuzer is it true that an action bundle should not depend on a binding bundle even in the case where its actions functioning are intentionally dependent on a properly configured and running binding, where each action call's first parameter is an item bound to the binding?

features/org.openhab.designer.feature/feature.xml
teichsta: yes, see https://github.com/openhab/openhab/issues/2407 where we discussed this issue

bundles/action/org.openhab.action.tinkerforge/META-INF/MANIFEST.MF
kaikreuzer: I am aware that it was not introduced by this PR and thus should also not block this PR. I just wanted to mention that there is a problem that should not have been merged in the first place and I hope there is already an existing issue entry to "work on it afterwards" - otherwise things tend to never get fixed.

features/org.openhab.designer.feature/feature.xml
kaikreuzer: #2407 is closed, I doubt anything will be done on it?
Here again, if there wasn't a dependency from the action to the binding, the ui dependency wouldn't also come in here.

bundles/action/org.openhab.action.tinkerforge/META-INF/MANIFEST.MF
teichsta: ok, see https://github.com/openhab/openhab/issues/2968

bundles/action/org.openhab.action.homematic/META-INF/MANIFEST.MF
kaikreuzer: > where each action call's first parameter is an item bound to the binding?

This imho mixes actions and bindings too heavily. An action should never really refer to an item where you need to know that a specific binding is configured to.

teichsta: see issue #2969

bundles/action/org.openhab.action.astro/META-INF/MANIFEST.MF
teichsta: see issue #2969

features/org.openhab.designer.feature/feature.xml
teichsta: added #2970 to track the removal of this dependency

gerrieg:
The homematic action requires a correctly configured homematic binding (no item). 
The astro action does not require a correctly configured binding, but uses the calculation methods in the binding.

If a action should not reference a binding, what's the correct way to implement this?

lolodomo:
I can confirm MiOS action is now visible in Designer 1.7.1.
Thank you

udo1toni:
I can confirm Squeezebox action is now visible in Designer 1.7.1
Thanks :)

PR #2973, opened 2015-07-27
added websocket functionality to /rest/items

querdenker2k:
added functionality to register to /rest/items for xml/json/jsonp for atmosphere (websocket) requests

### DIFF
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ResourceStateChangeListener.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ResourceStateChangeListener.java
@@ -54,6 +54,7 @@ abstract public class ResourceStateChangeListener {
 	
 	static ScheduledFuture<?> executorFuture;
 	
+	protected Item lastChange;
 	private Set<String> relevantItems = null;
 	private StateChangeListener stateChangeListener;
 	protected GeneralBroadcaster broadcaster;
@@ -149,6 +150,7 @@ abstract public class ResourceStateChangeListener {
 			}
 			
 			public void stateChanged(final Item item, State oldState, State newState) {
+				lastChange = item;
 				broadcaster.broadcast(item);
 //				Collection<AtmosphereResource> resources = broadcaster.getAtmosphereResources();
 //				if(!resources.isEmpty()) {
--- a/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ItemStateChangeListener.java
+++ b/bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ItemStateChangeListener.java
@@ -18,6 +18,7 @@ import org.openhab.core.items.Item;
 import org.openhab.io.rest.RESTApplication;
 import org.openhab.io.rest.internal.resources.ItemResource;
 import org.openhab.io.rest.internal.resources.ResponseTypeHelper;
+import org.openhab.ui.items.ItemUIRegistry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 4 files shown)


querdenker2k:
@digitaldan can you check this please

digitaldan:
Yep, I will review as soon as I can, thanks!

digitaldan:
Before I review, can you explain the lastItem variable and why its needed?

querdenker2k:
Yes, its needed because in the websocket callback i didn't have any chance 
to get the item which has changed. For the existing implementations the 
path was used for this, but for this new implementation the path is just 
/items.
Therefore i save the changed item.


Am 27. Juli 2015 10:55:04 nachm. schrieb Dan <notifications@github.com>:

teichsta:
jenkins, please rebuild

digitaldan:
Thanks, I'm a little hung up at work, I will review as soon as I can, thanks.

querdenker2k:
Just for your know, I'm in holidays for 3 weeks now and have no chance to do anything here, well i hope everything will be fine.

digitaldan:
Ok, I think I see, do you know what happens if more than one item change at the same time (or very close together)?  We had previous issues with this.

I think I should mention that Atmosphere is going away in openHAB 2 in favor of SSE (server side events) , the paperUI is a good example of how this works.  Mostly likely there will only be one last major release of openHAB 1.x, so I should mention anything that is done here will not carry over to the 2.x series.  I think the 1.x will focus on stability fixes and minors bugs, but new features will soon only be introduced into opehab 2.

querdenker2k:
Ok, but i didn't think that it helps me. There is no final release for openhab 2 so that i will not change my productive environment. Futhermore the documentation for openhab 2 is currently very weak. What does it mean soon? a week, a month, a year?
I didn't had problems with concurrent events for more then 100 items... so it would be nice if this feature can be merged.

teichsta:
> Thanks, I'm a little hung up at work, I will review as soon as I can, thanks.

@digitaldan any chance to dive into this PR? Best, Thomas E.-E.

digitaldan:
@teichsta Yep, I have some time this week to look.

bundles/io/org.openhab.io.rest/src/main/java/org/openhab/io/rest/internal/listeners/ResourceStateChangeListener.java
```diff
@@ -54,6 +54,7 @@
 	
 	static ScheduledFuture<?> executorFuture;
 	
+	protected Item lastChange;
```
digitaldan: Have you tested what happens if many things change at the same time?  Do you still get all the events?

querdenker2k: Yes, In openHAB are ~5 Items which are triggered by an addon this causes to ~5 new updates at the same time.

digitaldan: OK sounds good, since I don't really have a good way to test this without writing a WS client, I will trust you have thoroughly tested that both the websocket functionality and the existing rest functionality work correctly.  Thanks for your patience, its been a busy month.

digitaldan:
@teichsta this looks good to me, I think the changes are safe since the functionality should only change with websocket connections, something that is not supported today for this endpoint.

teichsta:
we should have this tested by a bigger group of people. So let's merge into master and see what they say. @querdenker2k could you please create an according post (https://community.openhab.org) and try to acquire more testers? Thanks, Thomas E.-E.

digitaldan:
Yep, I agree

querdenker2k:
Thanks for merging. Didn't know of this forum.
Done: https://community.openhab.org/t/new-websocket-for-registering-to-item-changes/2294

kaikreuzer:
> so I should mention anything that is done here will not carry over to the 2.x series.

This is imho very important. I am missing a bit the discussion on this issue why this feature has enough relevance to be included into a "severe-bug-fix only" release? @teichsta, has there been any discussion around this that I have missed?
Instead of asking many people to test it (and thus imply that this is an important new feature that the world should use), I rather think that a huge disclaimer should be posted (to the wiki as well) that this feature will only exist in 1.8 and won't be available anymore with 2.0 again.

querdenker2k:
This feature is available in OpenHAB 2. with SSE you can Register to all item changed AS well. I have tested it and it works like the same.

PR #2981, opened 2015-07-29
eBUS Major Update 1

csowada:
Major Update, add send support and rewritten item syntax

* New parameter `set` allows to set a value
* New syntax for property `id`, `command` and `set`
* Improved sender timings
* Improved cpu load for serial connections
* Restructured configurations
* Add new configurations for Vaillant VRC430,470,BAI00
* All offical configurations in english language
* Add CSV support to record raw telegrams
* Fix for openHAB 2.0 compatiblity
* Adjust some warning messages
* Workaround for serial close issue
* Smaller bugfixes and optimizations

PR #2982, opened 2015-07-29
eBUS Bugfix Update for 1.7.1

csowada:
-Fix wrong interpreted broadcasts as "no answer" telegrams
-optimized tcp network settings


teichsta:
Thanks for carving this out of #2981

PR #2984, opened 2015-07-29
support for external product database

paixaop:
Extract product database to the configurations folder so we don’t need to compile a new JAR for every product update.

Copied the database folder from inside the ZWave bundle to the openHab configurations folder.
A further enhancement would be to combine the two approaches. Load the database from the bundle, but search the configurations/zwave/database folder for extra products that the user can add.


cdjackson:
Sorry. As has been discussed a couple of times, I am not keen on this approach. This means that we have no control over the configuration, which can cause problems that are then impossible to identify.
Since we are close to the end of the OH1 zwave binding life, and OH2 is coming online, I don't intend to merge this. In OH2, the database will also not be external, and must be maintained within the binding (this is for ALL OH2 bindings) as again it's necessary in order for the binding to work.

PR #2988, opened 2015-07-30
Indicator class

paixaop:
Added support to the Indicator Command Class. Tested with Evolve LCD1. Button reports are working.

To use add a new Number item like so:

    Number Panel_GF_Kitchen_status   {zwave="11:0:command=INDICATOR"}

Then you can write a rule to check Button presses on the panel

    rule "Kitchen Panel Buttons"
    when
        Item Panel_GF_Kitchen_status changed
    then
        logInfo("Scenes", "Kitchen Button Panel Indicator Update")
        switch(Panel_GF_Kitchen_status.state) {
			case 0  : logInfo("Scene","No Buttons active")
			case 4  : logInfo("Scene","Button 1 Active")
			case 8  : logInfo("Scene","Button 2 Active")
			case 12 : logInfo("Scene","Button 1 and 2 Active")
		}
    end


bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommands.java
```diff
@@ -0,0 +1,18 @@
+/**
```
cdjackson: Why have you created this interface? I mean - what is the purpose - it's not used?

bundles/binding/org.openhab.binding.zwave/database/evolve/lcd1.xml
```diff
@@ -0,0 +1,647 @@
+<?xml version="1.0" encoding="utf-8"?>
```
cdjackson: Please remove this file.

bundles/binding/org.openhab.binding.zwave/database/products.xml
```diff
@@ -440,6 +440,15 @@
 			<Label lang="en">Wall Mounted Dimmer</Label>
```
cdjackson: Please revert this file

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
```diff
@@ -0,0 +1,154 @@
...
+import org.openhab.binding.zwave.internal.protocol.SerialMessage.SerialMessageType;
+import org.openhab.binding.zwave.internal.protocol.event.ZWaveCommandClassValueEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import com.thoughtworks.xstream.annotations.XStreamOmitField;
+
+/**
+ * Handles the Indicator command class.
```
cdjackson: Can you say what the command class does?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
```diff
@@ -0,0 +1,154 @@
...
+	@Override
+	public void handleApplicationCommandRequest(SerialMessage serialMessage,
+			int offset, int endpoint) {
+		logger.debug("NODE {}: Received Indicator Request", this.getNode().getNodeId());
+		int command = serialMessage.getMessagePayloadByte(offset);
+		switch (command) {
+			case INDICATOR_SET:
+				logger.debug("NODE {}: Indicator Set sent to the controller will be processed as Indicator Report", this.getNode().getNodeId());
+				// Process this as if it was a value report.
+				processIndicatorReport(serialMessage, offset, endpoint);
```
cdjackson: Are you sure this is correct, or is it something you copied from another class?

cdjackson:
Thanks for this. Just a few comments/questions :)

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
paixaop: In the Evolve LCD1 panel the indicator class is used to inform the controller and other panels about the state of the display. If you configure a button as a toggle button the display will invert,  i.e, go from back text on white background to white text on a black background, when the user taps it once, and go back to normal when the user taps it a second time. 

So say you use 4 panels to control kitchen lights from different locations, and button 1 is set to toggle the lights in every panel.
When the user taps button 1, in panel 1 the label will be inverted, indicating the light is on, and 
all other panels, included in the association group 16, will update their displays to reflect the label change, so they all sync up their label statuses through the indicator class.

Hope this makes it clear.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommands.java
paixaop: Honestly don't know. Being a newbie at this all I did was to copy what was implemented in the BasicClass and adapt it to the indicator class.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
paixaop: Copied from BasicClass, but it is working in my tests

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
cdjackson: Yep - thanks. What I meant is can you add a few lines about the class in the file - just so people reading it later know what it's about.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
cdjackson: Ok - how have you tested the binding receiving a SET? It would not normally receive this I think, which is why it's normally not implemented.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommands.java
cdjackson: Ok - then please delete it. There's no need to have this interface.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
paixaop: You should be able to set a device's indicator. Example set the panel label from an openHab rule.

paixaop:
I deleted the extra database files

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
cdjackson: Yes, but the binding shouldn't ever receive this message should it? If you SET the device indicator, then you SEND a SET - you don't receive it. If you request the state, then you RECEIVE a REPORT. There are only a couple of instances when you would ever reveice a SET - the BASIC command class does this because the BASIC command class is sent for a lot of different things, but I don't think any other command classes will ever receive the SET command unless something bad happens.

I could be wrong here though - do you actually see this command being received?

paixaop:
Merged all the latest changes into my fork, hopefully now it will merge cleanly.

cdjackson:
Thanks - I'm still waiting for the changes to be made before this can be merged (As above)

paixaop:
Chris Github now says the files will merge cleanly is there something else you need?

cdjackson:
Yes - I think there are still a number of outstanding issues before this can be resolved (as per the previous emails above).

paixaop:
Chris 

The function to set the indicator on the controller needs to stay so we can receive indicator updates and adjust the state of an indicator variable. So I would not remove that part of the code. 

Pedro

cdjackson:
As I said previously, I don't think the code is used - have you actually checked? The code that I mention is NOT used when you set the indicator - it's used if the binding RECEIVES a SET message - can you tell me when this will happen? Does it happen when you send a command? Have you tested this and can you confirm that it occurs as I don't think it should unless you are using this to set something inside the binding, which (I think) you aren't?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
```diff
@@ -0,0 +1,159 @@
...
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void handleApplicationCommandRequest(SerialMessage serialMessage,
+			int offset, int endpoint) {
+		logger.debug("NODE {}: Received Indicator Request", this.getNode().getNodeId());
+		int command = serialMessage.getMessagePayloadByte(offset);
+		switch (command) {
+			case INDICATOR_SET:
```
cdjackson: Just to be clear - this function is for receiving messages sent FROM the device - and as per my previous comments, I don't think the device will send a SET command and this should be removed unless you confirm that the device is sending SET commands.

bundles/binding/org.openhab.binding.zwave/database/products.xml
```diff
@@ -1,2600 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
cdjackson: It seems that in this latest version, every line of the products file is marked as changed. Please can you change this back so we don't loose the history?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveIndicatorCommandClass.java
paixaop: Chris

I tested this and the Evolve LCD1 sends indicator SET messages to all devices included in Association Group 17. This includes the binding. We have to leave this in.

cdjackson: Ok - thanks.  Let me know when you've made the other mods and we'll get this merged...

Cheers
Chris

cdjackson:
Since I see you're adding new classes, do you also intend to close this out as well?

paixaop:
Lets close this pull request and use the new one. I cannot clean this one up. Not well versed on git to be able to do it. 

Pedro

cdjackson:
Ok - that's fine. Please keep it separate from the Switch_All command class though...  Maybe we can finish that off first if you like?

paixaop:
Yes it is separate. None of the changes here are in the switch all class

paixaop:
a Clean Indicator Command Class implementation in pull request #3103 you can merge that one

PR #2996, opened 2015-08-01
Added devolo MT02648 to Z-Wave database

JensErat:


cdjackson:
Thanks - all looks good.

JensErat:
Do you want me to also add the other devices which are similar according to their product manual, but which I cannot verify afterwards (as I do not own them)?

cdjackson:
Up to you, but I don’t think it’s worth it unless you, or someone you know, is actually going to use the device soon? The problem with adding devices that aren’t tested is it can be difficult to know if it ever worked - the files themselves probably aren’t too bad, but the big issue I have is if we add something incorrectly to the products file, then it can be difficult to later work out what was tested when if someone adds a device with the same IDs...

PR #2998, opened 2015-08-02
Bticino Device : dimmed lights

e4glenight:
Hello !

Can we do something for Dimmer in Bticino Lights ? : 
Line 145  (file : org.openhab.binding.bticino-1.7.0-sources/org/openhab/binding/bticino/internal/BticinoDevice.java
>				switch (l_who) {
>				// Lights
>				case 1: {
>					if (OnOffType.ON.equals(command))
>						l_pr.addProperty("what", "1");
>					else
>						l_pr.addProperty("what", "0");
>					break;
>				}

This works sucessfully for turn on or turn off lights, but not for dimm them.

 For have some dimmable configuration, we should use what = 3 for 30%, for example (min : 0 = off / max 10 or 1 = on and each step is 10%)

Can we have one separate case for dimmed lights ? (with Dimmer attribut on .item file instead of Switch)
Or improve this part of code to use dimmer attribut ?

Thanks
(and worry about my java level... i'm begginer :))


teichsta:
@e4glenight you've created this PR by accident by assume. According to https://github.com/openhab/openhab/wiki/How-To-Contribute before opening issues here please your idea, potential issue in the Google Group.

Thanks, Thomas E.-E.

e4glenight:
Hello,

Sorry for opening one issues... i post again on google group :)

Thanks

PR #3011, opened 2015-08-05
Z-Wave Database Cleanup Aeon Door/Window Sensor

SkynetHome:
Edited for accuracy and portability into openhab2


cdjackson:
Same comment as the previous one - the commandClasses list is missing - this will stop the binding working properly with the device,

SkynetHome:
Re-committed and squashed!

PR #3012, opened 2015-08-05
Zwave Database Cleanup Aeon Minimote

SkynetHome:
Edited for readability and portability into OH2

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/aeon/minimote.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/aeon/minimote.xml
@@ -12,13 +12,14 @@
 		<Class><id>0x85</id></Class>
 		<Class><id>0x86</id></Class>
 	</CommandClasses>
-
 	<Configuration>
 		<Parameter>
 			<Index>241</Index>
 			<Type>list</Type>
 			<Default>0</Default>
 			<Size>1</Size>
+			<Label lang="en">Mode of Button 1 (upper left)</Label>
+			<Help lang="en">Defines the switching mode of Button 1</Help>
 			<Item>
 				<Value>0</Value>
 				<Label lang="en">Factory Default</Label>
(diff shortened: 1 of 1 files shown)


cdjackson:
Have you tested the devices you're changing? I'm worried that we may break something... For example, in this PR you've removed the commandClass list - while this isn't used in OH2 as I needed to work within the structure of the OH2 definitions, it is used in OH1 and without this list some things will stop working in the OH1 binding.

SkynetHome:
I am happy to leave the command classes in I just misunderstood the OH1 database wiki apparently to mean that they are only important in certain cases as I referenced on my other PR. I happy to leave them in. I was really intending to focus the clean up to missing parameters, parameters lacking min max values various values that are just wrong and cleaning up some the language in the help sections and adding CDATA where applicable to make way for OH2. If you feel this is all better done at a later date just say the word and I'll get back to working on my Java Development Classes. Just happy to help out here in any way I can but dont want to create more headache for you than its worth!

cdjackson:
No probs - and I appreciate the help. The problem is that I can't easily check these changes and as the whole file changes, it's hard to know what's what... If you're only changing the comments, then it's probably ok - especially so long as you check that Eclipse (or some other XML validator) shows the XML is ok (for example if you miss the ]]> off the end of the CDATA escaping, then it would screw the file).

So long as this is all you're changing, then I'm fine with it - otherwise it might be best to leave it to the OH2 conversion and do everything at once... Up to you though - I don't want to dampen your enthusiasm as it's all good work :)

SkynetHome:
Sounds good...I have been vigilant about closing tags and checking for irregularities. Now that i know to leave CommandClasses alone I would like to take a stab at straightening some of this out. If it becomes a headache just let me know and rest assured should a problem emerge from any changes of mine that you merge I will personally work quickly and diligently to get it straightened out, i'll not let that fall to you to wade through!

SkynetHome:
re-added command classes and squashed

Issue #3005, opened 2015-08-03
Z-Wave: ZW100-A Aeon Labs Gen5 MultiSensor 6 config parameters


pabloNZ:
Missing parameters: 
- 0x07(7) 
  Enable/disable the function of vibration sensor.
  Value =0, disable.
  Value= 1, enable.
  Default=1
  Size=1
- 0x64(100) 
  Set 101‐103 to default.
- 0x6E(110)
  Set 111‐113 to default.

Extra parameters present: 39, 45
These do not appear to be in the engineering spec sheet as per: https://github.com/robertvandervoort/SmartThings/blob/master/Aeon%20Multisensor%206/Engineering%20Spec%20-%20Aeon%20Labs%20MultiSensor%206.pdf
##

SkynetHome:
I will try and get this done tonight.

cdjackson:
@pabloNZ Please confirm if this can not we closed

pabloNZ:
Hi, sorry, I have had some family stuff pop up short notice. I pulled in the 1.8 snapshot last night and it appears to work the same as the old one, but with the added parameters. I'm still having other issues with the sensor (One of them won't accept changes to parameters) but can't see any issues with the changes made. Do you need any other feedback?

cdjackson:
I think there have been further changes to the MS6 config - can we close this?

pabloNZ:
I'm away from my setup for a couple of weeks I'm afraid so can't confirm if
recent nightly builds continue the good work you've been doing. If what I
saw late last year is anything to go by then you've got it covered.

9037568:
Presumably this is fixed. Please start a new ticket if you find additional problems.

PR #3013, opened 2015-08-05
Zwave Database Cleanup Aeon SES3

SkynetHome:
Edited SES3 adding missing configuration parameters, correcting values,
editing help and formatting for portability into OH2

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/aeon/ses3.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/aeon/ses3.xml
@@ -1,250 +1,280 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Product>
-        <Model>SES3</Model>
-        <Label lang="en">Smart Energy Switch 3rd Edition</Label>
-        <CommandClasses>
-                <Class><id>0x00</id></Class>
-                <Class><id>0x20</id></Class>
-                <Class><id>0x25</id></Class>
-                <Class><id>0x27</id></Class>
-                <Class><id>0x32</id></Class>
-                <Class><id>0x2C</id></Class>
-                <Class><id>0x2B</id></Class>
-                <Class><id>0x70</id></Class>
-                <Class><id>0x85</id></Class>
-                <Class><id>0x72</id></Class>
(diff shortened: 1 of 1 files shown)


cdjackson:
As this is a large change I missed the fact that you've removed the commandClasses list. This will stop this device working properly - can you add it back in please.

I'm not sure that changing files in OH1 is the right move - I don't want to break OH1, and I think we'd probably be better off doing these changes when we convert to OH2 rather than risk breaking devices in OH1...  IF you want to fix the formatting for devices you own, and can therefore test, then fine, but otherwise I think it's best to leave the changes to when we do the conversions...

SkynetHome:
Np will put the commandClasses back in. I thought I understood from the wiki that they had no effect. put them all in or none in and that only certain classes with values such as readonly or writeonly were necessary but. No worries I will replace the commandClasses!

cdjackson:
It's used in a couple of places - I forget exactly what, but I think if the command class list doesn't include certain values (eg associations) that it won't set the associations on startup... It isn't used for much, but it is used...

SkynetHome:
Np I am on it. While you are here did you write the binding for the Aeon Multi Sensor 6. There is an issue regarding it that I was working on. One parameter is in fact missing and I have added. It the parameters referring to disabling 101-103 show no values in the documentation and seem better left alone unless you know of standard values to put in there. The issue is #3005

SkynetHome:
recommitted changes but they are not showing up here from what I can see and I have no option to issue a new PR

SkynetHome:
@cdjackson It finally came around. New PR to set this straight is now open! Thanks for your patience! I'll try desperately to make sure that my love of working on this project does not become a thorn in your ass! :)

cdjackson:
No worries - I'm happy to help - you're helping me a lot, so I'm happy to help :)

SkynetHome:
LOL...well it makes me happy, that you are happy, to help me, help you, which in turn helps me. :) ......and I think that is my cue that I have stayed up to late again and need to get some sleep. I'll see if I cant manage to send more help your way tomorrow!

PR #3014, opened 2015-08-05
Edit Aeon Multi-Sensor 6

SkynetHome:
Added parameter 7 but am awaiting feedback on other issues. I am
hesitant to remove what may appear to be superfluous parameters as
documentation does refer to some of those values. As for adding a
parameter to set 101 - 103 to default there are no values given in the
documentation for doing that. I will await feedback on affecting those
changes as well. ref Issue #3005

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/aeon/zw100-a.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/aeon/zw100-a.xml
@@ -128,6 +128,21 @@
 			</Help>
 		</Parameter>
 		<Parameter>
+			<Index>7</Index>
+			<Type>list</Type>
+			<Default>1</Default>
+			<Size>1</Size>
+			<Label lang="en">Enable or Disable the function of the vibration sensor</Label>
+			<Item>
+				<Value>1</Value>
+				<Label lang="en">Disable</Label>
+			</Item>
+			<Item>
+				<Value>2</Value>
+				<Label lang="en">Enable</Label>
+			</Item>
+		</Parameter>
+		<Parameter>
 			<Index>39</Index>
 			<Type>short</Type>
 			<Size>1</Size>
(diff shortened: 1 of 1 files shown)


cdjackson:
I will merge this, but @pabloNZ please provide feedback ASAP that this parameter is working. The issue is that if this parameter isn't real, it may cause some problems with the initialisation of the device.

SkynetHome:
Thanks @cdjackson I will stay on top of this and the other cleanups I am working on. If any problems present from the changes made from the cleanup process I will rectify them immediately and then leave things as they are. Hopefully this will not occur and I can satisfy my OCD with a more uniform and streamlined product database!! I think uniformity in the database leads to better readability and hopefully stability across all products, and should make it easier for other users to create database entries for new products when the other devices in the database are good consistent examples to draw from.

cdjackson:
Thanks Mike - I completely agree and again really appreciate your help :)

SkynetHome:
@cdjackson While cleaning up and formatting the database, would you prefer an individual PR for each device file that I change or would you prefer a group at a time all in one PR?

SkynetHome:
@cdjackson LOL so I am impatient and went ahead issued individual PR's so that they would be easier to handle one by one. If you would prefer to get PR's with several files when I have time to do more than one I am happy to commit them all to one PR.

cdjackson:
No probs - single PRs are fine as it means we can approve them easier…

Just one further ‘warning’ - there is still some debate about adding the HTML tags. I thin this will be agreed, but see here -: https://www.eclipse.org/forums/index.php/t/1068570/ <https://www.eclipse.org/forums/index.php/t/1068570/>

SkynetHome:
Wow, interesting read, thanks for the link. I didnt realize I had stepped into such a highly debated topic. As I am currently not doing any web development I only ever come across markdown in comment sections like here on github or youtube. Where more often than not its an annoyance for me as it seems to do the exactly the opposite of what it claims. WYSIWYG is certainly not WYG  when you type something into a comment expecting it to look exactly as you typed it only to find out that what you typed is not just plain text but in fact code that is going to be interpreted as something you never intended it to be. That being said, I am sure my annoyance with markdown is a product of my own ignorance regarding it and not necessarily a shortcoming on its behalf. Again thanks for the link and the warning that things are still a bit up in the air regarding the transition from OH to OH2. I will reserve comment regarding OH2 as I dont want to rush to judgement regarding something in its alpha stage but I will say that thus far, from this end users perspective and very limited amount of time using it...I am not terribly excited. While I fully appreciate that a more "user friendly" ie less coding dependent platform will likely be embraced by a broader range of users, the control freak in me goes a little crazy when OH2 just starts creating things for me. Having to manually setup everything in openhab while less user friendly in the beginning, forced me to learn about the software, how it works, and how to manipulate it. It forced me to learn something new and in the process develop a much deeper understanding of not only the devices that I have deployed around my home themselves but the software that controls those devices. That makes me feel more in control than a system that does everything for you and allows a user to remain rather ignorant to whats really going on underneath. (Hence the screen name SkynetHome) But that's just me and I am quite certain that I dont represent the majority of homeowners that OH2 is targeting. But as to the topic at hand I will keep in mnd that what I am working on is not fully embraced. It appears that the debate is leaning toward the format I have been using so if you are comfortable merging the database files that I am creating then I am comfortable with the fact that things may move in a different direction in the future and I may need to redo what I am currently working on. I am also currently learning Java and C/C++ (for Arduino Development) soon to be working on C#, JavaScript, CSS and PHP so regardless of what direction openhab goes it should continue to be a project that I can employ my ever increasing knowledge of coding on and that seems like a win to me!

Issue #2842, opened 2015-07-02
Error when refreshing the item file


lolodomo:
While openHAB server is running, if I save my item file, I get this error again and again:

2015-07-02 22:32:23.802 [INFO ] [c.internal.ModelRepositoryImpl] - Refreshing model 'maison.items'
2015-07-02 22:32:24.705 [ERROR] [.o.m.c.i.folder.FolderObserver] - An unexpected exception has occured
java.lang.NullPointerException: null
        at org.openhab.model.item.binding.AbstractGenericBindingProvider.removeConfigurations(AbstractGenericBindingProvider.java:97) ~[na:na]
        at org.openhab.model.item.internal.GenericItemProvider.processBindingConfigsFromModel(GenericItemProvider.java:164) ~[na:na]
        at org.openhab.model.item.internal.GenericItemProvider.modelChanged(GenericItemProvider.java:390) ~[na:na]
        at org.openhab.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:159) ~[na:na]
        at org.openhab.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:100) ~[na:na]
        at org.openhab.model.core.internal.folder.FolderObserver.checkFolder(FolderObserver.java:142) ~[na:na]
        at org.openhab.model.core.internal.folder.FolderObserver.run(FolderObserver.java:99) ~[na:na]

It occurs with openHAB 1.7.

watou:
Looks like it might be the same as #1419.

hmerk:
See comment on #1419

Issue #1419, opened 2014-09-12
NPE when refresh item file


sebglon:
Openhab 1.5.0

When i change my item file, i get this error:

19:11:43.409 INFO  o.o.m.c.i.ModelRepositoryImpl[:97]- Refreshing model 'chaudiere.items'
19:11:43.526 ERROR o.o.m.c.i.f.FolderObserver[:105]- An unexpected exception has occured
java.lang.NullPointerException: null
	at org.openhab.model.item.binding.AbstractGenericBindingProvider.removeConfigurations(AbstractGenericBindingProvider.java:93)
	at org.openhab.model.item.internal.GenericItemProvider.processBindingConfigsFromModel(GenericItemProvider.java:164)
	at org.openhab.model.item.internal.GenericItemProvider.modelChanged(GenericItemProvider.java:379)
	at org.openhab.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:158)
	at org.openhab.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:99)
	at org.openhab.model.core.internal.folder.FolderObserver.checkFolder(FolderObserver.java:142)

teichsta:
see also #736

watou:
Just to note, PR #3016 might have corrected #2842, #3218, and this issue (#1419).  Could anyone run a recent nightly build and attempt to reproduce the problem?

steve-bate:
Only critical bug fixes will be made to the openHAB 1 core, designer tools and user interfaces since core development is focussed on openHAB 2.

PR #3016, opened 2015-08-05
Improve thread safety around contextMap

watou:
Trying to find the proper fix for #2842 and #1419.  Found lack of thread safety around protected `contextMap` member of `AbstractGenericBindingProvider`.  Also found bindings performing unnecessary and bug-inducing changes in their overrides of `processBindingConfiguration` calls.

Please review this code carefully (as I'm sure you would anyway).  I've added line comments in the changes to explain my thinking.  These changes are not tested, but here to spur others' thinking about the origins of #2842 and #1419.

### DIFF
--- a/bundles/binding/org.openhab.binding.autelis/src/main/java/org/openhab/binding/autelis/internal/AutelisGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.autelis/src/main/java/org/openhab/binding/autelis/internal/AutelisGenericBindingProvider.java
@@ -8,7 +8,6 @@
  */
 package org.openhab.binding.autelis.internal;
 
-import java.util.HashSet;
 import java.util.Set;
 
 import org.openhab.binding.autelis.AutelisBindingProvider;
@@ -68,9 +67,6 @@ public class AutelisGenericBindingProvider extends AbstractGenericBindingProvide
 	 */
 	@Override
 	public void processBindingConfiguration(String context, Item item, String bindingConfig) throws BindingConfigParseException {
-		super.processBindingConfiguration(context, item, bindingConfig);
-		
-		
 		//read values must contain a '.' like equipment.circuit1 or be lightcmd
 		if(bindingConfig.indexOf('.') < 0 && !bindingConfig.equalsIgnoreCase("lightscmd")){
 			logger.warn("Item {}'s configuration ({}) is not valid, please use the pattern 'parentType.childType' or lightscmd ", item.getName(), bindingConfig);
@@ -80,13 +76,8 @@ public class AutelisGenericBindingProvider extends AbstractGenericBindingProvide
 		AutelisBindingConfig config = new AutelisBindingConfig(bindingConfig);
 		
 		addBindingConfig(item, config);	
-		
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
+
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
 
 	/**
--- a/bundles/binding/org.openhab.binding.digitalstrom/src/main/java/org/openhab/binding/digitalstrom/internal/DigitalSTROMGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.digitalstrom/src/main/java/org/openhab/binding/digitalstrom/internal/DigitalSTROMGenericBindingProvider.java
@@ -9,9 +9,7 @@
 package org.openhab.binding.digitalstrom.internal;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Set;
 
 import org.openhab.binding.digitalstrom.DigitalSTROMBindingProvider;
 import org.openhab.binding.digitalstrom.internal.config.DigitalSTROMBindingConfig;
@@ -120,10 +118,5 @@ public class DigitalSTROMGenericBindingProvider extends AbstractGenericBindingPr
 			}
 		}
 		return deviceConsumptionItems;
-	}
-
-	@Override
-	public Set<Item> getItemNamesByContext(String context) {
-		return new HashMap<String, Set<Item>>(super.contextMap).get(context);
 	}
 }
--- a/bundles/binding/org.openhab.binding.dscalarm/src/main/java/org/openhab/binding/dscalarm/internal/DSCAlarmGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.dscalarm/src/main/java/org/openhab/binding/dscalarm/internal/DSCAlarmGenericBindingProvider.java
@@ -59,7 +59,6 @@ public class DSCAlarmGenericBindingProvider extends AbstractGenericBindingProvid
 	 * {@inheritDoc}
 	 */
 	public void processBindingConfiguration(String context, Item item, String bindingConfig) throws BindingConfigParseException {
-		super.processBindingConfiguration(context, item, bindingConfig);
 
 		String[] sections = bindingConfig.split(":");
 		
@@ -107,13 +106,8 @@ public class DSCAlarmGenericBindingProvider extends AbstractGenericBindingProvid
 		
 		DSCAlarmBindingConfig config = new DSCAlarmBindingConfig(dscAlarmDeviceType, partitionId, zoneId, dscAlarmItemType);
 		addBindingConfig(item, config);
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
 		
+		super.processBindingConfiguration(context, item, bindingConfig);
 		logger.debug("processBindingConfiguration(): Item added: {} - {}",item.getName(),bindingConfig);
 	}
 	
--- a/bundles/binding/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/config/EnoceanGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/config/EnoceanGenericBindingProvider.java
@@ -77,17 +77,6 @@ public class EnoceanGenericBindingProvider extends AbstractGenericBindingProvide
     }
 
     @Override
-    public void removeConfigurations(String context) {
-        Set<Item> configuredItems = contextMap.get(context);
-        if (configuredItems != null) {
-            for (Item item : configuredItems) {
-                items.remove(item.getName());
-            }
-        }
-        super.removeConfigurations(context);
-    }
-
-    @Override
     public EnoceanParameterAddress getParameterAddress(String itemName) {
         EnoceanBindingConfig config = (EnoceanBindingConfig) bindingConfigs.get(itemName);
         return config != null ? new EnoceanParameterAddress(EnoceanId.fromString(config.id), config.channel, config.parameter) : null;
--- a/bundles/binding/org.openhab.binding.freeswitch/src/main/java/org/openhab/binding/freeswitch/internal/FreeswitchGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.freeswitch/src/main/java/org/openhab/binding/freeswitch/internal/FreeswitchGenericBindingProvider.java
@@ -55,7 +55,6 @@ public class FreeswitchGenericBindingProvider extends AbstractGenericBindingProv
 	 */
 	@Override
 	public void processBindingConfiguration(String context, Item item, String bindingConfig) throws BindingConfigParseException {
-		super.processBindingConfiguration(context, item, bindingConfig);
 		
 		String[] configParts = bindingConfig.trim().split(":", 2);
 		
@@ -74,13 +73,7 @@ public class FreeswitchGenericBindingProvider extends AbstractGenericBindingProv
 		FreeswitchBindingConfig config = new FreeswitchBindingConfig(item.getName(),item.getClass(), type, argument);
 		
 		addBindingConfig(item, config);
-		
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
 
 	@Override
--- a/bundles/binding/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubGenericBindingProvider.java
@@ -62,7 +62,6 @@ public class HarmonyHubGenericBindingProvider extends AbstractGenericBindingProv
 	 */
 	@Override
 	public void processBindingConfiguration(String context, Item item, String bindingConfig) throws BindingConfigParseException {
-		super.processBindingConfiguration(context, item, bindingConfig);
 
 		Matcher m = bindingPattern.matcher(bindingConfig);
 
@@ -105,12 +104,7 @@ public class HarmonyHubGenericBindingProvider extends AbstractGenericBindingProv
 			HarmonyHubBindingConfig config = new HarmonyHubBindingConfig(type,qualifier,type.getLabel(),param1,param2, item.getClass());
 
 			addBindingConfig(item, config);		
-			Set<Item> items = contextMap.get(context);
-			if (items == null) {
-				items = new HashSet<Item>();
-				contextMap.put(context, items);
-			}
-			items.add(item);
+			super.processBindingConfiguration(context, item, bindingConfig);
 		} else {
 			throw new BindingConfigParseException("Config string did not match pattern { harmonyhub=\"<binding>[ (qualifier:)<binding> ...]\" }");
 		}
--- a/bundles/binding/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/OmniLinkGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/OmniLinkGenericBindingProvider.java
@@ -75,12 +75,7 @@ public class OmniLinkGenericBindingProvider extends
 				new Object[] { config.getObjectType(), config.getNumber() });
 		addBindingConfig(item, config);
 
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
 
 	@Override
--- a/bundles/binding/org.openhab.binding.onewire/src/main/java/org/openhab/binding/onewire/internal/OneWireGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.onewire/src/main/java/org/openhab/binding/onewire/internal/OneWireGenericBindingProvider.java
@@ -61,12 +61,11 @@ public class OneWireGenericBindingProvider extends AbstractGenericBindingProvide
 	 */
 	@Override
 	public void processBindingConfiguration(String pvContext, Item pvItem, String pvBindingConfig) throws BindingConfigParseException {
-		super.processBindingConfiguration(pvContext, pvItem, pvBindingConfig);
-
 		OneWireBindingConfig pvDevicePropertyBindingConfig = OneWireBindingConfigFactory.createOneWireDeviceProperty(pvItem, pvBindingConfig);
 
 		addBindingConfig(pvItem, pvDevicePropertyBindingConfig);
 
+		super.processBindingConfiguration(pvContext, pvItem, pvBindingConfig);
 	}
 
 	/* (non-Javadoc)
--- a/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelGenericBindingProvider.java
@@ -53,11 +53,11 @@ public class SatelGenericBindingProvider extends AbstractGenericBindingProvider 
 	public void processBindingConfiguration(String context, Item item, String bindingConfig)
 			throws BindingConfigParseException {
 		logger.trace("Processing binding configuration for item {}", item.getName());
-		super.processBindingConfiguration(context, item, bindingConfig);
 
 		SatelBindingConfig bc = this.createBindingConfig(bindingConfig);
 		logger.trace("Adding binding configuration for item {}: {}", item.getName(), bc);
 		addBindingConfig(item, bc);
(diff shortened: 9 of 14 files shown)


bundles/binding/org.openhab.binding.autelis/src/main/java/org/openhab/binding/autelis/internal/AutelisGenericBindingProvider.java
```diff
@@ -80,13 +76,8 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
-		
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
+
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
```
watou: Removed code duplicated what super.processBindingConfiguration had already done, so moved super call to end.

bundles/binding/org.openhab.binding.digitalstrom/src/main/java/org/openhab/binding/digitalstrom/internal/DigitalSTROMGenericBindingProvider.java
```diff
@@ -124,6 +124,6 @@ public void removeConfigurations(String context) {
 
 	@Override
 	public Set<Item> getItemNamesByContext(String context) {
-		return new HashMap<String, Set<Item>>(super.contextMap).get(context);
+		return contextMap.get(context);
 	}
```
watou: Replaced creation of duplicate contextMap and extraction of set from it, with simple extraction of set from contextMap.

bundles/binding/org.openhab.binding.dscalarm/src/main/java/org/openhab/binding/dscalarm/internal/DSCAlarmGenericBindingProvider.java
```diff
@@ -107,13 +106,8 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
 		addBindingConfig(item, config);
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
 		
+		super.processBindingConfiguration(context, item, bindingConfig);
 		logger.debug("processBindingConfiguration(): Item added: {} - {}",item.getName(),bindingConfig);
```
watou: Removed code duplicated what super.processBindingConfiguration had already done, so moved super call to end.

bundles/binding/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/config/EnoceanGenericBindingProvider.java
```diff
@@ -77,17 +77,6 @@ protected void addBindingConfig(Item item, BindingConfig config) {
...
-        if (configuredItems != null) {
-            for (Item item : configuredItems) {
-                items.remove(item.getName());
-            }
-        }
-        super.removeConfigurations(context);
-    }
-
-    @Override
     public EnoceanParameterAddress getParameterAddress(String itemName) {
```
watou: This override performed an unnecessary one-by-one removal of items from the contextMap, but introduced problem where listeners would not be called and items would not be removed from bindingConfigs map.  Removing method altogether should remove the bug.

bundles/binding/org.openhab.binding.freeswitch/src/main/java/org/openhab/binding/freeswitch/internal/FreeswitchGenericBindingProvider.java
```diff
@@ -74,13 +73,7 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
 		addBindingConfig(item, config);
-		
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
```
watou: Removed code duplicated what super.processBindingConfiguration had already done, so moved super call to end.

bundles/binding/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubGenericBindingProvider.java
```diff
@@ -105,12 +104,7 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
 
 			addBindingConfig(item, config);		
-			Set<Item> items = contextMap.get(context);
-			if (items == null) {
-				items = new HashSet<Item>();
-				contextMap.put(context, items);
-			}
-			items.add(item);
+			super.processBindingConfiguration(context, item, bindingConfig);
 		} else {
```
watou: Removed code duplicated what super.processBindingConfiguration had already done, so moved super call to end.

bundles/binding/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/OmniLinkGenericBindingProvider.java
```diff
@@ -75,12 +75,7 @@ public void processBindingConfiguration(String context, Item item,
...
 		addBindingConfig(item, config);
 
-		Set<Item> items = contextMap.get(context);
-		if (items == null) {
-			items = new HashSet<Item>();
-			contextMap.put(context, items);
-		}
-		items.add(item);
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
```
watou: super.processBindingConfiguration was never called, but method attempted to achieve equivalent result by hand.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelGenericBindingProvider.java
```diff
@@ -53,11 +53,11 @@ public void validateItemType(Item item, String bindingConfig) throws BindingConf
 	public void processBindingConfiguration(String context, Item item, String bindingConfig)
 			throws BindingConfigParseException {
 		logger.trace("Processing binding configuration for item {}", item.getName());
-		super.processBindingConfiguration(context, item, bindingConfig);
 
 		SatelBindingConfig bc = this.createBindingConfig(bindingConfig);
 		logger.trace("Adding binding configuration for item {}: {}", item.getName(), bc);
 		addBindingConfig(item, bc);
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
```
watou: Moved super.processBindingConfiguration to end of override, since previous steps could fail and we don't want invalid binding config objects in the system that were rejected.

bundles/binding/org.openhab.binding.onewire/src/main/java/org/openhab/binding/onewire/internal/OneWireGenericBindingProvider.java
```diff
@@ -61,12 +61,11 @@ public void validateItemType(Item pvItem, String pvBindingConfig) throws Binding
...
 	@Override
 	public void processBindingConfiguration(String pvContext, Item pvItem, String pvBindingConfig) throws BindingConfigParseException {
-		super.processBindingConfiguration(pvContext, pvItem, pvBindingConfig);
-
 		OneWireBindingConfig pvDevicePropertyBindingConfig = OneWireBindingConfigFactory.createOneWireDeviceProperty(pvItem, pvBindingConfig);
 
 		addBindingConfig(pvItem, pvDevicePropertyBindingConfig);
 
+		super.processBindingConfiguration(pvContext, pvItem, pvBindingConfig);
 	}
```
watou: Moved super.processBindingConfiguration to end of override, since previous steps could fail and we don't want invalid binding config objects in the system that were rejected.

bundles/binding/org.openhab.binding.xpl/src/main/java/org/openhab/binding/xpl/internal/XplGenericBindingProvider.java
```diff
@@ -159,6 +158,7 @@ else if (configParts[1].equalsIgnoreCase("COMMAND"))
 	    } 
 		
 		addBindingConfig(item, config);
+		super.processBindingConfiguration(context, item, bindingConfig);
 	}
```
watou: Moved super.processBindingConfiguration to end of override, since previous steps could fail and we don't want invalid binding config objects in the system that were rejected.

bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
```diff
@@ -42,7 +41,7 @@
 	private Set<BindingChangeListener> listeners = new CopyOnWriteArraySet<BindingChangeListener>();
 
 	/** caches binding configurations. maps itemNames to {@link BindingConfig}s */
-	protected Map<String, BindingConfig> bindingConfigs = new ConcurrentHashMap<String, BindingConfig>(new WeakHashMap<String, BindingConfig>());
+	protected Map<String, BindingConfig> bindingConfigs = new ConcurrentHashMap<String, BindingConfig>();
```
watou: Constructor was only copying, not wrapping, the new WeakHashMap, so removed.

bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
```diff
@@ -77,27 +76,31 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
-			contextMap.put(context, items);
+		synchronized (contextMap) {
+			Set<Item> items = contextMap.get(context);
+			if (items==null) {
+				items = new HashSet<Item>();
+				contextMap.put(context, items);
+			}
+
+			items.add(item);
 		}
```
watou: Needed to make contextMap.get and contextMap.put one atomic operation, so put in a synchronized block.

bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
```diff
@@ -77,27 +76,31 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
+		synchronized (contextMap) {
+			Set<Item> items = contextMap.get(context);
+			if(items!=null) {
+				for(Item item : items) {
+					// we remove all binding configurations for all items
+					bindingConfigs.remove(item.getName());
+					notifyListeners(item);
+				}
+				contextMap.remove(context);
 			}
```
watou: Needed to make contextMap.get and contextMap.remove one atomic operation, so put in a synchronized block.

watou:
This PR goes with [this Google Group thread](https://groups.google.com/forum/#!category-topic/openhab/p46JDUy3Ev8).

bundles/binding/org.openhab.binding.digitalstrom/src/main/java/org/openhab/binding/digitalstrom/internal/DigitalSTROMGenericBindingProvider.java
```diff
@@ -121,9 +119,4 @@ public void removeConfigurations(String context) {
 		}
 		return deviceConsumptionItems;
 	}
-
-	@Override
-	public Set<Item> getItemNamesByContext(String context) {
-		return new HashMap<String, Set<Item>>(super.contextMap).get(context);
-	}
-}
```
watou: Removed unused `getItemNamesByContext` method since it exposes non-threadsafe Set of Items.

bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
mrguessed: You have an internal/framework lock held when you make calls, via `notifyListeners`, to non-framework code.

It would be better if you bucketed _what_ to callout, while inside the framework lock, and then made the callouts themselves outside the lock...otherwise, depending upon the locking model of all of the called entities (Bindings), you could create a deadlock.


PS: You may want to run the Eclipse code-formatting function over the changes to align with the std openHAB uses.

watou: Thanks very much for having a look, @mrguessed.  I was hoping you would take an interest.  I will move the loop to call `notifyListeners` outside the lock as you suggest.  I didn't want to re-format the changes for fear it would hurt readability, but I see your point.  Any other thoughts/ideas/warnings most appreciated.

bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
```diff
@@ -77,27 +76,34 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
+		}
+		if (items != null) {
+			for (Item item : items) {
 				// we remove all binding configurations for all items
 				bindingConfigs.remove(item.getName());
 				notifyListeners(item);
 			}
-			contextMap.remove(context);
 		}
 	}
```
watou: Is this what you meant, @mrguessed ?

mrguessed: Generally, yes.  That said, I'm not sure how much cross-consistency is expected, by Core, between the `contextMap` and the `bindingConfig` entries.  

The change now has us removing the `bindingConfig` entry _after_ the `contextMap` entry, which reverses the logical ordering from the original code.  I'm not familiar enough with this layer to know if that's a problem or not.

watou: Thanks for taking another look.  I also don't know if there are side effects in changing the ordering, but what I do know is that now, if multiple threads are calling `removeConfigurations` with the same `context`, as I suspect is what is causing to the NPEs, only the first will have any effect, and the others will essentially be no-ops.  Unless there is an intervening `processBindingConfiguration`, in which case the next thread to call `removeConfigurations` will be iterating over a completely different `Set` from the previous thread to enter `removeConfigurations`.  In other words, I don't see how multiple threads can step on each other when contending for the `contextMap` after these changes.

There are most likely higher-level coordination issues when files are reloaded, but I believe (and tell me if I'm wrong) that at least this code, in isolation, appears threadsafe.

mrguessed: It's threadsafe, but my concern about remove [contextMap] before notify [bindingConfig] needs to be validated.  For example, with the changes in place someone previously using the `contextMap` object, via some sort of call into these Core methods during the `notifyListeners` phase might get a surprise ;)  

I don't know if that can actually happen, since I haven't looked over this part in that much detail, so it's theoretical at this point.

watou: Some bindings' BindingProvider implementations have a `getItem` method that references the `contextMap` to find an item by name, but I don't know if this code is ever reached from `notifyListeners` in Core.  My opinion, subject to @teichsta or @kaikreuzer saying otherwise, is that my changes are more correct, in that during a `removeConfigurations` call, bindings ought not be trying to locate items by name, since the bindings are in the process of being torn down.

watou:
It appears that the only reason current master works when *.items files are touched is _because of_ a thread concurrency issue in higher-level code.  Looking at the following sequence with the Ecobee binding, for example, the current Ecobee binding ought to create a NullPointerException every time:

* touch ecobee.items -> ... ->
 *  org.openhab.model.item.internal.GenericItemProvider.processBindingConfigsFromModel ->
  * org.openhab.model.item.binding.AbstractGenericBindingProvider.removeConfigurations ->
   * for each item in the contextMap set:
     * remove the item by name from bindingConfigs, THEN
     * notifyListeners, which calls bindingChanged in the EcobeeBinding, which then consults bindingConfigs for the item name, but it was just removed, so NPE.

So the only way this works is if EcobeeBinding.bindingChanged is never called, or some other thread has already added back the refreshed binding config object in a race condition.

In other words, while this PR fixes thread concurrency issues in AbstractGenericBindingProvider, there are higher-level side effects that actually depend on thread concurrency side effects for the current bindings to work.  I don't know to what degree these issues exist in OH2, but it's very seriously worth a look if there is no intention to fix these issues in OH1.

teichsta:
i'll have a look into this in the course of the next week. Thanks so much for all the analysis you've so far!

watou:
Hey @teichsta, I squashed this commit just now.  Please have a look when you can.  Thanks!

teichsta:
i had a look into the code and i agree with @watous changes. For the next weeks (after the merge) we should keep an eye on all issues that somehow "smell" like caused by this issue.

teichsta:
Thanks a million @watou and @mrguessed for driving this issue and the fruitful discussions around this PR!

watou:
@teichsta Thanks, Thomas.  Did you mean to close this PR without merge? :smile:

teichsta:
(thanks :blush:)

lolodomo:
Comparing 1.7.1 and current 1.8 snapshot, I can confirm that something was really fixed as the refresh of my items is now working well. I still see the Null Pointer Exception but apparently it does not break the reload process anymore.
Good.

watou:
Thank you for the report, @lolodomo.  Do you have an example of the NullPointerException stack trace against the 1.8 snapshot?  I'm really interested in seeing what it looks like, as I was expecting that there would be no exceptions thrown after the change.

lolodomo:
Here is the stack trace I get:

12:15:44.906 [ERROR] [.o.m.c.i.folder.FolderObserver:105  ] - An unexpected exception has occured
java.lang.NullPointerException: null
        at org.openhab.model.item.binding.AbstractGenericBindingProvider.removeConfigurations(AbstractGenericBindingProvider.java:104) ~[na:na]
        at org.openhab.model.item.internal.GenericItemProvider.processBindingConfigsFromModel(GenericItemProvider.java:164) ~[na:na]
        at org.openhab.model.item.internal.GenericItemProvider.modelChanged(GenericItemProvider.java:390) ~[na:na]
        at org.openhab.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:159) ~[na:na]
        at org.openhab.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:100) ~[na:na]
        at org.openhab.model.core.internal.folder.FolderObserver.checkFolder(FolderObserver.java:142) ~[na:na]
        at org.openhab.model.core.internal.folder.FolderObserver.run(FolderObserver.java:99) ~[na:na]

watou:
Thanks for the stack trace. I will investigate this shortly.

watou:
Hi @lolodomo, it's very odd that either `bindingConfigs` or `item` would be `null` at [line 104 in AbstractGenericBindingProvider.java](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java#L104).  I am running a private build from today's master branch, installed from .deb files on an RPi2/wheezy, including these addons:
```
org.openhab.action.ecobee-1.8.0-SNAPSHOT.jar
org.openhab.action.mail-1.8.0-SNAPSHOT.jar
org.openhab.action.twitter-1.8.0-SNAPSHOT.jar
org.openhab.binding.astro-1.8.0-SNAPSHOT.jar
org.openhab.binding.ecobee-1.8.0-SNAPSHOT.jar
org.openhab.binding.http-1.8.0-SNAPSHOT.jar
org.openhab.binding.mqtt-1.8.0-SNAPSHOT.jar
org.openhab.binding.mqttitude-1.8.0-SNAPSHOT.jar
org.openhab.binding.nest-1.8.0-SNAPSHOT.jar
org.openhab.binding.networkhealth-1.8.0-SNAPSHOT.jar
-rw-r--r-- 1 openhab openhab   10185 Oct 31 13:40 org.openhab.binding.ntp-1.8.0-SNAPSHOT.jar
-rw-r--r-- 1 openhab openhab 1618995 Oct 31 13:41 org.openhab.binding.xbmc-1.8.0-SNAPSHOT.jar
-rw-r--r-- 1 openhab openhab  613092 Oct 31 13:44 org.openhab.io.myopenhab-1.8.0-SNAPSHOT.jar
-rw-r--r-- 1 openhab openhab  832963 Oct 31 13:44 org.openhab.persistence.mysql-1.8.0-SNAPSHOT.jar
```
I have 5 .items files and when I `touch` each one, I don't get any exceptions.  So....
* are you using a recent snapshot build of runtime and all addons?
* which bindings are you running?
* which hardware?

lolodomo:
Hi @watou
I am running a 1.8 snapshot downloaded the 28th of October.

Here are my installed bindings (downloaded the same day):
org.openhab.action.mios-1.8.0-SNAPSHOT.jar
org.openhab.binding.astro-1.8.0-SNAPSHOT.jar
org.openhab.binding.freebox-1.8.0-SNAPSHOT.jar
org.openhab.binding.http-1.8.0-SNAPSHOT.jar
org.openhab.binding.hue-1.8.0-SNAPSHOT.jar
org.openhab.binding.mios-1.8.0-SNAPSHOT.jar
org.openhab.binding.netatmo-1.8.0-SNAPSHOT.jar
org.openhab.binding.ntp-1.8.0-SNAPSHOT.jar
org.openhab.binding.powermax-1.7.0-SNAPSHOT.jar
org.openhab.binding.rfxcom-1.8.0-SNAPSHOT.jar
org.openhab.binding.weather-1.8.0-SNAPSHOT.jar
org.openhab.persistence.logging-1.8.0-SNAPSHOT.jar
org.openhab.persistence.rrd4j-1.8.0-SNAPSHOT.jar

I have only 1 items file.
My hardware is a RaspBerry Pi.

watou:
Thank you, @lolodomo.  I see that before my code change, [what was line 97](https://github.com/openhab/openhab/blob/6fbe28aecaee6f56195e57d2b814a616aa27236c/bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java#L97) is the same as what is [now line 104](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java#L104), so the same code is failing:
```java
for (Item item : items) {
	// we remove all binding configurations for all items
	bindingConfigs.remove(item.getName());  // this line throws NullPointerException
	notifyListeners(item);
}
```
...except now I've made the modification to `contextMap` atomic and threadsafe before the NPE, and this seems to have improved matters for you.  So the NPE means either a `null` key was added to the `bindingConfigs` map, or the `bindingConfigs` map itself is `null`.  This seems to be a timing issue that appears on the slower single-core Raspberry Pi 1 but not on the much faster quad-core Raspberry Pi 2.  Interesting.  So this PR set out to fix handling of the `contextMap` and did, but another race condition seems to be in play.

lolodomo:
@watou: I will try to run OH (same bindings) with my Windows 7 PC to confirm that the NPE is not happening on a fast hardware.

watou:
@lolodomo: That would be interesting to learn whether that makes the difference.  The slower RPi 1 seems to show timing issues in some cases that faster systems don't show.  Ideally we could identify and fix them so system performance/number of cores makes no difference other than time.

PR #3023, opened 2015-08-06
Zwave GE Database Edit : zw30003

SkynetHome:
Added CDATA to help and formatted code.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/ge/zw3003.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/ge/zw3003.xml
@@ -4,14 +4,25 @@
 	<Endpoints>1</Endpoints>
 	<Label lang="en">In-Wall Dimmer</Label>
 	<CommandClasses>
-        <Class><id>0x20</id></Class>
-        <Class><id>0x26</id></Class>
-        <Class><id>0x27</id></Class>
-        <Class><id>0x72</id></Class>
-        <Class><id>0x86</id></Class>
-        <Class><id>0x70</id></Class>
+		<Class>
+			<id>0x20</id>
+		</Class>
+		<Class>
+			<id>0x26</id>
+		</Class>
+		<Class>
+			<id>0x27</id>
+		</Class>
+		<Class>
+			<id>0x72</id>
+		</Class>
+		<Class>
+			<id>0x86</id>
+		</Class>
+		<Class>
+			<id>0x70</id>
+		</Class>
 	</CommandClasses>
-	
 	<Configuration>
 		<Parameter>
 			<Index>3</Index>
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.zwave/database/ge/zw3003.xml
```diff
@@ -87,9 +87,11 @@
 			<Maximum>255</Maximum>
 			<Size>1</Size>
 			<Label lang="en">On/Off Command Dim Rate</Label>
-			<Help lang="en">This value indicates in 10 millisecond resolution, how often the dim level will change. For example, if you set this parameter to 1, then every 10ms the dim level will change. If you set it to 255, then every 2.55 seconds the dim level will change.</Help>
+			<Help lang="en"><![CDATA[This value indicates in 10 millisecond resolution, <br>
```
cdjackson: You shouldn't add ```<br>``` tags in places like this - just when you actually want to force an end-of-line. If you add breaks like this, then it means the browser/UI doesn't have the ability to format it as it sees fit - you've effectively hard formatted it.
So, only add the breaks where you actually want to enforce a new paragraph, or new point of some description.  Same comment further down in the file, but I won't repeat it...

SkynetHome:
I was using the break to keep the first line short and more readable, then breaking to the deeper explanation beneath that. Maybe I am confusing the formatting you wanted for labels with the format for help. Isn't the idea to keep the text in small contained boxes rather than long lines.  Is that format forced or should I be using a different tag in those places. I thought I would use "<p>" if I wanted to enforce a new paragraph.

cdjackson:
Keeping the first line short is good, however it needs to stand on its own and subsequent lines should also be left to themselves (I know, that probably makes no sense!).

So, first line should be a short description -:

```This value indicates in 10 millisecond resolution.<br>```

This stands by itself in that it's a complete sentence - it has a full stop....

Then, the next part should also stand alone, unless there's a need to break paragraphs...

```
This sets how often the dim level will change. For example, if you set this parameter to 1,
then every 10ms the dim level will change. If you set it to 255, then every 2.55 seconds the dim level will change
```

The way you have it, the second part gets broken - if the GUI is able to display longer lines, then it looks like everything is packed to the left, on the other hand if the GUI can only show shorter lines, then it wraps, and then hits your breaks half way through -:

```
This sets how often the dim level will

 change. For example, if you set this parameter to 1
```

Github probably won't show this as I want, but the point is that it's like hitting the enter key in a word processor - you set a break and you don't allow the WP to paginate as it likes and it all looks messy.
(I hope that makes sense?)

SkynetHome:
Got it, that makes perfect sense. Will work on concise but complete first lines and then make sure to break in necessary locations below that. I will likely have to over come my compulsion to keep all text on one screen without having to slide to read, but I'll make do. :)

SkynetHome:
Ignore that last part, clearly text can be input on multiple lines inside tags!

SkynetHome:
Lets see if that's better.

SkynetHome:
BTW do you like what format does to the Command Classes? Seems much more readable with open and closed tags on same line!

cdjackson:
Yes - I tend to agree wrt the way it's formatted the command classes. However in some instances there are things that go in the command classes, and then it does look better...

I think the standard format is useful though - even if it's not maybe the way I think looks neatest :)

PR #3030, opened 2015-08-06
Corrected the Modbus parameters in openhab_default.cfg

bakrus:
openhab_default.cfg is not consistent with
https://github.com/openhab/openhab/wiki/Modbus-Tcp-Binding

"modbus:slave1.host" seems not to work, while
"modbus:tcp.slave1.connection" in fact does.


teichsta:
Thanks @bakrus!

Issue #2943, opened 2015-07-20
PushNotification java.lang.NullPointerException: null


GASASB:
Hi All,

Started to have this problem after using some time using it. My setup is Raspberry Pi - Debian, AeonLab Zwave Controller and AeonLab door, window sensor. Every time my door opens I send a message through pushNotification. 

All of Sudden I started to get the followign error:

2015-07-19 21:12:12.118 [INFO ] [runtime.busevents             ] - Door_House state updated to CLOSED
2015-07-19 21:12:17.630 [INFO ] [runtime.busevents             ] - Door_House state updated to OPEN
2015-07-19 21:12:17.665 [WARN ] [.o.c.p.e.PersistenceExtensions] - There is no queryable persistence service registered with the name 'rrd4j'
2015-07-19 21:12:17.667 [WARN ] [.o.c.p.e.PersistenceExtensions] - There is no queryable persistence service registered with the name 'rrd4j'
2015-07-19 21:12:53.231 [ERROR] [g.openhab.io.net.http.HttpUtil] - Fatal transport error: java.net.SocketTimeoutException: Read timed out
2015-07-19 21:12:53.286 [WARN ] [o.a.pushover.internal.Pushover] - Can't parse response from Pushover: null
java.lang.NullPointerException: null
	at java.io.StringReader.<init>(StringReader.java:50) ~[na:1.8.0]
	at org.openhab.action.pushover.internal.Pushover.parseResponse(Pushover.java:339) [bundlefile:na]
	at org.openhab.action.pushover.internal.Pushover.pushover(Pushover.java:318) [bundlefile:na]
	at org.openhab.action.pushover.internal.Pushover.pushover(Pushover.java:102) [bundlefile:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0]
	at java.lang.reflect.Method.invoke(Method.java:483) ~[na:1.8.0]
	at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.invokeOperation(XbaseInterpreter.java:729) [org.eclipse.xtext.xbase_2.3.0.v201206120633.jar:na]
[...]

PR #3072, opened 2015-08-15
Pushover: Added A Clean Error Message to the Pushover Service API Not Responding

CrackerStealth:
When communication with the Pushover API fails, a null exception is through producing a confusing message with a useless stacktrace. Brought up in #2943 with discussions about Pushover service.


teichsta:
Thanks @CrackerStealth!

PR #3079, opened 2015-08-18
Added RGBController1 service alias to MIOS binding

### LINKED ISSUE #3078: Mios binding does not support RGBController1 service
There is no alias for urn:upnp-org:serviceId:RGBController1, which means you can't use RGB color bulbs in OpenHAB.

beowulfe:
Fixes #3078

Signed-off-by: Andy Lintner <dev@beowulfe.com> (github: beowulfe)

### DIFF
--- a/bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/config/ServiceAliases.properties
+++ b/bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/config/ServiceAliases.properties
@@ -109,3 +109,4 @@ Tuning1=urn:micasaverde-com:serviceId:Tuning1
 VideoAdjustment1=urn:micasaverde-com:serviceId:VideoAdjustment1
 Volume1=urn:micasaverde-com:serviceId:Volume1
 WMC1=urn:micasaverde-com:serviceId:WMC1
+RGBController1=urn:upnp-org:serviceId:RGBController1


beowulfe:
@mrguessed - FYI

bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/config/ServiceAliases.properties
```diff
@@ -109,3 +109,4 @@ Tuning1=urn:micasaverde-com:serviceId:Tuning1
 VideoAdjustment1=urn:micasaverde-com:serviceId:VideoAdjustment1
 Volume1=urn:micasaverde-com:serviceId:Volume1
 WMC1=urn:micasaverde-com:serviceId:WMC1
+RGBController1=urn:upnp-org:serviceId:RGBController1
```
mrguessed: Why is this in the upnp-org domain?  That's reserved strictly for use by standards emitted by the standards body.  I took a quick look, but didn't see that one.

beowulfe: Agreed - the plugin developer shouldn't have used that. However, it's the most popular RGB plugin for MIOS, and the one recommended for the new Aeotec bulb. https://apps.mios.com/plugin.php?id=6686

mrguessed:
See the discussion here for previous stuff added for openHAB-centric Color mgmt:
    http://forum.micasaverde.com/index.php/topic,30496.0.html

Short version: openHAB uses HSB for its color widget and not RGB...
Long version: see the thread for full details.  

Overall, I'm not opposed to adding other serviceIds, if they're standardized, but I'd be cautious about adding something in a non compliant UPnP domain.  We used to do that right at the beginning of Plugin creation (~5-6 yrs back) but it really shouldn't be done in anything new... And this seems new.

Apart from that, new aliases also need the associated .md file augmented in the same checkin, along with (preferably) an example or two on how to use it in the MiOS Examples page.  

The Item file generator XSL, in examples/, also needs to be augmented with the rules for conversion of MiOS Device service-variables to the associated (mapped) Item declarations you'd typically see.

Collectively, these help new folks as they transition.

note: the MiOS binding supports any serviceId, even without the ServiceAlias defined, those are just so you can use the shorthand in scripts etc.  In recent additions, I've tended to add these in pairs, one like the one you've added here, and another shorter one (like 'RGB') that's really easy to remember.

bundles/binding/org.openhab.binding.mios/src/main/java/org/openhab/binding/mios/internal/config/ServiceAliases.properties
mrguessed: Ok, a hyperlink that in the doc (see below) should at least cover it.

Double-check the GIT diff, it looks like it's got an issue with the EOL or EOF char.

beowulfe:
Ahh, interesting - I was getting a parser error when I tried to use the full serviceId, so I figured the Alias was a requirement. I'll play around with the syntax. I'm with you on avoiding adding an alias for an invalid use of the upnp domain. If the Alias isn't required, it may be best to just wait and see how the color stuff shakes out. I'll close this PR.

Regarding the RGB/HSV issue, I ended up using JS transforms to convert. Although, that introduced another issue I'd like your input on: https://github.com/openhab/openhab/issues/3081

mrguessed:
If you open a forum discussion on the parsing error you're seeing, I can take a look at that.  Just `@guessed` me on that forum thread as the new forum is missing the sub-classifications from the original google-groups forum.

btw, still happy to have this as an included serviceId alias, along with the other things mentioned above.  You'd likely need to include the .js files though, since the default mapping is non-obvious.

PR #3082, opened 2015-08-19
Binding for relay boards available from http://www.ucprojects.eu

rmichalak:


hakan42:
Honestly, the name "relayboard" is too generic for this binding. There are many more relayboards besides the one created by ucprojects (a polish company if I believe google translator)

Could you maybe prefix the name with some company name? I could not find any such information on the web page of ucprojects, though.

rmichalak:
So maybe I can rename it to ucproject-relayboard or ucproject-rb?
Which one you prefer? I will change the name and resolve conflicts once we agree on a name.

Oh, and thanks for digging this out - I thought that this is a long forgotten branch ;)

hakan42:
Please call it ucproject-relayboard and rebase the branch against the current master. 

After that, give me a short notice here and I will be happy to continue the review.

rmichalak:
Do you want me to rename java package to ucproject-relayboard as well? I would prefer to rename it to ucprelayboard - to not introduce very long package name with '-' in the middle. Same for a class names - I would add UCP prefix to them. What do you think?

teichsta:
> I would prefer to rename it to ucprelayboard
+1

hakan42:
Maybe call the package ucprelayboard and the binding ucp-relayboard? 

Or, if you like it more, name both the packet and the binding identically ucprelayboard?

hakan42:
And prefixing the classes with UCP is excellent assume well :smile:

teichsta:
> Maybe call the package ucprelayboard and the binding ucp-relayboard?

if i could choose i would always prefer names without dashes "-"

hakan42:
Ok, without dashes is the common preference then :smile:

rmichalak:
Done. Changed binding name to ucprelayboard and rebased.

hakan42:
* code looks good
* local build is successful
* merging now :smile:

hakan42:
@rmichalak , Thank you very much for your contribution. Could you please test this binding after tonights Jenkins build? If successful, an announcement posting on the forums would be nice :smile:

teichsta:
and please test your binding against the OH2 runtime (@hakan42 another requirement for new bindings :-))

rmichalak:
Verified on both OH and OH2 runtime. Created additional pull request #3509 to fix dependency and CRC calculation problem.

hakan42:
Why does this binding need javax.mail? I must admit I am not at my desktop right now and cannot read the full sources. Maybe if you could give me a short hint here?

hakan42:
Thank you for the confirmation. I will prepare a PR against OH2 later tonight.

teichsta:
@rmichalak after all could you please provide the according Wiki page until the weekend? Thanks, Thomas E.-E.

rmichalak:
Sure, I will try to do this before this weekend.

rmichalak:
Done: https://github.com/openhab/openhab/wiki/UCProjects.eu-Relay-Board-Binding

watou:
Looks great!  Very minor comment: Since `inverted=true` is optional, would the syntax explanation be a little better like:
```
{ucprelayboard="board=<name>;relay=<number>[;inverted=true]"}
```
(the second semicolon is only needed if you are specifying a third value)

rmichalak:
Good spot. Fixed.

PR #3086, opened 2015-08-21
MQTT Action bundle

kstaniek:
This is a small enhancement adding MQTT Action to OH. It can be used in rules to simply publish the MQTT message to topic using configured broker. I can see it useful to work with the project like mqttwarn simplifying MQTT message routing.

### DIFF
--- a/bundles/action/org.openhab.action.mqtt/src/main/java/org/openhab/action/mqtt/internal/MqttMessagePublisher.java
+++ b/bundles/action/org.openhab.action.mqtt/src/main/java/org/openhab/action/mqtt/internal/MqttMessagePublisher.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.action.mqtt.internal;
+
+import org.openhab.io.transport.mqtt.MqttMessageProducer;
+import org.openhab.io.transport.mqtt.MqttSenderChannel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class implementing Message Producer from MQTT IO service.
+ * 
+ * @author Klaudiusz Staniek
+ * @since 1.8.0
+ */
+public class MqttMessagePublisher implements MqttMessageProducer {
+
+	private static final Logger logger = LoggerFactory.getLogger(MqttMessagePublisher.class);
+
+	private MqttSenderChannel senderChannel;
+	
+	public void publish(String topic, String message) {
+		if (senderChannel == null) {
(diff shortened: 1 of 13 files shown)


watou:
Thanks for this!  It will be very useful.  You will probably be asked to make this new action visible in the openHAB Designer, and @teichsta's PR #2967 has good examples of what is required, in particular additions to the file features/org.openhab.designer.feature/feature.xml .

bundles/action/pom.xml
```diff
@@ -32,6 +32,7 @@
     <module>org.openhab.action.astro</module>
     <module>org.openhab.action.tinkerforge</module>
     <module>org.openhab.action.harmonyhub</module>
+    <module>org.openhab.action.mqtt</module>
   </modules>
 
-</project>
+</project>
```
teichsta: add newline

bundles/action/org.openhab.action.mqtt/src/main/java/org/openhab/action/mqtt/internal/MqttMessagePublisher.java
```diff
@@ -0,0 +1,43 @@
...
+	private MqttSenderChannel senderChannel;
+	
+	public void publish(String topic, String message) {
+		if (senderChannel == null) {
+			return;
+		}
+		try {
+			senderChannel.publish(topic, message.getBytes());
+		} catch (Exception e) {
+			logger.error("Error publishing message");
```
teichsta: please enhance the given error message with more context. Currently the message is that comprehensive as it could be.

teichsta:
@kstaniek thanks for this contribution. As @watou suggested already please also add it to the designers feature.xml. The action should be available in the binary build of the designer. Thanks, Thomas E.-E.

bundles/action/org.openhab.action.mqtt/src/main/java/org/openhab/action/mqtt/internal/MqttMessagePublisher.java
kstaniek: It is actually as comprehensive as the MqttSenderChannel interface description from org.openhab.io.transport.mqtt:
/**
	 * Send a message to the MQTT broker.
	 * 
	 * @param topic
	 *            Topic to publish the message to.
	 * @param message
	 *            message payload.
	 * @throws Exception
	 *             if an error occurs during sending.
	 */
The mqtt transport binding produces a detailed error messages into the log file in case of known issue. This exception handling is more to catch any unexpected situation or exception generated by mqtt pacho library. I can catch specifically the MqttException and log the output of getMessage() method. It it something that would help?
Please refer to http://www.eclipse.org/paho/files/javadoc/org/eclipse/paho/client/mqttv3/MqttException.html

bundles/action/pom.xml
kstaniek: I'm just curious why this new line is needed? Of course have no issue with this.

kstaniek:
@teichsta Thomas, could you please have a look if this is ok now?

teichsta:
everything fine now! Thanks a lot again …

Issue #3087, opened 2015-08-21
FS20 Binding - CULDeviceException: Can't instantiate CULHandler


wojtus:
Following the notes on https://github.com/openhab/openhab/wiki/FS20-Binding the configuration of openhab leeds to a ClassCastException in CULSerialHandlerImpl.java:120. The baudrate seems to be stored in the Map as an Integer and the unchecked cast fails. This is repeatable on openhub-1.7.1 on a raspbian-system (RasPi2) with OracleJdk7 & OpenJdk7.

CULSerialHandlerImpl.java
120: final String configuredBaudRate = (String) properties.get(KEY_BAUDRATE);
121: Integer tmpBaudRate=baudrateFromConfig(configuredBaudRate);

could get to:
Integer tmpBaudRate;
Object baudRateObject = properties.get(KEY_BAUDRATE);
if (baudRateObject instanceof String){
final String configuredBaudRate = baudRateObject;
tmpBaudRate=baudrateFromConfig(configuredBaudRate);
}
if (baudRateObject instanceof Integer){
tmpBaudRate= (Integer) properties.get(KEY_BAUDRATE);
}

If you need a patch or so for this - let me know.
Greetings!
Wojtek

The relevant Part of the stacktrace:
org.openhab.io.transport.cul.CULManager.createNewHandler(CULManager.java:161) ~[bundlefile:na]
	... 6 common frames omitted
Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
	at org.openhab.io.transport.cul.internal.CULSerialHandlerImpl.<init> CULSerialHandlerImpl.java:120) ~[na:na]
	... 11 common frames omitted
2015-08-21 17:23:26.828 [INFO ] [.service.AbstractActiveService] - FS20 Refresh Service has been started

watou:
Hi @wojtus, created PR #3088 to match this issue.  Please test a resulting build to verify it corrects the issue you see (I don't have the hardware here).

watou:
The JAR I built is [here](https://dl.dropboxusercontent.com/u/4286376/cul-issue-3087/org.openhab.io.transport.cul-1.8.0-SNAPSHOT.jar).  The .deb I built is [here](https://dl.dropboxusercontent.com/u/4286376/cul-issue-3087/openhab-addon-io-cul-1.8.0-SNAPSHOT.deb).

wojtus:
Hallo @watou !
[...]

PR #3088, opened 2015-08-21
[Cul] patch as described in #3087

watou:
Patch as described in #3087.  Not tested; only verified that `mvn clean install` succeeds.  @wojtus, please verify this fix is what you intended and that it works (i.e., test the bundle that contains this patch).  Thank you!

### DIFF
--- a/bundles/binding/org.openhab.binding.fs20/src/main/java/org/openhab/binding/fs20/internal/FS20Binding.java
+++ b/bundles/binding/org.openhab.binding.fs20/src/main/java/org/openhab/binding/fs20/internal/FS20Binding.java
@@ -151,7 +151,7 @@ public class FS20Binding extends AbstractActiveBinding<FS20BindingProvider>
 			Boolean configChanged = false;
 			String baudRateString = (String) config.get(KEY_BAUD_RATE);
 			if(StringUtils.isNotBlank(baudRateString)){
-				properties.put(KEY_BAUD_RATE, Integer.parseInt(baudRateString));
+				properties.put(KEY_BAUD_RATE, baudRateString);
 				configChanged = true;
 			}
 			
--- a/bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/CULSerialHandlerImpl.java
+++ b/bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/CULSerialHandlerImpl.java
@@ -24,11 +24,9 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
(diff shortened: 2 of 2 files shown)


teichsta:
any updates @wojtus?

bundles/binding/org.openhab.binding.fs20/src/main/java/org/openhab/binding/fs20/internal/FS20Binding.java
```diff
@@ -149,23 +149,30 @@ public void updated(Dictionary<String, ?> config)
...
+			 * "75", "110", "300", "1200", "2400", "4800", 
+			 * "9600", "19200", "38400", "57600", "115200"
+			 * @see org.openhab.io.transport.cul.internal.CULSerialHandlerImpl
+			 */
 			String baudRateString = (String) config.get(KEY_BAUD_RATE);
 			if(StringUtils.isNotBlank(baudRateString)){
-				properties.put(KEY_BAUD_RATE, Integer.parseInt(baudRateString));
+				properties.put(KEY_BAUD_RATE, baudRateString);
 				configChanged = true;
 			}
```
teichsta: since the valid baudRates are known couldn't we implement a validator in io.transport.serial so other serial based bundles could benefit as well?

teichsta:
Thanks @watou!

wojtus:
So the patch works for me. And im quite sure, that der master-branch of the Fs20 can not configure the parity and the baud rate of the serial device. 
To implement a validator? Yes it would be an idea. I do not now if other serial device bindings, that depends on the CUL-interface. But it wouldent be a bugfix anymore. Maybe is an idea for openhub2? I will look at it after the weekend.

wojtus:
I looked for the possibiity of putting some validation before opening the serial connection. 
In case of this bug such validation would not help: in issuse #3088 it was a clear bug in a code because of an unchecked cast. 
But still there woud be some possibility to validate the properties map.
FS20 depends on io.transport.serial through io.transport.cul. CULManager could validate the properties, before opening the serial device.
io.transport.serial does not contain any source code, but it would be a better place for such validation. To achive this some abstaction of a serial port would be needed. It could be a feature for openhab2. 
In openhab2 I could not find any sub-project for the cul-devices. Will there be any support for them in openhab2?

teichsta:
> In openhab2 I could not find any sub-project for the cul-devices. Will there be any support for them in openhab2?

yes, if you or probably @dereulenspiel ports them :-)

wojtus:
I created #3118 
@teichsta can you tag this feature request as an enhancement for openhab2.
To complete this issue the  https://github.com/openhab/openhab/wiki/FS20-Binding needs to be updated. What is the right time for this? After the 1.8.0 release? Or shoud I do it directly?

Porting CUL / FS20? I think, I could do this, there is not co much code in those projects. I need them working in OH2 :)

PR #3106, opened 2015-08-29
SupportSwitchAllCommandClass

paixaop:
Added support for Switch All Command Class.

To use add a new switch item 

    Switch AllOnOff "All On/Off" {zwave="4:command=SWITCH_ALL"}

When user toggles this switch an All On or All Off message will be sent to ALL Z-Wave nodes that support the Switch All Command Class 
The node ID used in the zwave binding MUST be the ID of any one node that supports the SwitchAll command class. 

Then on the sitemap you could do:

    Frame {
        Group item=gGF label="Ground Floor" icon="groundfloor"
        Group item=gFF label="First Floor" icon="firstfloor"
        Group item=Outdoor icon="garden" 
        Switch item=AllOnOff icon="" mappings=[ON="All On", OFF="All Off"]
    }

Added support for the SWITCH_ALL_SET message although it is not needed in OpenHab operation but would be needed to configure nodes.
How could it be added to Habmin so we can configure AllOn and AllOff membership of nodes like in all other Z-Wave Controllers?

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
@@ -25,6 +25,7 @@ import org.openhab.binding.zwave.internal.protocol.ZWaveDeviceClass;
 import org.openhab.binding.zwave.internal.protocol.ZWaveDeviceType;
 import org.openhab.binding.zwave.internal.protocol.ZWaveEventListener;
 import org.openhab.binding.zwave.internal.protocol.ZWaveNode;
+import org.openhab.binding.zwave.internal.protocol.SerialMessage;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveAssociationCommandClass;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveAssociationCommandClass.ZWaveAssociationEvent;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveBatteryCommandClass;
@@ -358,10 +359,10 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 	            if (switchAllCommandClass != null) {
 	            	record = new OpenHABConfigurationRecord(domain, "Switch All", "Switch All", false);
 	            	record.type = OpenHABConfigurationRecord.TYPE.LIST;
-					record.addValue("0x00", "Exclude from All On and All Off groups");
-					record.addValue("0x01", "Include in All On group");
-					record.addValue("0x02", "Include in All Off group");
-					record.addValue("0xFF", "Include in All On and All Off groups");
+					record.addValue("0", "Exclude from All On and All Off groups");
+					record.addValue("1", "Include in All On group");
+					record.addValue("2", "Include in All Off group");
+					record.addValue("255", "Include in All On and All Off groups");
 	            	records.add(record);
 	            }
 				
@@ -1110,7 +1111,7 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 				return;
 			}
 
-			// TODO: Should we check that the node is finished initialising
+			// TODO: Should we check that the node is finished initializing
 
 			ZWaveProductDatabase database = new ZWaveProductDatabase();
 			if (database.FindProduct(node.getManufacturer(), node.getDeviceType(), node.getDeviceId(), node.getApplicationVersion()) == false) {
@@ -1124,6 +1125,27 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 				}
 				if (splitDomain[2].equals("Location")) {
 					node.setLocation(value);
+				}
+				
+				if (splitDomain[2].equals("Switch All")) {
+					ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass) node
+							.getCommandClass(CommandClass.SWITCH_ALL);
+
+					if (switchAllCommandClass == null) {
+						logger.error("NODE {}: Error getting switchAllCommandClass in doSet", nodeId);
+						return;
+					}
+
+					
+					// Set the switch all mode
+					int mode = Integer.parseInt(value);
+					
+					PendingCfg.Add(ZWaveSwitchAllCommandClass.CommandClass.SWITCH_ALL.getKey(), node.getNodeId(), 0, mode);
+					
+					SerialMessage msg = switchAllCommandClass.setValueMessage(mode);
+					this.zController.sendData(msg);
+					// And request a read-back
+					this.zController.sendData(switchAllCommandClass.getValueMessage());
 				}
 				
 				// Write the node to disk
@@ -1305,6 +1327,18 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 
 			// Remove this node from the pending list
 			PendingCfg.Remove(ZWaveCommandClass.CommandClass.WAKE_UP.getKey(), event.getNodeId());
+			return;
+		}
+		
+		if (event instanceof ZWaveSwitchAllCommandClass.ZWaveSwitchAllModeEvent) {
+			ZWaveSwitchAllCommandClass.ZWaveSwitchAllModeEvent e = (ZWaveSwitchAllCommandClass.ZWaveSwitchAllModeEvent) event;
+			
+			// Write the node to disk
+			ZWaveNodeSerializer nodeSerializer = new ZWaveNodeSerializer();
+			nodeSerializer.SerializeNode(zController.getNode(event.getNodeId()));
+
+			// Remove this node from the pending list
+			PendingCfg.Remove(ZWaveCommandClass.CommandClass.SWITCH_ALL.getKey(), e.getNodeId());
 			return;
 		}
 
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
@@ -357,7 +357,7 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 
 				ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass)node.getCommandClass(ZWaveCommandClass.CommandClass.SWITCH_ALL);
 	            if (switchAllCommandClass != null) {
-	            	record = new OpenHABConfigurationRecord(domain, "Switch All", "Switch All", false);
+	            	record = new OpenHABConfigurationRecord(domain, "SwitchAll", "Switch All", false);
 	            	record.type = OpenHABConfigurationRecord.TYPE.LIST;
 					record.addValue("0", "Exclude from All On and All Off groups");
 					record.addValue("1", "Include in All On group");
@@ -1112,7 +1112,7 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 				return;
 			}
 
-			// TODO: Should we check that the node is finished initializing
+			// TODO: Should we check that the node is finished initialising
 
 			ZWaveProductDatabase database = new ZWaveProductDatabase();
 			if (database.FindProduct(node.getManufacturer(), node.getDeviceType(), node.getDeviceId(), node.getApplicationVersion()) == false) {
@@ -1128,7 +1128,7 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 					node.setLocation(value);
 				}
 				
-				if (splitDomain[2].equals("Switch All")) {
+				if (splitDomain[2].equals("SwitchAll")) {
 					ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass) node
 							.getCommandClass(CommandClass.SWITCH_ALL);
 
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
@@ -363,6 +363,7 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 					record.addValue("1", "Include in All On group");
 					record.addValue("2", "Include in All Off group");
 					record.addValue("255", "Include in All On and All Off groups");
+					record.value = String.valueOf(switchAllCommandClass.getMode());
 	            	records.add(record);
 	            }
 				
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
@@ -32,6 +32,7 @@ import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveCommandClas
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveCommandClass.CommandClass;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveConfigurationCommandClass.ZWaveConfigurationParameterEvent;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveConfigurationCommandClass;
+import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveSwitchAllCommandClass;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveVersionCommandClass;
 import org.openhab.binding.zwave.internal.protocol.commandclass.ZWaveWakeUpCommandClass;
 import org.openhab.binding.zwave.internal.protocol.event.ZWaveEvent;
@@ -353,6 +354,17 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 				record.value = node.getLocation();
 				records.add(record);
 
+				ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass)node.getCommandClass(ZWaveCommandClass.CommandClass.SWITCH_ALL);
+	            if (switchAllCommandClass != null) {
+	            	record = new OpenHABConfigurationRecord(domain, "Switch All", "Switch All", false);
+	            	record.type = OpenHABConfigurationRecord.TYPE.LIST;
+					record.addValue("0x00", "Exclude from All On and All Off groups");
+					record.addValue("0x01", "Include in All On group");
+					record.addValue("0x02", "Include in All Off group");
+					record.addValue("0xFF", "Include in All On and All Off groups");
+	            	records.add(record);
+	            }
+				
 				if(node.getManufacturer() != Integer.MAX_VALUE) {
 					if (database.FindProduct(node.getManufacturer(), node.getDeviceType(), node.getDeviceId(), node.getApplicationVersion()) == true) {
 						// Add links to configuration if the node supports the various command classes
@@ -391,6 +403,18 @@ public class ZWaveConfiguration implements OpenHABConfigurationService, ZWaveEve
 
 				record = new OpenHABConfigurationRecord(domain + "info/", "Information");
 				records.add(record);
+			} else if (arg.equals("switch_all/")) { 
+				/*ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass)node.getCommandClass(ZWaveCommandClass.CommandClass.SWITCH_ALL);
+	            if (switchAllCommandClass != null) {
+	                logger.debug("NODE {}: Supports Switch All Command Class Adding Switch All Configuration ", (Object)nodeId);
+	                record = new OpenHABConfigurationRecord(domain, "Mode", "Mode", true);
+	                record.type = OpenHABConfigurationRecord.TYPE.LIST;
+					record.addValue("0x00", "Exclude from All On and All Off groups");
+					record.addValue("0x01", "Include in All On group");
+					record.addValue("0x02", "Include in All Off group");
+					record.addValue("0xFF", "Include in All On and All Off groups");
+					records.add(record);
+	            }*/
 			} else if (arg.equals("info/")) {
 				record = new OpenHABConfigurationRecord(domain, "NodeID", "Node ID", true);
 				record.value = Integer.toString(node.getNodeId());
--- a/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveConverterHandler.java
+++ b/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveConverterHandler.java
@@ -287,6 +287,10 @@ public class ZWaveConverterHandler {
 			if (commandClassName.equalsIgnoreCase("info")) {
 				return infoConverter.getRefreshInterval();
 			}
+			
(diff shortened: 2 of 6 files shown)


bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
```diff
@@ -0,0 +1,207 @@
...
+        return result;
+    }
+
+    /**
+     * Create a new SwitchAll set message
+     * @param mode as (0x00 - Exclude, 0x01 Only All On, 0x02 Only All Off, 0xFF Both All on and All off)
+     * @return SerialMessage
+     */
+    public SerialMessage setValueMessage(int mode) {
+        if (!this.isGetSupported) {
```
cdjackson: Why is this here - this is the SET command?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
```diff
@@ -0,0 +1,207 @@
...
+                break;
+            default:
+                logger.warn(String.format("Unsupported Switch All mode 0x%02X.", mode));
+        }
+        ZWaveCommandClassValueEvent zEvent = new ZWaveCommandClassValueEvent(this.getNode().getNodeId(), endpoint, this.getCommandClass(), mode);
+        this.getController().notifyEventListeners(zEvent);
+    }
+
+    public SerialMessage getValueMessage() {
+        logger.debug("NODE {}: Creating new message for command SWITCH_ALL_GET", (Object)this.getNode().getNodeId());
```
cdjackson: This should have the ```if(getSupported)``` check.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveSwitchAllConverter.java
```diff
@@ -0,0 +1,119 @@
...
+        this.addStateConverter(new IntegerUpDownTypeConverter());
+        this.addCommandConverter(new BinaryOnOffCommandConverter());
+    }
+
+    /**
+	 * {@inheritDoc}
+	 */
+    @Override
+    public SerialMessage executeRefresh(ZWaveNode node, 
+    		ZWaveSwitchAllCommandClass commandClass, int endpointId, Map<String, String> arguments) {
```
cdjackson: Does it really make sense to poll this? IT's not polling the status of the switch, it's polling the status of the configuration (correct?). If so, I don't think it makes sense and I'd probably just remove the polling?
What do you think?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveSwitchAllConverter.java
```diff
@@ -0,0 +1,119 @@
...
+    	ZWaveCommandConverter<?,?> converter = this.getCommandConverter(command.getClass());
+        
+    	if (converter == null) {
+            logger.warn("NODE {}: No converter found for item = {}, endpoint = {}, ignoring command.", new Object[]{node.getNodeId(), item.getName(), endpointId});
+            return;
+        }
+        
+    	Integer cmd = (Integer)converter.convertFromCommandToValue(item, command);
+        
+    	if (cmd == 0x00) {
```
cdjackson: This always sends the command to the controller, which in turn always sends it to all nodes. While this might be the intention of the command class, in your example, you use 
```
Switch AllOnOff "All On/Off" {zwave="4:command=SWITCH_ALL"}
```

This configures node 4 - so I would have expected this command to be sent only to node 4? Wouldn't it make more sense to send to the node the user requested? If they want it sent to all nodes, then they can use the controller node id, and then you should add a check here to see if the node is to the controller (in which case do what you have here) or to another node, in which case you send just to that node?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
```diff
@@ -0,0 +1,207 @@
...
+    private static final int SWITCH_ALL_GET = 0x02;
+    private static final int SWITCH_ALL_REPORT = 0x03;
+    private static final int SWITCH_ALL_ON = 0x04;
+    private static final int SWITCH_ALL_OFF = 0x05;
+    
+    private static final int SWITCH_ALL_EXCLUDED = 0x00;
+    private static final int SWITCH_ALL_INCLUDE_ON_ONLY = 0x01;
+    private static final int SWITCH_ALL_INCLUDE_OFF_ONLY = 0x02;
+    private static final int SWITCH_ALL_INCLUDE_ON_OFF = 0xFF;
+    private boolean isGetSupported = true;
```
cdjackson: I suspect this isn't required at all since I doubt it makes sense for a device not to support the GET here since it's just a status readback. Maybe the best thing to do is to remove this?

cdjackson:
Thanks - just a few minor comments :)

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveSwitchAllConverter.java
paixaop: You're right don't think this is needed will remove.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveSwitchAllConverter.java
paixaop: Chris
you're right, I tried using the node 1 for the controller but the issue was that since the controller it self did not support the Switch All Command Class a check was failing and the command was not executed.
I will try to patch the converter handler to bypass the command class supported check to see if I can get around that.
Any suggestions on where to start?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
paixaop: Starting to get a better grasp on openHab and the zwave binding, today it is severely lacking the install/configuration stage of the Z-Wave network. Once it is running these are not needed but they would be for initial configuration. 
Any other home automation hub will support the installation and configuration of nodes, openHab on the other hand, without your HABmin does not. 
What's your suggestion as a path for development? Keep all install/configure functions on HABmin? or create full implementations of the command classes in openHab and export that to the HABmin GUI somehow?

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
paixaop: See comments bellow about implementing full command classes and openHAB vs HABmin

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveSwitchAllConverter.java
cdjackson: You should simply be able to do the following (in psuedo code)-:

```
if(node.getNodeId() == controller.getOwnNodeId()) {
    controller.OnOff()
}
else {
   get command class and call the allOn or allOff directly just for this node
   use one of the other command converters (eg BinarySwitch) as an example?
}
```

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
cdjackson: That's not correct - HABmin doesn't do the configuration - it's done in the binding during the initialisation of the device (the same as other devices should). HABmin just provides a tool to manually manipulate the configuration.

However, I'm not sure what this has to do with the ```isGetSupported``` flag. This flag is just used to set the database to not request GETS, but this command class doesn't really do gets in the way this is intended.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
cdjackson: I'm not sure what you mean? This is the SET command - why does it have the ```isGetSupported``` flag? This should be removed.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
paixaop: not directly related to the GET. What I meant was not binding initialization done at startup, I meant the user installing and configuring his or her network, set up associations, and in the Switch All case configure the mode in one of the 4 possible configurations of the command class,

    0 - excluded from all on and all off
    1 - include in all on only
    2 - include in all off only
    255 - include in all On and all Off 

How can we configure this in openHAB?

cdjackson: This sort of thing largely needs to be done manually. The binding automatically configures things like associations, but  I don't think the binding can take a guess at what the user wants for the all on class - can it?

It's relatively easy to add this configuration so that it appears in HABmin if that's what you want, but that's another issue...

paixaop: Agree and yes I think adding it to HABmin would make perfect sense. I tried last night but could not get around to it, as I'm pretty unfamiliar with the code base.

cdjackson: Just to note that you don't need to do anything to HABmin - this is ALL done in the binding. Take a look at the Configuration class.

paixaop:
Now it works for controller node, no need to use a specific node that supports SWITCH_ALL command class. Did all other fixes as well

paixaop:
Chris noticed that the AllOff message always fails the first time. Why do you think that is?

    18:32:15.278 [DEBUG] [.o.b.z.i.p.c.ZWaveCommandClass:224  ] - NODE 1: Creating new instance of command class SWITCH_ALL
    18:32:15.279 [DEBUG] [b.z.i.protocol.ZWaveController:473  ] - NODE 4: Supports Switch All Command Class Sending AllOff Message
    18:32:15.279 [DEBUG] [p.c.ZWaveSwitchAllCommandClass:189  ] - NODE 4: Switch All - Creating All Off message.
    18:32:15.279 [DEBUG] [o.b.z.i.protocol.SerialMessage:109  ] - NODE 4: Creating empty message of class = SendData (0x13), type = Request (0x00)
    18:32:15.279 [DEBUG] [b.z.i.protocol.ZWaveController:620  ] - Enqueueing message. Queue length = 1
    18:32:15.279 [INFO ] [runtime.busevents             :26   ] - AllOnOff state updated to OFF
    18:32:17.460 [ERROR] [WaveController$ZWaveSendThread:1309 ] - NODE 4: Timeout while sending message. Requeueing - 2 attempts left!
    18:32:17.460 [ERROR] [b.z.i.p.s.SendDataMessageClass:158  ] - NODE 4: Got an error while sending data. Resending message.
    18:32:17.460 [DEBUG] [b.z.i.protocol.ZWaveController:938  ] - Callback ID = 16
    18:32:17.460 [DEBUG] [b.z.i.protocol.ZWaveController:620  ] - Enqueueing message. Queue length = 2
    18:32:17.460 [DEBUG] [WaveController$ZWaveSendThread:1211 ] - Took message from queue for sending. Queue length = 1
    18:32:17.460 [DEBUG] [o.b.z.i.protocol.SerialMessage:233  ] - Assembled message buffer = 01 09 00 13 04 02 27 04 25 10 F5 
    18:32:17.460 [DEBUG] [WaveController$ZWaveSendThread:1268 ] - NODE 4: Sending REQUEST Message = 01 09 00 13 04 02 27 04 25 10 F5 
    18:32:17.465 [DEBUG] [eController$ZWaveReceiveThread:1481 ] - Receive Message = 01 04 01 13 01 E8 
    18:32:17.465 [DEBUG] [eController$ZWaveReceiveThread:1405 ] - Receive queue ADD: Length=1

paixaop:
Added switch all configuration to ZWaveConfiguration class and you can now set the switch all mode from HABmin

cdjackson:
> 
> 18:32:15.278 [DEBUG] [.o.b.z.i.p.c.ZWaveCommandClass:224  ] - NODE 1: Creating new instance of command class SWITCH_ALL
Why is this message logged? You shouldn’t ever create the command class on node 1 (I’m assuming node 1 is the controller?)


> 
> 18:32:15.279 [DEBUG] [b.z.i.protocol.ZWaveController:620  ] - Enqueueing message. Queue length = 1
> 18:32:15.279 [INFO ] [runtime.busevents             :26   ] - AllOnOff state updated to OFF
> 18:32:17.460 [ERROR] [WaveController$ZWaveSendThread:1309 ] - NODE 4: Timeout while sending message. Requeueing - 2 attempts left!
I don’t see the message being sent… Are your timeouts set to 2 seconds?

paixaop:
This message is logged when it the binding looks for a converter on `ZWaveConverterHandler::receiveCommand()` when it gets a command on the AllOn/AllOff Item.

I'm doing it this way

    if(node.getNodeId() == this.controller.getOwnNodeId() && commandClassName.equalsIgnoreCase("switch_all")) {
        commandClass = ZWaveCommandClass.getInstance(0x27, node, this.controller);
    }

Is there a better way to set the commandClass to ZWaveSwitchAllCommandClass?
Did it so we could use the controller node in the binding configuration of the AllOn/AllOff Switch item.

    Switch AllOnOff "All On/Off" {zwave="1:command=SWITCH_ALL"}

cdjackson:
> if(node.getNodeId() == this.controller.getOwnNodeId() && commandClassName.equalsIgnoreCase("switch_all")) {
    commandClass = ZWaveCommandClass.getInstance(0x27, node, this.controller);
}

Is the ```commandClassName.equalsIgnoreCase``` part necessary? I think the only way it can get to this code is if the command class already equals switch_all since that's taken care of in the converter handler (I think?)

> ZWaveCommandClass.getInstance(0x27, node, this.controller);

I wouldn't hardcode the reference here.

I'm getting very lost as to what the code now looks like - I'm not sure I'm a big fan of githubs review system as I can't find this code you're referring to. Is everything pushed to the server?

paixaop:
Did a new sync so everything should be in the repo. 

My thinking was to check if the node was the controller and the command class was the Switch All then set the converter to the ZWaveSwitchAllConverter

    ZWaveCommandClassConverter<ZWaveCommandClass> converter = 
				(ZWaveCommandClassConverter<ZWaveCommandClass>) getConverter(commandClass.getCommandClass());
	
Added the class check so that this new code path only works if the user adds an item with the 

    {zwave="1:command=SWITCH_ALL"}

If there are other Items that refer to the controller node we should not mess around with the converter handler.

Changes are around line `393` of `ZwaveConverterHandler.java`.

One question where do I set the ZWave timeouts?

cdjackson:
I’ve finally found this code - I’m really unsure why this needs to be in the converterHandler. I don’t see why there should be a need to change anything in the main handler - everything should be handled in the converter associated with the specific class. Can you tell me why you need to change this code in (for example) the handleRefresh method in the converter handler?

paixaop:
Did the change there when following your suggestion of using the controller node on the Item configuration, which makes perfect sense. When changing the node to 1 in the Item `{zwave="1:command=SWITCH_ALL"}` I was hitting the code path

    if (commandClass == null) {
        logger.warn("NODE {}: No converter found for item = {}, ignoring command.", node.getNodeId(), itemName);
        return;
    }

And the ZWaveSwitchAllConverter was never called.

cdjackson:
Ok - hadn’t thought of that :)

I’m not too keen on the way this is done though - I’m just worried that it’s not very extensible… I’ll take a look at this a bit later.

paixaop:
Chris any word on this?

cdjackson:
Sorry for the delay with this one...

I still don't know that I like the approach, but let's go with it for now... I have a few more comments :)

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
```diff
@@ -353,6 +355,18 @@ else if(node.getSendCount() > 20 && (node.getRetryCount() * 100 / node.getSendCo
 				record.value = node.getLocation();
 				records.add(record);
 
+				ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass)node.getCommandClass(ZWaveCommandClass.CommandClass.SWITCH_ALL);
```
cdjackson: I don't think putting this here is a good move - I think either in it's own group as you have commented out below, or in the config group. I'm thinking ahead to OH2 here...

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
```diff
@@ -1086,7 +1112,7 @@ public void doSet(String domain, String value) {
 				return;
 			}
 
-			// TODO: Should we check that the node is finished initialising
+			// TODO: Should we check that the node is finished initializing
```
cdjackson: :) It depends on what side of the pond you're on...

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
```diff
@@ -353,6 +355,18 @@ else if(node.getSendCount() > 20 && (node.getRetryCount() * 100 / node.getSendCo
 				record.value = node.getLocation();
 				records.add(record);
 
+				ZWaveSwitchAllCommandClass switchAllCommandClass = (ZWaveSwitchAllCommandClass)node.getCommandClass(ZWaveCommandClass.CommandClass.SWITCH_ALL);
+	            if (switchAllCommandClass != null) {
+	            	record = new OpenHABConfigurationRecord(domain, "Switch All", "Switch All", false);
```
cdjackson: It's preferable not to use spaces in the name. As elsewhere, I think it's better not to put this at the top level of the node.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/config/ZWaveConfiguration.java
```diff
@@ -1102,6 +1128,27 @@ public void doSet(String domain, String value) {
 					node.setLocation(value);
 				}
 				
+				if (splitDomain[2].equals("Switch All")) {
```
cdjackson: The name and domain should be aligned with the comments above...

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveConverterHandler.java
```diff
@@ -384,30 +389,34 @@ public void receiveCommand(ZWaveBindingProvider provider, String itemName, Comma
...
-					bindingConfiguration.getEndpoint());
-
-			if (commandClass == null) {
-				logger.warn(
-						"NODE {}: No command class found for item = {}. Class = {}({}), endpoint = {}. Ignoring command.",
-						node.getNodeId(), itemName, commandClassName, CommandClass.getCommandClass(commandClassName)
-								.toString(), bindingConfiguration.getEndpoint());
-				return;
+		
+		if(node.getNodeId() == this.controller.getOwnNodeId() && commandClassName.equalsIgnoreCase("switch_all")) {
```
cdjackson: You should move this around. By putting this check outside of the 'if commandClassName == null' check, if this is null if will cause an exception.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
cdjackson: I see this isGetSupported is still in the code - are you going to remove this?

cdjackson: Actually - ignore this comment - let's leave it in just in case...

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
```diff
@@ -0,0 +1,244 @@
...
+            this.isGetSupported = options.isGetSupported;
+        }
+        return true;
+    }
+    
+    /**
+     * get the Switch All mode
+     * @return the mode
+     */
+    public int getMode() {
```
cdjackson: I wonder if it wouldn't make sense to use an enum for `mode` to make it clearer for the upper layers.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
```diff
@@ -0,0 +1,244 @@
...
+    private static final int SWITCH_ALL_REPORT = 0x03;
+    private static final int SWITCH_ALL_ON = 0x04;
+    private static final int SWITCH_ALL_OFF = 0x05;
+    
+    private static final int SWITCH_ALL_EXCLUDED = 0x00;
+    private static final int SWITCH_ALL_INCLUDE_ON_ONLY = 0x01;
+    private static final int SWITCH_ALL_INCLUDE_OFF_ONLY = 0x02;
+    private static final int SWITCH_ALL_INCLUDE_ON_OFF = 0xFF;
+    private boolean isGetSupported = true;
+    private int mode = 0;
```
cdjackson: Should this be an int, or should we use an enum?

paixaop: I thought about it but wouldn't this cause some issues in converting from int to the enum and back?

paixaop:
Made the recommended changes, sorry for the delay.

bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveSwitchAllCommandClass.java
cdjackson: Any thoughts on this...

paixaop:
Chris anything needed before you can merge this?

cdjackson:
Yes, there's still the unanswered comment I had above about the mode and just wondering if it wouldn't make sense to use an enum rather than returning an integer.

paixaop:
Chris 

Sorry for the delay, move the mode to enum and added the node initialization code. Should be ready to merge.

PR #3109, opened 2015-08-30
add RFCW5 to the database

paixaop:
database files for the RFWC5 5-button scene controller device from Cooper

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/cooper/rfwc5.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/cooper/rfwc5.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>RFWC5</Model>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/database/cooper/rfwc5.xml
```diff
@@ -0,0 +1,77 @@
...
+			<id>0x85</id>
+		</Class>
+		<Class>
+			<id>0x86</id>
+		</Class>
+		<Class>
+			<id>0x87</id>
+		</Class>
+	</CommandClasses>
+	<Associations>
```
cdjackson: Are any of these association groups in use by the controller or are they just used to control other devices?  If they need to be configured to communicate with the controller, then you need to add ```<SetToController>true</SetToController>``` for the appropriate groups...

paixaop:
done, on group 255. hit commit too quickly :)

cdjackson:
Great - thanks :)

PR #3111, opened 2015-08-31
[Nest] Quiesce logging on status poll

watou:
If there is an error polling the Nest API (usually due to network issues), the failed poll would log as an ERROR and dump a stacktrace to openhab.log which is typically unhelpful.  Since network issues are usually temporary, this PR changes the log to WARN and the stacktrace is now not dumped (only the exception message is logged).  However, at the DEBUG level the stacktrace is dumped to the log for the non-typical failure situation.  @mrguessed, do you agree that this would be better?

### DIFF
--- a/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/NestBinding.java
+++ b/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/NestBinding.java
@@ -166,7 +166,11 @@ public class NestBinding extends AbstractActiveBinding<NestBindingProvider> impl
 				readNest(oauthCredentials);
 			}
 		} catch (Exception e) {
-			logger.error("Error reading from Nest:", e);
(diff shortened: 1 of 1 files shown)


mrguessed:
Ok with making it warn, instead of error, but would prefer to see only one log line in either case.

More like
```
if debug
   log.debug(...)
else
   log.warn(...)
```

It prevents log interleaving issues in multi-threaded envs and ensures there's only one log line per 'event'.  Not a hard rule but more a preference (vs 2 log lines in log file)

watou:
Thanks, Mark.  Good point.  How about this construct?
```
			if (logger.isDebugEnabled()) {
				logger.warn("Exception reading from Nest.", e);				
			} else {
				logger.warn("Exception reading from Nest: {}", e.getMessage());				
			}
```

mrguessed:
Sure, as long as the logger, at _warn_ level, emits the stack then that's the right construct...

PR #3114, opened 2015-08-31
Fix URL of GoogleTTS service after API changes

dominicdesu:
Fix URL of GoogleTTS by adding a `client` parameter to the URL as discussed in the [openHAB forum](https://groups.google.com/forum/#!msg/openhab/Sb8CuHDDCBk/O2OulKYCAwAJ) and on [stackoverflow](http://stackoverflow.com/a/31791632/3049015).


teichsta:
Thanks @dominicdesu

Issue #3090, opened 2015-08-23
Persisting Dimmer state erroneous with rrd4j persistence


aruder77:
I'm trying to persist a Dimmer item using rrd4j persistence. The item definition of the item is the following:

Dimmer  Shutter_AutoShutterPosition    "Auto-Mode P [%d %%]"
In the logs I see the save and restore of the item:

Stored 'Shutter_AutoShutterPosition' with state '66' in rrd4j database

Restored item state from 'Aug 21, 2015 11:08:00 PM' for item 'Shutter_AutoShutterPosition' -> '0.66000000000000003108624468950438313186168670654296875'
The dimmer state is apparently being saved and restored using different datatypes.
##

LKuech:
I stumbled upon the same issue as I was trying to make my rollershutters persistent (Because they loose there state each time when I change an item file otherwise).
I guess it is related to the PercentageType, because that is what both item type do have in common...

LKuech:
I did spend some time to find this bug.
It was there are 2 ways to fix it. I did choose the one that should keep the impact for existing rules near to zero by building just a special case for dimmer and rollershutter items during the restore of values.
I will test it a little bit and than do a pull request.

If you interested in the code, take a look at this branch:
https://github.com/LKuech/openhab/tree/rr4dj-restore-PercentType-Values

LKuech:
Pull request created:
https://github.com/openhab/openhab/pull/3122

Now we have to wait for one of the official devs...

J-N-K:
Isn't that fixed?

9037568:
Looks that way to me, too.

PR #3122, opened 2015-09-03
RRD4J: Handling of Dimmer and Rollershutter items during restore of values improved (Fix for #3090).

LKuech:


bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java
```diff
@@ -307,6 +309,10 @@ private State mapToState(double value, String itemName) {
 					return value==0.0d ? OnOffType.OFF : OnOffType.ON;
 				} else if(item instanceof ContactItem) {
 					return value==0.0d ? OpenClosedType.CLOSED : OpenClosedType.OPEN;
+				} else if(item instanceof DimmerItem || item instanceof RollershutterItem) {
+					// make sure Items that need PercentTypes instead of DecimalTypes
+					// do receive the right information
+					return new PercentType( (int) Math.round(value * 100) );
```
teichsta: why ```*100``` ?

LKuech: It seems the getStateAs method of DimmerItem and RollershutterItem do provide the state value to the persistence service divided by 100. Because of this the bug mentioned in issue #3090 is happening. 
If the rollershutter is on 70% while it was stored. It is just 0.7 in the database.

Changing the getStateAs methods seems to risky to me, has this might create impacts to existing rules and scripts which might use this methods.

teichsta:
Thanks @LKuech, also cherry-picked onto 1.8!

However could you please start a discussion about this getStateAs functionality in the Eclipse forum? Thanks!

PR #3125, opened 2015-09-04
Removed PersistentStateRestorer

### LINKED ISSUE #2661: Persistence API breaks compatibility to openHAB 2
In 1.6.1 a new service interface has been introduced: https://github.com/openhab/openhab/blob/master/bundles/core/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistentStateRestorer.java

This does was not synced to ESH and thus it potentially breaks the compatibility of persistence services with openHAB 2. 
I do not know the reasons for this change in the core runtime, but it might needs to be revisited as the parameter name "modelName" hints at the fact that there is a logical reference to DSL files here, which must not be the case from a core bundle (and hence breaks the architecture).

teichsta:
processed changes as suggested by @kaikreuzer in https://github.com/openhab/openhab/issues/2661#issuecomment-123662806. This should fix #2661 and also #1757


kgoderis:
Thomas, when would you integrate this commit so that I can update #3131 ?

teichsta:
now :-)

theoweiss:
@teichsta I think something went really wrong with this PR. I had a look into InfluxDBPersistenceService.java and it seems like you worked on a outdated version of the service and reverted a lot of changes.

teichsta:
thanks for noticing @theoweiss … i'll have a look at this issue soon!

teichsta:
oh yes, somehow i "managed" to unmerge some changes to InfluxDBService through this PR. I've created #3312 to fix this. Sorry for any inconvenience guys!

PR #3137, opened 2015-09-06
RFXCOM Add binding support for new sensors

Speederc:
Add :
- UV messages : UVN128,UVR128,UV138,UVN800,TFA
- Chime messages (only as input) : BYRONSX, BYRONMP001, SELECTPLUS, SELECTPLUS3, ENVIVO
- Average speed and temperature parameters for TFA Wind sensor


teichsta:
Thanks @Speederc !

PR #3150, opened 2015-09-10
Clean up bundle dependencies to make it compatible with openHAB 2.

### LINKED ISSUE #3149: Ecobee and Nest binding must not require io.rest bundle
Both the Ecobee and the Nest bundle currently have 
```
Require-Bundle: org.openhab.io.rest.lib
```
in their MANIFEST.MF. This prevents them from being used on openHAB 2.

As far as I can see, this can be simply removed, but I leave the change to the original authors, just to be sure.

kaikreuzer:
Fixes #3149


teichsta:
could you please have a look and merge? Thanks, Thomas E.-E.

watou:
I've built and tested resulting binding JARs on OH 1.7.1 on an RPi2 and they both performed properly.

PR #3151, opened 2015-09-10
[Ecobee] Added optional ecobee:timeout config param

watou:
Ecobee's API servers are producing very frequent HTTP request timeouts, so the previously hard-coded 10-second timeout is proving to be too short to complete many polls.  By adding 
```
ecobee:timeout=20000
```
to your openhab.cfg file, you will increase the HTTP request timeout from the default 10 seconds to 20 seconds.  There is no way to know if this is long enough for the HTTP request to succeed, but this gives the binding user the chance to extend the timeout as they see fit.

Also incorporated's @mrguessed's advice on having a single log message (with optional stack trace if in DEBUG logging mode).


watou:
[binding JAR for test](https://dl.dropboxusercontent.com/u/4286376/ecobee-config-timeout/org.openhab.binding.ecobee-1.8.0-SNAPSHOT.jar)
[action JAR that goes with if needed](https://dl.dropboxusercontent.com/u/4286376/ecobee-config-timeout/org.openhab.action.ecobee-1.8.0-SNAPSHOT.jar)

Issue #3064, opened 2015-08-14
ZWave: Add LS Control - E861C to database


CarstenMadsen:
Is it possible to add this temperature sensor to the databse, so I can use it in openhab.

http://www.pepper1.net/zwavedb/device/64

SkynetHome:
I will get this done.

CarstenMadsen:
How can I test this?

cdjackson:
Download the latest version from cloudbees

CarstenMadsen:
I have now tried with the latest version, but with no lock.
Are you sure that it has been added to the database.

I can't find the folder "lscontrol" under the zwave database.
https://openhab.ci.cloudbees.com/job/openHAB/org.openhab.binding$org.openhab.binding.zwave/ws/database/

cdjackson:
No - it doesn't look like it was added...

CarstenMadsen:
Can you add it?
Then I will try to test it.

cdjackson:
Can this be closed - the device has now been added?

cdjackson:
Device has been added but no feedback. If there is an issue, please reopen.

PR #3167, opened 2015-09-14
Added E861C

SkynetHome:
Added LS Control E861C per Issue #3064

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/lscontrol/e861c.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/lscontrol/e861c.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>E861C</Model>
+	<Label lang="en">Temperature Sensor</Label>
+	<CommandClasses>
+		<Class><id>0x20</id></Class>
+		<Class><id>0x80</id></Class>
+		<Class><id>0x86</id></Class>
+		<Class><id>0x84</id></Class>
+		<Class><id>0x70</id></Class>
+		<Class><id>0x31</id></Class>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/database/lscontrol/e861c.xml
```diff
@@ -0,0 +1,49 @@
...
+		<Parameter>
+			<Index>1</Index>
+			<Type>byte</Type>
+			<Default>0</Default>
+			<Size>1</Size>
+			<Label lang="en">Delta Percentage</Label>
+			<Help lang="en">2 % - 56 % -> Delta</Help>
+		</Parameter>
+		<Parameter>
+			<Index>2</Index>
```
cdjackson: I would make this a list with just the two options.

bundles/binding/org.openhab.binding.zwave/database/lscontrol/e861c.xml
```diff
@@ -0,0 +1,49 @@
...
+			<Type>byte</Type>
+			<Default>0</Default>
+			<Size>1</Size>
+			<Label lang="en">Celsius/Fahrenheit</Label>
+			<Help lang="en">0 - 64 -> Celsius : 64 - 127 -> Fahrenheit</Help>
+		</Parameter>
+		<Parameter>
+			<Index>3</Index>
```
cdjackson: I would make this a list with just the two options.

SkynetHome:
Its been a while since I looked at this, not sure I understand why I would create a list with two options for parameters that take a range of values.

cdjackson:
Because there's really only 2 values. 0-64 are the same, and 65 to 127 are the same, so just use 2 values - one for C and one for F.

PR #3172, opened 2015-09-16
rrd4j as dependency

FallenRiteMonk:
Signed-off-by: Florian Klement <fallenritemonk@gmail.com> (github: FallenRiteMonk)


kaikreuzer:
What is this good for?

FallenRiteMonk:
Updating to new version would be as simple as changing a number and removing binaries from the repo helps to prevent it from getting to big!

kaikreuzer:
No, it is not that easy, because the bundle won't compile within the IDE anymore. This is why we either have 3rd party libs within the target platform or bundles within a lib folder.

PR #3190, opened 2015-09-21
Enhance serial binding to support regex

csowada:
With this update you can use RegEx to parse serial data.

Example:

``Number SerialData	"Date [%.1f °C]"	{ serial='COM7,REGEX(D:1.*H:([0-9.]*))' }``


teichsta:
Thanks @csowada for this contribution. I am wondering if you are aware of the Transformation services which has been introduced to openHAB some time ago. In favour of implementing a "hardcoded" REGEX transformation i would rather suggest to make the transformation services available in the serial binding.

csowada:
Hello @teichsta , you're right I was aware of the service. But it's easy to change.
So I will change it.

BTW: The service doesn't cache the compiled regex objects, not optimal for performance if serial updates every n seconds ?!

teichsta:
Thanks @csowada! Please also update the wiki page accordingly!

csowada:
Ok, modified the wiki page.
But someone should update the transformation service wiki page.

TheNetStriker:
Is it possible to replace custom bytes in the received string using this regex function?

I've connected a serial port device that sends STX and EOT bytes to openhab. Openhab directly saves these bytes to the string. If such a value is saved I cannot call some of the rest services anymore.

E.g. the url http://127.0.0.1:8080/rest/items/serial_item throws the error "error on line 1 at column 126: PCDATA invalid Char value 2". The same error also occurs when calling  http://127.0.0.1:8080/rest/items.

I've compiled the new 1.8 version of the serial binding and set the serial configuration to serial="/dev/ttyUSB0@57600,REGEX(\u0002(.*)\u0004)", but the error still apperas. Is my regex wrong for this?

9037568:
This sounds like an xml parsing error.  U+0002 and U+0004 were not valid in
XML before 1.1...

I'd love to see the stack trace for this error.

TheNetStriker:
Sadly there is no detailed error in the log file. I've tried to add <logger name="org.openhab.io.rest" level="DEBUG"/> to logback.xml but still no detailed error. Can you tell me what I have to configure to get the stacktrace in my log?

9037568:
Have you tried running the "start_debug" script instead of plain "start"?

If that doesn't provide additional output, you can try adding this to your
logback.xml.
    <logger name="org.openhab.binding.serial" level="TRACE" />

I'll also note we may be sufficiently off-topic from the original
defect/discussion that we should take this to the forum...

TheNetStriker:
I've enabled debug mode and added the serial port logger, but still no stack trace of the error. I found out that the regex expression (.*) removes the STX and EOT characters, but the device also sends ACK bytes in between, so the error still occurs. I've opened a thread in the forum regarding this problem: https://community.openhab.org/t/error-in-rest-service-when-serial-port-sends-stx-ack-eot-bytes/3932

TheNetStriker:
I think I just found a bug in your code. If no pattern is defined nothing happened. I found out that entry.getValue().type.isInstance(StringItem.class) is never true. I've replaced it with entry.getValue().type == StringItem.class and now it's working. (Same for the SwitchItem)

marekhalmo:
@csowada could you look into [this issue please?](https://github.com/openhab/openhab-core/issues/128)

PR #3202, opened 2015-09-25
Add support for 'On/Off plug-in unit' in hue binding

tobiaswegner:
This patch adds support for hue devices that offer only on/off switching, no dimming. They deliver no brightness information so without this change, an exception is thrown. The patch returns 0 or 254 based on the "on" state of the hue device.

### DIFF
--- a/bundles/binding/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/data/HueSettings.java
+++ b/bundles/binding/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/data/HueSettings.java
@@ -151,9 +151,13 @@ public class HueSettings {
 			logger.error("Hue bridge settings not initialized correctly.");
 			return 0;
 		}
-		return (Integer) settingsData.node("lights")
-				.node(deviceId).node("state")
-				.value("bri");
+		Object bri = settingsData.node("lights").node(deviceId).node("state").value("bri");
+		if(bri instanceof Integer) {
+			return (Integer) bri;
+		} else {
+			//probably not dimmable, return on state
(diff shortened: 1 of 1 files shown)


kaikreuzer:
@tobiaswegner Just out of curiosity: What hue devices are there that aren't dimmable?

tobiaswegner:
Plug adapters like this http://www.busch-jaeger-katalog.de/6710-0-0004,artikel.html or wall mounted actuators like http://www.busch-jaeger-katalog.de/6710-0-0002,artikel.html that are connected e.g. to your multimedia system instead of a lamp.
I think even with the genuine Philips plug adapter http://www.philips.de/c-p/6916531PH/livingwhites-adapter dimming can be disabled via a mechanical switch if dimming makes no sense with the particular load.

kaikreuzer:
Ah, cool, I didn't know the Busch Jäger products yet! But 120 EUR for a smart plug? They should revisit their price list...

> I think even with the genuine Philips plug adapter http://www.philips.de/c-p/6916531PH/livingwhites-adapter dimming can be disabled

You are right, I am actually using quite some of them (but never tried out the switch mode and how that behaves).

tobiaswegner:
> Ah, cool, I didn't know the Busch Jäger products yet! But 120 EUR for a smart plug? They should revisit their price list...

Yes, they are only a few weeks old. They price is as expected from Busch-Jaeger, but at least it is noticeable lower than the KNX Powernet devices. Anyway, I agree, it is no bargain buy.

teichsta:
Thanks @tobiaswegner! Is that functionality already implemented in the ESH version of the binding? I know it's a rewrite but it might be worth having into that code. Would you mind @tobiaswegner checking double checking the ESH code?

PR #3203, opened 2015-09-25
Add BeNext Power Switch and Door Sensor as supported Z-Wave devices

randomname872634:
Add BeNext Power Switch as supported Z-Wave device, based on http://www.pepper1.net/zwavedb/device/299, but with Product ID 0x0101 instead of Product ID 0x0100

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/benext/doorsensor.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/benext/doorsensor.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>doorSensor</Model>
+	<Label lang="en">BeNext Door Sensor</Label>
+	<CommandClasses>
+		<Class>
+			<id>0x20</id>
+		</Class>		
+		<Class>
+			<id>0x30</id>
+		</Class>
+		<Class>
+			<id>0x31</id>
+		</Class>
+		<Class>
+			<id>0x70</id>
+		</Class>
+		<Class>
+			<id>0x71</id>
+		</Class>
+		<Class>
+			<id>0x72</id>
+		</Class>
+		<Class>
+			<id>0x73</id>
+		</Class>
+		<Class>
+			<id>0x84</id>
+		</Class>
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.zwave/database/benext/switch.xml
```diff
@@ -0,0 +1,44 @@
...
+		<Class>
+			<id>0x85</id>
+		</Class>
+		<Class>
+			<id>0x86</id>
+		</Class>
+	</CommandClasses>
+	<Associations>
+		<Group>
+			<Index>1</Index>
```
cdjackson: Do you expect this to be configured to talk to OH. If so, then you must set the `<SetToController>` to true

randomname872634:
You are correct, I do expect the device to talk back to OpenHAB, so I've corrected it. I've also added a BeNext Door Sensor device for which I already used the SetToController. I've tested both device configurations and they are working correctly.

bundles/binding/org.openhab.binding.zwave/database/benext/doorsensor.xml
```diff
@@ -0,0 +1,109 @@
...
+				<Value>2</Value>
+				<Label lang="en">If any mode other then 3, that value will be reported after a get but will be handled in SW as mode 1.</Label>
+			</Item>
+			<Item>
+				<Value>3</Value>
+				<Label lang="en">Z-Wave chip is always on to request e.g. version or manufacturer id.</Label>
+			</Item>
+			<Item>
+				<Value>4</Value>
+				<Label lang="en">If any mode other then 3, that value will be reported after a get but will be handled in SW as mode 1.</Label>
```
cdjackson: Why are 2 and 4 the same? If they really do do the same thing, then why not just have one of them in the list?

Ideally, it would be best to keep the option labels short so that they fit in the options box, and then add any additional text into the `<Help>` tag.  This comment is also true for the other long labels.

randomname872634:
I've squashed everything into two commits and added <help>-tags for the door sensor. I've been copy-pasting the manufactor documentation too much, sorry for that.

PR #3211, opened 2015-09-28
 modbus binding bug fix - connections were left open on command execution

paolodenti:


teichsta:
Thanks for this fix @paolodenti. Something seem to be messed up with the commits (87 commits in this PR). Could you please squash them into one?

paolodenti:
The real problem about the 87 commits is that in order to fix this modbus bug i branched my master that is having a pull request pending for a new binding. I started from there, removing the "binding" commits.
The evil comes from the fact that i used my master instead of a branch for the pull request and that the new binding is not yet pulled. And i cannot really squash them because they are intermixed in the master with other commits on the master branch.

So i so not really know how to proceed about squashing ... one way could be deleting the binding pull request, cleaning my master, and doing another pull request on a branch. And delete the modbus pull request and use another branch (from the clean master) just for this fix.

Unfortunately everything comes from the mistake of using the master for the pull request making impossible to squash the commits.

Any suggestion is welcome, please feel free to suggest a viable git path to follow.

teichsta:
ok, then the best would be to close this and create another PR from a specific branch. Thanks, Thomas E.-E.

paolodenti:
substituted by pull request #3228

Issue #658, opened 2013-12-19
modbus TCP - problem with multiple connections to same slave


uoOcF:
For each start-address a separate slave-entry in the openhab.cfg is needed and it seems like that the modbus-tcp-binding keeps a persistent tcp-connection open for each slave-entry, even if it is the same IP:PORT.

The Problem: The number of simultaneous tcp connections to slave devices is limited (i guess it depends on the device, in my case 2 connections are possible at the same time)

Suggestion:
Use only one permanent tcp-connection for all slaves which use the same IP:PORT for connecting.

(not sure how/if the serial connection suffers the same problems)

Discussion: https://groups.google.com/d/topic/openhab/9ty1C6zFs88/discussion

pjelonek:
I have the same issue.
PLC controller cannot handle to many connections and the binding code is opening new connections without closing the old ones (checked with tcpdump). After couple of pools cycles module stops to work.
Should be fixed as evidently bug.

wtrucci:
Same problem here.
There is a solution, tested and it works very well.
https://github.com/openhab/openhab/pull/3211/files

teichsta:
issue has been fixed through #3228 meanwhile

PR #3228, opened 2015-10-02
modbus binding bug fix - connections were left open on command execution

paolodenti:
substitues pull request #3211 / fix for #658


teichsta:
Thanks for this fix. Since i am no modbus i have absolutely no clue about the insights of this binding. To me it doesn't look like the connection is being opened within execute() so it doesn't sound to me good to close something which is not opened before. Could you shed some light on this? Can other users confirm  this fix is working?

wtrucci:
I confirm that this fix modbus connection.
This is a fix for the issue https://github.com/openhab/openhab/issues/658

paolodenti:
@teichsta 
connection is opened at the beginning of method but it is just never closed. issue #658, pointed out by @wtrucci, gives failure after too many connections for the device are opened and not closed or "timeouted" because of that

```
      //3. open the connection if not connected
      if (!m_Connection.isConnected()) {
        try {
          m_Connection.connect();
          m_IO = m_Connection.getModbusTransport();
        } catch (Exception ex) {
          throw new ModbusIOException("Connecting failed.");
        }
      }
```

teichsta:
Thanks @paolodenti!

PR #3233, opened 2015-10-02
modify Cosm Persistence for compatibility with Xively

fazioa:
modify Cosm Persistence for compatibility with actual Cosm, named Xively, where items alias ==null


teichsta:
Thanks for this update @fazioa! Since the different names are quite confusing, would you mind rename this persistence service completely and also change the according wiki page (see https://github.com/openhab/openhab/wiki/Cosm-Persistence)?

teichsta:
See also #334 it would be great if you could take over. Will you?

teichsta:
Thanks @fazioa!

PR #3241, opened 2015-10-03
Improve logging output for HueBinding when retrieving settings

dominicdesu:
The error described in https://community.openhab.org/t/hue-pairing-error/2969 is difficult to analyze because the logging output is not sufficient.
Therefore this pull request improves logging output when retrieving JSON settings from the hue bridge.


teichsta:
Thanks @dominicdesu!

PR #3251, opened 2015-10-05
Fix for DateTimeItem to accept DateTimeType commands.

druciak:
Some time ago I made DateTimeType a command, but forgot to add this type to accepted commands in DateTimeItem, which prevents sending commands as String. This PR fixes this issue.


teichsta:
Thanks @druciak! ESH is already up-to-date …

druciak:
Good to know, I am still a little behind regarding OH2. Thanks for merging this so quick.

staehler:
Is this bug fixed, as I have the same failure like udo1toni (#3282)?

watou:
I think this pull request #3251 is good but exposed a weak bit of code [here](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProvider.java#L338).  Issue #3282 is still open.

PR #3261, opened 2015-10-07
[Satel] Items initial poll fix

druciak:
This PR fixes the way initial refresh is done for items added after connection is established. 
Since items are updated only when appropriate state in the alarm system changes, items require initial poll of current state when new item added. This is done through additional flag in binding configuration.


teichsta:
Thanks @druciak!

PR #3262, opened 2015-10-07
Mail action: send message as UTF-8

druciak:
Sends message as UTF-8 which allows non-ascii characters in message body.


teichsta:
Thanks @druciak!

PR #3249, opened 2015-10-05
JSR223: Support for UpdatedEventTrigger

steve-bate:
This pull request adds an UpdatedEventTrigger to the JSR223 bundle. Most of the code to support it was already in place, but it was never completed. Previously, the ChangedEventTrigger was coded to respond to updated events but the trigger was not registered to respond to update events so it did not work as expected. I've added an UpdatedEventTrigger and am registering it to be evaluated on update events. I've also added the class to the predefined globals defined during script invocation.

Related discussion: https://community.openhab.org/t/jsr223-updatedeventtrigger/2908/2

PR #3263, opened 2015-10-08
JSR223: Support for UpdatedEventTrigger

steve-bate:
Second attempt. Commits are squashed. See discussion at #3249.


teichsta:
Thanks again @steve-bate!

PR #3273, opened 2015-10-10
Update products.xml

mcarlosro:
Changes the Nodon ID from the hexadecimal value to the integer value.


cdjackson:
This change is not correct.

PR #3284, opened 2015-10-12
Add Düwi ZW BJ ES 1000 a.k.a. ZWave.me 05457

mstormi:
ZWave.Me bought Düwi and apparently renamed their switch.
Still, I have a Düwi labeled version that has a different manufacturer + device ID.
Without defining it, habmin at present does not allow me to access the config parameters.
I hope it's ok to reference the existing zwave.me device file instead of copying it ?


mstormi:
Hmm, I just noticed that the device is already present in our database as a 'Reitz 054313', although that file does not provide any configuration options (which is why my habmin does not show it)
Reitz used to be the mother company of Düwi, but now they no longer are ... oh dear, what a mess.

As I guess we cannot have two entries in products.xml with the same device ID, and as the 'Reitz 054313' entry does not provide any configuration parameters whatsoever, I believe it makes sense to replace it with the zwave.me 05457 file.
Do you agree?

cdjackson:
With this change, you have left the reitz/054313.xml file in the folder but it is not linked. While this doesn't stop anything working, it will be confusing.

Either the reitz/054313.xml file should be removed, or you should copy the zwave.me file over the reitz/054313.xml file.

I think it's better to keep the reitz/054313.xml file an update it personally, and this is how it will work in OH2 (ie separate files per device).

mstormi:
I think it's better to remove the reitz file because others keep using the zwave.me 054313 and 05457
switches, and the authors of those .xml files will eventually update their files, but won't know that they would need to update the reitz file, too.
Also, I saw there already exists a 054313z.xml that I now changed the reference in products.xml to. The 054313 and 05457 files actually look the same, except for the ID. I believe it's just mechanically different, but zwave-wise identical versions of the switch, but I'm not 100% sure (else I'd vote to merge those, too).

cdjackson:
Hmmm - but are you sure that if one device gets updated, that the other will also be updated, or are they sold separately? It would be perferable I think to keep them separate - certainly in OH2, they will be different files since there is no centralised index.  I think it’s a bit ‘dodgy’ to assume that two devices are the same and if one is updates, that the other will also have its software updated - unless this is stated by the manufacturer of course.

mstormi:
Am 13.10.2015 um 22:20 schrieb Chris Jackson:
> Hmmm - but are you sure that if one device gets updated, that the 
> other will also be updated, or are they sold separately?
The device is still the same (because company was sold and HW has been 
relabeled), and the one with the reitz ID (which I have) won't be sold 
any more (hasn't been for a year or so, in fact).
> It would be perferable I think to keep them separate - certainly in 
> OH2, they will be different files since there is no centralised index. 
> I think it’s a bit ‘dodgy’ to assume that two devices are the same and 
> if one is updates, that the other will also have its software updated 
> - unless this is stated by the manufacturer of course.
True in general, but in this case I'm sure the HW is the same and my 
thinking was that if someone changes the zwaveme file, all reitz ID 
owners should profit from that, too.
I'm probably the only one anyway :-) Actually, I didn't think of new 
features but xml file error corrections. If you ever update the firmware 
to get new features, you'll get the newer zwaveme ID and keep
using the 'right' xml file from there on.

cdjackson:
If you keep two files, then if you update the firmware (assuming that is even possible - not that many devices support firmware updates) then you would change to the new device. Again, I think this supports keeping the two XML files separate.

mstormi:
alright then, I imported the zwaveme config into the reitz file. Ok now?

PR #3290, opened 2015-10-12
[Satel] action added, binding refactored.

druciak:
This PR introduces action bundle for Satel. Action bundle contains 3 actions at the moment:
- _satelIsConnected_ - returns _true_ if OH is connected to configured module
- _satelReadEvent_ - reads single record from alarm system's event log
- _satelReadDeviceName_ - reads device name to convert device ids to text

For this purpose Satel binding exposes a service which allows sending commands directly and synchronously to communication module and therefore it also needed bigger changes.


wtrucci:
Hi, thanks for this binding, I'm starting to buy and test a satel alarm system. With this new action is it possible to have a push (using pushover for example) on a specific event?
With Integra system is possible to use access control. Is it implemented in your binding?
B R
Walter

druciak:
Hi, it should be possible even without this action, but it really depends on what exactly you are going to achieve.
Regarding access control I believe it is implemented, at least something is. I don't have access control, so it is hard to say, but if it does not work for you, I will try to fix it with your help.

wtrucci:
Thanks a lot for you support on it. When I will have in my hand a Satel board I will test it.

druciak:
Hey, I made some changes in the code and now it builds successfully on 1.9. Also changed year in license info, but unfortunately I must remove dependency to action bundle, as it requires refactored version of the binding, so there will be another PR with just this small change in pom.xml.
Any chance to get this merged this year? ;)

druciak:
Fixed once again all the conflicts. Hoping to get this merged soon...

watou:
Hi @druciak, thanks for all the work here.  Could you verify that the code formatting is correct throughout?  I saw some diffs that didn't look right.  Do you have positive tester feedback from the community forum for test JARs of the refactored binding and action bundles?  If all clear, I will then begin reviewing your big refactor.  Thanks!

druciak:
I'll check the formatting in all the files, it looks like it changed at some point (tabs/spaces). Regarding stability of the binding and action it has been running for several months on my and few other persons installations without any troubles so it looks pretty good IMO.

druciak:
White spaces corrected, builds successfully, also make a quick test and works. I think you can start your review.

bundles/action/org.openhab.action.satel/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.openhab.action.satel</name>
+	<comment>This is the ${binding-name} binding of the open Home Automation Bus (openHAB)</comment>
```
watou: Not a binding, an action service.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
```diff
@@ -0,0 +1,190 @@
...
+            @ParamDoc(name = "eventIndex", text = "index of event to read, -1 for the most recent") int eventIndex) {
+        if (!satelIsConnected()) {
+            return null;
+        }
+        ReadEventCommand cmd = new ReadEventCommand(eventIndex);
+        if (!SatelActionService.satelCommModule.sendCommand(cmd)) {
+            logger.error("Unable to read record for given index: {}", eventIndex);
+            return null;
+        } else if (cmd.isEmpty()) {
+            logger.warn("No record under given index: {}", eventIndex);
```
watou: Why warn here, but error above?

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param deviceType type of the device
+     * @param deviceNumber number of the device
+     * @return string representing device's name or <code>null</code> if device is not present or an error occurred
+     */
+    @ActionDoc(text = "Returns name of a device.")
+    public static String satelReadDeviceName(
+            @ParamDoc(name = "deviceType", text = "type of the device, one of: PARTITION, ZONE, USER, EXPANDER, LCD, OUTPUT, TIMER, TELEPHONE, OBJECT") String deviceType,
+            @ParamDoc(name = "deviceNumber", text = "number of the device to read") int deviceNumber) {
+        if (!satelIsConnected()) {
+            return null;
```
watou: Why no log here but log in above method?

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/SatelActionService.java
```diff
@@ -0,0 +1,86 @@
...
+    public SatelActionService() {
+    }
+
+    public void activate() {
+    }
+
+    public void deactivate() {
+        // deallocate Resources here that are no longer needed and
+        // should be reset when activating this binding again
+    }
```
watou: If `policy="static"` in action.xml (see [here](https://felix.apache.org/documentation/subprojects/apache-felix-maven-scr-plugin/scr-annotations.html)) but `activate`/`deactivate` do nothing, perhaps it should be `policy="dynamic"`?  Also, remove boilerplate comment about "this binding".

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/SatelActionService.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @{inheritDoc}
+     */
+    @Override
+    public void updated(Dictionary<String, ?> config) throws ConfigurationException {
+        if (config != null) {
+            // action has no configuration, we just keep this method for the future
+        }
+    }
```
watou: No need to override `updated` method since not needed.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/SatelBindingConfig.java
```diff
@@ -1,146 +1,146 @@
...
+    protected State booleanToState(Item item, boolean value) {
+        if (item instanceof ContactItem) {
+            return value ? OpenClosedType.OPEN : OpenClosedType.CLOSED;
+        } else if (item instanceof SwitchItem) {
+            return value ? OnOffType.ON : OnOffType.OFF;
+        } else if (item instanceof NumberItem) {
+            return value ? DECIMAL_ONE : DecimalType.ZERO;
+        }
+
+        return null;
```
watou: Perhaps return `UnDefType.NULL` instead since all item types accept `UnDefType` states?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/ClearTroublesCommand.java
```diff
@@ -0,0 +1,31 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.satel.command;
+
+/**
+ * Command class for command that clear troubles memory.
+ *
+ * @author Krzysztof Goworek
+ * @since 1.7.0
```
watou: 1.9.0 since it's a new file?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/ControlCommand.java
```diff
@@ -0,0 +1,64 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.satel.command;
+
+import org.openhab.binding.satel.internal.protocol.SatelMessage;
+
+/**
+ * Base class for all commands that return result code in the response.
+ *
+ * @author kgoworek
```
watou: Krzysztof Goworek :-)

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/ControlObjectCommand.java
```diff
@@ -0,0 +1,63 @@
...
+import org.openhab.binding.satel.internal.protocol.SatelMessage;
+import org.openhab.binding.satel.internal.types.ControlType;
+
+/**
+ * Command class for commands that control (change) state of Integra objects,
+ * like partitions (arm, disarm), zones (bypass, unbypass) outputs (on, off,
+ * switch), etc.
+ *
+ * @author Krzysztof Goworek
+ * @since 1.7.0
```
watou: 1.9.0 since it's a new file?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/IntegraStatusCommand.java
```diff
@@ -0,0 +1,95 @@
...
+import org.openhab.binding.satel.internal.event.IntegraStatusEvent;
+import org.openhab.binding.satel.internal.protocol.SatelMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Command class for command that returns Integra RTC and basic status.
+ *
+ * @author Krzysztof Goworek
+ * @since 1.7.0
```
watou: 1.9.0 since it's a new file?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/NewStatesCommand.java
```diff
@@ -1,66 +1,69 @@
...
+     */
+    @Override
+    public boolean handleResponse(EventDispatcher eventDispatcher, SatelMessage response) {
+        if (super.handleResponse(eventDispatcher, response)) {
+            // dispatch event
+            eventDispatcher.dispatchEvent(new NewStatesEvent(response.getPayload()));
+            return true;
+        } else {
+            return true;
+        }
```
watou: Move `return true;` out of `if/else` block to make it clearer that it isn't a typo to return true in either case (if it's not a typo).

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/SatelCommandBase.java
```diff
@@ -1,181 +1,197 @@
...
+import org.openhab.binding.satel.internal.event.EventDispatcher;
+import org.openhab.binding.satel.internal.protocol.SatelMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Base class for all command classes.
+ *
+ * @author Krzysztof Goworek
+ * @since 1.9.0
```
watou: 1.7.0 (I think if the file existed since 1.7.0 but was refactored for 1.9.0, it's still correct to say the file has been around since 1.7.0).

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/SetClockCommand.java
```diff
@@ -0,0 +1,45 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+
+import org.apache.commons.lang.ArrayUtils;
+
+/**
+ * Command class for command to set RTC clock.
+ *
+ * @author Krzysztof Goworek
+ * @since 1.7.0
```
watou: 1.9.0 since new file.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelBinding.java
```diff
@@ -248,4 +247,68 @@ private synchronized boolean switchForceRefresh(boolean forceRefresh) {
...
+        }
+
+        while (!Thread.interrupted()) {
+            // wait for command state change
+            try {
+                synchronized (command) {
+                    command.wait(this.satelModule.getTimeout());
+                }
+            } catch (InterruptedException e) {
+                // ignore, we just leave the loop
```
watou: did you mean to leave the loop here?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
```diff
@@ -1,499 +1,494 @@
...
+                }
+            }
+
+            // return read message
+            return SatelMessage.fromBytes(baos.toByteArray());
+
+        } catch (IOException e) {
+            if (!Thread.interrupted()) {
+                logger.error("Unexpected exception occurred during reading a message", e);
+            }
```
watou: No log message needed for other IOExceptions?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
```diff
@@ -1,499 +1,494 @@
...
+                }
+            }
+            os.write(FRAME_END);
+            os.flush();
+            return true;
+
+        } catch (IOException e) {
+            if (!Thread.interrupted()) {
+                logger.error("Unexpected exception occurred during writing a message", e);
+            }
```
watou: No log message needed for other IOExceptions?

watou:
Very nice, comments are minor.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelBinding.java
druciak: Thread.interrupted() will return "true" in case InterruptedException is thrown, so we leave the loop in line 273.

bundles/action/org.openhab.action.satel/.project
druciak: This is generated comment, most of action bundles has tha same comment. Also (as far as I know) this is not shown anywhere, IMO not important, but will correct it.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
druciak: When senCommand returns "false" it does mean a failure, that's why we log an error. When it returns "true", but there is no data, it means that wrong index has been given and warn should be enough to tell the user something is broken in the rule.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
druciak: Could you please explain what you mean? Failures on sending command are logged. Only warn, but most common reason is invalid device number provided by the user. But if you think "error" would be better here, please let me know.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/SatelBindingConfig.java
druciak: In this case `null` is a marker that the method couldn't handle conversion. In such case the code just skip update of the item.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
watou: ok thanks.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/SatelActionService.java
druciak: None of the actions has `policy="dynamic"` and I don't feel like an expert here to be the only that decided to change default setting. ;) I just followed other action bundles and boilerplate code...

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
watou: I mean that a user would call `satelReadDeviceName`, but would get back `null` with no other feedback in the log or otherwise, leaving them to wonder why they got `null`.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/SatelBindingConfig.java
watou: ok thanks

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/SatelActionService.java
druciak: Well, if we want to implement `ManagedService` interface we need to implement this method. I left generated method as it might be usefull in the future and I don't think this matters. But if you think it does, I will remove this code.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelBinding.java
watou: The empty `catch` block is followed by the `switch` statement below, so the `while` loop isn't really left until more instructions are performed.  So it's confusing, even if it works correctly in practice.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/Satel.java
druciak: Ok, will add a log.

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/SatelActionService.java
watou: Understood.  At least if all the action bundles are wrong somehow, they can all be fixed at once. ;-)

bundles/action/org.openhab.action.satel/src/main/java/org/openhab/action/satel/internal/SatelActionService.java
watou: Definitely doesn't matter.  Even the `config != null` has a documentation value, so it's all good.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/ClearTroublesCommand.java
druciak: Moved from another package: https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/command/ClearTroublesCommand.java

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/ControlCommand.java
druciak: Corrected in all files.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/ControlObjectCommand.java
druciak: Moved: https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/command/ControlObjectCommand.java

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/IntegraStatusCommand.java
druciak: Moved: https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/command/IntegraStatusCommand.java

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/SatelCommandBase.java
druciak: This class was renamed also, as now SatelCommand is an interface, so I needed to change original SatelCommand to SatelCommandBase (which implements SatelCommand intf.). If you think it's OK to say it is available since 1.7.0, I will change it.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/SetClockCommand.java
druciak: Moved: https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/command/SetClockCommand.java

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/SatelCommandBase.java
watou: I think if the file existed since 1.7.0, even if its contents are different, then it makes sense to leave it as it was.  But we're definitely talking about very small stuff!

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelBinding.java
druciak: You are right, I am now looking at the whole method. Would changing the comment be enough (less confusing), or you would like to change the whole loop?

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/command/NewStatesCommand.java
druciak: I think it is a typo, corrected.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
druciak: Not sure what you mean by other IOExceptions? I catch them all here, but if thead has been interrupted likely it was the reason of thrown exception. I did extensive testing for networking failures and reconnection mechanism and I think I did this on purpose. Anyway if we are stopping communication (for any reason) exceptions at this point don't matter IMO.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
druciak: Same as above.

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/SatelBinding.java
watou: Please make it as understandable as possible, however you see fit! :-)

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
watou: "other IOExceptions" meant when the thread was interrupted, so I think you're saying that only when the IOException is not the result of an interrupted thread it needs to be logged.  If so, fair enough!

druciak:
I just pushed the commit again with all the corrections.

watou:
Thanks very much, @druciak.  I will monitor the next Cloudbees build [here](https://openhab.ci.cloudbees.com/job/openHAB1-Addons/).

watou:
The [nightly build](https://openhab.ci.cloudbees.com/job/openHAB1-Addons/lastSuccessfulBuild/) succeeded!  Could you verify the correct function of the resulting:
* [action bundle](https://openhab.ci.cloudbees.com/job/openHAB1-Addons/lastSuccessfulBuild/artifact/bundles/action/org.openhab.action.satel/target/)
* [binding bundle](https://openhab.ci.cloudbees.com/job/openHAB1-Addons/lastSuccessfulBuild/artifact/bundles/binding/org.openhab.binding.satel/target/)

druciak:
Just did a quick test and both seem to work correctly. Thanks for quick review and merge.

watou:
Thanks!

Issue #3194, opened 2015-09-22
onewire doesn't refresh some items after restart (bug?)


staehler:
Hi,

I have a reproducible misbehaviour with [onewire-binding](https://github.com/openhab/openhab/wiki/One-Wire-Binding) since some weeks. Unfortunately I cannot identify, which openhab release worked fine and which not. At the moment I run openhab 1.8 snapshot 1001.
After restarting openhab, every time the same **three windows** contacts remain *undefined*. Even when I manually open the window, nothing changes. When then the onewire.items file changes (even the timestamp with "*touch onewire.items*") the item state recovers. This behaviour is absolutely reproducible after every restart of openhab. This means, after the restart I need to "*touch onewire.items*". 
As mentioned, a openhab restart doesn't solve this problem, it's the cause for this problem.
```
Contact FensterGal "Galerie [MAP(de.map):%s]" (gGalerie,gFenster) { onewire="deviceId=12.2F38B6000000;propertyName=sensed.A;refreshinterval=40" }
Contact FensterSZ "Schlafzimmer [MAP(de.map):%s]" (gSchlafzimmer,gFenster) { onewire="deviceId=12.FD3CB6000000;propertyName=sensed.A;refreshinterval=31" }
Contact FensterWZre "Wohnzimmer rechts [MAP(de.map):%s]" (gWohnzimmer,gFenster) { onewire="deviceId=29.A47A0D000000;propertyName=sensed.0;refreshinterval=2" }
Contact FensterWZli "Wohnzimmer links [MAP(de.map):%s]" (gWohnzimmer,gFenster) { onewire="deviceId=29.A47A0D000000;propertyName=sensed.1;refreshinterval=2" }
Contact FensterGZ "Gästezimmer [MAP(de.map):%s]" (gGaestezimmer,gFenster) { onewire="deviceId=12.133CB6000000;propertyName=sensed.A;refreshinterval=60" }
Contact FensterBad "Bad [MAP(de.map):%s]" (gBad,gFenster) { onewire="deviceId=12.3130B6000000;propertyName=sensed.A;refreshinterval=30" }
Contact FensterKue "Küche [MAP(de.map):%s]" (gKueche,gFenster) { onewire="deviceId=12.8136B6000000;propertyName=sensed.A;refreshinterval=60" }
```
As the items-file works fine, after the restart, there must be a bug in the startup sequence.
[...]

PR #3303, opened 2015-10-17
Fix OneWireBinding, possible ClassCastException during startup

Dennis650:


Dennis650:
New PullRequest to fix #3194

PR #3306, opened 2015-10-17
Use https:// instead of http:// in forecast.io API

RolandRosenfeld:
The weather binding uses unencrypted http:// for the forecast.io API.
This stopped working two days ago.
Changing the URL from http:// to https:// solves the issue.
See also https://community.openhab.org/t/problem-with-weather-binding-and-forecast-io/3405


teichsta:
Thanks @RolandRosenfeld!

PR #3314, opened 2015-10-18
Add param 250 to Aeon Labs minimote

khendricks:
This param allows putting the remote into "Scene" mode which is necessary to use it properly with openHAB.


cdjackson:
Are you 100% sure this parameter exists - I don't see it in any manuals or databases? The problem is if we add it, and it's not there, then it will cause the device to go DEAD so I just want to be sure...

Thanks.
Chris

khendricks:
Yup!  After digging around I found [this](https://aeotec.freshdesk.com/support/solutions/articles/6000035629-in-depth-minimote-guide) article (they talk about param 250).

I applied it to my remote and started to finally see "SCENE_ACTIVATION" commands coming through when I pressed buttons.

cdjackson:
Perfect - thanks.

PR #3315, opened 2015-10-18
Added Leviton VRMX1-1LZ

davealden:


bundles/binding/org.openhab.binding.zwave/database/leviton/vrmx1-1lz.xml
```diff
@@ -0,0 +1,40 @@
...
+		<Class><id>0x73</id></Class>
+		<Class><id>0x77</id></Class>
+		<Class><id>0x86</id></Class>
+	</CommandClasses>
+	
+	<Configuration>
+
+	</Configuration>
+        <Associations>
+                <Group>
```
cdjackson: Are any of these association groups used to provide status information back to the controller? If so, you should add the SetToController tag.

davealden:
Hmm... I'm guessing I shouldn't have put the association groups in the config file (I had originally based this on the fact that the open-zwave project considers it to be similar to the vri10-1lz - not sure why I had 3 groups in there when the vri10-1lz only has one).  I see in the installation manual for the vri10-1lz it mentions an association group.  The vrmx1-1lz manual doesn't mention association groups at all - so I'm guessing I should take it out?

cdjackson:
I'm not completely sure I follow - if you mean that this device doesn't support 3 association groups, then yes, we should make sure the database is correct. Otherwise it can cause timeouts if we request groups etc that aren't really there.  If there is 1 group, and it is used to provide updates to the controller, then you should add the `<SetToController>` tag to the group.

davealden:
Sorry for not communicating well, I hadn't had enough caffeine before I 
wrote that.  :)

What I meant to say is that I don't know if it has association groups or 
not.  I've read everything I can find (the "installation guide" - if you 
can call it that, the database entry for the VRMX1 on 
http://products.z-wavealliance.org and various google searches). No 
where can I find it mentioning association groups (only scenes).

So I'm guessing I should just remove them?

cdjackson:
If you have the device already, then look in the /etc/zwave folder for an XML file for this device - it should say if the device supports associations (although it won’t say how many).

davealden:
Ahh.  It does support associations:

       <commandClass>ASSOCIATION</commandClass>
       <associationCommandClass>
         <version>1</version>
         <instances>1</instances>
         <configAssociations>
           <entry>
             <int>1</int>
             <associationGroup>
               <Index>1</Index>
               <Nodes>
                 <int>1</int>
               </Nodes>
             </associationGroup>
           </entry>
         </configAssociations>
         <maxGroups>1</maxGroups>
       </associationCommandClass>

So I should change the Associations section to just have 1 group and add 
the SetToController tag?

Any suggestions on what I should do with the Maximum tag?

cdjackson:
Sounds good - I would use 5 for the max number - it’s a common number, and probably more than you will use :)

davealden:
Ok - I believe I've updated the pull request correctly (I'm new to this).  Let me know if I broke something.  :)

cdjackson:
Thanks.

PR #3316, opened 2015-10-19
InsteonPLM binding bug fix: no wait for reply when sending X10 messages

berndpfrommer:
InsteonPLM binding bug fix: no wait for reply when sending X10 messages. This fixes the 10sec delay when sending X10 commands.

Here is a compiled jar file with the fix. Should be drop-in for OH 1.7 and 1.8:

https://drive.google.com/open?id=0B8OtzwATWQSWN0RTdG1Wb0Q5VUk


teichsta:
Thanks @berndpfrommer! Please give me sign when you think this one is good to merge …

berndpfrommer:
@teichsta , a user with X10 devices just reported they work, so this one can be merged!
Thanks for serving as our gatekeeper!

teichsta:
Thanks @berndpfrommer!

PR #3259, opened 2015-10-06
Fix incorrect pairing message for hue binding

dominicdesu:
Probably caused by a recent hue API change, the hue binding incorrectly shows the message that pairing would be necessary on each startup (as discussed [here](https://community.openhab.org/t/openhab-continiously-loosing-connection-to-hue-bridge/3013))

Instead of just parsing the json result for "error", this pull requests checks if the JSON response is the one defined by the API for authorization error.

PR #3323, opened 2015-10-19
Add logging output when authentication with hue bridge

dominicdesu:
Added logging output when openHAB tries to authenticate with the hue bridge. Should make it easier to analyse some recently reported issues (e.g. comments in pull request #3259)


teichsta:
Thanks @dominicdesu!

PR #3325, opened 2015-10-21
Node Naming Command Class Support

paixaop:
Add support for node naming command class so name and location can be written and read directly from nodes.

If a node does not support Node Naming Command Class then revert to previous behavior of storing name and location in the binding.


bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveNodeNamingCommandClass.java
```diff
@@ -0,0 +1,360 @@
...
+	 */
+	protected String getString(SerialMessage serialMessage, int offset) { 
+		int charPresentation = serialMessage.getMessagePayloadByte(offset + 1);
+		
+		// First 5 bits are reserved so 0 them
+		charPresentation = 0x07 & charPresentation;
+		
+		switch (charPresentation) {
+			case ENCODING_ASCII:
+				logger.info("NODE {} : Node Name is encoded with standard ASCII codes", this.getNode().getNodeId());
```
cdjackson: I think this should be a debug message - most users won't care about seeing this in their log.  Same for the next 3 instances.

cdjackson:
Sorry - I missed this request...

Just one small comment about logging level. Also, can you please run the formatter over the new file to get a consistent format.

Thanks
Chris

paixaop:
Chris

changed the debug messages to debug. Nice catch Thanks!

cdjackson:
Thanks.

PR #3328, opened 2015-10-21
Added job key to JSR223 timer trigger so reschedule works correctly.

steve-bate:
The JSR223 Quartz-based timer reschedule function was not setting the job key for the new trigger. This caused the Quartz reschedule to fail and the original timer timed out at the originally schedule time. Furthermore, no future reschedules worked so that after the original timeout there were no more callbacks.

From the Quartz docs...

 > Remove (delete) the Trigger with the given key, and store the new given one - which 
> must be associated with the same job (_the new trigger must have the job name & group specified_).

I observed the behavior while debugging a motion-triggered light rule that was not working correctly. With this change, the reschedule appears to work correctly.

### DIFF
--- a/bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/actions/Timer.java
+++ b/bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/actions/Timer.java
@@ -71,7 +71,10 @@ public class Timer {
 
 	public boolean reschedule(AbstractInstant newTime) {
 		try {
-			Trigger trigger = newTrigger().startAt(newTime.toDate()).build();
+			Trigger trigger = newTrigger()
+				.forJob(jobKey) // default group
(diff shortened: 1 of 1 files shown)


lewie:
Hi @steve-bate, 
why did you retract this issue?

Rgds
Helmut

steve-bate:
After more investigation, I believe that the real issue is that Quartz does not reschedule expired timers. I've run some tests to confirm this theory. When a timer expires, it's removed from the scheduler. The later reschedule doesn't find the original trigger in the scheduler and the rescheduling fails. Quartz returns a null from the reschedule function call but openHAB doesn't check it. The openHAB code assumes the reschedule succeeded and returns a success status to the caller.
 
I think the solution is to process the Quartz reschedule status and return the correct status code from openHAB. We should also update the documentation to warn people about attempting to reschedule expired timers.
 
I'm going create a new PR to do the reschedule status checks. Should I also make the change for the non-JSR223 timer implementation (which has the same issue)?

PR #3329, opened 2015-10-21
Updated CommandEventTrigger to support both wildcarded and specific c…

steve-bate:
…ommand values.

Updated RuleTriggerManager to pass command value to CommEventTrigger evaluations.
Updated RuleTriggerManager to not stop on first rule trigger for Change/Update/Command rules.

The UPDATE/CHANGE/COMMAND switch cases were previously combined and a null command was being passed to the CommandEventTrigger evaluator. I updated the code to pass the command value for the COMMAND trigger type. I also changed the CommandEventTrigger to match on a command value or to match any command on the item if the trigger command is null.

The purpose of the previous breaks in the rule evaluation code were not clear to me and I removed them. Wouldn't we want all triggered rules to execute based on a given trigger?

### DIFF
--- a/bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/engine/RuleTriggerManager.java
+++ b/bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/engine/RuleTriggerManager.java
@@ -183,13 +183,24 @@ public class RuleTriggerManager {
 			return timerEventTriggeredRules;
 		case UPDATE:
 		case CHANGE:
-		case COMMAND:
 			if (newType instanceof State) {
 				for (Rule rule : rules) {
 					for (EventTrigger t : rule.getEventTrigger()) {
 						if (t.evaluate(item, (State) oldType, (State) newType, null, triggerType)) {
 							result.add(rule);
 							break; // break from eventTrigger iteration
+						}
+					}
+				}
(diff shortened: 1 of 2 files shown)


steve-bate:
@smerschjohann, do you have any feedback about motivation for the rule evaluation loop break? Do you know of any issues with my modification? Thanks.

smerschjohann:
oh, did not see this pull request before.

So the change concerning the COMMAND trigger: I agree on this change, that was obviously a bug.

The removed break(s): If you remove the breaks from the inner loop, you will possibly add a rule multiple times to the later executed ruleSet, which is wrong.
A trigger is a trigger for a rule, one positive evaluated trigger is enough for the rule to be triggered, which is what we want, otherwise a rule is triggered multiple times on the same data.

So this break is mandatory and comparable to an OR-statement in programming languages (test(a) || test2(b)): if test(a) evaluates to true, test2(b) is not executed anymore as it is not necessary for the statement.

steve-bate:
@smerschjohann, thanks for reviewing the change. What you said makes sense. I've restored the breaks and updated the PR.

steve-bate:
Any other issues with the PR?

smerschjohann:
doesn't look that way, I think it's ready to merge

teichsta:
great, thanks @steve-bate and @smerschjohann for contributing and reviewing! Best, Thomas E.-E.

PR #3330, opened 2015-10-22
Add serial transport for benqprojector binding

plajjan:
This adds a serial transport for the benqprojector binding. Verified working with my Benq W1070 projector!

It is largely copied from the network transport but uses NRSerial to communicate directly over a serial port. The NRSerial library is already included in OpenHAB and used by other modules so I figured it was a no-brainer to rely on it over other modules (I actually thought of jssc before realising that something was included).

This is my first contribution to OpenHAB so apologies if I've missed to follow some contribution guidelines or similar.


teichsta:
PR lgtm, thanks @plajjan! I am looking to many further contributions :-)

PR #3332, opened 2015-10-22
InsteonPLM: fixed bug with handling of thermostat temperature status messages

berndpfrommer:
This pull request fixes a bug that causes outliers in thermostat temperature data due to incorrect parsing of status messages (0x6E).

I have tested the fix with my thermostat and the temperature status messages are now parsed correctly.


teichsta:
Thanks @berndpfrommer!

PR #3333, opened 2015-10-22
[z-wave] Database update (Qubino Flash 2 relays)

druciak:
Minor DB fix for Flash 2 relays.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhbd.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhbd.xml
@@ -54,7 +54,7 @@
 			<Index>10</Index>
 			<Type>list</Type>
 			<Default>255</Default>
-			<Size>1</Size>
+			<Size>2</Size>
 			<Label lang="en">Activate / deactivate functions ALL ON/ALL OFF</Label>			
 			<Item>
 				<Value>255</Value>


cdjackson:
Are you sure this is correct? It doesn't really look right, and the manual says it's only 1 byte.

druciak:
Well, I am not. ;-) I don't have this version of the module, friend of mine has.
Anyway originaly it was 1 byte (in the one I have it is 1 byte), but it seems it has changed in the newest version. At least after this correction my friend can change this setting, before it was impossible.

cdjackson:
So what you mean is that this will mean your friends works, but everyone else breaks :)

Ok, we need to think this through otherwise it won’t work for everyone else with this device. I find it strange that they would change a setting like this unless it has completely changed function as it only needs one byte.

Can your friend post a log that shows the initialisation of the device where the configuration is updated, versions are downloaded etc...

druciak:
But this is a new device and new db file. How would the change break anything if it didn't exists before? ;-)

cdjackson:
Ummm, I'm confused. It's not a new device or new database entry. You've changed 1 number in the existing file. I've had this device here for about 6 months or so, so it's certainly not new either.

druciak:
This is correction to file I added few days ago. I bet you have *zmnhba* (like I have) not *zmnhbd*, please double check.

druciak:
FYI: this is Z-Wave Plus version, but the version 1.0 states the param has 1 byte. However in version 1.1 it is already 2 bytes.

cdjackson:
Ok - you’re right…  Sorry.

I’ll merge.  If we get into problems with this parameter changing size then we might need to update the database with the version...

druciak:
From the version 1.1:
> Parameter no. 10 - Activate / deactivate functions ALL ON/ALL OFF
> Available configuration parameters (data type is 2 Byte DEC):
For some reason I cannot attach this doc here. It says the file must be PDF, but it is a PDF...

cdjackson:
you can email if you like (chris -at- cd-jackson.com).  Might be useful :)

Thanks.

druciak:
Sent. BTW: the db file needs corrections, I was preparing it basing on v1.0, so for example associations are wrong. Also there are many command classes I didn't add. Anyway the module works. ;-)

cdjackson:
Ok - no probs. I’ll expect another PR at some stage :)

Issue #3336, opened 2015-10-23
Withings binding / Item parsing error


Alexaas:
I get an error for the an item:

Binding configuration of type 'withings' of item 'weight' could not be parsed correctly.
java.lang.ArrayIndexOutOfBoundsException: 1

Item looks like that:
Number Weight "Gewicht [%.1f kg]" { withings = "weight" }

For me this looks like a bug  in method processBindingConfiguration of WithingsGenericBindingProvider:
**if (configElements.length == 1) {*
  **measureType = MeasureType.valueOf(configElements[1].toUpperCase());}**

Shouldn't it be a 0 in the bracket: configElements[0]??

Regards,
Alex
##

watou:
Right you are.  Added PR #3337 to index the single string.  I built a JAR with the change [here](https://dl.dropboxusercontent.com/u/4286376/withings-binding-error/org.openhab.binding.withings-1.8.0-SNAPSHOT.jar).  Please try it and report back here.  If it works, I will let @teichsta know it should be OK to merge.

Alexaas:
Hi, this is working now. However the binding does not store the authentication. Also not in 1.7.1. Any idea about that?

watou:
I don't know, @teichsta might, I think the [code here](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.withings/src/main/java/org/openhab/binding/withings/internal/api/WithingsAccount.java) might explain.  Is your authentication stored in openhab.cfg and is that file writeable by the user under which the openHAB runtime is running?

Alexaas:
The Userid is written into openhab.cfg. But nevertheless everytime I restart openhab it asks me again for authentication.

Alexaas:
it looks like the authentication is stored correctly in openhab.cfg. But how is it read if it is already there? I do not understand the code completely but for me it looks like the Authentication is just requested every time....

watou:
I see that @teichsta improved the code this past May, and has been improving the [wiki page](https://github.com/openhab/openhab/wiki/Withings-Binding/_compare/cffbfc5a6ce6fa91629cf117bd4e5cb9e3f5ab3b...4382f6d96a35f7588e7fb0969d1a5f65a5ad2c04) as recently as four days ago.  

[...]

PR #3337, opened 2015-10-23
Fixed bad array index in withings binding config.

watou:
Fixed bad array index as report in #3336.


watou:
Reported ok to merge [here](https://github.com/openhab/openhab/issues/3336#issuecomment-150668884).

teichsta:
Thanks @watou!

PR #3342, opened 2015-10-23
Tinkerforge binding: add I2CMode setting to the BrickletTemperature configuration

theoweiss:
This change was requested on the TinkerForge forum:
http://www.tinkerunity.org/forum/index.php?topic=3349.new
I'm awaiting feedback from there.


theoweiss:
Got feedback from the TF forum: it's playing well.

Issue #3338, opened 2015-10-23
Bugs in sendMail action


Dan1001:
There are at least two bugs in the sendMail action

First, image attachments arrive as a file called "Attachment" without a suffix/extension, and hence can't be opened on many systems.

Second, urls with basic authentication aren't accepted (i.e. username:password@), and give the error log I've copied below. Obviously a problem when using sendMail with webcam images.

(Opening up a new issue because the old one was closed, and I'm not sure it's monitored; apologies if that's wrong and I'm just duplicating posts)

thank you!

Dan

18:02:29.826 ERROR o.o.action.mail.internal.Mail[:97]- Error adding attachment to email.
org.apache.commons.mail.EmailException: Invalid URL set:http://admin:MYPASSWORD@192.168.1.145:80/dms.jpg
at org.apache.commons.mail.MultiPartEmail.attach(MultiPartEmail.java:367)
at org.apache.commons.mail.MultiPartEmail.attach(MultiPartEmail.java:311)
at org.openhab.action.mail.internal.Mail.sendMail(Mail.java:93)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
Caused by: java.io.IOException: Server returned HTTP response code: 401 for URL: http://admin:MYPASSWORD@192.168.1.145:80/dms.jpg
at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1627)
at java.net.URL.openStream(URL.java:1037)
at org.apache.commons.mail.MultiPartEmail.attach(MultiPartEmail.java:362)
at org.apache.commons.mail.MultiPartEmail.attach(MultiPartEmail.java:311)
at org.openhab.action.mail.internal.Mail.sendMail(Mail.java:93)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

watou:
[...]

PR #3348, opened 2015-10-24
Add filename to email attachment

9037568:
Fix the mail sending action so that the correct filename of the attachment is used.

Fixes the attachment name issue discussed in #3338


watou:
I was making my pull request #3349 without realizing you had already submitted one.  I submitted mine because it should also allow multiple attachments and does file name parsing differently.

teichsta:
close this one in favour of #3349 (which seem to be more comprehensive. Nevertheless thanks for your contribution @9037568!

PR #3359, opened 2015-10-27
Added error checking for JSR223 timer reschedule.

steve-bate:
The JSR223 TImer management code was not checking the Quartz reschedule return value when returning openHAB status. This change adds the check. 

Quartz will not reschedule a trigger that has expired. Previously, the failed Quartz reschedule would not result in a failed status from openHAB.

Note, there is a similar issue with the Timer management code used by Xtext rules.


teichsta:
Thanks @steve-bate! Any chance to suggest a solution for the Timer issue?

steve-bate:
Thanks for merging the PR. I appreciate the significant effort you put into this project!

Do you mean the issue with the Timer management code used by Xtext rules? If so, it's exactly the same as it was with the JSR223 TImer reschedule status non-handling. It appears the JSR223 Timer code was partially copy-and-pasted from the Xtext-related code. I'd be happy to submit a PR for the same fix to the Xtext Timer reschedule if you'd like but I'd need some testing assistance since I've only used JSR223.

PR #3361, opened 2015-10-27
Z-wave Qubino Flush Dimmer Plus ZMNHDD

evansj:
Just a Z-wave database update.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -2392,6 +2392,15 @@
 		</Product>
 		<Product>
 			<Reference>
+				<Type>1</Type>
+				<Id>51</Id>
+			</Reference>
+			<Model>ZMNHDD</Model>
+			<Label lang="en">Flush Dimmer Plus</Label>
+			<ConfigFile>qubino/zmnhdd.xml</ConfigFile>
+		</Product>
+		<Product>
+			<Reference>
 				<Type>5</Type>
 				<Id>1</Id>
 			</Reference>
--- a/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhdd.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhdd.xml
@@ -1,0 +1,465 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>ZMNHDD</Model>
+	<Label lang="en">Flush Dimmer</Label>
+	<CommandClasses>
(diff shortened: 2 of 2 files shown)


evansj:
Since creating the pull request I've cleaned up my repo a bit so the PR only contains one commit now.

druciak:
Hi @evansj, do you have this device? I have one, but it the doc there are more configuration parameters. I would correct the db, but I don't want to break anything. For example in my doc (from 08.10.2015) there are 2 parameters for enabling I2 and I3, 100 for I2 and 101 for I3. 
Could you please doble check?

evansj:
I have three, although I've only installed one so far. The docs I used only have parameter 100. http://www.vesternet.com/downloads/dl/file/id/637/product/1559/z_wave_qubino_flush_dimmer_plus_manual.pdf
Do you have a link to your version of the doc? I'm not at home at the moment so I won't be able to check my devices (and the paper version of the manual) until tomorrow.

druciak:
Here's the link from their website: http://qubino.com/download/1293/, but I had to register first.
This is even newer doc than I received with my device.

druciak:
This is link to downloasd page: http://qubino.com/downloads-pdf/
I am hoping the difference is in the docs only, not in the firmware.

druciak:
Added #3372. Tested on my Qubino Dimmer, works well.

evansj:
The paper manual that came with my dimmers finishes with the line:

**NOTE** User manual is valid for module with SW version S1 (SW version is part of P/N)! Example: P/N ZMNHDDx H1**S1**P1

This matches the part number of the devices I have. The manual is dated 08.10.2015.

The PDF downloaded from Qubino ends with exactly the same statement, but the filename implies it's for version 1.1. It has a date of 13.10.2015.

The one on Vesternet's site also ends with exactly the same sentence, but the date is different again! :) This time it's 12.05.2015.

I'm compiling the latest version of the code with @druciak changes, I'm sure it will work fine.

Incidentally the one I've already installed is reporting:

Device ID: 51
Device Type: 1
Version: 4
Protocol Version: 4.5
Application Version: 1.1

I'm sure the other two will be the same. I'll be installing them this afternoon.

druciak:
Yeah, it's a mess with their docs. I have exactly the same values in my device.
Please let us know if my change works well with all your devices.

evansj:
Your changes are working perfectly on all three of my dimmers. Do the device version numbers etc. that I posted above match yours?

druciak:
Yep, they are the same.

evansj:
Sorry, just read your comment again and realised you already answered the question about the version numbers. :)

PR #3322, opened 2015-10-19
InfluxDB persistence for influxdb versions >= 0.9

theoweiss:
Influxdb database versions 0.9 and newer have a completely changed api and are not backward compatible.
This is a new influxdb persistence service plug-in which supports influxdb databases >=0.9. It is based on the existing influxdb persistence service that only supports version <0.9, but has some fundamental changes.
I decided to put this functionality in it's own plug-in in order to not force our users to upgrade their influxdb databases when upgrading to openHAB 1.8. Even more because influxdb until now has no upgrade path from influxdb 0.8 to 0.9, which means people would have to drop their databases.

Regards,
Theo

PR #3362, opened 2015-10-28
Fix implementation of DimmerItem#getStateAs

abrenk:
DimmerItem.getStateAs(PercentType) always returned PercentType.ZERO as
returned by its parent class SwitchItem.

### DIFF
--- a/bundles/core/org.openhab.core.library.test/src/test/java/org/openhab/core/library/items/DimmerItemTest.java
+++ b/bundles/core/org.openhab.core.library.test/src/test/java/org/openhab/core/library/items/DimmerItemTest.java
@@ -1,0 +1,73 @@
+package org.openhab.core.library.items;
+
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.library.types.PercentType;
+import org.openhab.core.types.State;
+
+/**
+ * @author Andreas Brenk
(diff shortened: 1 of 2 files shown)


teichsta:
Thanks @abrenk! Could you please a Test for DimmerItem (see org.openhab.core.library.test bundle)

abrenk:
Now with `DimmerItemTest`.

theoweiss:
I've tested if this PR will fix the Dimmer issue reported with #3322: yes it will ;-)

Thanks @abrenk

teichsta:
Thanks also from my side @abrenk!

kaikreuzer:
BEWARE: This is a change in the core runtime which MUST NOT be merged unless the change is synced to ESH.

kaikreuzer:
@teichsta, could you please follow up on this?

teichsta:
Oh yes, you're right @kaikreuzer. I am sorry for that … seem i was in merge mood this day. I've created (and merged) #3379 which reverts this PR and back ported the ESH enhancements of DimmerItem to OH1.

kaikreuzer:
Ok, thanks, @teichsta. So the problem was actually already addressed in ESH? I didn't look at the details yet myself...

teichsta:
codewise i would say yes. I hope @theoweiss or @abrenk will confirm …

abrenk:
I can take a look at it. But I'd argue that the unit test I wrote for this pull request was more complete and clear than the one @teichsta "merged" (i.e. replaced mine with) from ESH.

By the way: the diffs of e.g. 22aa812 are hard to read because a different code formatter is used: openHAB uses tabs to indent, ESH seems to use four spaces. Would be nice to reformat the files when copying code between the two.

abrenk:
Okay, the backport also fixes the problem. You can find my test in 1cae516.

teichsta:
great, thanks for the update!

PR #3363, opened 2015-10-29
Added try/catch for rule exceptions in the JSR223 ScriptWatcher.

steve-bate:
Otherwise, the script watcher thread dies. This causes rules to no no longer be reloaded automatically. Currently, the workaround is to restart the bundle.

### DIFF
--- a/bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/engine/scriptmanager/ScriptUpdateWatcher.java
+++ b/bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/engine/scriptmanager/ScriptUpdateWatcher.java
@@ -101,7 +101,12 @@ public class ScriptUpdateWatcher implements Runnable {
 					}
 				}
 
-				scriptManager.scriptsChanged(addedScripts, removedScripts, modifiedScripts);
+				try {
+					scriptManager.scriptsChanged(addedScripts, removedScripts, modifiedScripts);
+				}
+				catch (Exception ex) {
+					logger.error("Error during script change processing", ex);
+				}
 
 				boolean valid = key.reset();
 				if (!valid) {


teichsta:
Thanks @steve-bate! Please find my review comment inline …

bundles/core/org.openhab.core.jsr223/src/main/java/org/openhab/core/jsr223/internal/engine/scriptmanager/ScriptUpdateWatcher.java
```diff
@@ -101,7 +101,12 @@ public void run() {
 					}
 				}
 
-				scriptManager.scriptsChanged(addedScripts, removedScripts, modifiedScripts);
+				try {
+					scriptManager.scriptsChanged(addedScripts, removedScripts, modifiedScripts);
+				}
+				catch (Exception ex) {
+					logger.error("Error during script change processing", ex);
```
teichsta: decrease to WARN since it doesn't have impact on the platform stability

steve-bate: I agree that it doesn't impact platform stability given this fix for the thread exit, but it does indicate an actual (vs. potential) error in one of the JSR223 Rule instances. Also, the current log level is consistent with the log level for an exception thrown in the execute method of a Rule (RuleExecutionRunnable.java:38). That said, please confirm that you want the message at WARN level and I'll make the change. 

For future reference, is there any developer documentation to guide logger level usage?

teichsta:
> Also, the current log level is consistent with the log level for an exception thrown in the execute method of a Rule (RuleExecutionRunnable.java:38). That said, please confirm that you want the message at WARN level and I'll make the change.

ok, consistence is a good argument :-)

> For future reference, is there any developer documentation to guide logger level usage?

no, but it would be great to have added to https://github.com/openhab/openhab/wiki/How-To-Contribute. Would you mind creating a proposal and which we then discuss at https://community.openhab.org. Could  also find it's way to http://www.eclipse.org/smarthome/documentation/community/contributing.html after all.

teichsta:
Thanks @steve-bate!

PR #3376, opened 2015-11-01
Try to avoid null items being added to contextMap items set.

watou:
This change is intended to stop what appears to be the adding of a `null` `Item` to the `contextMap` in AbstractGenericBindingProvider.  This appears to only happen on Raspberry Pi A/B/B+ and possibly other slow machines due to a race condition that shows itself there and not the much faster Raspberry Pi 2 or other faster machines.

See [this comment](https://github.com/openhab/openhab/pull/3016#issuecomment-152816629) to PR #3016 based on logs from @lolodomo.

Note that throwing the `BindingConfigParseException` if `null` was passed for `item` will stop the `null` from ending up in the set of items, but it will essentially move the issue closer to its cause.  The other possible PR was to carefully bypass any `null` `Item` in the `contextMap` sets to avoid the NPE of the longstanding bug, but that approach would keep us from ever understanding the original cause of the apparent race condition.

If you are OK with this approach, @teichsta, then hopefully @lolodomo can try a new build after the merge, and see if it helps us identify the problem at its source.

### DIFF
--- a/bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
+++ b/bundles/model/org.openhab.model.item/src/org/openhab/model/item/binding/AbstractGenericBindingProvider.java
@@ -72,6 +72,10 @@ public abstract class AbstractGenericBindingProvider implements BindingConfigRea
 	 * {@inheritDoc}
 	 */
 	public void processBindingConfiguration(String context, Item item, String bindingConfig) throws BindingConfigParseException {
+		if( item == null ) {
+			throw new BindingConfigParseException("item is not permitted to be null");
+		}
+
 		if( context == null ) {
 			throw new BindingConfigParseException("context is not permitted to be null for item "+item);
 		}


lolodomo:
Can we have a merge for this PR please ?
If not, @watou can you provide the updated jar file for testing ?

watou:
Despite all my fiddling with openHAB, I don't know how to take a *core* JAR and patch it into an already installed openHAB runtime.  Perhaps it's as easy as replacing the JAR, but my one attempt at that failed.  My hope was/is that this PR is merged, and the next nightly build (or full build you do from `master`) is installed and rigorously tested on the slow machine where the stacktraces are being generated.  The reason that immediate testing is needed is that this PR should move the source of the `null` further up the foodchain, exposing (hopefully) the root cause so we can properly resolve the problem.  I want @teichsta's agreement that we can proceed this way, or perhaps we can just figure out how to test this change without the likely ripple effect.

cdjackson:
It’s easy - just remove it from the system runtime folder, and put the new version into the addons folder…

watou:
Thank you, Chris!  I will supply the JAR to @lolodomo so he can proceed that way.

watou:
@lolodomo [Here](https://dl.dropboxusercontent.com/u/4286376/contextMap-no-null-items/org.openhab.model.item-1.8.0-SNAPSHOT.jar) is the JAR to test on the system that's generating the stacktraces in question.  Let us know what you see!

lolodomo:
I first installed the last 1.8 build. As expected, the same error is still produced.
Then I tried to replace your jar file in my folder server/plugins but unfortunately after that openHAB is not starting properly anymore. No more logs after loading my sitemaps.
Can we really handle that like that without a full distribution jar file ?

watou:
I think Chris suggested removing the equivalent JAR from its current location, and putting the new JAR  in your addons folder.

cdjackson:
Yes - this has always worked for me in the past…

lolodomo:
Ok it works, thank you.
I let you know the result in 5 minutes.

lolodomo:
Ok, I see nothing interesting in my logs but I can confirm that the NPE exception has now disappeared with this fix and item reloading is done.
So I will encourage validating this PR.
@watou: bravo

watou:
Thank you for testing, @lolodomo.  Assuming your testing went through the same process as before, this would suggest that the `BindingConfigParseException` is being thrown when a `null` `Item` is being passed to `processBindingConfiguration` but nothing is being logged about it in the call stack above it.  I would like it if you could continue exercising openHAB on this same system with this replacement JAR for a few days, and if, after then, you still haven't seen any negative side effects, I will wholeheartedly agree that this PR can be merged.  Thanks again.

lolodomo:
No problem, my production environment is now running with your JAR file.

watou:
Excellent; thanks.  After a few days and hopefully a few cycles that would normally load items files, please remind me to merge if all still looks good.  Maybe this issue is finally put to bed?  (Famous last words.)

lolodomo:
Looking at the code, I see that processBindingConfiguration is called inside internalDispatchBindings.
internalDispatchBindings could be called with a null item only in methods dispatchBindingsPerItemType and dispatchBindingsPerType.
In the two cases, item is build by calling createItemFromModelItem.
Maybe you could add a test if resulting item is null before calling internalDispatchBindings in these 2 methods ?
You could even add warning information if null to uderstand what lead to this case ?

watou:
I see that [`processBindingConfigsFromModel`](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/internal/GenericItemProvider.java#L170) already had a `null` check before calling [`internalDispatchBindings`](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/internal/GenericItemProvider.java#L323), but [`dispatchBindingsPerItemType`](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/internal/GenericItemProvider.java#L286) and [`dispatchBindingsPerType`](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/internal/GenericItemProvider.java#L309) did not, even though all three calls immediately get their `Item` from [`createItemFromModelItem`](https://github.com/openhab/openhab/blob/master/bundles/model/org.openhab.model.item/src/org/openhab/model/item/internal/GenericItemProvider.java#L177).  So I am adding  `null` checks around the other two calls to `internalDispatchBindings` to this PR.  Thanks for the suggestion to dig deeper!

watou:
@lolodomo : [new JAR to test](https://dl.dropboxusercontent.com/u/4286376/contextMap-no-null-items/org.openhab.model.item-1.8.0-SNAPSHOT.jar) with the two `null` checks added, please!

watou:
@lolodomo Please hold off -- this created a problem in my testing.  I will have to dig deeper!

watou:
Unfortunately this PR as it currently stands creates a new kind of startup timing problem, so it shouldn't be merged until it's understood and resolved.  Sorry!

watou:
@lolodomo, please revert the test JAR and open an issue describing your remaining issue and stacktrace, and I will try a new PR to more "gently" address it without side effects.

Issue #3317, opened 2015-10-19
MiOS Binding - Tracking last used pin on locks


silencery:
I'd like to see if there is a method to obtain the last used pin on a door lock (Specifically, in my case, i'm using a Schlage BE469)?

Vera reports this as sl_UserCode service ID. For instance:
`<state id="80" service="urn:micasaverde-com:serviceId:DoorLock1" variable="sl_UserCode" value="UserID="1" UserName="John""/>`

Using the item file generator out of the box, the sl_UserCode is set as a number by default, which is the wrong datatype in my case. Updating the item to string is the first step. However, after taking this first step, some transformation still needs to be done since the default output is not really usable. With the preceding example, the item value would look like this:

`UserID="1" UserName="John"`

Any suggestions on capturing either just the UserID or Username?

mrguessed:
@silencery,
The bug part of this is that the MiOS Item Generator is using _Number_ instead of _String_... when it's a _String_ value in the MiOS source material (an encoded one)

The "what to do" part of this should be opened as a discussion on http://community.openhab.org, since a straight "it's a String" will only fix the immediate issue.

Likely it'll need a REGEX to pull out the `UserID` and `UserName` parts, and it seems like the Item Generator should do that, but let's use the Community forum to have that part of the discussion.


_Original MiOS Discussion_: http://forum.micasaverde.com/index.php/topic,34125.0.html

mrguessed:
@teichsta,
Can you assign this to me?  I won't be able to do it for a week or so, but I don't have permissions to self-assign issues from the openHAB repo.

teichsta:
sure :-) thanks for taking over!

mrguessed:
@teichsta, 
This bug can be closed, the base merge has the necessary fixes.

mrguessed:
This issue was fixed by the base PR #3377

PR #3377, opened 2015-11-02
MiOS Binding: Issue #3317 Tracking last used pin on locks.

mrguessed:
Bugfix for Issue #3317 Tracking last used pin on locks.

MiOS Item Generator (XSLT) incorrectly generated the `sl_UserCode` UPnP attribute as `Number` type instead of `String` type.

Fix involves:
a) Emitting original value as `String` type, which is of the form:

        "UserID=".+" UserName=".*"

b) Emitting new Item declaration `*UserCode_userid` representing the _UserID_ value encoded in the original value (`Number`).

c) Emitting new Item declaration `*UserCode_username` representing the _UserName_ value encoded in the original value (`String`).

d) When transforming multiply-bound source values, avoid destroying the original/untransformed value.

e) Tighten the parse logic to work better with Transformations (like `REGEX`) that have complex parameters.

Note: Online docs will be updated once the PR is in the codebase.  The fix isn't isolated to the MiOS Item Generator code, but it helped me find an issue found in the MiOS Binding, which needed to be fixed for it all to work together cleanly


teichsta:
Thanks @mrguessed!

mrguessed:
Thanks @teichsta, appreciate the merge!

PR #3381, opened 2015-11-03
MiOS Binding: Improve defaults for MiOS Item Generator tool.

mrguessed:
The MiOS Item generator uses an XSLT file, in conjunction with core defaults in the MiOS Binding, to aid users converting from MiOS Systems to openHAB.

This change improves the defaults outlined in the generator so that more cases are converted over (particularly Sonos, RFXCOM Plugins) without need to hand-edit the files after generation.

Modifications are as a result of the work done in:

*  https://community.openhab.org/t/more-dumb-questions-rfx-door-bells-mios-binding-vera-scenes/3874


bundles/binding/org.openhab.binding.mios/examples/scripts/miosTransform.xslt
```diff
@@ -124,13 +125,16 @@ String   <xsl:value-of select="$DeviceNameFixed"/>DeviceStatus "<xsl:value-of se
 <xsl:when test="@service = 'urn:micasaverde-com:serviceId:AlarmPartition2'"      >AlarmPartition2</xsl:when>
 <xsl:when test="@service = 'urn:micasaverde-com:serviceId:Camera1'"              >Camera1</xsl:when>
 <xsl:when test="@service = 'urn:cd-jackson-com:serviceId:SystemMonitor'"         >SystemMonitor</xsl:when>
+<xsl:when test="@service = 'urn:cd-jackson-com:serviceId:Config'"                >CDJConfig</xsl:when>
```
teichsta: are these two cd-jackson-com services intended to be in the xslt? They look like user specific config to me, are they?

watou: When writing plugins for MiOS/Vera, the plugin author has to come up with unique service names.  In this case our own Chris Jackson was the plugin author.  So it's not user-specific after all.

mrguessed: Exactly.  All MiOS Plugins create a UPnP service domain to work in., in order to store their state.

You'll see there categories, stand upnp-org ones, micasaverde-com ones and then custom ones by private developers.

When they're common in use, they get added to the item generator so it makes it easier for people to work with them (and have them auto migrate correctly instead of just 'String' maps... Which is the default

teichsta:
> Exactly. All MiOS Plugins create a UPnP service domain to work in., in order to store their state.

ok, thanks for the explanation! Was just wondering …

PR #3388, opened 2015-11-05
Base64 support and error comparing classes

TheNetStriker:
I’ve added support for Base64 communication because some bytes from serial devices are causing problems in the rest api: (see https://community.openhab.org/t/error-in-rest-service-when-serial-port-sends-stx-ack-eot-bytes/3932/2)

I also fixed an error when no regex is used. (entry.getValue().type.isInstance(SwitchItem.class) was never true)


steve-bate:
@teichsta. I've tested this PR (stepped through it in a debugger while handling a BASE64 configuration) and it works for me as well.

watou:
Minor comment: the use of `Boolean base64` instead of `boolean base64` could lead to the possibility of a `null` value for `base64`, since it's an object, not a native type.  Java is kindly boxing and unboxing `true` and `false` into `Boolean.TRUE` and `Boolean.FALSE` for you, but if you don't need an object version of a boolean value, it might be better to use the native type `boolean` throughout instead of the object class `Boolean`.

teichsta:
PR lgtm, thanks! Could you please squash the commits into one @TheNetStriker?

TheNetStriker:
Sorry, seams I did make a mistake merging the commits and deleted the pull request. I've created a new one: https://github.com/openhab/openhab/pull/3420

PR #3396, opened 2015-11-08
InsteonPlm: Added support for leak sensor heartbeat messages.

steve-bate:
I've tested this over the last several days (only one heartbeat/day). The item status is being updated (based on events.log) and the lastheardfrom timestamps are being updated correctly. This feature is useful for the leak sensor because there is no battery life indicator and it sends very few messages unless there is a status change. Thanks to @berndpfrommer for his help.


berndpfrommer:
Steve, thank you for your contribution!

For lack of a leak detector I cannot test this PR. @steve-bate , please confirm that you have tested the code successfully in the present form, and that it handles heartbeat and status messages correctly.

The code path affected by this PR will only be executed when a leak sensor is present and configured, i.e. risk of breaking other parts of the InsteonPLM binding is very low.

I reviewed the code and (given confirmation from Steve that he has tested) recommend to merge it as is.

teichsta:
it seems as @steve-bate tested the PR already (see his comment). Thanks @berndpfrommer for reviewing and @steve-bate for contributing!

berndpfrommer:
Shame on me for not reading Steve's first two sentences!

PR #3398, opened 2015-11-08
[XMBC] Added Input.ExecuteAction to XBMC binding.

msiedlarek:
Trivial addition that opens many new possibilities (http://kodi.wiki/view/JSON-RPC_API/v6#Input.Action).


teichsta:
Thanks @msiedlarek!

Issue #3201, opened 2015-09-24
TCP binding: An issue with "tcp:postamble" in "openhab.cfg"


lawrizs:
I am using TCP binding to send a command, which has to end with a "new line" symbol. I set tcp:postamble=\r\n or tcp:postamble="\r\n" or tcp:postamble='\r\n' and always end-up with the "explicit preamble" such as 1;0;1;0;2;0\r\n1;0;1;0;2;0\r\n

I would expect "\r\n" to be converted into symbols for a new line. It this an issue, or just my wrong assumption? I expect this is the same for "tcp:preamble".

steve-bate:
This doesn't appear to be specific to the TCP binding. Looking at the code, it appears the org.openhab.config.core.ConfigDispatcher does not preprocess character sequences like "\n" and "\r" (as would happen in a Java properties file).

villaocho:
I confirm the issue.
it does send the information as string.
TCPDUMP confirm the characters.

smee204:
I have also hit this problem. This worked ok on 1.7.0. It works until commit #2683.

steve-bate:
The rationale for the #2683 PR is short on details but it seems like it was a response to issue #2688.

watou:
I added PR #3406 in hopes of fixing this issue.  Could people test [this JAR](https://dl.dropboxusercontent.com/u/4286376/tcp-escape-amble/org.openhab.binding.tcp-1.8.0-SNAPSHOT.jar) to see if the `preamble` and `postamble` config strings are being parsed correctly now?

smee204:
The new jar works for me.

watou:
Thank you for testing it, @smee204.  @lawrizs, @villaocho and @steve-bate, do you concur?

steve-bate:
I haven't had a chance to test it but the diffs look good to me.

PR #3406, opened 2015-11-11
[TCP-Binding] correctly parse postamble and preamble strings

watou:
Use Apache `StringEscapeUtils.unescapeJava` to correctly parse the preamble and postamble config strings.  To address issue #3201.


teichsta:
lgtm, thanks @watou!

PR #3410, opened 2015-11-12
Fixes for PointType; added distanceFrom(otherPoint)

watou:
Addition:
* `distanceFrom(PointType otherPoint)` returns distance in meters from this point to another point.  Useful for geofencing and for other purposes in rules.

Fixes:
* `toString()` method would return nicely formatted string version of `PointType`, but the formatted version was useless from the REST API, and therefore unusable by clients.  Now returns 3 decimal numbers separated by commas without rounding or other formatting.  @clinique, is this OK with you?
* The `String` constructor would take any non-empty string and construct the object regardless of the string contents if it had no commas in the string.  Following the example of the `HSBType` `ComplexType`, the string is now sanity checked in the constructor, and `IllegalArgumentException`s are thrown if the string is not in the valid "lat,long[,alt]" format. 
* The `valueOf` method must be static, since it's expected to be called in [`org.openhab.core.types.TypeParser`](https://github.com/openhab/openhab/blob/master/bundles/core/org.openhab.core/src/main/java/org/openhab/core/types/TypeParser.java#L39) using reflection.  (There is still an issue with actually using `TypeParser`, since it uses a list of Types and the first successful call to `XXXType.valueOf` determines the type returned.  For example, if `PointType` appeared before `StringType` in the list (which it would have to be), then any string that happened to be of an acceptable format to `PointType` would "win", even if a `StringType` was preferred.)

### DIFF
--- a/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
+++ b/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
@@ -13,6 +13,7 @@ import java.util.Formatter;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
+import org.apache.commons.lang.StringUtils;
 import org.openhab.core.types.Command;
 import org.openhab.core.types.ComplexType;
 import org.openhab.core.types.PrimitiveType;
@@ -74,15 +75,18 @@ public class PointType implements ComplexType, Command, State {
 	}
 
 	public PointType(String value) {
-		if (!value.isEmpty()) {
+		if (StringUtils.isNotBlank(value)) {
 			String[] elements = value.split(",");
 			if (elements.length >= 2) {
-				canonicalize(new DecimalType(elements[0]), new DecimalType(
-						elements[1]));
+				canonicalize(new DecimalType(elements[0]), new DecimalType(elements[1]));
 				if (elements.length == 3) {
 					setAltitude(new DecimalType(elements[2]));
 				}
+			} else {
+				throw new IllegalArgumentException(value + " is not a valid PointType syntax");
 			}
+		} else {
+			throw new IllegalArgumentException("Constructor argument must not be blank");
 		}
 	}
 
@@ -113,6 +117,25 @@ public class PointType implements ComplexType, Command, State {
 	}
 
 	/**
+	 * Return the distance in meters from otherPoint, ignoring altitude. This algorithm also
+	 * ignores the oblate spheroid shape of Earth and assumes a perfect sphere, so results
+	 * are inexact.
+	 *
+	 * @param otherPoint
+	 * @return distance in meters
+	 * @see <a href="https://en.wikipedia.org/wiki/Haversine_formula">Haversine formula</a>
+	 */
+	public DecimalType distanceFrom(PointType otherPoint) {
+		double dLat  = Math.toRadians(otherPoint.latitude.doubleValue() - this.latitude.doubleValue());
+		double dLong = Math.toRadians(otherPoint.longitude.doubleValue() - this.longitude.doubleValue());
+		double a = Math.pow(Math.sin(dLat / 2D), 2D) + Math.cos(Math.toRadians(this.latitude.doubleValue()))
+				* Math.cos(Math.toRadians(otherPoint.latitude.doubleValue()))
+				* Math.pow(Math.sin(dLong / 2D), 2D);
+		double c = 2D * Math.atan2(Math.sqrt(a), Math.sqrt(1D - a));
+		return new DecimalType(WGS84_a * c);
+	}
+
+	/**
 	 * <p>
 	 * Formats the value of this type according to a pattern (@see
 	 * {@link Formatter}). One single value of this type can be referenced by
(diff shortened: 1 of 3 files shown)


clinique:
@watou : yes, that is fine to me but are you sure I did not already put distanceFrom in the lib ? Sorry, I have headaches catching up with what I think, what I do, what I push to OH1, ESH... will check but nearly sure it already exists somewhere.
For toString(), very fine with this.

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
```diff
@@ -113,6 +118,24 @@ public DecimalType getGravity() {
 	}
 
 	/**
+	 * Return the distance in meters from otherPoint, ignoring altitude. This algorithm also
```
robnielsen: Can you specify which algorithm that is being used?

watou: It's based on the [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula).

watou:
@clinique I don't see a `distanceFrom` method in the [ESH version](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/PointType.java), but I will add an issue to add it there.  I did see that `valueOf` was fixed there already, and I am taking the `toString()` method from ESH back to OH1 so they are the same (optional output of altitude if it's non-zero).

bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PointType.java
robnielsen: cool, I haven't used that one before. Might be useful to update the javadoc with this info.

watou: Your question prompted me to do so.  Thanks for suggesting!

clinique:
distanceFrom is in LocationItem https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/items/LocationItem.java

watou:
Thank you, @clinique -- I should have looked there as well.  I did not expect to find it in an item class.  Is there a way of thinking about where such a method should live?  My thinking was it should live in `PointType`, but you may have more background knowledge about why having it as a method of the item would make more sense.  Also, the `PointType` class has a `getGravity` method, so it seemed to be more natural to put distanceFrom there as well.  Regards, John

clinique:
Yes, because gravity only depends on the PointType itself

watou:
I would argue that `distanceFrom` also only depends on the `PointType` itself (specifically, this and another instance of `PointType` only).  The implementation also relies on the `WGS84_a` constant (equatorial radius) as does `getGravity`.  So I would ask @kaikreuzer and @teichsta to tell me what is a good way to think about it.  Sorry for duplicating your work, @clinique, but also the process will at least educate me about what the architects' thinking is as well.  Regards, John

teichsta:
> So I would ask @kaikreuzer and @teichsta to tell me what is a good way to think about it.

kaikreuzer:
Yes, I can see @watou point here and would agree that this is probably the better place.
I am not sure if anybody is using this method on the LocationItem in ESH yet - if so, moving the method would break this. But if we decide to move it, then it should be done rather sooner than later.
@clinique would you be ok with this as well or do you have any good arguments why it should stay on the item?

clinique:
@kaikreuzer , @watou , no issue to move it. I'm sure I had a very good reason to put it there when I did it, but as I can't put the finger on it anymore, the reason might not be so good :)

watou:
Thank you, @clinique.  I changed `LocationItem.distanceFrom(PointType)` to `LocationItem.distanceFrom(LocationItem)`, and it delegates to `PointType.distanceFrom(PointType)`.  Also updated the `LocationItemTest` but didn't add `PointTypeTest` as it's redundant.  This is a breaking change for any OH1.7 users who might be using it, but with an easy fix.  Likewise for any ESH users when a matching code change is made there.  This PR will also allow any REST API clients to use `PointType` states against 1.8, since it's `toString()` method now matches the ESH version.

I can set up an ESH development environment and submit a PR there, or I could leave it someone else.  Whatever you think.

clinique:
I've got an ESH dev env - no pb to push the PR

watou:
I think this PR is ready for 1.8, so please push a matching PR in ESH and thank you again, @clinique .

watou:
The ESH bug I opened is [482188](https://bugs.eclipse.org/bugs/show_bug.cgi?id=482188).

teichsta:
@watou once the ESH PR is merged please go ahead and merge this PR as well. Thanks, Thomas E.-E.

watou:
OK, will do.  Thanks.

kaikreuzer:
https://github.com/eclipse/smarthome/pull/520 has been merged. Thanks!

teichsta:
Thanks @watou!

PR #3413, opened 2015-11-13
Netatmo: Don't use deprecated method devicelist 

robnielsen:
The method devicelist is deprecated, use the method getstationsdata instead.


teichsta:
great, thanks!

PR #3420, opened 2015-11-14
Base64 support and error comparing classes

TheNetStriker:
I’ve added support for Base64 communication because some bytes from
serial devices are causing problems in the rest api: (see
https://community.openhab.org/t/error-in-rest-service-when-serial-port-s
ends-stx-ack-eot-bytes/3932/2)

I also fixed an error when no regex is used.
(entry.getValue().type.isInstance(SwitchItem.class) was never true)


csowada:
For me the bugfix looks good. But I can't check Base64 support.

steve-bate:
I tried this and it worked for me.

hakan42:
Seems fine to me. @TheNetStriker , would you perhaps add some comments to the wiki page how the Base64 support in the binding is supposed to be used?

TheNetStriker:
Thanks, I've just added a short description how to use this in the wiki.

PR #3424, opened 2015-11-15
benq: Allow minimum source number 0

plajjan:
The first valid number is actually 0 so we have to be inclusive when
checking the range of valid numbers. The code would previously consider
1 to be first lowest valid number.

It's possible to set the source either referring to its numeric index or using a string and while using a string has worked, the code to map index into a string was broken as it would consider 1 to be the lowest valid number when it in fact should be 0.


teichsta:
great, thanks @plajjan!

PR #3434, opened 2015-11-18
insteonplm: Added On/Off Outlet and related CommandHandler changes

SwissKid:
Added the features necessary for the insteon on/off outlet. Tested, verified working.

### DIFF
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/CommandHandler.java
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/CommandHandler.java
@@ -148,20 +148,35 @@ public abstract class CommandHandler {
 		LightOnOffCommandHandler(DeviceFeature f) { super(f); }
 		@Override
 		public void handleCommand(InsteonPLMBindingConfig conf, Command cmd, InsteonDevice dev) {
-			try {
+			try {				
+				int ext  = getIntParameter("ext", 0);
+				int direc = 0x00;
+				int level = 0x00;
+				Msg m = null;
 				if (cmd == OnOffType.ON) {
-					int level = getMaxLightLevel(conf, 0xff);
-					Msg m = dev.makeStandardMessage((byte) 0x0f, (byte) 0x11, (byte) level,
-								s_getGroup(conf));
-					dev.enqueueMessage(m, m_feature);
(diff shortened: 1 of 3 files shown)


berndpfrommer:
Thanks for contributing a new device to the InsteonPLM binding.
Here are a few comments:

1) Can you set the CRC before calling dev.enqueueMessage()? That way it's thread safe or enqueueMessage() can make a deep copy should it be necessary in later implementations.


					m = dev.makeExtendedMessage((byte) 0x0f, (byte) direc, (byte) level, data);
					dev.enqueueMessage(m, m_feature);
					logger.info("{}: was an extended message for device {}", nm(), dev.getAddress());
					if (ext == 1) m.setCRC();
					else if (ext == 2) m.setCRC2();

2) I suggest having just one logging line for both std and ext messages, with a ternary ? operator in the argument to distinguish std vs ext. Put the logging line right before the call to enqueueMessage(), and  make it read something like "sending...".
3) One can argue validly that camel case is more readable, but since none of the other feature keywords uses camel case, please change topOutlet -> topoutlet, bottomOutlet -> bottomoutlet.
4) Please update the insteonplm wiki to add the new device, and include a configuration example for the top_outlet and bottom_outlet.

Looks good otherwise, thanks again!

Bernd

teichsta:
any update @SwissKid?

SwissKid:
Totally missed the emails in response! I'll try to take a look at finishing those tonight.
1)Total oops on the CRC bit, that'll be easy. 
2) I'll check out the ternary operator tonight, the rest seems easy.
3) Sure, easy enough
4) Sure - just the file in docs/?

berndpfrommer:
4) Actually the wiki pages in docs/ are frozen to OH 1.7. Please update the wiki page (that one is not versioned). Once OH 1.8 is released we'll freeze the wiki page and put it under docs again. Thanks!

SwissKid:
Where is that file? I don't see it on my end... I might just be missing a general github thing.

SwissKid:
Did the other changes - I'll do a squash of all of it once we're happy :smile:

berndpfrommer:
This is the page that you should edit:
https://github.com/openhab/openhab/wiki/Insteon-PLM-Binding

From my side the code is now good to go!

SwissKid:
Awesome. Would you like me to squash, or do you guys want to keep the commits separate?

SwissKid:
Updated, squashed, rebased with current master (for easy merge) and updated the wiki page :smile:

teichsta:
great, thanks guys!

PR #3435, opened 2015-11-18
Added new Sallegra Binding

Satelco:


watou:
Thank you for this contribution.  I have looked at the [wiki page](https://github.com/openhab/openhab/wiki/Sallegra-Binding).  A few initial comments:

- For the sake of consistency with other bindings, I prefer config parameters to be of the form:
```
sallegra:livingroom.password=admin
```
instead of
```
sallegra:livingroom:password=admin
```
(It's one of those small differences that could trip up users.)
- In the code, you use `@note` and `@date` Javadoc tags, but those aren't really Javadoc tags.  Perhaps the `@note` tags could be part of the `@author` tags, and `@date` is unneeded, since the `@since` tags should suffice.
- Using the openHAB code formatter in Eclipse on all *.java files now would help in future reviews and maintenance.
- I will make inline comments when I am able to start reviewing the code in depth.

Thanks again for this contribution!

Satelco:
Hi @watou 

Thank you for your feedback. Ok, that all should be easily fixable. I think we will wait until someone reviewed the code so we can make all changes in one run.

watou:
Sounds good!

bundles/binding/org.openhab.binding.sallegra/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,30 @@
...
+ org.openhab.core.binding,
+ org.openhab.core.events,
+ org.openhab.core.items,
+ org.openhab.core.library.items,
+ org.openhab.core.library.types,
+ org.openhab.core.types,
+ org.openhab.io.net.http,
+ org.openhab.model.item.binding,
+ org.osgi.framework,
+ org.osgi.service.cm;version="1.4.0",
```
hakan42: If the specific version is really required, please document in about.html. If not, please remove version definition.

bundles/binding/org.openhab.binding.sallegra/src/main/java/org/openhab/binding/sallegra/internal/SallegraBinding.java
```diff
@@ -0,0 +1,353 @@
...
+		 * do we have a binding for this item at all?
+		 */
+		if (providesBindingFor(itemName)) {
+			/*
+			 * go through all the providers and look for a BindingConfig
+			 */
+			for (SallegraBindingProvider provider : providers) {
+				SallegraBindingConfig bindingConfig = provider.getBindingConfigFor(itemName);
+				
+				if (bindingConfig == null) {
```
hakan42: instead of (bindingConfig == null) continue; you could write if (bindingConfig != null) { your-business-code-here }.

At least for me, this would be more readable :smile:

Satelco:
Should be all fixed now.

hakan42:
Hi @Satelco , 

could you please rebase the PR against the current master and squash your commits into one?

Thanks,
Hakan

Issue #3437, opened 2015-11-18
MapDB Persistence wont restore PercentType with non-Integer values


mrguessed:
If you have a `PercentType` based Item, and you're using _MapDB Persistence_, then the restoration operation can fail if the original (pre-persisted) value in non-Integer.

The openHAB `PercentType` is derived from `DecimalType`, so all numerics should be supported.

When this failure occurs, it causes openHAB to loop indefinitely since the entire Persistence-restore fails, and triggers openHAB to go into a loop - never starting up.

For discovery, here's the type of Exception trace that can occur:
```
11:41:51.892 DEBUG o.o.c.p.i.PersistenceManager[:377]- Restored item state from 'Nov 18, 2015 10:25:01 AM' for item 'FamilyRoomZoneTripped' -> 'CLOSED'
11:41:51.900 DEBUG o.o.c.p.i.PersistenceManager[:377]- Restored item state from 'Nov 18, 2015 10:26:00 AM' for item 'FamilyRoomZoneArmed' -> 'ON'
11:41:51.930 ERROR o.o.m.c.i.f.FolderObserver[:105]- An unexpected exception has occured
java.lang.NumberFormatException: For input string: "26.0000000000000008881784197001252323389053344726562500"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Integer.parseInt(Integer.java:580)
        at java.lang.Integer.parseInt(Integer.java:615)
        at org.openhab.persistence.mapdb.internal.MapDBitemSerializer.deserialize(MapDBitemSerializer.java:65)
        at org.openhab.persistence.mapdb.internal.MapDBitemSerializer.deserialize(MapDBitemSerializer.java:1)
        at org.mapdb.BTreeMap$NodeSerializer.deserialize(BTreeMap.java:449)
        at org.mapdb.BTreeMap$NodeSerializer.deserialize(BTreeMap.java:288)
        at org.mapdb.Store.deserialize(Store.java:297)
        at org.mapdb.StoreDirect.get2(StoreDirect.java:475)
        at org.mapdb.StoreWAL.get2(StoreWAL.java:336)
        at org.mapdb.StoreWAL.get(StoreWAL.java:320)
        at org.mapdb.Caches$HashTable.get(Caches.java:246)
        at org.mapdb.EngineWrapper.get(EngineWrapper.java:58)
        at org.mapdb.BTreeMap.get(BTreeMap.java:602)
[...]

PR #3438, opened 2015-11-18
MapDB Persistence: PercentType restore needs to allow Decimal values (Issue #3437)

mrguessed:
openHAB Issue #3437

In openHAB, `PercentType` is derived from `DecimalType`.  The
MapDB Persistence is pre-parsing the value, assuming it's an `Integer`,
and failing when it wasn't an Integer.

In some cases, this would cause the openHAB startup process to enter a never ending loop, since the Persistence-Restore functionality was failing.

With this change, we simply hand the String value to `PercentType.valueOf()` and let it perform any necessary conversions.

### DIFF
--- a/bundles/persistence/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDBitemSerializer.java
+++ b/bundles/persistence/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDBitemSerializer.java
@@ -26,10 +26,10 @@ import org.openhab.core.types.State;
 
 /**
  * Serializer to serialize items to and from Mapdb format
- * 
+ *
  * @author Jens Viebig
  * @since 1.7.0
- * 
+ *
  */
 public class MapDBitemSerializer implements Serializer<MapDBItem>, Serializable {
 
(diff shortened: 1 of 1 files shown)


mrguessed:
@JSurf,
You you mind taking a first pass review?

lewie:
@mrguessed,
try to use https://github.com/openhab/openhab/pull/3248 instead. The problem is the conversation between 'double' and 'BigDecimal'.
More Infos:  https://github.com/openhab/openhab/wiki/JDBC-Persistence#number-precision.

Rgds
Helmut

mrguessed:
@lewie,
Actually I do want MapDB, and don't need the extra components that JDBC would introduce (I already use RRD for my numeric values, and don't require history otherwise)

This PR is specifically to address the data handling in the MapDB persistence binding, where it (externally) attempts to convert to an int, where the input value needn't necessarily be an int. (See the diffs, from my fix above, for reference)

lewie:
@mrguessed,
you are right!
Missed some details. 

Thanks for the clarification
Helmut

JSurf:
Pull request looks good to me. Perhaps für consistency DecimalType And HSBType then should use valueOf aswell ? At least JPA and mysql binding also use the same code is it also a problem there ?

mrguessed:
@JSurf,
I converted the remaining types to use `*Type.valueOf(...)` instead of `new *Type(...)` and updated the PR's Branch. 

`DecimalType` was technically the only one that warranted this conversion, but the others were done for consistency (even though they internally just `return new *Type(...)`

If JPA and/or mysql handle `PercentType` in the way MapDB Persistence did, then there's a potential problem lurking in that code if they're not scrubbing the value upon save (I haven't looked).  Did you file a ticket for them and/or open a forum discussion for them (I don't use either of these Bindings)

The other changes were cosmetic, and not an issue.

teichsta:
Thanks @mrguessed!

@lewie could you please have a look at the JPA binding and follow-up on @JSurfs comment?

Thanks, Thomas E.-E.

Issue #3432, opened 2015-11-17
Designer does not recognize the transform action in rule


watou:
[User reported](https://community.openhab.org/t/retrieving-a-result-from-a-command-line-https-curl-for-use-in-a-rule/4285/8?u=watou) on Windows 10 (64Bit).  I can reproduce on Mac 10.10.5.

steve-bate:
PR #3441

teichsta:
#3341 has been merged!

watou:
Reported fixed [here](https://community.openhab.org/t/retrieving-a-result-from-a-command-line-https-curl-for-use-in-a-rule/4285/10?u=watou).

PR #3441, opened 2015-11-19
Fix for #3432

watou:
The `transform` action was not being recognized in the Designer.  This PR adds 
`Bundle-ActivationPolicy: lazy` to bundles/core/org.openhab.core.transform/META-INF/MANIFEST.MF, and adds ActionDoc and ParamDoc annotations to the `transform` static method.

Tested on Designer on Mac.  Not tested in runtime or on Windows.


teichsta:
Thanks @watou!

PR #3444, opened 2015-11-19
allow lower chart scale values

CWempe:
I just realized the scale factor for the chart servlet is restrictet from 0.5 to 1.0.
Is there a particular reason for that?

I am using a Sasmung Galaxy S6 with a 575 dpi display.
I can use a scale of 0.5, but it would be more readable with a value of ~0.3 - 0.4.

The next generation of smartphones will be having evan higher resolution display (e.g. Sony Z5 Premium)

### DIFF
--- a/bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java
+++ b/bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java
@@ -309,7 +309,7 @@ public class ChartServlet extends HttpServlet implements ManagedService {
 		}
 		if(properties.get("scale") != null) {
 			scale = Double.parseDouble((String)properties.get("scale"));
-			if(scale < 0.5)
+			if(scale < 0.1)
 				scale = 1.0;
 		}
 	}
--- a/bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java
+++ b/bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java
@@ -309,8 +309,10 @@ public class ChartServlet extends HttpServlet implements ManagedService {
 		}
 		if(properties.get("scale") != null) {
 			scale = Double.parseDouble((String)properties.get("scale"));
-			if(scale < 0.1)
+			// Set scale to normal if the custom value is unrealisticly low
+			if(scale < 0.1) {
 				scale = 1.0;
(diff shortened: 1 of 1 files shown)


bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java
```diff
@@ -309,7 +309,7 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
 		}
 		if(properties.get("scale") != null) {
 			scale = Double.parseDouble((String)properties.get("scale"));
-			if(scale < 0.5)
+			if(scale < 0.1)
 				scale = 1.0;
```
teichsta: more expected would be to set the value to 0.1, wouldn't it?

bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java
```diff
@@ -309,7 +309,7 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
 		}
 		if(properties.get("scale") != null) {
 			scale = Double.parseDouble((String)properties.get("scale"));
-			if(scale < 0.5)
+			if(scale < 0.1)
```
teichsta: please add curly braces (i know there weren't there before)

teichsta:
Thanks for this contribution @CWempe. Please some minor comments inline. Best, Thomas E.-E.

teichsta:
arg … i pushed the wrong button as i didn't want to merge this PR before having the same change merged into ESH!

kaikreuzer:
@teichsta It is difficult to sync this change to ESH, because https://github.com/openhab/openhab/pull/785 is still missing there as a prerequisite.

watou:
I can create a sync PR to bring #785 to ESH if no one else is doing it.

kaikreuzer:
@watou Thanks for the offer, probably best to check with @cdjackson who does it. 
Looking at the [history of ChartServlet](https://github.com/openhab/openhab/commits/9e1843efbfd1ebb0f19c20d55c1481e11d9c05a9/bundles/ui/org.openhab.ui/src/main/java/org/openhab/ui/internal/chart/ChartServlet.java) it seems to me as if many more PRs than just #785 got lost on the way.

Remember to include "Also-by" footers for any contributor who worked on the code and that we will need an Eclipse CLA for all of them...

Issue #3390, opened 2015-11-06
NTP date reverts to uninitialized state on items file update


9037568:
When a change is made to the items file in configurations, the existing date items (via the NTP binding) revert to an uninitialized state.

Reproduced in openhab versions:
1.6.2
1.7.0 (snapshot)

steve-bate:
PR has been merged.

PR #3450, opened 2015-11-21
Fix for #3390. Refreshes NTP-bound items when binding changes (items reloaded).

steve-bate:
I extracted part of the execute method and called it from the bindingChanged callback. I manually tested it by updating my items file and verifying the NTP-bound item was updated immediately after the items file was reloaded rather than waiting for the next timer callback. Re: #3390

### DIFF
--- a/bundles/binding/org.openhab.binding.ntp/src/main/java/org/openhab/binding/ntp/internal/NtpBinding.java
+++ b/bundles/binding/org.openhab.binding.ntp/src/main/java/org/openhab/binding/ntp/internal/NtpBinding.java
@@ -24,6 +24,7 @@ import org.apache.commons.net.ntp.NTPUDPClient;
 import org.apache.commons.net.ntp.TimeInfo;
 import org.openhab.binding.ntp.NtpBindingProvider;
 import org.openhab.core.binding.AbstractActiveBinding;
+import org.openhab.core.binding.BindingProvider;
 import org.openhab.core.library.types.DateTimeType;
 import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.cm.ManagedService;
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.ntp/src/main/java/org/openhab/binding/ntp/internal/NtpBinding.java
```diff
@@ -83,19 +84,23 @@ public void execute() {
...
-			for (String itemName : provider.getItemNames()) {
-				
-				TimeZone timeZone = provider.getTimeZone(itemName);				
-				Locale locale = provider.getLocale(itemName);
-				
-				Calendar calendar = Calendar.getInstance(timeZone, locale);
-				calendar.setTimeInMillis(networkTimeInMillis);
-
-				eventPublisher.postUpdate(itemName, new DateTimeType(calendar));
+			for (String itemName : provider.getItemNames()) {
```
hakan42: Please run eclipse formatter to get rid of trailing tabs

bundles/binding/org.openhab.binding.ntp/src/main/java/org/openhab/binding/ntp/internal/NtpBinding.java
```diff
@@ -83,19 +84,23 @@ public void execute() {
...
+				updateTime(provider, itemName, networkTimeInMillis);
 			}
 		}
 		
 	}
+
+	private void updateTime(NtpBindingProvider provider, String itemName,
+			long networkTimeInMillis) {
+		TimeZone timeZone = provider.getTimeZone(itemName);				
+		Locale locale = provider.getLocale(itemName);
```
hakan42: Please run eclipse formatter to get rid of trailing tabs

hakan42:
@steve-bate , if you could remove the trailing slashes, and squash the result into one commit, this change looks good to me and could be merged. Thank you for your contribution :smile:

steve-bate:
It appears those tabs were in the original file. How did you detect them? Will you point me to the Eclipse formatted configuration to use?

hakan42:
I pulled your PR and ran a "git diff" in my colorized shell. Which has show me the tabs.

The complete OH2 development environment ( https://community.openhab.org/t/setting-up-a-development-environment/2445 ) contains the code formatter configuration too.

On the other hand, @kaikreuzer mentioned in another PR that we should maybe *not* reformat older OH1 code with the new OH2 settings as this would change too much of the code, making reviews harder to do.

So, all that remains to do on this PR seems to be to remove the three tab characters :smile:

kaikreuzer:
> we should maybe not reformat older OH1 code with the new OH2 settings

Correct, you wouldn't be able to distinguish between code changes and format changes.
Once we declare the OH1 IDE setup obsolete and we have all developers move over to OH2, we should probably run a one time code formatter over all OH1 addons.

steve-bate:
The tabs at the end of the line in the extracted method are removed. Commits squashed.

hakan42:
Thank you for your contribution :smile:

PR #3457, opened 2015-11-23
Updated Lightwave Binding including:

foxy82:
1) Improved the message sending and receiving so both can happen at the same time
2) Added support for Relays
3) Added support for Energy Monitor
4) Added support for status message from lightwave wifi link
5) Tidied code and improved inheritence


teichsta:
@foxy82 since you changed a lot in the PR can anyone from the community confirm the binding is still working? Whom shall we mention here in order to get user feeback? Did you check your binding against OH2?

foxy82:
Hi,

Yeah a lot changed as the original binding had some serious issues.
There is a thread here where people are testing this:
https://community.openhab.org/t/lightwaverf-binding-updated/4446/46

SupraWez, tinkerfailure, and morak can confirm they have tested it.

Also another thread here: https://community.openhab.org/t/lightwaverf-heating-controls/4432/13
Which barneyd can comment on testing. 

I've not tested it against OH2 as I don't have an OH2 installation.

tinkerfailure:
Yeah, I've been using the updates as foxy82 has been developing them. I've just installed the latest snapshot a couple of hours ago, so I'll keep an eye on it over the next 24 hours. I'm running OH 1.7.1.

teichsta:
ok, sounds god already. If any of you could please test against OH2 i would be happy to merge this PR.

teichsta:
> I've not tested it against OH2 as I don't have an OH2 installation.

could you please have a look into this the next days? Appreciate your feedback!

foxy82:
I see this didn't make it in to 1.8.0 due to the fact we don't have anyone to test on OH2. Essentially this now means that anyone that gets the latest firmware upgrade from Lightwave can't use this binding. 

Whilst OH2 isn't production stable it seems wrong to have a requirement that a binding being released for OH1 has to pass testing in OH2. It's essentially meant that people on OH1 are now suffering due to the fact OH2 isn't ready...

PR #3458, opened 2015-11-23
Weather: Removed wind post processing, not necessary

gerrieg:


teichsta:
Thanks @gerrieg! Just out of curiosity why did you originally added the preprocessing?

gerrieg:
i thought, if wind speed is null or 0, then there is also no wind direction and i set both to null. But that's not so optimal for the UI.

PR #3463, opened 2015-11-24
A fix to prevent item state updates if a homematic devices is not reachable.

csowada:
+ Add a AutoUpdateBindingProvider
+ Copy private method postUpdate() from AutoUpdateBinding
+ Add boolean return value to HomematicCommunicator to identifiy
failures


csowada:
This is a small fix, maybe there is is a easier solution.

teichsta:
@gerrieg would you be so kind and have a quick look (or can you refer to others being able to?)

csowada:
Is there a better solution than copying the function postUpdate(,,,) from org.openhab.core.autoupdate.internal.AutoUpdateBinding ? I dont' like double code.

watou:
@csowada, could you explain in greater detail what this PR is trying to achieve/problem you want to solve (ignoring for the moment how you are trying to achieve it)?  Is there a link to a discussion thread regarding this subject you could add?

csowada:
@watou Regarding your question, no there is no public thread or discussion.

If you send a command through a switch item to the binding and there is a communication error, the item is nonetheless set to a new value.

This fix is intended to avoid this behaviour. In this case the binding should decide if the communication was successful or not and set the new item value.

I use the "AutoUpdateBinding" service to disable this for this binding, but I have copied a private code block from AutoUpdateBinding. That is maybe not optimal.

watou:
Thanks for the reply.  I don't disagree that the autoupdate behavior may be considered incorrect when one (some? all?) connected bindings fails to propagate the command, but I do disagree that the approach to address it would be per-binding code changes.  It seems to me to be an architecture-level issue.  Could you not simply advise users of this, or any, binding to add `, autoupdate="false"` to their `{ binding section }`, and raise the subject with the architects or with community at large?

csowada:
I agree with you that this behavior is not optimal. Disabling AutoUpdate only works if there is only one binding per item. But this is mostly the case.

If you add all items ``autoupdate="false``, then you must also apply rules to all items. Also not very nice. But I understand if this change is not suitable for the standard.

But we should discuss the design issue and to develop a user-friendly solution. For a normal user it is not clear why items are set in openhab, although there was an error.

Should I start a thread in our community forum?

watou:
> Should I start a thread in our community forum? 

By all means!

teichsta:
any progress @csowada? Can you please refer to the according community thread here?

csowada:
I started a open discussion here https://community.openhab.org/t/general-discussion-about-autoupdate-of-items/4678
But I think this change should be decided by @gerrieg and the openhab team.
I accept if this change does not fit into the openhab 1 framework.

teichsta:
it would be great if you could briefly comment on this issue @gerrieg. Thanks, Thomas E.-E.

gerrieg:
Sure, sorry, i overlooked the issue.
While this PR is OK for Homematic, i think we need a global solution for all bindings, see my comment in the discussion. If this PR will be merged, we have the first inconsistency

csowada:
I personally would close this PR and address this general issue to a further release.

teichsta:
> Sure, sorry, i overlooked the issue.

no problem.

> If this PR will be merged, we have the first inconsistency

ok, so no good idea to start that in the OH1 version of the binding.

> I personally would close this PR and address this general issue to a further release.

yes, probably directly in the OH2 version? Would that make sense?

Issue #728, opened 2014-01-10
REGEX expression in TCP binding


bertalf:
There seems to be an issue with REGEX expressions in items with the TCP binding. When adding {tcp="<[IP:PORT:'REGEX(([-0-9]*))']"} to an item, I get the error message: "given transformation function 'REGEX(([-0-9' does not follow the expected pattern '<function>(<pattern>)' at org.openhab.binding.tcp.protocol.internal.TCPBinding.splitTransformationConfig(TCPBinding.java:206)"

It seems to me that openHAB interpretes the ']' as closing bracket of the binding rather than as part of the REGEX expression.

Also when entering a REGEX expression like {tcp="<[IP:PORT:'REGEX((\d*))']"} I get an error in the openHAB designer which interpretes the '\' incorrectly.

steve-bate:
The transformation part of the ACTION and STATUS_CONFIG_PATTERN regex patterns shouldn't use a non-greedy match. It prematurely matches the ] in the REGEX transformation specification.

steve-bate:
@kgoderis What was the reason for the optional single quotes around the transformation?

steve-bate:
Closing issue for merged PR.

PR #3464, opened 2015-11-24
Fix for #728. REGEX Expression in TCP binding.

steve-bate:
Modified the TcpBinding regular expressions to better handle binding delimiter
characters in a REGEX transformation. Modified the binding parsing code to
handle the different matching groups. Changed a Command StringItem comparison
from an identity check (==) to "equals". During testing I found that sending
commands through the REST API didn't work because the StringItem instances
where equal but not the same instance. Re: #728


teichsta:
Thanks @steve-bate!

PR #3468, opened 2015-11-26
caldav binding 26/11

querdenker2k:

PR #3469, opened 2015-11-26
Included CalDAV binding into the build process

hakan42:


hakan42:
This is in addition to #3468

PR #3475, opened 2015-11-27
fix permission and content directory issues when used within apt installation

theoweiss:
This PR is related to https://github.com/openhab/openhab/issues/3451
Changes:
1/ use smarthome.userdata directory for storing dbx files
2/ smarthome/openhab.configdir as default contentdir


theoweiss:
@teichsta could you please review these changes (your the initial contributor :-) ).
Especially getConfigDirFolder() holds a compromise for  OH2 compatibility, which will deprecate as soon as OH1 is EOL. May be there is a better solution for handling the smarthome/openhab.configdir properties?

teichsta:
changes lgtm, thanks @theoweiss!

PR #3476, opened 2015-11-27
apt installation: openhab user as owner of /etc/openhab/configurations directory

theoweiss:
This PR is related to https://github.com/openhab/openhab/issues/3451

With this change the openhab user will be the owner of the /etc/openhab/configurations directory and all files below. Thus bindings will be able to change the openhab configuration. It means somewhat lesser security, but better usability.


teichsta:
Thanks @theoweiss!

PR #3477, opened 2015-11-27
sapp binding: added signed intepretation for words and bytes

paolodenti:
This version adds capability to interpret words and bytes as signed.

Existing unsigned bindings data types are:

* \* (unsigned word, 0 to 65535)
* H (unsigned high byte, 0 to 255)
* L (unsigned low byte, 0 to 255)

New signed correspondent data types bindings are:

* + (signed word, -32768 to 32767)
* H+ (unsigned high byte, -128 to 127)
* L+ (unsigned low byte, -128 to 127)


hakan42:
Looks fine to me.

PR #3485, opened 2015-11-28
Log execution failures in ExecUtil.

steve-bate:
Currently, the ExecUtil silently ignores exceptions caught by the Apache Commons Exec utilities. The `executor.execute` does not throw an exception so the existing implementation doesn't log the error. The exception is returned in the result handler. The change in this PR looks for an exception in the result handler and logs a warning with the exception message. This will apply to both the Exec transformation service and the Exec action and will help users diagnose problems with the external programs.

### DIFF
--- a/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/exec/ExecUtil.java
+++ b/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/exec/ExecUtil.java
@@ -137,7 +137,12 @@ public class ExecUtil {
 			resultHandler.waitFor();
 			int exitCode = resultHandler.getExitValue();
 			retval = StringUtils.chomp(stdout.toString());
-			logger.debug("exit code '{}', result '{}'", exitCode, retval);
+			if (resultHandler.getException() != null) {
+				logger.warn(resultHandler.getException().getMessage());
+			}
+			else {
+				logger.debug("exit code '{}', result '{}'", exitCode, retval);
+			}
 
 		} catch (InterruptedException e) {
 			logger.error("Timeout occured when executing commandLine '"


kaikreuzer:
Could you please contribute the same as a PR to https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/exec/ExecUtil.java? See https://www.eclipse.org/smarthome/documentation/community/contributing.html#conventions-for-pull-requests for the details on how to do this. Thanks!

steve-bate:
@kaikreuzer, I don't currently have an ESH development environment and the Eclipse IDE set up instructions on the site seem to be a bit out of date. Some of screenshots were different than the screens I saw in the latest version of the Mac installer. It didn't ask for github credentials and didn't clone the ESH repository. I double-checked the settings (selected the Eclipse / Runtime and Designer entry). I cloned the repository myself and imported the projects into Eclipse but I currently have 20K+ errors to diagnose. :-) I'm not going to have time to work through all the issues and do the paperwork to submit an ESH PR in the near future. It might be better for one of the existing ESH developers to merge this change into that repository.

kaikreuzer:
Ok, I have added it to ESH (https://github.com/eclipse/smarthome/pull/550) and the compat layer in OH2 (https://github.com/openhab/openhab2/pull/499).

Nonetheless, we should get you up to speed for OH2 development - actually the IDE setup is supposed to be much easier than for OH1.

> the Eclipse IDE set up instructions on the site seem to be a bit out of date

This should not be the case. Maybe it is lacking some infos for certain situations. Let's try to find out:

> It didn't ask for github credentials

This sounds as if you have used the installer before. If you entered your credentials once, it will never ask again (but you can make them appear by selecting "show all variables" on the lower left.

> didn't clone the ESH repository
> I double-checked the settings (selected the Eclipse / Runtime and Designer entry)

Did you not only double check, but also double click the entry? This is the step that is usually missed...

steve-bate:
Thanks for the tips. I'll try it again. I don't remember using the installer before, but it's possible I did at some point. I think you are correct about the double-click.

kaikreuzer:
Btw, for OH2 development (and required changes in ESH), I would recommend this IDE setup https://github.com/openhab/openhab2/blob/master/docs/sources/development/ide.md, where in step 4 you will also find an option "openHAB 2 with ESH", which sets up the workspace with both ESH and OH2 bundles. This isn't yet documented as I am still doing changes to it, but this will eventually be the IDE I ask openHAB developers to set up for themselves.

steve-bate:
@kaikreuzer, the OH2 installer worked great. *Much* easier than OH1 IDE setup. Thanks!

PR #3486, opened 2015-11-28
Added support for regex substitutions to Regex transformation

steve-bate:
This modification allows the Regex transformation to do substitutions. A substitution pattern is specified using a syntax similar to sed, Python, Perl, etcetera. The substitution will be performed if the regex looks like "s/<pattern>/<substitution>/". The substitution can use capturing groups using the "$n" syntax supported by Java regular expressions. By default, the substitution is applied to the first match. If a "g" is appended to the regex (e.g., "s/([0-9]+)000/$1K/g"), then the substitution if applied to all matches in the source string.

### DIFF
--- a/bundles/core/org.openhab.core.transform.test/src/test/java/org/openhab/core/transform/internal/RegExTransformationServiceTest.java
+++ b/bundles/core/org.openhab.core.transform.test/src/test/java/org/openhab/core/transform/internal/RegExTransformationServiceTest.java
@@ -69,5 +69,21 @@ public class RegExTransformationServiceTest extends AbstractTransformationServic
 		Assert.assertEquals("8", transformedResponse);
 	}
 	
+	@Test
+	public void testTransformByRegex_substituteFirst() throws TransformationException {
+		// method under test
+		String transformedResponse = processor.transform("s/^OP:(.*?),ARG:(.*)$/$1($2)/", "OP:SetMode,ARG:42");
+		
+		// Asserts
+		Assert.assertEquals("SetMode(42)", transformedResponse);
+	}
 
+	@Test
+	public void testTransformByRegex_substituteAll() throws TransformationException {
+		// method under test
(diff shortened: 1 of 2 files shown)


steve-bate:
@kaikreuzer Using the OpenHAB2/ESH Eclipse setup I've applied these changes to the ESH code. However, I'm having trouble running the unit tests in org.eclipse.smarthome.transform.regex.test from Eclipse. It seems to be missing an org.hamcrest.core dependency. I also tried running the unit tests using Maven but I'm not trusting the results. The Maven test phase passes, but it still passes even if I intentionally break one of the unit tests.

For the Eclipse project plugin dependency, is that generated from the Maven pom or in some other way? The plugin manifest specifies junit 4.0.0 but the Plug-in Dependencies list in the Package Explorer says it's using junit 4.11.0 (which requires the hamcrest jar).

Any suggestions?

kaikreuzer:
>  I'm having trouble running the unit tests in org.eclipse.smarthome.transform.regex.test from Eclipse

I never tried running the tests from the openHAB 2 IDE - it is well possible that something is missing in the target platform there, which exists in the ESH target platform.

> The Maven test phase passes, but it still passes even if I intentionally break one of the unit tests.

This is the ESH Maven build, right? That is interesting - so if we have tests that actually aren't executed, please enter an issue at https://github.com/eclipse/smarthome/issues for follow up!

> For the Eclipse project plugin dependency, is that generated from the Maven pom or in some other way? 

No, not generated, it is manually maintained. Having 4.0.0 im the manifest means that this is the minimal version - if 4.11 is resolved, this is hence fine.

teichsta:
@steve-bate the changes lgtm.

However since this changes the core.transform services i would like to ask you to provide the same changes to ESH as well (please refer to the according PR here). Once the ESH PR has been merged we can go ahead with this one. Thanks, Thomas E.-E.

steve-bate:
I've had trouble running the ESH regex unit tests. I'll reopen this PR once I have a working ESH development environment.

Issue #2847, opened 2015-07-04
MPD binding notification


zergasso:
In my openhab installation mpd binding updates info about track and artist played only once at system start, then nothing more even if a play/pause or previous/next is triggered on sitemap.
Mpd server is on another (virtual) machine and the others clients (linux, android) in the network update info regularly on track changing.

item config is pretty straightforward as it is essentially copy/pasted from documentation of the MPD binding bundle:
```
String CurrentTrack    "Current track [%s]" { mpd="TITLE:avserver:tracktitle" }
String CurrentArtist    "Current artist [%s]" { mpd="ARTIST:avserver:trackartist" }
String ConcatInfo       "Now playing [%s]"

Switch Mpd_avserver_StartStop       "Start/Stop" 	  (Audio)  { mpd="ON:avserver:play, OFF:avserver:stop" }
Switch Mpd_avserver_NextPrev        "Track control"   (Audio)  { mpd="ON:avserver:next, OFF:avserver:prev" } 
Dimmer Mpd_avserver_VolumeControl   "Volume [%d%%]"   (Audio)  { mpd="INCREASE:avserver:volume_increase, DECREASE:avserver:volume_decrease, PERCENT:avserver:volume" }
``` 
rule config:
```
rule "concat"
		when
  			Item CurrentTrack received update or
  			Item CurrentArtist received update
		then
  			ConcatInfo.postUpdate(CurrentTrack.state.toString + " / " + CurrentArtist.state.toString)
end
```
sitemap config:
```
Text label="Musica" icon="media_player1"
					{
						Text item=CurrentTrack
						Text item=CurrentArtist
						Text item=ConcatInfo
						Switch item=Mpd_avserver_StartStop mappings=[OFF="Pause", ON="Play"]
						Switch item=Mpd_avserver_NextPrev  mappings=[OFF="Previous", ON="Next"]
						Slider item=Mpd_avserver_VolumeControl
					}
```
some log:
```
tail -f /var/log/openhab/openhab.log | grep mpd
2015-07-04 16:19:59.433 [DEBUG] [o.o.b.mpd.internal.MpdActivator] - MPD binding has been started.
2015-07-04 16:19:59.451 [DEBUG] [o.o.binding.mpd.internal.MpdBinding] - MPD binding connecting to players
2015-07-04 16:19:59.719 [DEBUG] [o.o.binding.mpd.internal.MpdBinding] - MPD binding connecting to players
[...]

PR #3500, opened 2015-12-02
mpd: fix artist and track not updated

stefanroellin:
It may happen that a song changes, but the id stays the same (e.g. for streams).
Therefore, test whether the title or the artist has changed instead of
the id (which is of no interest anyway).


stefanroellin:
When I wrote the patch, I was not aware of issue #2847, which could be fixed with this pull request.

teichsta:
Thanks @stefanroellin for this contribution! Does this address any other open issue regarding MPD?

stefanroellin:
I am not aware of further issues that could be solved by this patch.

teichsta:
ok, thanks @stefanroellin! @stefanroellin could you please also follow-up on #2847?

PR #3504, opened 2015-12-04
[mqtt] Make inbound messages sensitive to item's accepted types and commands

watou:
This PR is for discussion and review only at this point.  It relates to [this discussion](https://community.openhab.org/t/strange-string-behaviour-on-on-off/4735/1).

### DIFF
--- a/bundles/binding/org.openhab.binding.mqtt.test/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.mqtt.test/META-INF/MANIFEST.MF
@@ -10,3 +10,4 @@ Require-Bundle: org.junit;bundle-version="4.8.1"
 Bundle-ClassPath: lib/mockito-all-1.9.5.jar,
  lib/slf4j-jdk14-1.6.4.jar,
  .
+Import-Package: org.openhab.core.library.items
--- a/bundles/binding/org.openhab.binding.mqtt.test/src/test/java/org/openhab/binding/mqtt/internal/MqttItemConfigTest.java
+++ b/bundles/binding/org.openhab.binding.mqtt.test/src/test/java/org/openhab/binding/mqtt/internal/MqttItemConfigTest.java
@@ -11,6 +11,8 @@ package org.openhab.binding.mqtt.internal;
 import static org.junit.Assert.assertEquals;
 
 import org.junit.Test;
+import org.openhab.core.library.items.NumberItem;
+import org.openhab.core.library.items.SwitchItem;
 import org.openhab.model.item.binding.BindingConfigParseException;
 
 /**
@@ -22,7 +24,7 @@ public class MqttItemConfigTest {
 	@Test
 	public void canParseInboundConfig() throws BindingConfigParseException {
 
-		MqttItemConfig c = new MqttItemConfig("myItem",
+		MqttItemConfig c = new MqttItemConfig(new NumberItem("myItem"),
 				"<[publicweatherservice:/london-city/temperature:state:default]");
 		assertEquals(0, c.getMessagePublishers().size());
 		assertEquals(1, c.getMessageSubscribers().size());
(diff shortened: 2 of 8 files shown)


steve-bate:
Thanks for creating the PR. To continue the discussion, I see at least a couple of issues with this approach. The first is that the item class could change if an item is redefined. This could lead to failed updates for similar reasons as the current implementation. The second issue is that some item types (e.g.,  StringItem) don't have a default constructor. 

I still like the idea of looking up the item in the registry. However, a drawback of that approach is the name-based lookup. The current lookup implementation is O(n) and it would be done again when the event is published. To be efficient for large item sets, the binding might need to maintain it's own hashed index of items and process item registry listener callbacks.

watou:
>  The first is that the item class could change if an item is redefined.

I am fairly sure that this is handled, as the `MqttMessageSubscriber` objects which hold the item class are recreated when items are redefined.  Please tell me where in the code I'm wrong!

> The second issue is that some item types (e.g., StringItem) don't have a default constructor

Oh yes, that's a problem! :) But since all items type have a one-arg constructor, that's easily fixed!

> I still like the idea of looking up the item in the registry.

What makes that approach preferable, or one that would duplicate a subset of the item registry in a hashset in the binding?  If we know each in-binding's item's class like in the PR, the work is already done.

Once my PR is working, it will have the advantage of having a shorter code path and producing less garbage to collect than current.

Also... I know that the MQTT event bus binding is broken with this PR, but I think I can fix it so it works more flexibly than current, so, for example, the publishing openHAB could have a String item publish over MQTT to a subscribing openHAB that has a different item type, like Location or Number, instead of it being hardcoded based on what can be parsed, as it is currently.

Thanks very much for your input; I look forward to more!

watou:
> Oh yes, that's a problem! :) But since all items type have a one-arg constructor, that's easily fixed!

Or better yet, I can just keep copies of the accepted type and command lists in the `MqttMessageSubscriber` object from the current `Item`, so no `Item` needs to be constructed at all.  This will make the subscribing code path less costly.  I will update the PR soon.

watou:
Might have to inspect `ItemRegistry` to get `MqttEventBusBinding` to work as intended, then I will test this PR thoroughly.

steve-bate:
My earlier comment about item redefinition was based on a misperception. You had said that a binding should never hold onto an item because of differing lifecycles. However, I see now that binding change listeners can update their item-related information when the items are redefined/reloaded. I don't currently use MQTT so I didn't have a test environment set up to step through it in a debugger. I've created one now and tried your changes and they are working well for me.

I think this will be a significant improvement to the binding. Thanks!

watou:
That's great to hear, Steve.  The current PR doesn't build, but I will complete it make sure a JAR gets a thorough test by a few community members including myself. I use the MQTT message bus binding "in production" as well so we should be covered.

steve-bate:
Yeah, to do my testing I commented the code that wasn't compiling in the bus binding.

watou:
I am not aware of any issues with this PR.  I am soliciting feedback from anyone using either or both of the MQTT items or event bus bindings by testing [this JAR](https://dl.dropboxusercontent.com/u/4286376/mqtt-types/org.openhab.binding.mqtt-1.8.0-SNAPSHOT.jar) or building this PR on your own.

watou:
@dvanherbergen Hi Davy, since you are involved again with openHAB/Eclipse SmartHome, I thought I would run this PR by you, since you were the original author of the MQTT binding.  This PR appears to work without problems and solve the original issue, has been user tested, and I have no doubts about it.  But if you have any comments, please let me know because I intend to merge it this weekend.  Thank you.

dvanherbergen:
Thanks for asking. I don't see any immediate issues with the PR.

watou:
Thank you, Davy.

PR #3509, opened 2015-12-05
fixed ucprelayboard binding dependencies and CRC calculations

rmichalak:


hakan42:
Why does this binding need javax.mail? I must admit I am not at my desktop right now and cannot read the full sources. Maybe if you could give me a short hint here?

rmichalak:
It doesn't. I just removed this dependency. Copy-paste error ;)

hakan42:
Thank you for the explanation  :smile:

PR #3515, opened 2015-12-06
InsteonPLM: Include known features when an item is configured with an unknown feature

robnielsen:
Improve error logging by logging the known features when an unknown feature is accidentally used. The previous error message was:

    item foo references unknown feature: dimmer, item disabled!

It now is:

    item foo references unknown feature: dimmer, item disabled! Known features for F00.00.15 are: fastonoff, keypadbuttonA, keypadbuttonB, keypadbuttonC, keypadbuttonD, lastheardfrom, loaddimmer, manualchange.

@berndpfrommer does this look OK?

### DIFF
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/InsteonPLMActiveBinding.java
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/InsteonPLMActiveBinding.java
@@ -9,7 +9,9 @@
 package org.openhab.binding.insteonplm;
 
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Dictionary;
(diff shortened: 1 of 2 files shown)


berndpfrommer:
@robnielsen couldn't you just use this as a test whether a feature is a group feature or not:

    public boolean isFeatureGroup() {
    		 return (!m_connectedFeatures.isEmpty()); 
    }

That would avoid having to specify that explicitly in the handlers. The feature groups (i.e. features that have features connected to them) should never be referenced in the items file, so I think that test should be sufficient.

On the cosmetic side, I would prefer "feature group" over "group feature" :)

Other than that the PR looks good to me.

robnielsen:
Thanks for the suggestion @berndpfrommer. I've made the changes and squashed the commits.

teichsta:
Thanks @robnielsen for this contribution and @berndpfrommer for reviewing the code! Appreciate your support!

PR #2567, opened 2015-05-09
New binding to control RC switches

mroeckl:
This OpenHAB binding can control RC switches on a Raspberry Pi via a 433Mhz RF module connected to one of the IO ports of the Raspberry Pi.

The IO port can be configured in the OpenHAB configuration file. Items can be set up like this:

*Switch   SleepingRoomLight   {rcswitch="11111:4"}*

*'11111'* being the group address and *'4'* being the device address.

A discussion is already ongoing in the OpenHAB issues/form: [RC Switch binding to control RF transmitters](https://github.com/openhab/openhab/issues/1993) and [Binding to control RF Power Switches](https://groups.google.com/forum/#!category-topic/openhab/features/_TPI5STjP8s)

PR #3518, opened 2015-12-07
Add about page for googleTTS addon including 3rd party license information

dominicdesu:
As suggested by @kaikreuzer in #2567


hakan42:
Looks good to me.

PR #3524, opened 2015-12-09
[sapp binding] - binding not included in addons build

paolodenti:
distribution/pom.xml is missing sapp binding - binding is not included in build


hakan42:
Oops, sorry :smile:

paolodenti:
thanks hakan

PR #3528, opened 2015-12-10
Add TechST500 Binding

snoerenberg:
Binding for TechST500
http://www.techsterowniki.pl/eng/tech-st500-produkty


watou:
I see this is closed, but I did notice that you submitted the PR from your fork's `master` branch, which will lead to inconveniences later as the PR is reviewed.  I recommend creating two branches in your fork, both off latest `master`, like
```
git checkout master
git checkout -b temperlan-binding
git checkout master
git checkout -b techst500-binding
```
and then adding the respective files to each branch and submitted two PRs from those branches.  That way, you can add further commits on those branches without confusion.

PR #3539, opened 2015-12-12
Added Chamberlain MyQ Binding

computergeek1507:
@teichsta Chamberlain MyQ Binding is ready for review

### DIFF
--- a/bundles/binding/org.openhab.binding.myq/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.myq/META-INF/MANIFEST.MF
@@ -1,0 +1,44 @@
+Manifest-Version: 1.0
+Private-Package: org.openhab.binding.myq.internal
+Ignore-Package: org.openhab.binding.myq.internal
+Bundle-License: http://www.eclipse.org/legal/epl-v10.html
+Bundle-Name: openHAB myq Binding
+Bundle-SymbolicName: org.openhab.binding.myq
+Bundle-Vendor: openHAB.org
+Bundle-Version: 1.8.0.qualifier
+Bundle-ManifestVersion: 2
+Bundle-Description: This is the myq binding of the open Home Aut
+ omation Bus (openHAB)
+Import-Package: com.google.common.base,
+ com.google.common.io,
+ org.apache.commons.httpclient,
+ org.apache.commons.httpclient.auth,
+ org.apache.commons.httpclient.methods,
+ org.apache.commons.httpclient.params,
+ org.apache.commons.httpclient.util,
+ org.apache.commons.io,
+ org.apache.commons.lang,
+ org.apache.commons.lang.builder,
+ org.apache.commons.logging,
+ org.codehaus.jackson,
+ org.codehaus.jackson.annotate,
+ org.codehaus.jackson.map,
+ org.codehaus.jackson.map.annotate,
+ org.openhab.core.binding,
+ org.openhab.core.events,
+ org.openhab.core.items,
+ org.openhab.io.net.http,
+ org.openhab.core.library.items,
+ org.openhab.core.library.types,
+ org.openhab.core.types,
+ org.openhab.model.item.binding,
+ org.osgi.framework,
+ org.osgi.service.cm,
+ org.osgi.service.component,
+ org.osgi.service.event,
+ org.slf4j
+Export-Package: org.openhab.binding.myq
+Bundle-DocURL: http://www.openhab.org
+Bundle-RequiredExecutionEnvironment: J2SE-1.5
+Service-Component: OSGI-INF/binding.xml, OSGI-INF/genericbindingprovider.xml
+Bundle-ClassPath: .
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/MyqBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/MyqBindingProvider.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.myq;
+
+import java.util.List;
+
+import org.openhab.core.binding.BindingProvider;
+import org.openhab.binding.myq.internal.MyqBindingConfig;
+
+/**
+ * This interface is implemented by classes that can provide mapping information
+ * between openHAB items and Chamberlain Garage Door Openers.
+ * 
+ * I used the Hue binding as my example so most of the code is the same
+ * 
+ * @author Scott Hanson
+ * @since 1.8.0
+ */
+public interface MyqBindingProvider extends BindingProvider {
+
+	/**
+	 * Returns the configuration for the item with the given name.
+	 * 
+	 * @param itemName
+	 * @return The configuration if there is an item with the given name, null
+	 *         otherwise.
+	 */
+	public MyqBindingConfig getItemConfig(String itemName);
+
+	/**
+	 * Returns a list of item names
+	 * 
+	 * @return List of item names mapped to a Chamberlain myQ Binding
+	 */
+	public List<String> getInBindingItemNames();
+}
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorData.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorData.java
@@ -1,0 +1,90 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.myq.internal;
+
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.regex.Pattern;
+
+import org.codehaus.jackson.JsonNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This Class parses the JSON data and creates a LinkedList of Garage Door
+ * Openers
+ * <ul>
+ * <li>devices: LinkedList of Devices</li>
+ * </ul>
+ * 
+ * @author Scott Hanson
+ * @since 1.8.0
+ */
+public class GarageDoorData {
+	static final Logger logger = LoggerFactory.getLogger(GarageDoorData.class);
+
+	/**
+	 * Matching pattern for door device types, some models have spaces in them,
+	 * some don't
+	 */
+	private static final Pattern DEVICE_TYPE_PATTERN = Pattern
+			.compile("Garage\\s?Door\\s?Opener");
+
+	LinkedList<GarageDoorDevice> devices = new LinkedList<GarageDoorDevice>();
+
+	/**
+	 * Constructor of the GarageDoorData.
+	 * 
+	 * @param rootNode
+	 *            The Json node returned from the myq website.
+	 */
+	public GarageDoorData(JsonNode rootNode) throws IOException {
+		if (rootNode.has("Devices")) {
+			JsonNode node = rootNode.get("Devices");
+			if (node.isArray()) {
+				logger.trace("Chamberlain MyQ Devices:");
+				int arraysize = node.size();
+				for (int i = 0; i < arraysize; i++) {
+					int deviceId = node.get(i).get("MyQDeviceId").asInt();
+					String deviceName = node.get(i).get("SerialNumber")
+							.asText();
+					String deviceType = node.get(i).get("MyQDeviceTypeName")
+							.asText();
+
+					if (DEVICE_TYPE_PATTERN.matcher(deviceType).matches()) {
+						JsonNode attributes = node.get(i).get("Attributes");
+						if (attributes.isArray()) {
+							int attributesSize = attributes.size();
+							for (int j = 0; j < attributesSize; j++) {
+								String attributeName = attributes.get(j)
+										.get("AttributeDisplayName").asText();
+								if (attributeName.contains("doorstate")) {
+									int doorstate = attributes.get(j)
+											.get("Value").asInt();
+									logger.trace(
+											"DeviceID: {} DeviceName: {} DeviceType: {} Doorstate : ",
+											deviceId, deviceName, deviceType,
+											doorstate);
+
+									devices.add(new GarageDoorDevice(deviceId,
+											deviceType, deviceName, doorstate));
+									break;
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	public GarageDoorDevice getDevice(int index) {
+		return index >= devices.size() ? null : devices.get(index);
+	}
+}
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorDevice.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorDevice.java
@@ -1,0 +1,127 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.myq.internal;
+
+/**
+ * This Class holds the Garage Door Opener Device data.
+ * <ul>
+ * <li>DeviceId: DeviceId from API, need for http Posts</li>
+ * <li>DeviceType: MYQ Device Type. GarageDoorOpener or Gateway I've seen</li>
+ * <li>DeviceName: Serial number of device I think</li>
+ * <li>Status: Garage Door Opener "doorstate" Attribute</li>
+ * </ul>
+ * 
+ * @author Scott Hanson
+ * @author Dan Cunningham
+ * @since 1.8.0
+ */
+class GarageDoorDevice {
+	private int deviceId;
+	private String deviceType;
+	private String deviceName;
+	private GarageDoorStatus status;
+
+	public GarageDoorDevice(int deviceId, String deviceType, String deviceName,
+			int status) {
+		this.deviceId = deviceId;
+		this.deviceType = deviceType;
+		this.deviceName = deviceName;
+		this.status = GarageDoorStatus.GetDoorStatus(status);
+	}
+
+	public int getDeviceId() {
+		return this.deviceId;
+	}
+
+	public String getDeviceType() {
+		return this.deviceType;
+	}
+
+	public String getDeviceName() {
+		return this.deviceName;
+	}
+
+	public GarageDoorStatus getStatus() {
+		return this.status;
+	}
+
+	public enum GarageDoorStatus {
+		OPEN("Open", 1), CLOSED("Closed", 2), PARTIAL("Partially Open/Closed",
+				3), OPENING("Opening", 4), CLOSING("Closing", 5), UNKNOWN(
+				"Unknown", -1);
+
+		/**
+		 * The label used to display status to a user
+		 */
+		private String label;
+		/**
+		 * The int value returned from the MyQ API
+		 */
+		private int value;
+
+		private GarageDoorStatus(String label, int value) {
+			this.label = label;
+			this.value = value;
+		}
+
+		/**
+		 * Label for the door status
+		 * 
+		 * @return human readable label
+		 */
+		public String getLabel() {
+			return label;
+		}
+
+		/**
+		 * Int value of the door status
+		 * 
+		 * @return int value of the door status
+		 */
+		public int getValue() {
+			return value;
+		}
+
+		/**
+		 * Is the door in a closed or closing state
+		 * 
+		 * @return is closed or is closing
+		 */
+		public boolean isClosedOrClosing() {
+			return (this == CLOSED || this == CLOSING);
+		}
+
+		/**
+		 * Is the door in motion
+		 * 
+		 * @return door in motion
+		 */
+		public boolean inMotion() {
+			return (this == OPENING || this == CLOSING);
+		}
+
+		/**
+		 * Lookup a door status by its int value
+		 * 
+		 * @param value
+		 * @return a door status enum
+		 */
+		public static GarageDoorStatus GetDoorStatus(int value) {
+			for (GarageDoorStatus ds : values()) {
+				if (ds.getValue() == value)
+					return ds;
+			}
+			return UNKNOWN;
+		}
+	}
+
+	public String toString() {
+		return this.deviceName;
+	}
+}
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/InvalidDataException.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/InvalidDataException.java
@@ -1,0 +1,18 @@
+package org.openhab.binding.myq.internal;
+
+import java.io.IOException;
+
+/**
+ * Throw if the data we are parsing in not what we are expecting for input.
+ * 
+ * @author Dan Cunningham
+ *
+ */
+public class InvalidDataException extends IOException {
+
+	private static final long serialVersionUID = 1L;
+
+	public InvalidDataException(String message) {
+		super(message);
+	}
+}
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/LoginData.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/LoginData.java
@@ -1,0 +1,58 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.myq.internal;
+
+import java.io.IOException;
+import java.util.Map;
+
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This Class parses the JSON data and throws if the login request was not
+ * successful.
+ * <ul>
+ * <li>securityToken: securityToken from login request</li>
+ * </ul>
+ * 
+ * @author Scott Hanson
+ * @since 1.8.0
+ */
+public class LoginData {
+	static final Logger logger = LoggerFactory.getLogger(LoginData.class);
+
+	String securityToken;
+
+	/**
+	 * Constructor of the LoginData.
+	 * 
+	 * @param loginData
+	 *            The Json node from the myq website.
+	 */
+	@SuppressWarnings("unchecked")
+	public LoginData(JsonNode root) throws IOException, InvalidDataException {
+		ObjectMapper mapper = new ObjectMapper();
+		Map<String, Object> treeData = mapper.readValue(root, Map.class);
+		Object data = treeData.get("SecurityToken");
+		if (data == null)
+			throw new InvalidDataException(
+					"Could not find SecurityToken in JSON data");
+		securityToken = data.toString();
+		logger.debug("myq securityToken: {}", securityToken);
+	}
+
+	/**
+	 * @return Login SecurityToken
+	 */
+	public String getSecurityToken() {
+		return securityToken;
+	}
+}
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyQResponseCode.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyQResponseCode.java
@@ -1,0 +1,78 @@
+package org.openhab.binding.myq.internal;
+
+/**
+ * Enum of error codes from the MyQ API
+ * 
+ * @see http
+ *      ://chamberlain.custhelp.com/app/answers/detail/a_id/4922/~/what-are-the
+ *      -myq-error-codes%3F
+ * @author Dan Cunningham
+ *
+ */
+public enum MyQResponseCode {
+
+	OK(0,"OK"),
+	LOGIN_ERROR(-3333,
+			"Login error. Please login again."), // token has expired
+	ACCOUNT_INVALID(203,
+			"The username or password you entered is incorrect. Try again."),
+	ACCOUNT_NOT_FOUND(204,
+			"The username was not found or is locked out."),
+	ACCOUNT_LOCKED_PENDING(205,
+			"This user will be locked out."),
+	ACCOUNT_LOCKED(207,"This user is locked out."),
+	GATEWAY_OFFLINE(223,"Gateway is Offline."),
+	GATEWAY_LEARNMODE(224,"Gateway is in learn mode."),
+	DEVICE_LEARNMODE(305,"The device is currently in Learn Mode."),
+	DEVICE_NOT_RESPONDING(308,
+			"The device is not responding. Please check that the device is powered and in range."),
+	DEVICE_OFFLINE(309,
+			"The gateway or hub is offline. Please check the power and network connections."),
+	UNKNOWN(-1,"Unknow resonse");
+
+	private int code;
+	private String desc;
+
+	/**
+	 * Creates a new Response code object from the numeric code
+	 * 
+	 * @param code
+	 * @param desc
+	 */
+	private MyQResponseCode(int code, String desc) {
+		this.code = code;
+		this.desc = desc;
+	}
+
+	/**
+	 * The human readable description of the error
+	 * 
+	 * @return
+	 */
+	public String getDesc() {
+		return desc;
+	}
+
+	/**
+	 * The error code number from the API.
+	 * 
+	 * @return
+	 */
+	public int getCode() {
+		return code;
+	}
+
+	/**
+	 * Return a MyQResponseCode from a given code number
+	 * 
+	 * @param code
+	 * @return
+	 */
+	public static MyQResponseCode fromCode(int code) {
+		for (MyQResponseCode rc : values()) {
+			if (rc.getCode() == code)
+				return rc;
+		}
+		return UNKNOWN;
+	}
+}
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
@@ -1,0 +1,400 @@
+﻿/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.myq.internal;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.openhab.binding.myq.MyqBindingProvider;
+import org.openhab.binding.myq.internal.MyqBindingConfig;
+import org.openhab.binding.myq.internal.GarageDoorDevice.GarageDoorStatus;
+import org.apache.commons.lang.StringUtils;
+import org.openhab.core.binding.AbstractBinding;
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.library.types.PercentType;
+import org.openhab.core.library.types.StringType;
+import org.openhab.core.library.types.OpenClosedType;
+import org.openhab.core.library.types.UpDownType;
+import org.openhab.core.types.State;
+import org.openhab.core.types.Command;
+import org.openhab.core.types.UnDefType;
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * this class polls the Chamberlain MyQ API and sends updates to the event bus
+ * of configured items in openHAB
+ * 
+ * @author Scott Hanson
+ * @author Dan Cunningham
+ * @since 1.8.0
+ */
+public class MyqBinding extends AbstractBinding<MyqBindingProvider> {
+	private static final Logger logger = LoggerFactory
+			.getLogger(MyqBinding.class);
+
+	/**
+	 * The BundleContext. This is only valid when the bundle is ACTIVE. It is
+	 * set in the activate() method and must not be accessed anymore once the
+	 * deactivate() method was called or before activate() was called.
+	 */
+	@SuppressWarnings("unused")
+	private BundleContext bundleContext;
+
+	/**
+	 * The myqData. This object stores the connection data and makes API
+	 * requests
+	 */
+	private MyqData myqOnlineData = null;
+
+	/**
+	 * the refresh interval which is used to poll values from the myq server
+	 * (optional, defaults to 60000ms)
+	 */
+	private long refreshInterval = 60000;
+
+	/**
+	 * We use our own polling service so we can adjust the polling rate during
+	 * periods of activity when a device has been sent a command or is in motion
+	 */
+	private ScheduledExecutorService pollService = Executors
+			.newSingleThreadScheduledExecutor();
+
+	/**
+	 * The regular polling task
+	 */
+	private ScheduledFuture<?> pollFuture;
+
+	/**
+	 * This task will reset the poll interval back to normal after a rapid poll
+	 * cycle
+	 */
+	private ScheduledFuture<?> pollResetFuture;
+
+	/**
+	 * When polling quickly, how often do we poll
+	 */
+	private static int RAPID_REFRESH = 2000;
+
+	/**
+	 * Cap the time we poll rapidly to not overwhelm the servers with api
+	 * requests.
+	 */
+	private static int MAX_RAPID_REFRESH = 30 * 1000;
+
+	/**
+	 * If our login credentials are invalid then we will stop api requests until
+	 * our configuration is changed
+	 */
+	private boolean invalidCredentials;
+
+	/**
+	 * Called by the SCR to activate the component with its configuration read
+	 * from CAS
+	 * 
+	 * @param bundleContext
+	 *            BundleContext of the Bundle that defines this component
+	 * @param configuration
+	 *            Configuration properties for this component obtained from the
+	 *            ConfigAdmin service
+	 */
+	public void activate(final BundleContext bundleContext,
+			final Map<String, Object> configuration) {
+		this.bundleContext = bundleContext;
+		modified(configuration);
+	}
+
+	/**
+	 * Called by the SCR when the configuration of a binding has been changed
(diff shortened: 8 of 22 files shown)


bundles/binding/org.openhab.binding.myq/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,44 @@
+Manifest-Version: 1.0
+Private-Package: org.openhab.binding.myq.internal
+Ignore-Package: org.openhab.binding.myq.internal
+Bundle-License: http://www.eclipse.org/legal/epl-v10.html
+Bundle-Name: openHAB myq Binding
+Bundle-SymbolicName: org.openhab.binding.myq
+Bundle-Vendor: openHAB.org
+Bundle-Version: 1.0.0.qualifier
```
watou: Change to 1.8.0.qualifier

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/MyqBindingProvider.java
```diff
@@ -0,0 +1,41 @@
...
+	 */
+	public MyqBindingConfig getItemConfig(String itemName);
+
+	/**
+	 * Returns a list of item names
+	 * 
+	 * @return List of item names mapped to a Chamberlain myQ Binding
+	 */
+	public List<String> getInBindingItemNames();
+}
```
watou: please add newline

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorData.java
```diff
@@ -0,0 +1,89 @@
...
+				for (int i = 0; i < arraysize; i++) {
+					int deviceId = node.get(i).get("MyQDeviceId").asInt();
+					String deviceName = node.get(i).get("SerialNumber")
+							.asText();
+					String deviceType = node.get(i).get("MyQDeviceTypeName")
+							.asText();
+
+					if (deviceType.contains("Garage")
+							&& deviceType.contains("Door")
+							&& deviceType.contains("Opener")) {
```
watou: Suggest instead using a static final java.util.regex.Pattern to match these three strings in case they change later.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorDevice.java
```diff
@@ -0,0 +1,127 @@
...
+					return ds;
+			}
+			return UNKNOWN;
+		}
+	}
+
+	public String toString() {
+		return this.DeviceName;
+	}
+}
```
watou: please add newline

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/MyqBindingProvider.java
computergeek1507: @watou sorry I do not understand where you want the newline, at the end of the file?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/LoginData.java
```diff
@@ -0,0 +1,58 @@
...
+	 * @param loginData
+	 *            The Json string as it has been returned myq website.
+	 */
+	@SuppressWarnings("unchecked")
+	public LoginData(JsonNode root) throws IOException {
+		ObjectMapper mapper = new ObjectMapper();
+		Map<String, Object> treeData = mapper.readValue(root, Map.class);
+		Object data = treeData.get("SecurityToken");
+		if(data == null)
+			throw new IOException("Could not find SecurityToken in JSON data");
```
watou: Create a new subclass of Exception instead of using IOException here, since it's not an I/O issue.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/LoginData.java
```diff
@@ -0,0 +1,58 @@
...
+		logger.debug("myq securityToken: {}", securityToken);
+	}
+
+	/**
+	 * @return Login SecurityToken
+	 */
+	public String getSecurityToken() {
+		return securityToken;
+	}
+}
```
watou: please add newline

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyQResponseCode.java
```diff
@@ -0,0 +1,66 @@
...
+	 public int getCode(){
+		 return code;
+	 }
+	 
+	 /**
+	  * Return a MyQResponseCode from a given code number
+	  * @param code
+	  * @return
+	  */
+	 public static MyQResponseCode FromCode(int code){
```
watou: Please use `fromCode` method naming style.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/MyqBindingProvider.java
watou: Yes, at the end of the file, thanks!

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorData.java
digitaldan: I'll get this one

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/LoginData.java
digitaldan: I'll get this one

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyQResponseCode.java
digitaldan: I can grab this one

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/LoginData.java
digitaldan: @watou I could make a custom exception for this, but nothing needs to do anything different if the response is malformed VS something else that throws an IOException deeper down the network stack.  By adding a new exception type I would be requiring other parts of the code to catch additional exception types that they don't care about.  This seemed to be the cleanest way to me as I considered it an data input problem, is there something else that could work?

watou: I think it would improve readability to formally use an exception type that represents the class of exception this is, which could be about malformed JSON.  I worry that higher catch blocks could take an IOException at face value that there are network issues, when in fact the I/O is fine but the JSON was not as expected.  For example, if a Java library class unexpectedly threw an IOException when there way no I/O problem, a catch block could potentially tear down and reconnect a socket that was not the problem at all.  Can higher catch blocks catch any shared parent class like Exception?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,369 @@
...
+ * this class polls the Chamberlain MyQ API and sends updates to the event bus
+ * of configured items in openHAB
+ * 
+ * @author Scott Hanson
+ * @author Dan Cunningham
+ * @since 1.8.0
+ */
+public class MyqBinding extends AbstractBinding<MyqBindingProvider> {
+	private static final Logger logger = LoggerFactory
+			.getLogger(MyqBinding.class);
```
watou: Please the openHAB Eclipse formatter over all Java files (120 max line length, etc.)

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,369 @@
...
+
+	/**
+	 * The BundleContext. This is only valid when the bundle is ACTIVE. It is
+	 * set in the activate() method and must not be accessed anymore once the
+	 * deactivate() method was called or before activate() was called.
+	 */
+	@SuppressWarnings("unused")
```
watou: Is this still needed?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,369 @@
...
+		if (pollResetFuture != null && !pollResetFuture.isCancelled())
+			pollResetFuture.cancel(true);
+	}
+
+	/**
+	 * Poll for device changes
+	 */
+	private void poll() {
+		if (invalidCredentials || this.myqOnlineData == null) {
+			return;
```
watou: Might be a good spot for a `logger.debug` message.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBindingConfig.java
```diff
@@ -0,0 +1,46 @@
...
+	 * <li>ContactStatus</li>
+	 * </ul>
+	 */
+	public enum ITEMTYPE {
+		Switch, Rollershutter, StringStatus, ContactStatus
+	};
+
+	int deviceIndex;
+	ITEMTYPE type;
+}
```
watou: please add newline

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBindingConfig.java
```diff
@@ -0,0 +1,46 @@
...
+	 * <li>StringStatus</li>
+	 * <li>ContactStatus</li>
+	 * </ul>
+	 */
+	public enum ITEMTYPE {
+		Switch, Rollershutter, StringStatus, ContactStatus
+	};
+
+	int deviceIndex;
+	ITEMTYPE type;
```
watou: Instead of an `ITEMTYPE` enum, just keep the `Class<? extends Item>` of the item, so you're not creating an enum that partially mirrors the various item types in openHAB.

watou: ...or possibly better, save a copy of `item.getAcceptedDataTypes()`, and when it's time to post an update, loop through the list until you find a type you support.  That way you're independent of which kind of item it is you're posting to.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,222 @@
...
+
+	/**
+	 * Send Command to open/close garage door opener with MyQ API Returns false
+	 * if return code from API is not correct or connection fails
+	 * 
+	 * @param deviceID
+	 *            MyQ deviceID of Garage Door Opener.
+	 * @param state
+	 *            Desired state to put the door in, 1 = open, 0 = closed
+	 * @param attemps
```
watou: please correct

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}
+
+	@SuppressWarnings("serial")
+	public class InvalidLoginException extends Exception {
+		public InvalidLoginException(String message) {
+			super(message);
+		}
+	}
+}
```
watou: please add newline

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,222 @@
...
+				throw new IOException("Request Failed: " + rc.getDesc());
+			}
+
+		} catch (JsonProcessingException e) {
+			throw new IOException("Could not parse response", e);
+		}
+	}
+
+	@SuppressWarnings("serial")
+	public class InvalidLoginException extends Exception {
```
watou: should this be a static nested class instead of an inner class?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqGenericBindingProvider.java
```diff
@@ -0,0 +1,110 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public String getBindingType() {
+		return "myq";
+	}
+
+	/**
+	 * @{inheritDoc
```
watou: should be `{@inheritDoc}`

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqGenericBindingProvider.java
```diff
@@ -0,0 +1,110 @@
...
+
+		if (item instanceof SwitchItem) {
+			config.type = ITEMTYPE.Switch;
+		} else if (item instanceof RollershutterItem) {
+			config.type = ITEMTYPE.Rollershutter;
+		} else if (item instanceof ContactItem) {
+			config.type = ITEMTYPE.ContactStatus;
+		} else if (item instanceof StringItem) {
+			config.type = ITEMTYPE.StringStatus;
+		}
```
watou: As mentioned before, instead of `ITEMTYPE`, consider
```
config.acceptedDataTypes = new ArrayList<Class<? extends State>>(item.getAcceptedDataTypes());
```
and stay unaware of the different item classes altogether.  Just post an update by looping the array and finding a type you support.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqGenericBindingProvider.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	public List<String> getInBindingItemNames() {
+		List<String> inBindings = new ArrayList<String>();
+		for (String itemName : bindingConfigs.keySet()) {
+			inBindings.add(itemName);
+		}
+		return inBindings;
+	}
+}
```
watou: please add newline

distribution/openhabhome/configurations/openhab_default.cfg
```diff
@@ -2180,3 +2180,12 @@ tcp:refreshinterval=250
...
 # ucprelayboard:refresh=
+
+############################## Chamberlain MyQ Binding ##############################
+#
+# Data refresh interval in ms (optional, defaults to 60000)
+# myq:refresh=60000
+
+# Chamberlain MyQ Username and Password
+# myq:username=
+# myq:password=
```
watou: please add newline

computergeek1507:
@watou I think @digitaldan and I have integrated most comments you suggested. Please let me know if we missed something. Thanks.

watou:
Thanks very much for the binding.  I know lots of people use/used the Vera plugin to talk to the myq garage door opener, so this should be pretty popular!

I would like to take one more look in the next day or two, and thanks for the changes!

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorData.java
```diff
@@ -0,0 +1,81 @@
...
+				logger.trace("Chamberlain MyQ Devices:");
+				int arraysize = node.size();
+				for (int i = 0; i < arraysize; i++) {
+					int deviceId = node.get(i).get("MyQDeviceId").asInt();
+					String deviceName = node.get(i).get("SerialNumber")
+							.asText();
+					String deviceType = node.get(i).get("MyQDeviceTypeName")
+							.asText();
+
+					if (deviceType.matches("Garage\\s?Door\\s?Opener")) {
```
watou: This is expensive compared to having
```
private static final Pattern DEVICE_TYPE_PATTERN = Pattern.compile("Garage\\s?Door\\s?Opener");
//...
if (DEVICE_TYPE_PATTERN.matcher(deviceType).matches()) { 
//...
```

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBindingConfig.java
```diff
@@ -0,0 +1,30 @@
...
+ * <li>The type is the item type</li>
+ * <ul>
+ * 
+ * @author Scott Hanson
+ * @since 1.8.0
+ */
+public class MyqBindingConfig implements BindingConfig {
+
+	int deviceIndex;
+	Item type;
```
watou: Bindings should avoid holding onto `Item` objects because their lifecycle is independent of bindings.  I would replace this member with
```
List<Class<? extends State>> acceptedDataTypes;
```

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqGenericBindingProvider.java
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/**
+	 * Parse item type to see what the action is used for
+	 */
+	private MyqBindingConfig parseBindingConfig(Item item, String bindingConfig)
+			throws BindingConfigParseException {
+		final MyqBindingConfig config = new MyqBindingConfig();
+
+		config.type = item;
```
watou: Recommend
```
config.acceptedDataTypes = new ArrayList<Class<? extends State>>(item.getAcceptedDataTypes());
```

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,374 @@
...
+			GarageDoorData garageStatus = myqOnlineData.getGarageData();
+
+			for (MyqBindingProvider provider : providers) {
+				for (String mygItemName : provider.getInBindingItemNames()) {
+					MyqBindingConfig deviceConfig = getConfigForItemName(mygItemName);
+
+					if (deviceConfig != null) {
+						GarageDoorDevice garageopener = garageStatus
+								.getDevice(deviceConfig.deviceIndex);
+						if (garageopener != null) {
```
watou: Recommend
```
State newState = UnDefType.UNDEF;
for (Class<? extends State> type : deviceConfig.acceptedDataTypes) {
  if (StringType.class == type) {
    // newState = new StringType...
    break;
  }
  else if (OpenClosedType.class == type) {
    // newState = OpenClosedType...
    break;
  }
  else {
    // etc...
  }
}
eventPublisher.postUpdate(mygItemName, newState);
```
this way you don't care about the kind of item, only the state types it supports.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,207 @@
...
+public class MyqData {
+	static final Logger logger = LoggerFactory.getLogger(MyqData.class);
+
+	private String userName;
+	private String password;
+	private String sercurityTokin;
+
+	private Properties header;
+	private final String webSite = "https://myqexternal.myqdevice.com";
+	private final String appId = "JVM/G9Nwih5BwKgNCjLxiFUQxQijAebyyg8QUHr7JOrP+tuPb8iHfRHKwTmDzHOu";
```
watou: Is there any risk that some other API user could use and abuse this appId, as opposed to having the binding user register his/her own app and add it to configuration in openhab.cfg?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,207 @@
...
+ * @author Scott Hanson
+ * @author Dan Cunningham
+ * @since 1.8.0
+ */
+public class MyqData {
+	static final Logger logger = LoggerFactory.getLogger(MyqData.class);
+
+	private String userName;
+	private String password;
+	private String sercurityTokin;
```
watou: `securityToken` possibly? :)

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,207 @@
...
+
+		logger.debug("Requsting URL {}", url);
+
+		String dataString = executeUrl(method, url, header,
+				payload == null ? null : IOUtils.toInputStream(payload),
+				payloadType, 5000);
+
+		logger.debug("Received MyQ  JSON: {}", dataString);
+
+		if (dataString == null)
```
watou: Always use `{` and `}` even for one-liners.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqGenericBindingProvider.java
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void processBindingConfiguration(String context, Item item,
+			String bindingConfig) throws BindingConfigParseException {
+		super.processBindingConfiguration(context, item, bindingConfig);
```
watou: move to end of method in case your code throws a `BindingConfigParseException`, in which case we don't want to pollute the internal openHAB data structures with the bad data.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,374 @@
...
+	 * @{inheritDoc
+	 */
+	@Override
+	public void internalReceiveCommand(String itemName, Command command) {
+		super.internalReceiveCommand(itemName, command);
+		logger.debug("MyQ binding received command '{}' for item '{}'",
+				command, itemName);
+		if (myqOnlineData != null) {
+			computeCommandForItem(command, itemName);
+		}
```
watou: Might want to log a WARN if you are going to drop the command due to no `myqOnlineData`.

watou:
Please add to `distribution/pom.xml`.  I'm going to tag this PR awaiting-feedback, but if other maintainers would add review comments, please untag it while reviewing and retag after.  Thanks!

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorData.java
digitaldan: yep

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
digitaldan: Unfortunately we have to use an existing app id that's out in the wild.  This is also what the smart things plugin is doing and I'm assuming the vera plugin as well.  Not ideal, I know.

digitaldan:
@watou thanks for spending the time to thoroughly review the binding.  @computergeek1507 I have a PR around the regex pattern, I'll leave the rest up to you so we don't step on each others toes ;-)   If you want me to fix anything else please let me know!

computergeek1507:
@watou I hope my latest commit is what you were thinking with the "acceptedDataTypes" stuff. @digitaldan Can you verify this new code still works with your setup. It works with my server but I just want another set of eyes to verify my code isn't crazy. Thanks everyone for the help.

digitaldan:
@computergeek1507 I just pulled and compiled your latest changes and all looks good, I'm out of town and its late here so I didn't actually open or close my door, but the door status is updating fine.

computergeek1507:
@digitaldan Thanks for the look. I only really changed the status code so as long as it working I think it's good.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/GarageDoorDevice.java
```diff
@@ -0,0 +1,127 @@
...
+ * 
+ * @author Scott Hanson
+ * @author Dan Cunningham
+ * @since 1.8.0
+ */
+class GarageDoorDevice {
+	private int DeviceId;
+	private String DeviceType;
+	private String DeviceName;
+	private GarageDoorStatus Status;
```
watou: variable names should not begin with upper-case letters

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/InvalidDataException.java
```diff
@@ -0,0 +1,17 @@
+package org.openhab.binding.myq.internal;
+
+import java.io.IOException;
+
+/**
+ * Throw if the data we are parsing in not what we are expecting for input. 
+ * @author Dan Cunningham
+ *
+ */
+public class InvalidDataException extends IOException {
```
watou: Is `InvalidDataException` conceptually a subclass of `IOException`?  Since the transport guarantees that delivered data is not corrupted due to network problems, it seems to me that invalid data could not be the result of an I/O problem.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,381 @@
...
+										break;
+									}
+								} else if (PercentType.class == type) {
+									if (garageopener.getStatus() != GarageDoorStatus.UNKNOWN) {
+										newState = new PercentType(50);
+										break;
+									}
+								}
+							}
+							if (newState != UnDefType.UNDEF) {
```
watou: Minor point: Why not post the UNDEF update?  That would give feedback to the user that the binding is explicitly reporting "I don't know how to update the item you've bound to me."

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,381 @@
...
+									if (garageopener.getStatus() == GarageDoorStatus.CLOSED) {
+										newState = UpDownType.DOWN;
+										break;
+									} else if (garageopener.getStatus() == GarageDoorStatus.OPEN) {
+										newState = UpDownType.UP;
+										break;
+									}
+								} else if (PercentType.class == type) {
+									if (garageopener.getStatus() != GarageDoorStatus.UNKNOWN) {
+										newState = new PercentType(50);
```
watou: Just curious: How does a 50% update derive from the garage door status?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,208 @@
...
+public class MyqData {
+	static final Logger logger = LoggerFactory.getLogger(MyqData.class);
+
+	private String userName;
+	private String password;
+	private String sercurityToken;
+
+	private Properties header;
+	private final String webSite = "https://myqexternal.myqdevice.com";
+	private final String appId = "JVM/G9Nwih5BwKgNCjLxiFUQxQijAebyyg8QUHr7JOrP+tuPb8iHfRHKwTmDzHOu";
```
watou: Is this safe to hard-code an app ID in the binding?  What external dependencies are implied by tying the binding code to this particular app ID?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,208 @@
...
+	 */
+	private synchronized JsonNode request(String method, String url,
+			String payload, String payloadType, boolean retry)
+			throws IOException, InvalidLoginException {
+
+		logger.debug("Requsting URL {}", url);
+
+		String dataString = executeUrl(method, url, header,
+				payload == null ? null : IOUtils.toInputStream(payload),
+				payloadType, 5000);
```
watou: Is there any chance at all that a 5-second timeout will in some case be insufficient?  If so, could this be a binding config value in openhab.cfg / myq.cfg ?

watou:
Thanks for the updates.  I've added some final comments and questions.  Among the others, I'm concerned about the hard-coded `appID` and who controls it, and if it would not be better to document in the wiki how the user is empowered to create and use their own appID via openhab.cfg / myq.cfg.  

If other maintainers want to chime in, please do.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/InvalidDataException.java
digitaldan: This is how the Jackson json parser implements its exceptions ( like JsonParseException) if the JSON is malformed or unexpected. I liked this pattern and thought it fit this use case well as the developer can be very specific about the exception they want to catch or not.

watou: sounds good, Dan -- thanks.

digitaldan:
@watou thanks, regarding the app Id... I have gone back and forth about this.  The API only allows clients with the correct app id to access the API, users can not make up there own.  Furthermore Chamberlain seems to occasionally stop known Id's from working.  I feel like we are kinda walking on egg shells with them, and we are mindful of raising attention if they start getting requests from us with random app ids.  On the other hand, if they do depreciate an id I could see it being nice for users to update the binding on their own until we can release a updated version, this would be an advanced option and I would only want users to use this temporarily.  But on the other other hand, if they have an old ID in their config but get a new version of the binding with an updated id, then we start getting into a bad state.   

That was a very long way of me saying I think we should stick with it in the binding, I could be persuaded to make it an optional config override, but I worry over time that could actually hurt users when newer version of the binding come out that require an even newer id.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
digitaldan: We know when the door is in motion or is stuck in in a half open state, we don't know exactly what percentage the door is open or closed. Roller shutter items afaik only support OPEN/CLOSE and percentage.  I put this there so that roller shutter items could display the correct state icon, so in my sitemap the  icon goes from close -> halfway open -> open.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
digitaldan: see my comment at the end

watou:
Hi Dan, at the very least I think users should be able to supply their own app ID in the config file, and not rely on one that is registered to and controlled by a third party person or company.  I also think that whatever steps you or Scott went through to obtain an app ID ought to be well documented in the wiki page (see the [Ecobee](https://github.com/openhab/openhab/wiki/Ecobee-Binding) and [Nest](https://github.com/openhab/openhab/wiki/Nest-Binding#binding-configuration) binding wiki pages by example).

> I worry over time that could actually hurt users when newer version of the binding come out that require an even newer id.

What code dependencies exist between an app ID and later versions of the binding?

Have you or Scott written a wiki page for the binding yet?

Thanks again for all the good work!

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
watou: As the `RollershutterItem` is currently defined
```
		acceptedDataTypes.add(UnDefType.class);
		acceptedDataTypes.add(UpDownType.class);
		acceptedDataTypes.add(PercentType.class);
```
the `UpDownType` appears earlier in the list of accepted data types, so I don't think that code will be hit, at least not for Rollershutter items.

watou: Never mind -- I see now that the UpDownType check will be skipped and match the PercentType test on the next loop.

digitaldan:
@watou that's part of the problem, the only way to obtain an app id has been by sniffing traffic from android or IOS clients ;-(  Sad I know.  Chamberlain is very, well....,  unfriendly with the developer community.  I had to set up a reverse ssl proxy for my phone to get a valid ID.  As far as I know there may only be a half dozen or less (just a guess) working app id's. I feel strongly we should control the app id in the binding and not put the burden on the end user to figure out which ids work.  Also the last time ( 2 weeks ago) they expired the ID we were using they also modified their api to reject requests which were missing some seemingly unimportant http headers, in this case being able to change the ID would not have helped as it also signaled a breaking change in their API. 

If you still feel strongly about it, it's not a lot of work to put in, I just worry that if a user needs to change the id, more will be needed anyway to make the binding work. And if they do change it, updating the binding which depends on a newer id may then break it for the users who over wrote the id in their config (and will need to go back and comment their change out).

digitaldan:
@watou  There is a wiki!  @computergeek1507  has a page here https://github.com/computergeek1507/openhab/wiki/Chamberlain-MyQ-Binding  which he has been keepig up to date.

watou:
@digitaldan Now I understand where you're coming from, and why this could be a moving target.  Here is my thinking about it:

1. Absolutely put an explanation similar to what you just provided above in the wiki.  Users should know that the vendor could break the binding without notice, and possibly in such a way that it can't be quickly repaired and code changes might be required.
1. Make the appID configurable from the .cfg file (or if configuration is missing, try to use the one in the code) because there is a chance that a new app ID would make the binding work again, thus avoiding the need for a code change, and the vendor may yet decide to work with 3rd party developers and expect app IDs to be kept secure by developers (or in openHAB's case, to have individual users get their own app IDs).
1. When the vendor invalidates an app ID, how are they not breaking their Android and iOS clients in the process?  Do the app clients simply fail and need an update from the app store, or do they retrieve an app ID on a secure channel, or have a time-limited list of app IDs embedded in the apps, or what exactly?  Is it possible to work similarly to the apps, assuming they work better than simply failing?

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
computergeek1507: @watou My intention is to use a switch item as a "push button" and trigger opening or closing event. Since I don't want the switch state to update I didn't add it to the if else statements and it will always be UNDEF.

watou: Why not have a Switch item track an ON or OFF state, where you give the meaning that ON means open and OFF means closed (or the reverse if you prefer).  I think it would be of benefit to be able to use a Switch item for both state and commands, and not require an additional Contact item just to be able to do both state and commands.

The CMD servlet and REST API support the TOGGLE function, the UI can give you a toggle using visibility= for two switches in the sitemap, and rules can inspect the current Switch state in order to send the opposite ON or OFF command.  So having a "no-state" switch is not necessary.  Do you disagree?

computergeek1507: I didn't know what state to put the switch in when the garage is partially open but I could just say its on for partially open.

watou: Since that is the same logic as for OpenClosedType and it makes perfect sense, I agree! :-)

digitaldan:
@watou 
1. I will update the wiki with a summary of the appid
2.  I have a PR open on @computergeek1507 repo to address the appId and timeout overrides in the config file.
3. I'm assuming they force users to update the app if the id gets retired.  To tell you the truth, I only started working on this 3 weeks ago so I am going on what I have read from reports on github projects and also over at the smarthings community ( which also has a community supported plugin).  There is not a lot of transparency on how the appid works.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,393 @@
...
+
+		if (invalidCredentials || deviceConfig == null) {
+			return;
+		}
+
+		try {
+			GarageDoorData garageStatus = myqOnlineData.getGarageData();
+			GarageDoorDevice garageopener = garageStatus
+					.getDevice(deviceConfig.deviceIndex);
+			if (garageopener != null) {
```
watou: As mentioned earlier, recommend against implementing toggle logic in binding.  So briefer code:
```
if (command.equals(OnOffType.ON) || command.equals(UpDownType.UP)) {
    myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 1);
    beginRapidPoll(true);
} else if (command.equals(OnOffType.OFF) || command.equals(UpDownType.DOWN)) {
    myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 0);
    beginRapidPoll(true);
} else {
    logger.warn("Unknown command {}", command);
}
```

computergeek1507: switch items have a toggle command and I don't know what type that will be.

computergeek1507:
@watou I hope I have changed everything you have made recommendations on. @digitaldan I updated the wiki to explain all the optional parameters better. If you have any more edits, please feel free.

digitaldan:
@computergeek1507 the explanation on the wiki looks great, thanks for updating it!

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,389 @@
...
+								}  else if (OnOffType.class == type) {
+									if (garageopener.getStatus() == GarageDoorStatus.CLOSED) {
+										newState = OnOffType.OFF;
+										break;
+									} else {
+										newState = OnOffType.ON;
+										break;
+									}
+								} else if (PercentType.class == type) {
+									if (garageopener.getStatus() != GarageDoorStatus.UNKNOWN) {
```
digitaldan: I don't think this is right, it should be
```
} else if (PercentType.class == type) {
									if(garageopener.getStatus() == GarageDoorStatus.OPEN){
										newState = new PercentType(0);
									} else if (garageopener.getStatus() == GarageDoorStatus.CLOSED){
										newState = new PercentType(100);
									} else if (garageopener.getStatus() != GarageDoorStatus.UNKNOWN) {
										//if we are not open, closed or unknow, we are partially opened
										newState = new PercentType(50);
										break;
									}
								} 
```

digitaldan: That didn't format as I thought it would, let me do a quick pr ;-)

computergeek1507: like this:
if (PercentType.class == type) {
									if (garageopener.getStatus() == GarageDoorStatus.CLOSED) {
										newState = PercentType.HUNDRED;
										break;
									} else if (garageopener.getStatus() == GarageDoorStatus.OPEN) {
										newState = PercentType.ZERO;
										break;
									}
									else if (garageopener.getStatus() != GarageDoorStatus.UNKNOWN) {
										newState = new PercentType(50);
										break;
									}
								}

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,400 @@
...
+										newState = OnOffType.OFF;
+										break;
+									} else {
+										newState = OnOffType.ON;
+										break;
+									}
+								} else if (PercentType.class == type) {
+									switch (garageopener.getStatus()) {
+									case OPEN:
+										newState = new PercentType(0);
```
watou: `PercentType.ZERO` would mean one less object creation.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,400 @@
...
+										break;
+									case CLOSED:
+										newState = new PercentType(100);
```
watou: `PercentType.HUNDRED` would mean one less object creation.

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -0,0 +1,400 @@
...
+		} catch (InvalidLoginException e) {
+			logger.error("Could not log in, please check your credentials.", e);
+			invalidCredentials = true;
+		} catch (IOException e) {
+			logger.error("Could not connect to MyQ service", e);
+		}
+	}
+
+	/**
+	 * @{inheritDoc
```
watou: Use `{@inheritDoc}`

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqData.java
```diff
@@ -0,0 +1,229 @@
...
+	 * Retrieves garage door device data from myq website, throws if connection
+	 * fails or user login fails
+	 * 
+	 */
+	public GarageDoorData getGarageData() throws InvalidLoginException,
+			IOException {
+		logger.debug("Retreiveing door data");
+		String url = String.format(
+				"%s/api/v4/userdevicedetails/get?appId=%s&SecurityToken=%s",
+				WEBSITE, appId, getSecurityToken());
```
watou: Shouldn't `appId` and `SecurityToken` values be URL-encoded (or any URL parameters that could potentially contain illegal characters elsewhere in this source file)?  I see the default app ID has `/` and `+` in them, for example.  The appId might be only accidentally working at present.

watou:
I think we're just about there.  Thank you for all your efforts in contributing this new binding.  Could you take a final sweep through for any remaining small issues, make sure the Eclipse formatter has been applied to all Java files, rebase against master to resolve conflicts, do final testing, and squash into one commit?

teichsta:
> I think we're just about there. Thank you for all your efforts in contributing this new binding. Could you take a final sweep through for any remaining small issues, make sure the Eclipse formatter has been applied to all Java files, rebase against master to resolve conflicts, do final testing, and squash into one commit?

+1

computergeek1507:
@watou @teichsta I used the the Eclipse formatter and rebase against master. I have been running the binding on my main openHAB 1.7.1 server with zero issues for the last couple weeks and I have also tested it with the openHAB 2.0 snapshot. I squash my commits and have made a wiki here:
https://github.com/computergeek1507/openhab/wiki/Chamberlain-MyQ-Binding

If there is anything else I need to do, please let me know.

Thanks!

teichsta:
great thanks! @watou would you like to go ahead an merge?

watou:
All merged.  Could @computergeek1507 or @digitaldan create the wiki page in the openhab/openhab/wiki, hook it into the sidebar, and add a note in the wiki page that the binding is available as of openHAB 1.8?  Thanks for all the hard work and persistence through the review process!

teichsta:
> Thanks for all the hard work and persistence through the review process!

+1 great work, guys!

computergeek1507:
@watou Wiki Page is created 
https://github.com/openhab/openhab/wiki/Chamberlain-MyQ-Binding

watou:
Very nicely done!  Thank you.

PR #3541, opened 2015-12-13
[SAPP binding] bug: some events from device are not managed

paolodenti:
Some events do not trigger eventPublisher update


hakan42:
Paolo knows his binding best and is probably one of the few people having the hardware to really test this.

Codewise, LGTM.

paolodenti:
thanks @hakan42

wtrucci:
Tested and the problem is solved.

Thanks @paolodenti @hakan42

PR #3549, opened 2015-12-14
a few bug fixes to make the addons work on openHAB 2 (Karaf)

kaikreuzer:

### DIFF
--- a/bundles/action/org.openhab.action.xmpp/META-INF/MANIFEST.MF
+++ b/bundles/action/org.openhab.action.xmpp/META-INF/MANIFEST.MF
@@ -21,7 +21,13 @@ Import-Package: org.apache.commons.io,
  org.osgi.framework,
  org.osgi.service.cm,
  org.osgi.service.component,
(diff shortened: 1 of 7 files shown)


maggu2810:
Seems your formatter does not work.

kaikreuzer:
It worked and I had to disable it, because it changed the complete files. Remember that we are here at openHAB 1, where hardly any code complies to the openHAB 2 formatter!

maggu2810:
> we are here at openHAB 1

Didn't realize that.

teichsta:
since we didn't require any special formatter other than the Eclipse build-in one. Shall we apply the formatter to all files once? And require the ESH formatter after that?

kaikreuzer:
> Shall we apply the formatter to all files once? And require the ESH formatter after that?

Probably after the 1.8 release, when we remove the runtime & designer code and switch to the openHAB 2 IDE.

teichsta:
Thanks, Kai!

PR #3550, opened 2015-12-15
add ct-32 to zwave database

airix1:
Add ct-32 to zwave database

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -1201,14 +1201,6 @@
 				<Type>2002</Type>
 				<Id>0100</Id>
 			</Reference>
-			<Reference>
-				<Type>2002</Type>
-				<Id>0101</Id>
-			</Reference>
-			<Reference>
-				<Type>0002</Type>
-				<Id>0100</Id>
-			</Reference>
 			<Model>CT32</Model>
 			<Label lang="en">Z-Wave Thermostat</Label>
 			<ConfigFile>rtc/ct32.xml</ConfigFile>
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -1196,6 +1196,23 @@
 			<Label lang="en">Z-Wave Thermostat</Label>
 			<ConfigFile>rtc/ct100.xml</ConfigFile>
 		</Product>
+		<Product>
+			<Reference>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/database/products.xml
```diff
@@ -1196,6 +1196,23 @@
 			<Label lang="en">Z-Wave Thermostat</Label>
 			<ConfigFile>rtc/ct100.xml</ConfigFile>
 		</Product>
+		<Product>
```
cdjackson: How do you know all three of these codes are for this device? Generally, I would prefer unless you're absolutely sure (ie have tested) that these codes are all associated with this device, that we just use the one(s) you have tested. Otherwise, if there's a mistake it's hard to work out what's wrong if someone later comes along with a device using these codes.

bundles/binding/org.openhab.binding.zwave/database/rtc/ct32.xml
```diff
@@ -0,0 +1,221 @@
...
+			<Item>
+				<Value>1</Value>
+				<Label lang="en">Function is enabled.</Label>
+			</Item>
+		</Parameter>
+	</Configuration>
+
+	<Associations>
+		<Group>
+			<Index>1</Index>
```
cdjackson: Since this group looks like it's to advise the controller about state, please add the SetToController tag to true.

cdjackson:
Thanks for producing the file - just a couple of small comments and I'll merge.

Cheers
Chris

bundles/binding/org.openhab.binding.zwave/database/products.xml
airix1: You are probably right, may be errors in the pepper db. The pepper db shows 2 versions (http://www.pepper1.net/zwavedb/device/701 and http://www.pepper1.net/zwavedb/device/764) for the US and EU radio versions.  Those are the numbers I used.  I've only tested one, the US version but it matches the EU product id..
However, every now and again mine shows up as id 100, type 002 for some reason during initialisation which is why that was added - but that may be another story.

bundles/binding/org.openhab.binding.zwave/database/rtc/ct32.xml
airix1: Done.

cdjackson:
I suspect you need to push your changes still as the commit history hasn't changed, and, well, the files haven't changed :)

airix1:
Changes pushed

cdjackson:
Thanks.

PR #3551, opened 2015-12-15
Fixes #3454: Use userdata for token path when running in OH2

### LINKED ISSUE #3454: Twitter action is not compatible with openHAB 2
See https://community.openhab.org/t/twitter-binding-in-openhab2/3149.

The exceptions seems to suggest that the action tries to store a token in the etc folder - but this does not exist in openHAB 2. It should instead persist it in the userdata folder, like other addons do it, see e.g. https://github.com/openhab/openhab/blob/master/bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java#L317

hakan42:


watou:
Will this continue to work in OH1 without losing track of where Twitter tokens are already stored in existing OH1 installations?

kaikreuzer:
@watou Yes, this is the path for OH1: https://github.com/openhab/openhab/pull/3551/files#diff-caaecd5047ca8612d61565fea44d101cR187

@hakan42 lgtm, thanks!

hakan42:
Yes, as the directory detection depends on the system property "smarthome.userdata" to be existent.

PR #3553, opened 2015-12-16
[Telegram action] enhancement: added send photo capability

paolodenti:
Added method to send pictures to Telegram.
Pictures are downloaded from a URL and sent to Telegram
An optional caption can be added to the picture.

Can be used to send surveillance systems screenshots.

Usage example
```
rule "Send telegram with image"
when
    Item Light_GF_Living_Table changed
then
    sendTelegramPhoto("bot1", "http://www.openhab.org/assets/images/openhab-logo-top.png", "sent from openHAB")
end
```
The third  parameter (caption) can be null (no caption is sent)

### DIFF
--- a/bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
+++ b/bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
@@ -8,12 +8,18 @@
  */
 package org.openhab.action.telegram.internal;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
+
+import javax.imageio.ImageIO;
+import javax.imageio.ImageReader;
+import javax.imageio.stream.ImageInputStream;
 
 import org.apache.commons.httpclient.DefaultHttpMethodRetryHandler;
 import org.apache.commons.httpclient.Header;
@@ -22,7 +28,13 @@ import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.HttpException;
 import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.httpclient.NameValuePair;
+import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.commons.httpclient.methods.PostMethod;
+import org.apache.commons.httpclient.methods.multipart.ByteArrayPartSource;
+import org.apache.commons.httpclient.methods.multipart.FilePart;
+import org.apache.commons.httpclient.methods.multipart.MultipartRequestEntity;
+import org.apache.commons.httpclient.methods.multipart.Part;
+import org.apache.commons.httpclient.methods.multipart.StringPart;
 import org.apache.commons.httpclient.params.HttpMethodParams;
 import org.apache.commons.io.IOUtils;
 import org.openhab.core.scriptengine.action.ActionDoc;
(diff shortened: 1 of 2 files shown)


bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
```diff
@@ -130,4 +143,107 @@ static public boolean sendTelegram(@ParamDoc(name = "group") String group,
...
+		try {
+			postMethod.getParams().setContentCharset("UTF-8");
+			postMethod.getParams().setSoTimeout(HTTP_TIMEOUT);
+			postMethod.getParams().setParameter(HttpMethodParams.RETRY_HANDLER,
+					new DefaultHttpMethodRetryHandler(3, false));
+			Part[] parts = new Part[caption != null ? 3 : 2];
+			parts[0] = new StringPart("chat_id", groupTokens.get(group)
+					.getChatId());
+			parts[1] = new FilePart("photo", new ByteArrayPartSource(
+					String.format("image.%s", imageType), imageFromURL));
```
watou: Suggest something like what I added to the mail action in order to use the actual file name from the URL, or fall back to a generic name:
```
String fileName = photoURL.replaceFirst(".*/([^/?]+).*", "$1");
fileName = StringUtils.isNotBlank(fileName) ? fileName : String.format("image.%s", imageType);
```

paolodenti:
@watou 
If you are suggesting to get the name from the url, in my opinion this is not a good idea.
The url usually is something like http://..../get.php?date=11122015&height=24&width=250&res=high&....
In general the content type of the image is sent along the http headers and not in the name itself.

My code makes an analysis of the actual bytes received in order to guess the content type received, so that the telegram is happy with the file extension. And the name itself is not used by telegram apart from getting the extension, it is there just because required from the ByteArrayPartSource.

So in my opinion my approach is fine.

bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
paolodenti: @watou 
If you are suggesting to get the name from the url, in my opinion this is not a good idea.
The url usually is something like http://..../get.php?date=11122015&height=24&width=250&res=high&....
In general the content type of the image is sent along the http headers and not in the name itself.

My code makes an analysis of the actual bytes received in order to guess the content type received, so that the telegram is happy with the file extension. And the name itself is not used by telegram apart from getting the extension, it is there just because required from the ByteArrayPartSource (if you use the FilePart instead of ByteArrayPartSource a name is not even required).
If you make a "save as" from telegram desktop the actual filename proposed is not the sent one but it is something like 83456634889_1232656545353543282341.jpg

So in my opinion my approach is fine.

watou: Yes, your correct use of the file name extension to match the image type will make the generic "image" name unimportant to users.  This is a different scenario from attaching any arbitrary file to an email, where the name can actually matter.  Thanks for the detailed response.

paolodenti: thanks!

wtrucci:
Tested sending a jpg photo from my mobotix cam! Very fast usefull!!!

Thanks @paolodenti for your great work

watou:
Will a 2-second HTTP timeout be sufficient if the user attempts to send a very, very large image?  Aside from that question, this PR lgtm.

paolodenti:
@watou 
John, after your suggestion I went through he documentation and I understood that the soTimeout is not a connection timeout (as I wrongly supposed) but is it a data transfer timeout.
So, yes, it is low. I will put it to something quite higher.

Thanks for the very useful suggestion.

Do you want me to squash the 2 commits?

paolodenti:
@watou 
a different timeout for pictures has been set to 10 seconds.
Moreover a timeout was missing from http get for img download (it was left to default, no timeout), it has been set to 10 seconds, like for http post to telegram servers.

I squashed the 2 commits into 1

Thanks

watou:
Thanks @paolodenti!

PR #3555, opened 2015-12-16
Fixed session timeout problem (important for 1.8.0 release)

ollie-dev:
This commit fixes a sometimes happening timeout problem and is very important for a stable binding in OH 1.8.0. So please include it in the release.


teichsta:
Thanks @ollie-dev !

PR #3566, opened 2015-12-18
Added PersistenceService interface

derekadams:
Added PersistenceService interface since the store method is not called if it is not included. This may also cause problems with other persistence plugins.


teichsta:
are you really sure? Since QueryableService extends PersistenceService i am wondering if this can be case?

derekadams:
I would have assumed the same thing, but I have tried it both ways a number of times and it does cause a problem. I'm definitely not an OSGI expert, but my guess is that it's not actually introspecting the class to see if the interface is implemented and it's not smart enough to infer it through inheritance. Again, this is all just speculation. I am 100% sure that removing the line causes the plugin to quit working though.

watou:
The persistence bundles caldav, db4o, influxdb, jdbc, and jpa all provide both `QueryablePersistenceService` and `PersistenceService` in their service components, while all the other persistence bundles only provide the `PersistenceService` interface, even though many of them can be queried.  It would therefore seem like there is an inconsistency across all of these bundles in how they expose the interfaces they implement.  @teichsta, what if anything do you recommend?

watou:
I don't see in the code any reason for the
```xml
<provide interface="org.openhab.core.persistence.QueryablePersistenceService"/>
```
line.  I don't think the the persistence manager deals in those, only `PersistenceService`s.  Does anyone know of a case where the SCR deals in `QueryablePersistenceService`s?

cdjackson:
I would think it’s reasonable that PersistenceService should always be used - it is a ‘persistence service’ service after all.

teichsta:
@derekadams could you please check whether removing `<provide interface="org.openhab.core.persistence.QueryablePersistenceService"/>` works, too?

teichsta:
any update @derekadams?

derekadams:
Sorry for the late response @teichsta . I was out of town for the holiday. I removed the QueryablePersistenceService line and everything still works fine. Should I commit the change?

teichsta:
yes, please!

derekadams:
Ok, it should be good to go.

teichsta:
> Ok, it should be good to go.

yes, almost … would you please finally squash the commits into one?

teichsta:
squashed and merged manually (see https://github.com/openhab/openhab/commit/4adf1986795412f572a6fea035ef49a267107a9b)

PR #3568, opened 2015-12-20
[ecobee] Added quick poll after update, command or action.

watou:
*Summary:* added optional `ecobee:granularity` and `ecobee:quickpoll` config properties; polling the Ecobee API will occur much sooner after the binding has sent an update, command or action call to Ecobee in order to give quicker feedback to the binding.

Changed the polling logic in the Ecobee binding to invoke the `execute()` method based on optional `ecobee:granularity` config property (default 5000ms).  On each of those, check to see if it's time to poll the Ecobee API.  Normal polling continues to use the optional `ecobee:refresh` config property (default 180000ms).  However, after a successful update, command or action call to the Ecobee API, the next poll will occur after `ecobee:quickpoll` milliseconds have passed (optional, defaults to 6000).  So existing configurations are compatible, but the binding will perform a poll about 6-11 seconds after a successful write to the Ecobee API.

### DIFF
--- a/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
+++ b/bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
@@ -20,6 +20,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.prefs.Preferences;
 
 import org.apache.commons.beanutils.ConvertUtils;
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
```diff
@@ -238,27 +255,32 @@ protected String getName() {
...
+						logger.debug("Nothing to retrieve for '{}'; skipping thermostat retrieval.",
+								oauthCredentials.userid);
 						continue;
 					}
-				}
 
-				readEcobee(oauthCredentials, selection);
+					if (oauthCredentials.noAccessToken()) {
+						if (!oauthCredentials.refreshTokens()) {
+							logger.warn("Periodic poll skipped for '{}'.", oauthCredentials.userid);
```
teichsta: it might be helpful to the reason why polling is skipped (due to a missing token)

distribution/openhabhome/configurations/openhab_default.cfg
```diff
@@ -2050,20 +2050,29 @@ tcp:refreshinterval=250
...
 ################################ Ecobee Binding #######################################
 #
+# the private API key issued be Ecobee to use the API (required, replace with your own)
+# ecobee:appkey=9T4huoUXlT5b5qNpEJvM5sqTMgaNCFoV
+
+# the application scope used when authorizing the binding
+# choices are smartWrite,smartRead, or ems, or multiple (required, comma-separated, no spaces)
+# ecobee:scope=smartWrite
+
+# Rate at which to check if poll is to run, in ms (optional, defaults to 5000)
```
teichsta: "in ms" (add unit)

teichsta:
lgtm … just one small remark which might help debugging the binding.

bundles/binding/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/EcobeeBinding.java
watou: I rely on the `refreshTokens()

watou:
I will merge now and thanks, @teichsta.

PR #3590, opened 2015-12-23
InsteonPLM: removed comments after port configuration in openhab_default.cfg

berndpfrommer:
Several people did not remove the comment behind the port configuration line in openhab_default.cfg. Fixed this to avoid users reporting the port does not work.


robnielsen:
lgtm

teichsta:
Thanks guys!

PR #3593, opened 2015-12-23
Weather: Fixed exception if degree is not in a valid range

gerrieg:
should fix https://community.openhab.org/t/error-from-weather-binding/5305/2


watou:
Could it be that -12 could mean 348 degrees?

gerrieg:
No, this is clearly the index from the directions array. For my location the degree is 135 now for example.

teichsta:
Thanks @gerrieg!

PR #3603, opened 2015-12-28
Add ZWave configuration for Greenwave PowerNode NS-310

dominicdesu:
Signed-off-by: Dominic Lerbs <dominicdesu@hallojapan.de>

The Greenwave PowerNode NS310 seems to be a new version of the NS210. The device reports itself as NS210 (version 4), however according to the manual it's called NS310 (also see [this zwaveeurope manual](http://manuals.zwaveeurope.com/make.php?lang=en&type=mini&sku=GWRENS310-F)).
Also it seems that the configuration parameters and association groups between these two versions are quite different.

I have created this zwave device file using the NS-310 manual and tested it with an actual device (well except for some functionality like the over-current protection)

I was not sure about the NS210 devices but I assume they would use a version <=3, as the NS310 is version 4.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/greenwave/ns310.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/greenwave/ns310.xml
@@ -1,0 +1,157 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>NS310</Model>
+	<Label lang="en">Smart PowerNode</Label>
+	<CommandClasses>
+		<Class><id>0x20</id></Class>
+		<Class><id>0x25</id></Class>
+		<Class><id>0x27</id></Class>
+		<Class><id>0x32</id></Class>
+		<Class><id>0x56</id></Class>
+		<Class><id>0x70</id></Class>
+		<Class><id>0x71</id></Class>
+		<Class><id>0x72</id></Class>
+		<Class><id>0x75</id></Class>
+		<Class><id>0x85</id></Class>
+		<Class><id>0x86</id></Class>
+		<Class><id>0x87</id></Class>
+	</CommandClasses>
+	
+	<Configuration>
+		<Parameter>
+			<Index>0</Index>
+			<Type>int</Type>
+			<Minimum>0</Minimum>
+			<Maximum>100</Maximum>
+			<Size>1</Size>
+			<Label lang="en">Min. variation of load current</Label>
+			<Help lang="en">Minimum variation in load current before a message is sent. Value in percent (30 => 30%)</Help>
+		</Parameter>
+	
+		<Parameter>
+			<Index>1</Index>
+			<Type>int</Type>
+			<Default>2</Default>
+			<Minimum>-1</Minimum>
+			<Maximum>127</Maximum>
+			<Size>1</Size>
+			<Label lang="en">No communication light</Label>
+			<Help lang="en">After how many minutes the GreenWave device should start flashing if the controller didn't communicate with this device</Help>
+		</Parameter>
+	
+		<Parameter>
+			<Index>2</Index>
+			<Type>list</Type>
+			<Default>0</Default>
+			<Size>1</Size>
+			<ReadOnly>true</ReadOnly>
+			<Item>
+				<Value>-128</Value>
+				<Label lang="en">Black</Label>
+			</Item>
+			<Item>
+				<Value>-127</Value>
+				<Label lang="en">Green (1)</Label>
+			</Item>
+			<Item>
+				<Value>-126</Value>
+				<Label lang="en">Dark Blue (2)</Label>
+			</Item>
+			<Item>
+				<Value>-125</Value>
+				<Label lang="en">Red (3)</Label>
+			</Item>
+			<Item>
+				<Value>-124</Value>
+				<Label lang="en">Yellow (4)</Label>
+			</Item>
+			<Item>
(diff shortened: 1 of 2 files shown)


dominicdesu:
@cdjackson I have one issue, maybe you can help:
Trying to reset the total power comsuption shows an error in the log file, even though it actually works.

Item:
```
Switch BRPowerNodeTotalReset "Stromverbrauch Reset"    <energy>{zwave="9:command=METER,meter_reset=true", autoupdate="false" }
```

Log file extract:
```
2015-12-28 12:03:18.616 [INFO ] [runtime.busevents             ] - BRPowerNodeTotalReset received command ON
...
2015-12-28 12:03:23.615 [ERROR] [WaveController$ZWaveSendThread] - NODE 9: Timeout while sending message. Requeueing - 2 attempts left!
2015-12-28 12:03:23.622 [TRACE] [b.z.i.p.s.SendDataMessageClass] - NODE 9: Handling failed message.
2015-12-28 12:03:23.623 [ERROR] [b.z.i.p.s.SendDataMessageClass] - NODE 9: Got an error while sending data. Resending message.
```

Please see [here](https://gist.github.com/dominicdesu/1073fe1bef9fd8635a91) for the complete log.
Any ideas?

cdjackson:
This is likely caused by the device not responding to requests - it’s only sending the data due to an association configuration.

To fix this, you need to disable requests in the database with the following XML -:

<Class>
	<id>0x32</id>
        <isGetSupported>false</isGetSupported>
</Class>

cdjackson:
Are you sure about the version numbers, and the fact that the different products are using the same device ID / Type? I'm not saying it's wrong, just that it's different than I've seen in the past...

dominicdesu:
@cdjackson I tried adding the isGetSupported=false but nothing changed.
The reason is probably because the [ZWaveMeterCommandClass.getResetMessage()](https://github.com/openhab/openhab/blob/0ca751504d050b1059988f47280dc89e832e3e54/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveMeterConverter.java#L126) is called, which does not handle the isGetSupported flag.

Actually, the device responds to requests for the current power consumption and for the total power consumption. This is working fine during the refresh, for example.

The following messages are sent when triggering the reset:
1. Send METER_RESET message: fails (after 3 retries and about 15 seconds)
2. Send METER_GET for current consumption (W): success
3. Send METER_GET for total consumption (kWh): success

I digged a little deeper and saw that we receive a message of type RESPONSE after sending the METER_RESET message:
```
2015-12-28 16:58:31.895 [Thread-35] [DEBUG] [b.z.i.protocol.ZWaveController] - Message: class = SendData (0x13), type = Response (0x01), payload = 01 
2015-12-28 16:58:31.896 [Thread-35] [TRACE] [b.z.i.protocol.ZWaveController] - Incoming Message type = RESPONSE
```

There is a nice [comment in ZWaveController](https://github.com/openhab/openhab/blob/0ca751504d050b1059988f47280dc89e832e3e54/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java#L1210) explaining that the transaction might be finished after the RESPONSE is received from the controller and that the serial message class processor would set transactionCompleted. 
Currently, the [SendDataMessageClass.handleResponse()](https://github.com/openhab/openhab/blob/0ca751504d050b1059988f47280dc89e832e3e54/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/SendDataMessageClass.java#L38) method is called but nothing more is done.

Might this be related?

dominicdesu:
> Are you sure about the version numbers, and the fact that the different products are using the same device ID / Type? I'm not saying it's wrong, just that it's different than I've seen in the past...

Yes, this is somehow strange. I checked in HABmin, which reports type=2 and id=2. I assume this information is coming from the actual device? Any other way to find out the actual values? The device package has a GWRENS310-F label and the manual also is for a NS310-F ...

cdjackson:
> On 28 Dec 2015, at 17:10, Dominic Lerbs <notifications@github.com> wrote:
> 
> There is a nice comment in ZWaveController explaining that the transaction might be finished after the RESPONSE is received from the controller and that the serial message class processor would set transactionCompleted. 
> Currently, the SendDataMessageClass.handleResponse() method is called but nothing more is done.
> 
Maybe, but not directly here. If this is the problem, then the fix will be in the METER class. When you request a transaction, you also state what it is that is needed to complete the transaction and that’s what my comment is related to...

The problem is probably wth this line -:
		SerialMessage result = new SerialMessage(this.getNode().getNodeId(), SerialMessageClass.SendData,
				SerialMessageType.Request, SerialMessageClass.ApplicationCommandHandler, SerialMessagePriority.Set);


It probably shouldn’t use the ApplicationCommandHandler..

However, why is it even sending this message? It shouldn’t do this very often should it - it should only happen when you manually reset the meter?

dominicdesu:
@cdjackson you are right, this only happens when I trigger a reset of the total power consumption. So if it nothing I can do in the device file, should I create a seperate ticket for it?
As I am still experimenting with the device and can reproduce the issue, should I try to change something in the code you mentioned above and see if it is resolved?

cdjackson:
Let’s create a separate issue then we can merge this…

dominicdesu:
@cdjackson: Done

PR #3614, opened 2015-12-28
added definition of Karaf features

kaikreuzer:
Equivalent to https://github.com/openhab/openhab2/pull/580 - Karaf features are now defined, where the source code of the bundle is found as well.


features/openhab-addons-external/src/main/resources/conf/mail.cfg
```diff
@@ -0,0 +1,2 @@
+# EnOcean USB adapter serial port
+#serialPort=
```
watou: mail?

kaikreuzer: good catch! Fixed it...

PR #3617, opened 2015-12-29
Update default config for BenQ binding

plajjan:
Since we actually have a working direct serial binding for BenQ
projectors it's only appropriate we also update the default
configuration to reflect this!


hakan42:
Nice, thank you.

Issue #3579, opened 2015-12-21
Insteonplm: Feature Request: Set ramp rate for device


SwissKid:
For InteonPLM, when I send a command to it it updates immediately. However, in many cases I want to dim or brighten to that setting, not immediately switch.

First feature I'd like is being able to put the ramp rate in the insteonplm="" line.

Second feature would be having the ability to trigger it within a rule special. The "user story" I'm thinking of using this is:

    As a person, I'd like to use my insteon light as a compliment to my alarm to wake me up in the morning. 
    I'd like it to ramp up slowly over a period of at least 30 seconds, rather than immediately switch to full brightness.

This could be done by either having a second item for that same light that just had a different ramp rate, or by being able to specify the ramp rate in the rule action.

steve-bate:
Have you tried doing this with a Dimmer item and rules?

SwissKid:
Can i specify a specific ramp rate for the dimmer item anywhere? I don't see anywhere to specify anything except a max brightness.... I'd need to be able to do something like

    Dimmer d1 "dimmer 1" {insteonplm="xx.xx.xx:F00.00.11#dimmer,ramprate=20"}

And that's not possible with current configs from what I can tell. 
Currently, it's using the "on" command (though i feel as if it's using "fast on" when I set a specific level). From Insteon docs:
This command tells the device to turn its load ON to the specified On-Level at its Local Ramp Rate (Ramp
Rate only used for dimmable devices). 

I'm asking to be able to use "Ramp Rate ON"

http://cache.insteon.com/developer/2672-222dev-052014-en.pdf
page 24

This would allow me to specifically slowly dim a room in some instances.

steve-bate:
Maybe @berndpfrommer can comment, but I don't think that's currently possible directly with the binding although it would be an interesting feature (I have some of those bulbs too).

berndpfrommer:
Should not be too hard to implement. This document is pretty good:

http://www.madreporite.com/insteon/ramprate.htm

I suggest to introduce a new feature, and map it to a switch. It really is a separate command, so tying it into the "dimmer" feature is probably not a good idea. How about this syntax:

Switch rampRateSwitch "dimmer 1 ramp" {insteonplm="xx.xx.xx:F00.00.11#ramponoff,ramptime=20,ramplevel=55"}

[...]

PR #3621, opened 2015-12-29
Insteon: Added Ramp Dimmer Support

steve-bate:
See #3579. I've tested this manually with my Insteon PLM and 2672-222 bulb.

### DIFF
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
@@ -382,6 +382,74 @@ public abstract class MessageHandler {
 	}
 
 	/**
+	 * This message handler processes replies to Ramp ON/OFF commands.
+	 * Currently, it's been tested for the 2672-222 LED Bulb. Other
+	 * devices may use a different pair of commands (0x2E, 0x2F). This
+	 * handler and the command handler will need to be extended to support
+	 * those devices.
+	 */
+	public static class RampDimmerHandler extends MessageHandler {
+		private int onCmd;
+		private int offCmd;
+		
+		RampDimmerHandler(DeviceFeature p) { 
+			super(p);
+			// Can't process parameters here because they are set after constructor is invoked.
+			// Unfortunately, this means we can't declare the onCmd, offCmd to be final.
+		}
+		
+		@Override
+		public void setParameters(HashMap<String, String> params)
+		{
+			super.setParameters(params);
+			onCmd = getIntParameter("on", 0x2E);
+			offCmd = getIntParameter("off", 0x2F);
+		}
+		
+		@Override
+		public void handleMessage(int group, byte cmd1, Msg msg,
+				DeviceFeature f, String fromPort) {
+			if (isMybutton(msg, f)) {
+				if (cmd1 == onCmd) {
+					int level = getLevel(msg);
+					logger.info("{}: device {} was switched on using ramp to level {}.", 
+							nm(), f.getDevice().getAddress(), level);
+					if (level == 100) {
+						f.publish(OnOffType.ON, StateChangeType.ALWAYS);						
+					}
+					else {
+						// The publisher will convert an ON at level==0 to an OFF.
+						// However, this is not completely accurate since a ramp
+						// off at level == 0 may not turn off the dimmer completely
+						// (if I understand the Insteon docs correctly). In any case,
+						// it would be an odd scenario to turn ON a light at level == 0
+						// rather than turn if OFF.
+						f.publish(new PercentType(level), StateChangeType.ALWAYS);
+					}
+				}
+				else if (cmd1 == offCmd) {
+					logger.info("{}: device {} was switched off using ramp.", nm(),
+							f.getDevice().getAddress());
+					f.publish(new PercentType(0), StateChangeType.ALWAYS);
+				}
+			} else {
+				logger.debug("ignored message: {}", isMybutton(msg,f));
+			}
+		}
+
+		private int getLevel(Msg msg) {
+			try {
+				byte cmd2 = msg.getByte("command2");
+				return (int) Math.round(((cmd2 >> 4) & 0x0f) * (100/15d));
+			}
+			catch (FieldException e) {
+				logger.error("Can't access command2 byte", e);
+				return 0;
+			}
+		}
+	}
+
+	/**
 	 * A message handler that processes replies to queries.
 	 * If command2 == 0xFF then the light has been turned on
 	 * else if command2 == 0x00 then the light has been turned off
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/resources/device_features.xml
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/resources/device_features.xml
@@ -18,6 +18,23 @@
 	<poll-handler>NoPollHandler</poll-handler>
 </feature>
 
+<feature name="RampDimmer">
+	<message-dispatcher>DefaultDispatcher</message-dispatcher>
+	<!-- cmd1 defaults to 0x2E, 0x2F -->
+	<!-- default dispatcher uses 0x19 for lookup key instead of cmd1 -->
+	<message-handler cmd="0x19">RampDimmerHandler</message-handler>
+	<command-handler command="PercentType">RampPercentHandler</command-handler>
+	<command-handler command="OnOffType">RampOnOffCommandHandler</command-handler>
+</feature>
+
+<feature name="RampDimmer_3435">
+	<message-dispatcher>DefaultDispatcher</message-dispatcher>
+	<!-- default dispatcher uses 0x19 for lookup key instead of cmd1 -->
+	<message-handler cmd="0x19" on="0x34" off="0x35">RampDimmerHandler</message-handler>
+	<command-handler command="PercentType"  on="0x34" off="0x35">RampPercentHandler</command-handler>
+	<command-handler command="OnOffType" on="0x34" off="0x35">RampOnOffCommandHandler</command-handler>
+</feature>
+
 <feature name="ManualChange">
 	<message-dispatcher>DefaultDispatcher</message-dispatcher>
 	<message-handler cmd="0x17">StartManualChangeHandler</message-handler>
--- a/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/CommandHandler.java
+++ b/bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/CommandHandler.java
@@ -9,6 +9,8 @@
 package org.openhab.binding.insteonplm.internal.device;
 
 import java.io.IOException;
+import java.lang.Math;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Timer;
 import java.util.TimerTask;
@@ -211,6 +213,50 @@ public abstract class CommandHandler {
 				logger.error("{}: command send message creation error ", nm(), e);
 			}
 		}
+	}
+
+	public static class RampOnOffCommandHandler extends RampCommandHandler {
+		RampOnOffCommandHandler(DeviceFeature f) { 
+			super(f); 
+		}
+		
+		@Override
+		public void handleCommand(InsteonPLMBindingConfig conf, Command cmd, InsteonDevice dev) {
+			try {
+				if (cmd == OnOffType.ON) {
+					double ramptime = getRampTime(conf, 0);
+					int ramplevel = getRampLevel(conf, 100);
+					byte cmd2 = encode(ramptime, ramplevel);
+					Msg m = dev.makeStandardMessage((byte) 0x0f, getOnCmd(), (byte) cmd2,
+								s_getGroup(conf));
+					dev.enqueueMessage(m, m_feature);
+					logger.info("{}: sent ramp on to switch {} time {} level {} cmd1 {}", 
+							nm(), dev.getAddress(), ramptime, ramplevel, getOnCmd());
+				} else if (cmd == OnOffType.OFF) {
+					double ramptime = getRampTime(conf, 0);
+					int ramplevel = getRampLevel(conf, 0 /*ignored*/);
+					byte cmd2 = encode(ramptime, ramplevel);
+					Msg m = dev.makeStandardMessage((byte) 0x0f, getOffCmd(), (byte) cmd2,
+									s_getGroup(conf));
+					dev.enqueueMessage(m, m_feature);
+					logger.info("{}: sent ramp off to switch {} time {} cmd1 {}", 
+							nm(), dev.getAddress(), ramptime, getOffCmd());
+				}
+				// expect to get a direct ack after this!
+			} catch (IOException e) {
+				logger.error("{}: command send i/o error: ", nm(), e);
+			} catch (FieldException e) {
+				logger.error("{}: command send message creation error ", nm(), e);
+			}
+		}
+		
+		private int getRampLevel(InsteonPLMBindingConfig conf, int defaultValue)
+		{
+			HashMap<String, String> params = conf.getParameters();
+			return params.containsKey("ramplevel")
+					? Integer.parseInt(params.get("ramplevel"))
+					: defaultValue;
+		}		
 	}
 
 	public static class ManualChangeCommandHandler extends CommandHandler {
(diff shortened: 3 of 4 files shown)


teichsta:
@berndpfrommer could you please have a look at this one?

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
```diff
@@ -381,6 +381,29 @@ public void handleMessage(int group, byte cmd1, Msg msg,
 		}
 	}
 
+	public static class RampOnOffSwitchHandler extends MessageHandler {
```
robnielsen: Shouldn't this be RampOnOffDimmerHandler?

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
```diff
@@ -381,6 +381,29 @@ public void handleMessage(int group, byte cmd1, Msg msg,
...
+		RampOnOffSwitchHandler(DeviceFeature p) { super(p); }
```
robnielsen: Can you create separate handlers? One for off and another for on?

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
steve-bate: Please see the discussion in #3579 about dimmer versus switch.

robnielsen:
It appears that the 2672-222 ramp rate is different than other devices. This is from page 5 of http://cache.insteon.com/developer/2672-222dev-052014-en.pdf
> The Ramp Rate and relative On Level of the INSTEON LED Bulb are different from
other INSTEON dimming devices. See the Appendix for differences. 

Looking at some of the other devices that support dimming at http://www.insteon.com/developer/#devdocs, it appears they use different commands to set ramp rates.

steve-bate:
I kept the 0x34 and 0x35 registrations in case the issue with 0x19 is resolved in the future. It seems like the 0x19 registration could potentially cause more confusion since it's not a valid response to the ramp commands.

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
robnielsen: But this is not a switch, it's a dimmer.

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/device/MessageHandler.java
steve-bate: The same handler is invoked using pseudo-key 0x19 (DIRECT ACK handling in MessageDispatcher) for both 0x34 and 0x35 responses.

robnielsen: How would this work with other devices that use 0x2E for ramp on and 0x2F for ramp off? see: http://www.insteon.com/developer/#devdocs.

robnielsen: Oops, I meant http://www.madreporite.com/insteon/commands.htm.

teichsta:
> I got flagged as a robot ... tss. Hopefully this post goes through.

great, to have you back @berndpfrommer :-)

So what is the conclusion @steve-bate @robnielsen?

bundles/binding/org.openhab.binding.insteonplm/src/main/resources/device_features.xml
```diff
@@ -18,6 +18,15 @@
 	<poll-handler>NoPollHandler</poll-handler>
 </feature>
 
+<feature name="RampOnOff">
+	<message-dispatcher>DefaultDispatcher</message-dispatcher>
+	<message-handler cmd="0x19">RampOnOffSwitchHandler</message-handler>
```
robnielsen: Isn't this ignored by the rampOnOffSwitchHandler? I only see 0x34 and 0x35 in the Handler.

robnielsen:
The convention is to have separate handlers for on/off, even if the code is similar, along with using the word Dimmer for dimmer specific handlers. For example: LightOnDimmerHandler, LightOffDimmerHandler, LightOnSwitchHandler, LightOffSwitchHandler

bundles/binding/org.openhab.binding.insteonplm/src/main/resources/device_features.xml
steve-bate: Please see my earlier answer regarding 0x19 and also the discussion in the linked feature request. Message handlers are looked up using a key. Sometimes the key is cmd1 and sometimes, as in this case, it is not. Take a look at the the DIRECT ACK handling in the MessageDispatcher that I mentioned earlier. There is a comment that says cmd1 is useless for DIRECT ACK, but this is not always the case. In this scenario, the lookup key is changed to 0x19 but cmd1 remains a valid value (0x34 or 0x35). The XML message-handler entries for 0x34 and 0x35 could actually be removed from device_features.xml although the feature definition would then break if the DIRECT ACK handling is fixed in the future. In any case, the handler instance must process the cmd1 values of 0x34 and 0x35. If this is not clear, maybe @berndpfrommer can describe it better.

steve-bate:
@robnielsen, I understand there is a dominant convention although it's not always followed. For example, see SwitchRequestReplyHandler that handles both ON and OFF states. I started with the convention you described but the DIRECT ACK translation of cmd1 to 0x19 forced me to use a pattern similar to the SwitchRequestReplyHandler.

As for the dimmer versus switch discussion, I'll let you and @berndpfrommer debate that one. This PR is based on Bernd's suggested implementation.

robnielsen:
@steve-bate, I'll wait for @berndpfrommer to weigh in on my comments. It might be a better idea to remove 0x34 and 0x35 from the device_features.xml and add a comment why they are not needed. 

Also what about my comments about other devices that use 0x2e and 0x2f, how would those be supported? For example, most of the dimmer devices at http://www.insteon.com/developer/

steve-bate:
I removed the message handlers elements in device_features.xml and added a comment. Hopefully someone will remember to scan the code looking for 0x19 usage if the DIRECT ACK behavior is modified/fixed in the future.

I've also added comments to the message and command handler classes about device support. I hadn't responded to your questions about protocols for other devices because I wasn't sure how to respond. This PR is clearly labeled as supporting a specific device and the feature has only been configured for that device. Also, that's the device I have for testing purposes and where I know it works. I can think of ways to support other protocols and devices, but since I can't test those devices I believe it would not be a good idea for me to support those devices in this PR. However, it would be easy for someone else to extend the handler implementation to support other devices that they have.

berndpfrommer:
I think the code can go in as is, in particular now that the extraneous handlers for 0x34 and 0x35 have been removed.

All points raised by all participants are valid ones, in particular Rob's reference to other devices that use other commands for ramp rates.

However, I think we should hold off on supporting devices that we can't test. Insteon has consistently thrown me curve balls whenever I tried to design based on their specs. There have been so many "gotcha" moments I can't even begin to tell. Stuff like devices not responding to queries as they should, commands simply not making it through when you send them too close back-to-back. None of that mentioned in the spec. The InsteonTerminal was developed just for the purpose of fooling around with the devices to see what they actually send back. The thermostat support involved creative merging of several specs. Then cryptic stuff like this (RR ON spec on one of the devices):On level = 16*On + 0F, RR = 2*RR+1? What, a recursive relationship???? End of rant.

About the hack with 0x19: in the beginning I thought all direct commands for all devices would behave like that, and since I wanted to get something off the ground without reading all the specs of all devices out there, we have the 0x19 hack now that confuses developers. At some point I should clean it up, but that's definitely not  part of this commit.

When a new device comes along that needs to be supported we'll look at what the differences are, and try to generalize the code then.  I find it easier to look at duplicated code and consolidate it than to look at specs and generalize. That's the path I have pursued with the message and command handlers by allowing parameters in the xml file. I suggest we do the same with the ramp commands. I expect that down the road there won't even be a RampOnOffSwitchHandler, but just a "SuperFlexibleHandler" that does the translation for many other scenarios as well. That SuperFlexibleHandler will then have a set of obscure flags and parameters in the xml file that people will have to first understand before they can change them. So that abstraction needs to be well thought out and documented, which is why I want to push this step out until we have more devices and features supported. For new contributors it's almost easier to cut-and-paste an existing handler than to deal with an unwieldy parameterized monster that absorbs the quirks of several devices.

robnielsen:
Thanks for responding @berndpfrommer, The 0x34 and 0x35 handlers were confusing, now it looks better so I agree with you about pushing it out.  BTW, I am the one who added the 2672-222 about a year ago and I also have a bunch of 2466D dimmers, and few 2334-232 KeypadLink dimmers that i could try and add support.

BTW, it appears that the KeypadLink devices support ramp rates for the buttons (see http://cache.insteon.com/developer/2334-xxx2486xdev-042013-en.pdf). This document is clearer since it uses Ramp Rate instead of RR :)

steve-bate:
I have some changes that would allow the RampOnOff handler cmd1 values to be parameterized in the XML. The defaults are 0x2E (on) and 0x2F (off) since that seems to be the standard cmd1 values for ramp ON/OFF. I have no idea why the 2672-222 uses reserved 0x34 and 0x35 cmd1 values. In any case, these handlers may work with other devices without code changes if others want to test those devices. Would you like me to include those changes in this PR?

robnielsen:
@steve-bate, that sounds like a good idea, once the PR is merged I can try it with my devices and see if they work or not and create a new PR with hopefully only XML updates.

steve-bate:
I went ahead and committed the changes (without squashing for now). If we don't want to include the latest changes for some reason I'll reset the commit. I also added ramp support for the 2674-222 Recessed LED Bulb (which I have). It uses the standard 0x2E/0x2F so I've been able to test both variations of the ramp feature XML specification.

Here's another Insteon odd thing. The link to the developer documentation for the 2674-222 actually shows the PDF for the 2672-222. I would have though that meant they use the same commands and were just packaged differently. However, the specs are clearly different.

steve-bate:
Based on the comment from @robnielsen, I'll go ahead and squash the commits.

robnielsen:
Thanks @steve-bate!

steve-bate:
One more thing... :-) I just added a handler for ramp dimming (not committed yet). It works, but there are some quirks. For example, if a light is off and you send a command to ramp to X% for 1 minute then it will work as expected (really, this is the current switch-oriented behavior). However, if the light is on at 50% and you give a dimmer command of 100, it *appears* (based on observation) that the ramp time may still be from 0 to 100% (vs. 50% to 100%). If so, you wouldn't see any change for 30 seconds and then it would ramp from 50% to 100% in the next 30 seconds. The rampdimmer doesn't replace the existing dimmer feature so maybe the risk of strange behavior is low. @berndpfrommer and @robnielsen, what do you think?

berndpfrommer:
Cool. I wouldn't even call that a quirk. I wonder what you do about the status of the rampdimmer. Is it just a command handler with no feedback, or does it poll and get the current level, like the "dimmer" feature?

steve-bate:
Good question. It currently just processes the direct ack for the command. Would the same FlexPollHandler that the GenericDimmer uses also work for the RampDimmer? If so, I can add that to the feature definition.

robnielsen:
It does sound cool, lets add it :)

steve-bate:
The latest commit removes the "ramponoff" feature and adds a "rampdimmer" feature. I added a ramp percent command handler and created an abstract base class for code shared between the on/off handler and the percent handler. I also added a rampdimmer feature to the 2466D device as requested by Rob.

robnielsen:
Hi steve,  I tried you latest changes and they are close for the KeypadLink, but I think I'll wait until your PR is merged and then I'll finish it up. Also, can we change the feature name from 'rampdimmer' to 'ramp'? This sounds more consistent with the others, like 'fastonoff' and 'manualchange'.

I also tried both of my 2672-222 LED bulb, and I'm seeing both the command and state change:
>2015-12-31 10:17:12 - mKitchenLampRamp received command ON
2015-12-31 10:17:12 - mKitchenLampRamp state updated to ON

But the bulbs are not turning on. This is what shows up in the logs:

>2015-12-31 10:17:12 INFO  o.o.b.i.InsteonPLMActiveBinding[:125]- Item: mKitchenLampRamp got command ON
2015-12-31 10:17:12 DEBUG o.o.b.i.i.device.InsteonDevice[:168]- processing command ON features: 5
2015-12-31 10:17:12 DEBUG o.o.b.i.i.d.CommandHandler[:585]- ramp encoding: time 15.0 insert 4 d1 8.5 d2 4.0 ramp 11
2015-12-31 10:17:12 DEBUG o.o.b.i.i.device.InsteonDevice[:379]- qe taken off direct: RampDimmer_3435(1:2:1) OUT:Cmd:0x62|toAddress:21.F6.AF|messageFlags:0x0F=DIRECT:3:3|command1:0x34|command2:0xFB|
2015-12-31 10:17:12 DEBUG o.o.b.i.i.device.InsteonDevice[:399]- next request queue processed in 2000 msec, quiettime = 2000
2015-12-31 10:17:12 DEBUG o.o.b.i.internal.driver.Port[:382]- writing (500): OUT:Cmd:0x62|toAddress:21.F6.AF|messageFlags:0x0F=DIRECT:3:3|command1:0x34|command2:0xFB|
2015-12-31 10:17:12 INFO  o.o.b.i.i.d.CommandHandler[:233]- RampOnOffCommandHandler: sent ramp on to switch 21.F6.AF time 15.0 level 100 cmd1 52
2015-12-31 10:17:12 DEBUG o.o.b.i.InsteonPLMActiveBinding[:581]- got msg: IN:Cmd:0x50|fromAddress:21.F6.AF|toAddress:24.11.9D|messageFlags:0x2B=ACK_OF_DIRECT:3:2|command1:0x34|command2:0xFB|
2015-12-31 10:17:12 DEBUG o.o.b.i.i.d.MessageDispatcher[:134]- 21.F6.AF:RampDimmer_3435 DIRECT_ACK: q:QUERY_PENDING cmd: 80
2015-12-31 10:17:12 DEBUG o.o.b.i.i.d.MessageDispatcher[:139]- changing key to 0x19 for msg IN:Cmd:0x50|fromAddress:21.F6.AF|toAddress:24.11.9D|messageFlags:0x2B=ACK_OF_DIRECT:3:2|command1:0x34|command2:0xFB|
2015-12-31 10:17:12 INFO  o.o.b.i.i.d.MessageHandler[:414]- RampDimmerHandler: device 21.F6.AF was switched on using ramp.
2015-12-31 10:17:12 DEBUG o.o.b.i.i.device.DeviceFeature[:264]- 21.F6.AF:RampDimmer_3435 publishing: ON
2015-12-31 10:17:12 DEBUG o.o.b.i.i.d.MessageDispatcher[:159]- defdisp: 21.F6.AF:RampDimmer_3435 set status to: QUERY_ANSWERED
2015-12-31 10:17:12 DEBUG o.o.b.i.i.device.DeviceFeature[:264]- 21.F6.AF:GenericLastTime publishing: 2015-12-31T10:17:12
2015-12-31 10:17:14 DEBUG o.o.b.i.i.d.RequestQueueManager[:132]- device queue for 21.F6.AF is empty!

Any ideas? @berndpfrommer, maybe you have suggestions since you also have a 2672-222 LED bulb that I sent you last year.

berndpfrommer:
@steve-bate I would not poll. Extra load on the network, unless we use feature groups. But in fact we need to think what the state of the "rampdimmer" (Rob, I prefer it over "ramp" because it is more descriptive) reflects. The "dimmer" feature already captures the light level, so what is the state of "rampdimmer" supposed to represent?

The approach most consistent with the rest of the binding would be to have the state of "rampdimmer" reflect whatever was acked last, i.e. have the RampDimmerHandler extract the light level from the command2 field of the ack message. It already sends On/Off down the OpenHab bus as it should. It's pretty much up to you if you want to support this. I think very few people will actually be interested in using the state of "rampdimmer" at all.  In summary, no polling, and if you want to be fancy, extract the light level from the ack message (tacitly dropping the "rate", which is also acked).

steve-bate:
I removed the polling and updated the RampDimmerHandler to post the ack'ed level. Commits have been squashed. Thanks for the time you and Rob have invested in reviewing this PR. Happy New Year!

robnielsen:
You are welcome. Happy New Year!

teichsta:
great work guys, thanks!

Issue #3619, opened 2015-12-29
Add to Z-Wave Database


drewwaite:
Please add these z-wave devices to the database:

1. Fibaro System  ID:3000 Type:c00
Fibaro Smoke Sensor
New: FGSS001 
Appears to be identical to FGSS101

2. Aeon Labs ID:59 Type:202
Recessed Door Sensor - ZW089-B
Identical to the ZW089 already in the database.

Thanks,
Andrew

SkynetHome:
How do I get rid of all of the other fing commits. They are showing up in the new request I was about to pull?

cdjackson:
It’s probably easiest to just change to master and create a new branch and make the change there since it’s only a small change.

SkynetHome:
Thats what I did but every time I go to pull it yanks all these commits from branches that I have deleted
just did git reset --hard and git -f -d and now things look good on my end I guess I just push that back up and hopefully it gets rid of those unwanted commits

SkynetHome:
nothing I do gets rid of these unwanted commits. And here I though was going to come on here and make a quick PR and get back to work.

cdjackson:
Have you made these changes directly into master, or merged them into master already? Somehow you presumably need to get a clean master branch...

SkynetHome:
I fetched and merged earlier today because the commits I made last night couldnt be merged. Then I changed branches committed and pulled. Thats when I noticed all of the crap I waded through last night got pulled into today's request. I try to rebase but only the latest commit shows up.  I try to just create a different branch from master and make changes there but for some reason nothing updates in eclipse so I cant make changes as they are already there.

SkynetHome:
I guess at some point changes were made to master

cdjackson:
It sounds like you merged your changes into master, rather than master into your new branch. You might need to roll back, although this might be a bess… might be better to download again to reset...

SkynetHome:
by download again do you mean reclone my fork and reset eclipse with that

SkynetHome:
I just deleted my fork and reforked...will clone and start over from there

cdjackson:
Can this be closed?

PR #3625, opened 2015-12-30
Added ZW089B

SkynetHome:
Copied existing and added reference for -B model to products.xml per
Issue #3619


SkynetHome:
ahh will repair or delete and re-pull

PR #3642, opened 2015-12-30
Compare strings with equals method

coolweb:


watou:
Thanks!  Does the resulting JAR still work properly for you?  Let me know and I'll merge this.

coolweb:
@watou Ok, I tested it now and it works.

Issue #3638, opened 2015-12-30
Stale Bundle-Versions


watou:
Which of these needs to be updated to `1.8.0.qualifier`, @teichsta?
```bash
$ grep -R "Bundle-Version" * | grep -v target | grep -v "1.8.0.qualifier"
bundles/archetype/org.openhab.archetype.action/src/main/resources/archetype-resources/META-INF/MANIFEST.MF:Bundle-Version: 1.3.0.qualifier
bundles/archetype/org.openhab.archetype.binding/src/main/resources/archetype-resources/META-INF/MANIFEST.MF:Bundle-Version: ${version}.qualifier
bundles/binding/org.openhab.binding.lightwaverf.test/META-INF/MANIFEST.MF:Bundle-Version: 1.7.0.qualifier
bundles/binding/org.openhab.binding.octoller/META-INF/MANIFEST.MF:Bundle-Version: 1.8.0
bundles/binding/org.openhab.binding.samsungac.test/META-INF/MANIFEST.MF:Bundle-Version: 1.6.0.qualifier
bundles/binding/org.openhab.binding.yamahareceiver/META-INF/MANIFEST.MF:Bundle-Version: 1.7.1.qualifier
bundles/designer/org.openhab.designer.core/resources/MANIFEST.MF:Bundle-Version: 1.0.0
bundles/io/org.openhab.io.rest.test/META-INF/MANIFEST.MF:Bundle-Version: 1.0.0.qualifier
bundles/io/org.openhab.io.transport.cul.test/META-INF/MANIFEST.MF:Bundle-Version: 1.7.0.qualifier
bundles/model/org.openhab.model.codegen/META-INF/MANIFEST.MF:Bundle-Version: 1.0.0.qualifier
```

teichsta:
fixed with #3646 and #3670

PR #3646, opened 2015-12-30
Update stale Bundle-Versions (for #3638)

watou:
To address #3638


teichsta:
Thanks @watou!

teichsta:
see also my PR #3670 which should solve the root cause of this problem

watou:
Thank you, Thomas!

PR #3650, opened 2015-12-31
[build fix] Bump java compiler version to 1.7

dvanherbergen:
Signed-off-by: Davy Vanherbergen <davy.vanherbergen@gmail.com>


dvanherbergen:
Not yet needed as openHAB 1.8 release will still use 1.6.

Issue #3662, opened 2016-01-01
Add to Z-Wave Database


drewwaite:
It appears that I have found another one. It seems to be the AU versions!

Can you please add the Fibaro RGBW Controller to the database. The XML file is the same as fgrgbw.xml.

[ID:3000, Type:900]

If you can point in the direction of what to do I can probably add the information myself.

Thanks again.

Andrew

cdjackson:
You need to update the products.xml file - open this file in the database folder (in the zwave binding) - search for the device - add an extra reference (should be obvious what to do as you’ll see the references there) - save, and generate a pull request on git so we can get it added to the binding…

If you don’t do this though, please say so so that we can update it (i.e. I’ll assume you’re going to create this PR unless you say otherwise).

Chris

drewwaite:
Not sure I did this right... I think I have created a pull request but it seems to be wanting to change more than 1 file!

cdjackson:
I assume this is working and can be close?

drewwaite:
Yes.This issue can be closed.

PR #3664, opened 2016-01-01
Fibaro RGBW

drewwaite:
Issue #3662

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -1657,6 +1657,10 @@
 				<Type>0900</Type>
 				<Id>4000</Id>
 			</Reference>
+			<Reference>
+				<Type>3000</Type>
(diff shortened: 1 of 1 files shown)


drewwaite:
Attempt #2. That looks better.

cdjackson:
The PR looks better (thanks),  but I don't think you've put the reference in the right place? If this is meant to be for the RGB, then it's not correct as you've put it in with the roller shutter...

cdjackson:
Oh - to fix this, don't change the PR - just move the lines to the right place (assuming it's wrong), recommit and push - it will update the PR automatically.

drewwaite:
Opps. My error. I didn't realise the "item label" was at the bottom. Fixed.

cdjackson:
It is in this case, and probably most cases, but you need to look at the structure of the XML - the reference needs to go inside the Product tag for the right product - it’s not about the order or the elements…

cdjackson:
Thanks.

PR #3666, opened 2016-01-01
Remove already renamed configs, fix generated id-lists and generator.

csowada:
This is a small fix that updates the documentation. The setter ids are wrong. Also fixed the document generator.
Removed already renamed config (starts with underscore)  files.


teichsta:
Thnanks @csowada!

PR #2908, opened 2015-07-12
Added support for Energenie PWM-LAN Measurement Device

chris-kuhr:

PR #3671, opened 2016-01-01
Add support for energie measuerement on PWM-LAN devices. Replaces #2908

hmerk:
Signed-off-by: hmerk <hans-joerg.merk@t-online.de>


hmerk:
@teichsta I think the merge conflict results in #2908 is not closed.
Could you check this please.

PR #3679, opened 2016-01-02
remove the special toString() in order to sync with ESH

teichsta:


kaikreuzer:
This is imho a backward compatibility breaking change, which I'd suggest to revert. See also https://github.com/openhab/openhab/pull/3576#issuecomment-168414518.

teichsta:
ok, reverted this PR. Nevertheless i scanned through the code and it seems that codewise we were not going to expect any problems. For rules you will be right for sure …

PR #3682, opened 2016-01-02
InsteonPLM: add support for rampdimmer with KeypadLinc dimmers

robnielsen:
Added rampdimmer support for the KeypadLinc dimmers described in http://cache.insteon.com/developer/2672-222dev-052014-en.pdf. The button check in RampDimmerHandler was causing the state to not get updated for the load button. It is not needed since button commands for the non-load buttons are sent to groups, not to the actual device.

@steve-bate and @berndpfrommer can you please review?


berndpfrommer:
Looks good to me, can go in.

teichsta:
@robnielsen is this a fix which should go into 1.8 or can we consider this an enhancement?

robnielsen:
@teichsta It is an enhancement to https://github.com/openhab/openhab/pull/3621 for the issue https://github.com/openhab/openhab/issues/3579. I could go either way, the change is low risk.

teichsta:
thanks … also cherry-picked onto 1.8 in order to complement the existing enhancement.

PR #3684, opened 2016-01-02
Modbus binding: coil commands executed no matter current state

ssalonen:
Ensures that coil commands are written to the modbus slave no matter what the current state is.

Furthermore, now handling nicely if no poll is done but we receive a command.

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
@@ -91,14 +91,14 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
             if (provider.providesBindingFor(itemName)) {
                 ModbusBindingConfig config = provider.getConfig(itemName);
                 ModbusSlave slave = modbusSlaves.get(config.slaveName);
-                slave.executeCommand(command, config.readRegister, config.writeRegister);
+                slave.executeCommand(command, config);
             }
         }
     }
 
     /**
      * Posts update event to OpenHAB bus for "holding" type slaves
-     * 
+     *
      * @param binding ModbusBinding to get item configuration from BindingProviding
      * @param registers data received from slave device in the last pollInterval
      * @param itemName item to update
@@ -116,7 +116,7 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
             String slaveValueType = modbusSlaves.get(slaveName).getValueType();
             double rawDataMultiplier = modbusSlaves.get(slaveName).getRawDataMultiplier();
 
-            State newState = extractStateFromRegisters(registers, config.readRegister, slaveValueType);
+            State newState = extractStateFromRegisters(registers, config.readIndex, slaveValueType);
             /* receive data manipulation */
             State newStateBoolean = provider.getConfig(itemName)
                     .translateBoolean2State(!newState.equals(DecimalType.ZERO));
@@ -170,7 +170,7 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
 
     /**
      * Posts update event to OpenHAB bus for "coil" type slaves
-     * 
+     *
      * @param binding ModbusBinding to get item configuration from BindingProviding
      * @param registers data received from slave device in the last pollInterval
      * @param item item to update
@@ -180,7 +180,7 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
             if (provider.providesBindingFor(itemName)) {
                 ModbusBindingConfig config = provider.getConfig(itemName);
                 if (config.slaveName.equals(slaveName)) {
-                    boolean state = coils.getBit(config.readRegister);
+                    boolean state = coils.getBit(config.readIndex);
                     State currentState = provider.getConfig(itemName).getItemState();
                     State newState = provider.getConfig(itemName).translateBoolean2State(state);
                     if (!newState.equals(currentState)) {
@@ -193,7 +193,7 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
 
     /**
      * Returns names of all the items, registered with this binding
-     * 
+     *
      * @return list of item names
      */
     public Collection<String> getItemNames() {
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusGenericBindingProvider.java
@@ -29,15 +29,15 @@ import org.slf4j.LoggerFactory;
  *
  * 1) single coil/register per item
  * Switch MySwitch "My Modbus Switch" (ALL) {modbus="slave1:5"}
- * 
+ *
  * This binds MySwitch to modbus slave defined as "slave1" in openhab.config reading/writing to the coil 5
  *
  * 2) separate coils/registers for reading and writing
  * Switch MySwitch "My Modbus Switch" (ALL) {modbus="slave1:<6:>7"}
- * 
+ *
  * In this case coil 6 is used as status coil (readonly) and commands are put to coil 7 by setting coil 7 to true.
  * You hardware should then set coil 7 back to false to allow further commands processing.
- * 
+ *
  * @author Dmitry Krasnov
  * @since 1.1.0
  */
@@ -127,10 +127,13 @@ public class ModbusGenericBindingProvider extends AbstractGenericBindingProvider
     public class ModbusBindingConfig implements BindingConfig {
 
         /**
-         * readRegister and writeRegister store references to the register in device data space
+         * Index to read, relative to start
          */
-        int readRegister;
-        int writeRegister;
+        int readIndex;
+        /**
+         * Index to write, relative to start
+         */
+        int writeIndex;
 
         /**
          * Name of the ModbusSlave instance to read/write data
@@ -195,8 +198,8 @@ public class ModbusGenericBindingProvider extends AbstractGenericBindingProvider
                 String[] items = config.split(":");
                 slaveName = items[0];
                 if (items.length == 2) {
-                    readRegister = Integer.valueOf(items[1]);
-                    writeRegister = Integer.valueOf(items[1]);
+                    readIndex = Integer.valueOf(items[1]);
+                    writeIndex = Integer.valueOf(items[1]);
                 } else if (items.length == 3) {
                     assignRegisters(items[1]);
                     assignRegisters(items[2]);
@@ -216,9 +219,9 @@ public class ModbusGenericBindingProvider extends AbstractGenericBindingProvider
          */
         private void assignRegisters(String item) throws BindingConfigParseException {
             if (item.startsWith("<")) {
-                readRegister = Integer.valueOf(item.substring(1, item.length()));
+                readIndex = Integer.valueOf(item.substring(1, item.length()));
             } else if (item.startsWith(">")) {
-                writeRegister = Integer.valueOf(item.substring(1, item.length()));
+                writeIndex = Integer.valueOf(item.substring(1, item.length()));
             } else {
                 throw new BindingConfigParseException("Register references should be either :X or :<X:>Y");
             }
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
@@ -11,6 +11,7 @@ package org.openhab.binding.modbus.internal;
 import java.util.Collection;
 
 import org.openhab.binding.modbus.ModbusBindingProvider;
+import org.openhab.binding.modbus.internal.ModbusGenericBindingProvider.ModbusBindingConfig;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.IncreaseDecreaseType;
 import org.openhab.core.library.types.OnOffType;
@@ -36,6 +37,7 @@ import net.wimpi.modbus.msg.WriteMultipleRegistersRequest;
 import net.wimpi.modbus.msg.WriteSingleRegisterRequest;
 import net.wimpi.modbus.procimg.InputRegister;
 import net.wimpi.modbus.procimg.Register;
+import net.wimpi.modbus.procimg.SimpleRegister;
 import net.wimpi.modbus.util.BitVector;
 
 /**
@@ -86,10 +88,10 @@ public abstract class ModbusSlave implements ModbusSlaveConnection {
 
     /**
      * A multiplier for the raw incoming data
-     * 
+     *
      * @note rawMultiplier can also be used for divisions, by simply
      *       setting the value smaller than zero.
-     * 
+     *
      *       E.g.:
      *       - data/100 ... rawDataMultiplier=0.01
      */
@@ -108,24 +110,23 @@ public abstract class ModbusSlave implements ModbusSlaveConnection {
     /**
      * writes data to Modbus device corresponding to OpenHAB command
      * works only with types "coil" and "holding"
-     * 
+     *
      * @param command OpenHAB command received
-     * @param readRegister data from readRegister are used to define value to write to the device
-     * @param writeRegister register address to write new data to
+     * @param config
      */
-    void executeCommand(Command command, int readRegister, int writeRegister) {
+    void executeCommand(Command command, ModbusBindingConfig config) {
         if (ModbusBindingProvider.TYPE_COIL.equals(getType())) {
-            setCoil(command, readRegister, writeRegister);
+            setCoil(command, config);
         }
         if (ModbusBindingProvider.TYPE_HOLDING.equals(getType())) {
-            setRegister(command, readRegister, getStart() + writeRegister);
+            setRegister(command, config);
         }
     }
 
     /**
      * Calculates boolean value that will be written to the device as a result of OpenHAB command
      * Used with item bound to "coil" type slaves
-     * 
+     *
      * @param command OpenHAB command received by the item
      * @return new boolean value to be written to the device
      */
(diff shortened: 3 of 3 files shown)


ssalonen:
Fixes #3685

ssalonen:
Review needed especially in `setCoil`... I can't get gist of the original implementation -- especially when readRegister != writeRegister.

teichsta:
@paolodenti @bakrus @nickma82 could you probably provide some feedback?

paolodenti:
@teichsta this is what I think about it

@ssalonen
Could you please explain better the modification?

From what I can understand just looking quickly at the code this is basically what changed:

* **before**: `setRegister` and `setCoil` check against value in repository. If value is different with respect to the one stored in the repository, the coil/register is set
* **after**: `setRegister` and `setCoil` set value without checking against value in repository.

If I am right, and I am not sure about that, I think the previous implementation is better. That is a common philosophy: do the polling, keep the updated values in repo, execute remote updates just if new value is different from previously polled value.
IMHO Your modification should be done just is polling is disabled. But documentation does not talk about disabling polling and after a short look at the code it looks to me that poling cannot be disabled.

I personally think that your modification could break previous behaviours that rely on the fact that coil/register update is executed just if value changes. Such a modification could potentially be useful (if you are doing that I guess it is useful for some scenarios) but as an extension to the binding instead of a modification to the previously implemented behaviour.

Please forgive me if I did not understand your modifications in detail.

ssalonen:
Thank you very much @paolodenti for taking time to comment on the changes. You are on the right tracks but I find it bit more complicated than that. 

Let me describe how the current implementation (**before**) works per my understanding:

**when binding receives command corresponding to an item that matches *holding register***:
- value is written to modbus straight away, no checks are made whether value has changed etc. In other words, command is simply carried out and written to the modbus slave. 
- If the command is is of type `IncreaseDecreaseType` or `UpDownType`, previously cached/polled value is used to compute the updated value: `newValue = prevValue + delta` (where `delta=+1` or `delta=-1` depending on command)
- See [ModbusSlave.setRegister](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L173) for details.

**when binding receives command corresponding to an item that matches *coil***:
- [Command is compared to cached/polled value](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L152) first. If boolean command matches previously polled value, command is not processed and nothing is written to the slaves.
- If command does not match polled value (i.e. user really tries to *change* to value of the coil), we value boolean `x` to the modbus slave. 
  - For some reason I can't understand, the value `x` depends on the command's status (see [this](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L158) and [this](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L160)). 
    - if command corresponds to boolean `true`, coil is set with value `x=true`
    - if command corresponds to boolean `false`
      - coil is set with value `x=true` [only if read location is different from write location](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L160) (note! even though command was `false`, value of `true` is written). 
       - Otherwise, coil gets value of `x=false`.


This PR only changes the behavior of coils such that

**when binding receives command corresponding to an item that matches *coil***:
- command is carried out and coil value is set to the boolean matching the command. 


Note that polling is also separate from this PR. State updates are not sent to event bus unless the value changes.

The motivation behind the binding is to streamline & verify the implementation before making [some larger changes I'm planning](https://community.openhab.org/t/current-modbus-changes-from-ssalonen/5356/). I actually found the issue during writing tests for the current implementation. I felt that the current implementation violates the logic described in the [wiki guide](https://github.com/openhab/openhab/wiki/How-To-Implement-A-Binding) (emphasis mine):

>The purpose of a binding is to translate between events on the openHAB event bus and an external system. This translation should happen "stateless", i.e. the binding must not access the ItemRegistry in order to get hold of the current state of an item. Likewise, **it should not itself keep states of items in memory.**

I think it is very valid to keep the binding as stateless as possible. It is entirely possible that the slave coil is updated without explicit advice from openhab, and the the cached state in the binding does not even match the reality. Before proposing this PR, I tried to look other bindings for conventions and did not find similar logic in those. Please also note that the registers are handled differently even in the modbus binding.

You pointed out the valid concern about the backwards compatibility, but at least there is a simple workaround using rules. I'm not sure if performance is such an issue since this PR only affects *command*s, not the polling. I understand that polling is very time critical (let's say with polling period of 200ms), but do our users have tight loop with commands emitted?

Once again thanks for the comments and looking for more feedback on this.

nickma82:
@ssalonen Am I right, that this effects 'Write Single Coil' (command code five) *only*??
@ssalonen Have you tested that?

@paolodenti 
>From what I can understand just looking quickly at the code this is basically what changed:
>
>    before: setRegister and setCoil check against value in repository. If value is different with respect to the one stored in the repository, the coil/register is set
>    after: setRegister and setCoil set value without checking against value in repository.

That's how I understand it too.

**Analysis of the existing code**
IMHO the condition 
> This translation should happen "stateless"

from the [wiki guide](https://github.com/openhab/openhab/wiki/How-To-Implement-A-Binding), is offended by the [current check](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L156) what the previous state was, as @ssalonen pointed out.
And I too find that [these lines](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L156-L162) are looking strange, because they are leading to different external behaviour between `setCoil(..)` and `setRegister(..)`.

@paolodenti 
> I personally think that your modification could break users installations that rely on the fact that coil/register update is executed just if value changes

I think so too, but I think that this would be the wrong usage?? On the other hand if one is willing to send the same command over and over again (e.g. to enable a timeout over and over again), this is currently not possible.

Lastly, what distracts me the most is [this weird line](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L160), (@ssalonen pointed that out too `For some reason I can't understand, the value x depends on the command's status`) which should be changed.

**What I think about the changes**
* I do like the interface changes of `setCoil(..)` and `setRegister(..)` and 
* I like that he cleaned up the [weird line](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L160)
* I also think that it's better to send the command out every time (which is what the proposed change does), but if think that this is a little bit of a political question.

ssalonen:
@nickma82 @paolodenti yes, the change really affects only coil command -- holding register does not have any checks for the `!old_value.equals(new_value)` in master branch. Please tell me if I missed the comparison of old value vs new value with holding register items -- I can't find such comparison, not when writing commands to slaves.

The setRegister modifications are basically refactoring and cleaning up (previously `NullPointerException` can be thrown since `storage = null` before first poll completes. PR fixes the situation by error logging and not requiring storage to be non-null -- if the command does not really require it, see comments below). 

Here's walkthrough of master branch code (the way I see it): For the register behaviour in master branch, see [setRegister L195](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L195). Yes the old/cached/polled value is read on [line 181](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L181), but it is not used at all (unless we have INCREASE/DECREASE or UP/DOWN command). `ModbusSlave.setRegister` is called by [`ModbusSlave.executeCommand`](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L117) which does no check on the actual values. [`ModbusSlave.executeCommand`](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java#L117) is called by [`ModbusBinding.internalReceiveCommand`](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java#L99) which once again quite straightforward passthrough. 

I have tested this in my integration tests (test data [here](https://github.com/ssalonen/openhab/blob/modbus-binding-tests/bundles/binding/org.openhab.binding.modbus.test/src/test/java/org/openhab/binding/modbus/internal/WriteRegistersTestParameters.java#L353) and test case [here](https://github.com/ssalonen/openhab/blob/modbus-binding-tests/bundles/binding/org.openhab.binding.modbus.test/src/test/java/org/openhab/binding/modbus/internal/WriteRegistersTestCase.java#L180), please ignore the "knownFailure" naming, it's not failing). Sorry for the messy tests... cleanup would be in order. 

Indeed this is a political question and relates to more like conventions how the openhab bindings should behave. Perhaps some of the core maintainers (how can I higlight those guys? who are they) can comment?

I do believe that not all binding can be fully stateless (e.g. handling of INCREASE command in modbus binding), but avoiding it as much as possible is something to aim for.

watou:
I am not sure how the following applies to this binding or any existing constraints, but in general, caching item state should be avoided.

Caching item state in a binding can be motivated by wanting to improve efficiency, but can also lead to problems like:

* Other external systems can independently change the real item's state, meaning your cache is inherently unreliable.
* A user sent a command but it wasn't sent by the binding.  (If a rule, script, or API client wants to send a command, the binding should try to send it.)
* Suppressing updates when the value is unchanged deprives the system of knowing the most recent time when the item's state was accurate.  If you want to know if an item changed value, listen for changes, not updates.

One kind of efficiency that the binding writer might be pursuing is in the case where, for some possibly valid reason, the polling cycle is very short, like 200ms (but system dependent), and it's easy to see that the event bus could be overwhelmed with processing "update is the same" messages.  In that case, some other mechanism other than caching can be used, like quantizing the updates over time, so no item will be updated more often than T.  But you would want to evaluate the reasons for such a high polling rate in the first place.

paolodenti:
I have been thinkng deeply to what @nickma82 wrote

> On the other hand if one is willing to send the same command over and over again (e.g. to enable a timeout over and over again), this is currently not possible.

And I am not sure of what I am writing below because both @nickma82 and @ssalonen share the same opinion about that.

However I have to say that I do not agree with their consideration. If a command is not going to be effective in any way, it just should not be executed. And if we set an already set modbus coil, or if we set a holding register with the same already kept value, no effect is generated (apart of the secondary effect of overloading the communication channel).

An objection that could be done to my point is that the coil state could really be in a different state from the one stored in the item repository because of the polling delay. But IMHO this is pointless; the only state we should take into account is the repository one; it is the state over which all the rules, commands, scripts, ... are operating, even if potentially 200ms late. The real modbus state should not be taken into consideration (and this "problem" is anyway unavoidable with modbus tcp, a polling is just necessary, no direct update can be implemented)

So, I think that is for any scenario (that I personally do not see) such a "forced command" execution is needed or at least useful, it should be carried out with an additional binding implementation instead of substituting the actual one.

But ... I agree that this is basically a political decision.

ssalonen:
I wanted to document steps to assert my claim:
> yes, the change really affects only coil command -- holding register does not have any checks for the !old_value.equals(new_value) in master branch. Please tell me if I missed the comparison of old value vs new value with holding register items -- I can't find such comparison, not when writing commands to slaves.

Here's the tests done with Modbus TCP using the server provided by the binding.

**Start modbus tcp slave (server)**

The binding provides one:
`java -classpath bundles/binding/org.openhab.binding.samsungac.test/lib/slf4j-simple-1.7.5.jar:bundles/io/org.openhab.io.rest.lib/lib/slf4j-api-1.6.1.jar:/path/to/org.openhab.binding.modbus-1.7.1.jar  net.wimpi.modbus.cmd.TCPSlaveTest 5000´

Server listens for port 5000 with your hostname (not localhost, nor 127.0.0.1). Coils are initialized to 1, while holding register is initalized to integer 251.


**Configure openhab**

openhab.cfg:
```
// read one holding register every 0.5s
modbus:poll=500
modbus:tcp.slave1.connection=YOURHOSTNAME:5000
modbus:tcp.slave1.id=15
modbus:tcp.slave1.start=0
modbus:tcp.slave1.length=1
modbus:tcp.slave1.type=holding

modbus:tcp.slave1coil.connection=YOURHOSTNAME:5000
modbus:tcp.slave1coil.id=15
modbus:tcp.slave1coil.start=0
modbus:tcp.slave1coil.length=1
modbus:tcp.slave1coil.type=coil
```

items:
```
// interpret holding register as integer
Number RegItem "RegItem [%d]" (ALL) {modbus="slave1:0"}
SwitchItem CoilItem "CoilItem [%d]" (ALL) {modbus="slave1coil:0"}
```

rules:
```
// write integer value 251 to holding register every 2s
rule "Write value to modbus"
    when
        Time cron "0/2 * * * * ?"
    then
        sendCommand(RegItem, 251)
        sendCommand(CoilItem, true)
    end
```


start_debug.sh:
```
15:05:49.260 [DEBUG] [m.r.internal.engine.RuleEngine:77   ] - Started rule engine
15:05:52.260 [DEBUG] [o.b.m.internal.ModbusActivator:30   ] - Modbus binding has been started.
15:05:52.261 [DEBUG] [i.internal.GenericItemProvider:341  ] - Start processing binding configuration of Item 'RegItem (Type=NumberItem, State=Uninitialized)' with 'ModbusGenericBindingProvider' reader.
15:05:52.262 [DEBUG] [i.internal.GenericItemProvider:341  ] - Start processing binding configuration of Item 'CoilItem (Type=SwitchItem, State=Uninitialized)' with 'ModbusGenericBindingProvider' reader.
15:05:52.274 [DEBUG] [.modbus.internal.ModbusBinding:282  ] - modbusSlave 'slave1' instanciated
15:05:52.275 [DEBUG] [.modbus.internal.ModbusBinding:282  ] - modbusSlave 'slave1coil' instanciated
15:05:52.276 [DEBUG] [.modbus.internal.ModbusBinding:344  ] - config looked good, proceeding with slave-connections
15:05:52.278 [INFO ] [.service.AbstractActiveService:169  ] - Modbus Polling Service has been started
15:05:52.283 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:52.285 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:52.786 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:52.787 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:53.287 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:53.288 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:53.788 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:53.789 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:54.290 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:54.290 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:54.791 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:54.792 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:55.292 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:55.293 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:55.793 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:55.794 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:56.294 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:56.295 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:56.796 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:56.796 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:57.297 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:57.297 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:57.798 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:57.799 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:58.299 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:58.300 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:58.717 [INFO ] [c.internal.ModelRepositoryImpl:80   ] - Loading model 'default.rules'
15:05:58.800 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:58.800 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:58.821 [DEBUG] [.r.i.engine.RuleTriggerManager:438  ] - Scheduled rule Write value to modbus with cron expression 0/2 * * * * ?
15:05:58.832 [DEBUG] [.o.m.r.i.engine.ExecuteRuleJob:53   ] - Executing scheduled rule 'Write value to modbus'
15:05:58.907 [DEBUG] [riptExtensionClassNameProvider:63   ] - Script actions have changed: HTTPActionService, AudioActionService, TransformationActionService, ExecActionService, PingActionService, 
15:05:59.058 [INFO ] [runtime.busevents             :22   ] - RegItem received command 251
15:05:59.059 [DEBUG] [.b.modbus.internal.ModbusSlave:216  ] - ModbusSlave: FC6 ref=0 value=251
15:05:59.084 [INFO ] [runtime.busevents             :22   ] - CoilItem received command ON
15:05:59.301 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:59.302 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:05:59.802 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:05:59.803 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:06:00.003 [DEBUG] [.o.m.r.i.engine.ExecuteRuleJob:53   ] - Executing scheduled rule 'Write value to modbus'
15:06:00.008 [INFO ] [runtime.busevents             :22   ] - RegItem received command 251
15:06:00.008 [DEBUG] [.b.modbus.internal.ModbusSlave:216  ] - ModbusSlave: FC6 ref=0 value=251
15:06:00.014 [INFO ] [runtime.busevents             :22   ] - CoilItem received command ON
15:06:00.304 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:06:00.305 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:06:00.805 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:06:00.806 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:06:01.306 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
15:06:01.307 [INFO ] [runtime.busevents             :26   ] - RegItem state updated to 251
15:06:01.807 [INFO ] [runtime.busevents             :26   ] - CoilItem state updated to ON
```

From the log you get entries like `ModbusSlave: FC6 ref=0 value=251` which indicate that the holding value is being written to the slave over and over again. Note that the coil is not written since the command equals polled value (1).


If we change the rule to write value 0 to coil, value is written (once) since it differs from the polled value:
```
rule "Write value to modbus"
    when
        Time cron "0/2 * * * * ?"
    then
        sendCommand(RegItem, 251)
        sendCommand(CoilItem, OFF)
    end
```

```
# ./start_debug.sh|grep FC
15:11:32.266 [DEBUG] [.b.modbus.internal.ModbusSlave:216  ] - ModbusSlave: FC6 ref=0 value=251
15:11:32.294 [DEBUG] [.b.modbus.internal.ModbusSlave:246  ] - ModbusSlave: FC05 ref=0 value=false
15:11:34.007 [DEBUG] [.b.modbus.internal.ModbusSlave:216  ] - ModbusSlave: FC6 ref=0 value=251
15:11:36.007 [DEBUG] [.b.modbus.internal.ModbusSlave:216  ] - ModbusSlave: FC6 ref=0 value=251
```

ssalonen:
Rebased against master

wtrucci:
Please add missing row added from @paolodenti  for closing connection. Now after your modify the binding is not working anymore for multiple connections.

ssalonen:
@wtrucci I believe you are referring to [to this discussion](https://community.openhab.org/t/modbus-connection-problem/6108)? If you don't mind I think it should be a separate pull request. If I have time tomorrow I shall work on it.

wtrucci:
Yes exactly. 
Thanks

Walter

ssalonen:
Anything needed here from me still? Based on the good discussion above, do you think this PR will get accepted as is?

teichsta:
> Anything needed here from me still? Based on the good discussion above, do you think this PR will get accepted as is?

i am going to step through the argumentation thread and try to understand your arguments.

ssalonen:
@teichsta any update on this? I'm happy to help if needed.

teichsta:
@ssalonen i'd gone through the comments above. To make a long story short: i would follow your proposed change in order to send commands to the binding event though it's state didn't change. Please note that i didn't cherry-pick this change onto 1.8 due to the extensive discussion here. I would rather like others to test the changes rather than breaking the binding in a Bugfix release.

Thanks for all your efforts guys!

PR #3692, opened 2016-01-02
[nest] Added optional nest:timeout config parameter

watou:
[A user reports](https://community.openhab.org/t/how-to-access-the-nest/4310/13?u=watou) and I also see occasional request timeouts trying to query the Nest API.  The default 10-second timeout is occasionally not enough time to successfully query the API.  So this PR adds an optional `nest:timeout` config parameter to openhab.cfg, so the user can override the default 10000ms timeout with a different timeout value.

@teichsta, sorry for the late addition.  If you would prefer that this is deferred, OK!


teichsta:
lgtm … please go ahead an merge!

watou:
Thanks @teichsta!

PR #3702, opened 2016-01-04
Add Telldus door sensor

Drassen:
Created from http://www.pepper1.net/zwavedb/device/789/789-0176-0005-0001-03-04-05-01-01.xml
And tested of course.

Thanks!


bundles/binding/org.openhab.binding.zwave/database/telldus/tzdw-100.xml
```diff
@@ -0,0 +1,160 @@
...
+    <Class><id>0x27</id></Class>
+    <Class><id>0x20</id></Class>
+    <Class><id>0x25</id></Class>
+    <Class><id>0x26</id></Class>
+  </CommandClasses>
+  <Associations>
+    <Group>
+      <Index>1</Index>
+      <Maximum>1</Maximum>
+      <Label lang="en">Lifeline: The sensor will send it's status to the node in this group whenever the status of the sensorchanges.</Label>
```
cdjackson: Please break these labels into shorter names as they won't render very nicely in the UI. The label should be very short - in this case, just use 'Lifeline'. Then move the explanation into a separate Help element.

Same for the next associations...

Drassen:
Done!

egila:
Hi, is there any any possibility to get this included in the 1.8 release?

PR #3712, opened 2016-01-05
Corrected parameter 10 for Qubino devices

skenhall:
Update of configuration parameter 10 from 1 byte to 2 byte according to Qubino documentation. (Verified by Qubino support.)
Answer from Qubino support:
"According z-wave standards 1 Byte DEC is used for range -127 to +127, so as parameter 10 is using values up to 255, it is correct to be 2 Byte DEC."

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhaa.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhaa.xml
@@ -72,7 +72,7 @@
 			<Index>10</Index>
 			<Type>list</Type>
 			<Default>255</Default>
-			<Size>1</Size>
+			<Size>2</Size>
 			<Label lang="en">Activate / deactivate functions ALL ON/ALL OFF</Label>			
 			<Item>
 				<Value>255</Value>
--- a/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhad.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhad.xml
@@ -72,7 +72,7 @@
 			<Index>10</Index>
 			<Type>list</Type>
 			<Default>255</Default>
-			<Size>1</Size>
+			<Size>2</Size>
 			<Label lang="en">Activate / deactivate functions ALL ON/ALL OFF</Label>			
 			<Item>
 				<Value>255</Value>
--- a/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhba.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/qubino/zmnhba.xml
@@ -69,7 +69,7 @@
 			<Index>10</Index>
 			<Type>list</Type>
 			<Default>255</Default>
-			<Size>1</Size>
+			<Size>2</Size>
 			<Label lang="en">Activate / deactivate functions ALL ON/ALL OFF</Label>			
 			<Item>
 				<Value>255</Value>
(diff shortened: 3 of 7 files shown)


cdjackson:
Please can you confirm this is tested?

While I agree with Qubino that the parameter by the zwave standard is a signed value, many devices use it unsigned (it really doesn't matter as it is device dependant). This in itself is not justification for it to be 2 bytes (ie just because you want a value >127 doesn't mean you must use 2 bytes).

I expect this is correct, so you probably think I'm being a pain (sorry), but if this isn't correct, it will cause problems so I'd prefer to confirm.

Thanks.

skenhall:
Hi!
First of all this issue caught my eye as an inconsistency between different Qubino devises in the database - Look for instance at the "zmnhbd.xml" file that already sets parameter 10 as 2 bytes.
My primary intension with this pull request was to get all the Qubino devices in sync. (Based on the answer from Quoins support.)
But I'm happy to have this tested properly, but I might need your help on the way! :)
I haven't issued any ALL ON/OFF commands yet. The ALL ON / ALL OFF command classes are not listed as supported command classes on the Wiki (https://github.com/openhab/openhab/wiki/Z-Wave-Binding), but there seems to be an implementation of them in "ZWaveSwitchAllCommandClass.java" on GitHub.
I'm not sure on how I should issue these ALL ON / ALL OFF commands from the controller.Could you please give me some advice on this?
Since I'm not able (yet) to build openHAB myself, my plan for testing is based on the current latest build of 1.8 and do the following steps on three of my Qubino modules (with all are defined as 1 byte for parameter 10):Send ALL ON commandAll Qubino nodes should turn on since "255=ON+OFF Active" is default value and I haven't changed them yet.Send ALL OFF commandAll Qubino nodes should turn off since "255=ON+OFF Active" is default value and I haven't changed them yet.Set configuration parameter 10 to "0=ON inactive, OFF inactive" for one node (node A)Set configuration parameter 10 to "1=ON inactive, OFF active" for one node (node B)Set configuration parameter 10 to "2=ON active, OFF inactive" for one node (node C)Send ALL ON commandOnly Qubino node C should turn on.Manually turn on node A and BSend ALL OFF commandOnly Qubino node B should turn off.
Please give med some advice on sending ALL ON/OFF commands and I'll try to perform these tests.
Best Regards,Per
PS: I quickly looked through "ZWaveSwitchAllCommandClass.java" and I don't really get the implementation where SWITCH_ALL_GET logs "Command {} not implemented.", but there still is a method "public SerialMessage getValueMessage()"... But that is perhaps a bit too much at the moment for me :)

Date: Tue, 5 Jan 2016 10:21:45 -0800

cdjackson:
I note that the definitions in OZW also have 1 byte as the length, so I don't want to merge this until this is confirmed or it will cause problems for people with this device.

I'm not convinced that these should be 2 bytes. The OZW files have (apparently) been produced by Qubino itself, so they should be correct -:
https://groups.google.com/forum/#!topic/openzwave/0XMSa6d9hxo

Please can you test these before we merge them.

REgarding your question about SWITCH_ALL_GET - if the controller received this, it would mean that other devices were asking the contoller for information (which it doesn't support since it's not a switch!) where the getValueMessage is to send a request to the device.

Regarding the use of SWITCH_ALL, I've never used it however I would expect it can be used in the same ways as other command classes - eg by defining a switch item that is bound to this command class.

Please can we discuss switch_all somewhere else - I don't like mixing completely different issues in the same issue - it will mean things get lost.

druciak:
The size depends on the version of the device. I wouldn't touch the files only because they said so. I've got some of Qubino device and thay work perfectly, so unless you are able to test all of them, please leave the definition as it is. I can be wrong, but 2 bytes are present only in the newest Z-Wave plus devices, as far as I remeber I changed one PLUS definition from 1 byte to 2 bytes some time ago. This was at the beginning 1 byte because older version of the manual stated this is 1 byte parameter, but in newer version of the doc it became 2 bytes param. 
But again - old devices work very well with this param defined as 1 byte and unless you test that 2 bytes works as well, I wouldn't touch this.

PR #3719, opened 2016-01-06
Fix a bug preventing proper logout from energenie devices

hmerk:
As reported in the community, there is a bug in senLogOut() which will be fixed with this PR.
The bug only occurs when using the WebInterface or Smartphone App while openHAB is running.
Cause I only use openHAB for controlling my energenie devices, this never happend in my environment.

Signed-off-by: hmerk <hans-joerg.merk@t-online.de>


teichsta:
also cherry-picked into 1.8

PR #3720, opened 2016-01-06
Added Ecobee and Nest to features.

watou:
Adding to 2.0 compatible action and bindings as in #3614.

PR #3722, opened 2016-01-07
Corrected Ecobee action dependency

watou:
I think this fixes the build problem I introduced with #3720.  I did a full mvn clean install locally and it gave
```
[INFO] Verification of feature openhab-action-ecobee/1.8.0.SNAPSHOT succeeded
```

### DIFF
--- a/features/openhab-addons/src/main/feature/feature.xml
+++ b/features/openhab-addons/src/main/feature/feature.xml
@@ -6,6 +6,7 @@
     <feature name="openhab-action-ecobee" description="Ecobee Action" version="${project.version}">
         <feature>openhab-runtime-base</feature>
         <feature>openhab-runtime-compat1x</feature>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
If the action has a dependency on the binding then this is indeed required. The good thing is that openHAB 2 will then automatically install it for you :-)

watou:
Thank you, @kaikreuzer.  Could you cherry pick it into the 1.8 branch, too? (I managed to break that build, too.)

I know there's a good argument to refactor the action/binding with an io bundle, but that would have badly complicated the usage of the action bundle.

teichsta:
> Could you cherry pick it into the 1.8 branch, too? 

done!

kaikreuzer:
Thanks. Just launched the Cloudbees build and wanted to see that it is green first.

PR #3731, opened 2016-01-08
Souliss - Fix typical T18 state mapping - fixes  #3730

### LINKED ISSUE #3730: Souliss binding: T18 states are not mapped correctly
I'm using Souliss and OpenHAB to control a latching switch (Typical T18).

Unfortunately the returned code from Souliss are 0x24 (OFF) & 0x23 (ON) as implemented in

https://github.com/souliss/souliss/blob/friariello/base/T1n.cpp#L878
https://github.com/souliss/souliss/blob/friariello/Typicals.h#L84

xalopp:


teichsta:
Thanks @xalopp also cherry-picked into 1.8

PR #3735, opened 2016-01-08
Adding support for multiple station positions for Netatmo binding

jannegpriv:
When testing the Netatmo binding on branch 1.8 with my 2 different Netatmo stations, one at my
home and the other at our summer house, I discovered that the code handled several
devices correctly but does not support several station positions.
Station position contains long/lat and altitude for a station using a PointType and when I tested I saw that
I got the same altitude for both my stations which is not correct.

The change adds the possibility to have several station positions since the binding already supports several devices.

The change also updates the getstationsdata.json with content from a query returning station data from 2 different stations/devices.
It also corrects the getstationsdata.json for the module type NAModule4 (extra indoor module) which contains CO2 and not Noise KPI in the dashboard_data.
It also adds some lacking KPIs in dashboard_data structures, e.g. temp_trend and pressure_trend.

The JUnit-tests are updated to verify altitude from both stations/devices. I know that addition doesn't verify the code change but it verifies that the code correctly parses a getstationdata.json containing content from 2 different stations.

I've tested the code change locally and altitudes for my 2 stations are now displayed correctly.

This is my first ever contribution so please be nice to me if I've screwed something up :-)

### DIFF
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
@@ -266,8 +266,7 @@ public class NetatmoBinding extends
 									if (stationPositions.get(device) == null) {
 										DecimalType altitude = DecimalType.ZERO;
 										if (device.getAltitude() != null) {
-											altitude = new DecimalType(Math.round(unitSystem.
-													convertAltitude(device.getAltitude())));
+											altitude = new DecimalType(device.getAltitude());
 										}
 										stationPositions.put(device, new PointType(
 												new DecimalType(new BigDecimal(device.getLatitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
@@ -282,7 +281,8 @@ public class NetatmoBinding extends
 										state = stationPositions.get(device).getLongitude();
 										break;
 									case ALTITUDE:
-										state = stationPositions.get(device).getAltitude();
+											state = new DecimalType(Math.round(unitSystem.
+													convertAltitude(stationPositions.get(device).getAltitude().doubleValue())));
 										break;
 									case WIFISTATUS:
 										state = new DecimalType(
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
@@ -77,7 +77,7 @@ public class NetatmoBinding extends
 	 */
 	private long refreshInterval = 300000;
 
-	private PointType stationPosition = null;
+	private Map<Device, PointType> stationPositions = new HashMap<Device, PointType>();
 
 	private Map<String, OAuthCredentials> credentialsCache = new HashMap<String, OAuthCredentials>();
 
@@ -262,35 +262,34 @@ public class NetatmoBinding extends
 						case STATIONNAME:
 							for (Device device : oauthCredentials.getStationsDataResponse
 									.getDevices()) {
-								if (stationPosition == null) {
-									DecimalType altitude = DecimalType.ZERO;
-									if (device.getAltitude() != null) {
-										altitude = new DecimalType(Math.round(unitSystem.
-												convertAltitude(device.getAltitude())));
-									}
-									stationPosition = new PointType(
-											new DecimalType(
-													new BigDecimal(device.getLatitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
-											new DecimalType(new BigDecimal(device.getLongitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
-											altitude);
-								}
 								if (device.getId().equals(deviceId)) {
+									if (stationPositions.get(device) == null) {
+										DecimalType altitude = DecimalType.ZERO;
+										if (device.getAltitude() != null) {
+											altitude = new DecimalType(Math.round(unitSystem.
+													convertAltitude(device.getAltitude())));
+										}
+										stationPositions.put(device, new PointType(
+												new DecimalType(new BigDecimal(device.getLatitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
+												new DecimalType(new BigDecimal(device.getLongitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
+												altitude));
+									}
 									switch (measureType) {
 									case LATITUDE:
-										state = stationPosition.getLatitude();
+										state = stationPositions.get(device).getLatitude();
 										break;
 									case LONGITUDE:
-										state = stationPosition.getLongitude();
+										state = stationPositions.get(device).getLongitude();
 										break;
 									case ALTITUDE:
-										state = stationPosition.getAltitude();
+										state = stationPositions.get(device).getAltitude();
 										break;
 									case WIFISTATUS:
 										state = new DecimalType(
 												device.getWifiLevel());
 										break;
 									case COORDINATE:
-										state = stationPosition;
+										state = stationPositions.get(device);
 										break;
 									case STATIONNAME:
 										state = new StringType(
--- a/bundles/binding/org.openhab.binding.energenie/src/main/java/org/openhab/binding/energenie/internal/EnergenieBinding.java
+++ b/bundles/binding/org.openhab.binding.energenie/src/main/java/org/openhab/binding/energenie/internal/EnergenieBinding.java
@@ -223,7 +223,7 @@ public class EnergenieBinding extends AbstractActiveBinding<EnergenieBindingProv
 
 	private void sendLogOut(String pmsId) {
 		String pmsIp = pmsIpConfig.get(pmsId);
-		String url = "http://"+pmsIp;
+		String url = "http://"+pmsIp+"/login.html";
 
 		try {
 			HttpUtil.executeUrl("POST", url, timeout);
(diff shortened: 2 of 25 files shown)


bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -262,35 +262,34 @@ protected void execute() {
...
-													new BigDecimal(device.getLatitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
-											new DecimalType(new BigDecimal(device.getLongitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
-											altitude);
-								}
 								if (device.getId().equals(deviceId)) {
+									if (stationPositions.get(device) == null) {
+										DecimalType altitude = DecimalType.ZERO;
+										if (device.getAltitude() != null) {
+											altitude = new DecimalType(Math.round(unitSystem.
+													convertAltitude(device.getAltitude())));
```
watou: Note that a valid PointType expresses its altitude in meters above sea level.  So converting the unit system of altitude before constructing the PointType is invalid if the unit conversion is not guaranteed to result in meters above sea level.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -262,35 +262,34 @@ protected void execute() {
...
-										state = stationPosition.getLatitude();
+										state = stationPositions.get(device).getLatitude();
 										break;
 									case LONGITUDE:
-										state = stationPosition.getLongitude();
+										state = stationPositions.get(device).getLongitude();
 										break;
 									case ALTITUDE:
-										state = stationPosition.getAltitude();
+										state = stationPositions.get(device).getAltitude();
```
watou: consider doing the unit conversion here instead?

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
jannegpriv: I've not changed where unit conversion is done, but if the current implementation is faulty I can of course change it.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
jannegpriv: Sure, I can do it here instead.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
watou: I would recommend that.  The third DecimalType argument to the PointType constructor must be meters above sea level, not feet or another unit.  Thanks.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -262,35 +262,34 @@ protected void execute() {
...
+									if (stationPositions.get(device) == null) {
+										DecimalType altitude = DecimalType.ZERO;
+										if (device.getAltitude() != null) {
+											altitude = new DecimalType(Math.round(unitSystem.
+													convertAltitude(device.getAltitude())));
+										}
+										stationPositions.put(device, new PointType(
+												new DecimalType(new BigDecimal(device.getLatitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
+												new DecimalType(new BigDecimal(device.getLongitude()).setScale(6, BigDecimal.ROUND_HALF_UP)),
+												altitude));
```
watou: While you're at it, I don't see the reason to potentially decrease the resolution of a GPS coordinate with `.setScale(6, BigDecimal.ROUND_HALF_UP)`.  If that makes sense in some non-obvious way, I suppose leave it, but my preference is to not misplace the coordinate through unnecessary rounding.

robnielsen: The resolution isn't decreased. It's exactly the same resolution that is provided by Google Maps which Netatmo uses when setting location. 6 digits of resolution is 1m. Please leave as is.

watou: @robnielsen If you know the resolution isn't decreased, then you must know the data before it arrives in the future, and there is no reason to round, either.
@jannegpriv Please feel free to keep this PR focused on the fixes you are intending to provide, and thanks for your contribution.

jannegpriv:
@robnielsen I've not changed the location resolution and since my pull request was aiming at supporting multiple station positions, I will not do that change.

The last commit moves the place for altitude unit conversion as @watou suggested.
Can you please review that change?

watou:
@jannegpriv Your changes look good to me.  Could you squash your commits into 1?  I  don't know if there is time to include this in 1.8; that's for @teichsta to decide.  Thanks again for your fixes!

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
robnielsen: @jannegpriv, please do not make the change that @watou is suggesting. As I stated please leave as it was since it was this way before your PR.  @watou, it appears that Netatmo is sending the data in a way that introduces extra digits, and the changes that @jannegpriv are modifying more than the PR is adding in new functionality.

robnielsen:
@jannegpriv, please revert the change that @watou suggests. 6 digits of precision is all that they can get from Google Maps when you select the location. For some reason they must be storing in a way that introduces extra digits.

watou:
@robnielsen Are you saying Netatmo is putting bad data after the sixth decimal place?  How do you know 1) where Netatmo gets its geocoded coordinates, or 2) that losing precision of a GPS coordinate from what was provided by the API is a good idea?  I still don't understand your motivation here (you could just let us know in plain English), but I won't pollute @jannegpriv's PR any further with this subject.

robnielsen:
@watou, I have articulated my reasons. Clearly you haven't configured a Netatmo device, or you'd understand what I've said. As the active maintainer of the Netatmo binding, I'd hope that my input would have some value.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
teichsta: Guys, thanks for arguing so hard!

My standpoint is quite simple: we should work with the values that comes from the API. Whatever value with whatever number of digits is being delivered, it's good enough. Any other position than one being delivered by the API is unexpected. I don't see a reason to round here. However, since the code was already there i would tend to leave the scaling in and change that with another PR.

Please forgive me that i didn't observed (and discussed) that before. I must have overlooked this piece of code.

teichsta:
@watou @robnielsen see my comments above. I would like to process this PR (once the commits have been squashed) and create a follow-up PR to remove the scaling here.

@jannegpriv i would tend to not merge this PR into 1.8 anymore since it is a clean enhancements. What do you think?

jannegpriv:
I think I might know why Netatmo API is returning different amount of decimals for long/lat for different stations.
When configuring position in the Netatmo App or website you can either let Netatmo use the Smartphones GPS coordinates or manually enter the position using Google maps. If you don't configure position manually, the [Netatmo manual](http://www.vesternet.com/downloads/dl/file/id/283/product/1299/netatmo_connected_weather_station_faq.pdf) in chapter 11.3 states that position is automatically configured using the smartphones internal GPS.

I've done some tests on my local Netatmo station and when configuring location manually using Google maps and entering my street name, then I get the following json for place:
  "place": {
          "altitude": 58,
          "city": "Huddinge",
          "country": "SE",
          "improveLocProposed": true,
          "location": [
            18.0006368,
            59.2230361
          ],
          "timezone": "Europe\/Stockholm"
        },       

When instead of entering the street name manually in Google maps, pressing the "Would you like to use your current location" then letting Netatmo take the position from my MBP I get the following json
 "place": {
          "altitude": 58,
          "city": "Huddinge",
          "country": "SE",
          "improveLocProposed": true,
          "location": [
            18.001556832332,
            59.222217282513
          ],
          "timezone": "Europe\/Stockholm"
        },
 
MBP is using WiFi-triangulation so there is some room for fault marginals in that location.

Doing the same in my iPhone Netatmo App using iPhones internal GPS chip give the location I get the following json:
 "place": {
          "altitude": 58,
          "city": "Huddinge",
          "country": "SE",
          "improveLocProposed": true,
          "location": [
            18.001464010777,
            59.222091894443
          ],
          "timezone": "Europe\/Stockholm"
        },

So Google maps is using 6 decimals but auto-configuring using computer/smartphone seems to give more decimals, in this cases 12 decimals.

jannegpriv:
@teichsta That's fine with me, just glad if someone else in the future is helped with the added support. :-)

I will squash the commits into one. 
NOTE: The last commit does not alter how long/lat is scaled, only when the altitude unit conversion is being made.

watou:
Closing #3735 because #3743 makes identical changes in one commit.  Thank you, @jannegpriv!

robnielsen:
Thanks @teichsta, your standpoint makes sense, and is basically what I was asking for this PR.

Issue #3700, opened 2016-01-03
Popp  (POPE004001) Z-Wave Smoke Detector and Siren


wdriessen:
Please add the folowing device to the database:
http://www.pepper1.net/zwavedb/device/725

cdjackson:
Please advise if this can be closed once tested.

svenschaefer74:
I just tested it with the latest binding from https://openhab.ci.cloudbees.com/job/openHAB/1121/. I excluded the sensor und re-included it but the log still states that there is no db entry.

    2016-01-10 10:05:29.471 [DEBUG] [.z.i.config.ZWaveConfiguration:276 ]- NODE 35: No database entry: POPP [ID:201,Type:100]

I've also attached the newly generated xml file:
[node35.xml.txt](https://github.com/openhab/openhab/files/84034/node35.xml.txt)

I found one difference in the products.xml file where I originally used three digits for deviceid and devicetype where the other POPP devices used four. This might be relevant but I'm not sure. I could issue a pull request if this would be relevant.

cdjackson:
Please can you provide a log file on startup.  The 3 versus 4 characters won’t make any difference as the database interprets this as a number, not a string.

svenschaefer74:
Here is the log file:
[zwave.log.txt](https://github.com/openhab/openhab/files/84072/zwave.log.txt)

Additional info: 
- With the snapshot binding from January 3rd everything was fine.
- The device that got lost is a Fibaro FGS222. The only related issue I've seen was #3710.

cdjackson:
So this isn’t related to the POPP smoke detector then?  We probably should track the issue with the FGS222 separately…

cdjackson:
For the POPP, can you delete node35.xml and restart and resend the log please.

svenschaefer74:
Sorry. My fault.

The device we're talking here is a missing entry in the zwave db for the POPP 004001 smoke detector. I shouldn't have mentioned the FGS222 here. The FGS is related to issue #3742. I'll move the additional info from above to the right plave. Sorry again.

svenschaefer74:
Done.
[node35.xml.txt](https://github.com/openhab/openhab/files/84077/node35.xml.txt)

svenschaefer74:
I reviewed the products.xml file again and found that I mixed up the type and the id of the device. *ugh* I patched my local binding and now the HABmin interface shows me the right information besides the device class which is GARAGE_DOOR but this might be correct in terms of zwave terminology.

[...]

PR #3736, opened 2016-01-09
Added POPP 004001 to resolve issue #3700

svenschaefer74:
Added POPP 004001 Smoke detector with internal Siren to zwave database.

Alll data was pulled from Pepper1.


cdjackson:
#3700

PR #3737, opened 2016-01-09
Added cause of ConnectException to the log call

kwave:
As it was impossible to determine the cause of a failed connection, I added the (formerly discarded) exception to the debug call so the user is able to see it.


teichsta:
Thanks @kwave, also cherry-picked into 1.8

PR #3779, opened 2016-01-13
Remove the switch_all command class from the default command classes …

cdjackson:
…for switches

This isn't a mandatory class and is causing problems with devices that do not support this class.

@xsnrg FYI - regarding the garage door sensor


xsnrg:
Thanks!

cdjackson:
@teichsta Can this be picked into 1.8.1 please

Issue #3784, opened 2016-01-13
[serialbinding] Multiple serial items with regex causes warn messages "Unable to convert regex result ..."


csowada:
If you use multiple items in your configuration with the new REGEX parameter then it causes warn messages like:

``20:14:28.684 WARN  o.o.b.s.internal.SerialDevice[:244] - Unable to convert regex result '%SERIALDATA%' for item %ITEMNAME% to number``

We have to check if the translation service input is unequal to the result or the serial binding. Without this check the item will be set with the original serial data or causes a exception for numbers.

csowada:
Okay, as suggested in the PR I will fix the translation service for regex directly.

PR #3785, opened 2016-01-13
Fix for #3784 serial binding regex service.

csowada:
Just add a line to check transformation sercive result

### DIFF
--- a/bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialDevice.java
+++ b/bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialDevice.java
@@ -237,14 +237,18 @@ public class SerialDevice implements SerialPortEventListener {
 								} else {
 									try {
 										String value = transformationService.transform(entry.getValue().pattern, result);
-										if(entry.getValue().type.equals(NumberItem.class)) {
-											try {
-												eventPublisher.postUpdate(entry.getKey(), new DecimalType(value));
-											} catch (NumberFormatException e) {
-												logger.warn("Unable to convert regex result '{}' for item {} to number", new String[] { result, entry.getKey()});
+										
+										// check if regex service transformed the data
+										if(!value.equals(result)) {
+											if(entry.getValue().type.equals(NumberItem.class)) {
+												try {
+													eventPublisher.postUpdate(entry.getKey(), new DecimalType(value));
+												} catch (NumberFormatException e) {
+													logger.warn("Unable to convert regex result '{}' for item {} to number", new String[] { result, entry.getKey()});
+												}
+											} else {
+												eventPublisher.postUpdate(entry.getKey(), new StringType(value));
 											}
-										} else {
-											eventPublisher.postUpdate(entry.getKey(), new StringType(value));
 										}
 										
 									} catch (TransformationException e) {


watou:
It is valid for a transform to return an identical string to the one it was passed.

csowada:
For the translation service it is okay, but it makes no sence for regex in the binding.In this case I will filter the serial input for a specified part of the data. I we found nothing, we can ignore the serial data for that item.

csowada:
Here the related topic from the communit >> https://community.openhab.org/t/regex-help-needed-serial-binding/5449/15

watou:
One could argue that it is incorrect that the regex transform returns the whole string when the capture fails (instead of throwing TransformationException as it should), but I don't think adding the special case `if(!value.equals(result)) { ... }` in the binding is the way to address it.  With your change, any regex transform that intends to capture the entire string will return false for the new `if(!value.equals(result)) { ... }` section, which means they will be dropped and no state update posted.  This would lead to unwanted side effects for other valid use cases.

In OH1, RegexTransformation returns the source string.  In ESH, it returns null.  I think both should instead throw TransformationException.

csowada:
I wanted to correct an error seen in my eyes here. I've added the basic changes for RegEx with this Commit (https://github.com/openhab/openhab/commit/12b88aeffb2095ce1517ae324638449b25f12d82). After the commit, I was asked if I could use the RegEx Transfomation Service. I then changed it (https://github.com/openhab/openhab/commit/87124d0e8a6a55bec35d83089168ce262a55bdb9) and have not noticed the changed behavior.

I would see the REGEX parameter just as __a__ parameter of the serial bindings. In this case REGEX has nothing to do with transformation, even if it internally uses this service.

Here, an example of serial data via the interface. The gateway received data and sends each line to the Serial Binding.

    {"id": 5, "temp": 23.1, "hum": 61, "RSSI": -67}
    {"id": 6, "temp": 22.6, "hum": 59, "RSSI": -59}
    {"id": 7, "temp": 20.6, "hum": 55, "RSSI": -61}

In this case I use RegEx to filter for the correct ``id`` and then extract the value for example ``temp``. If it is the wrong ``id`` I ignore the string.

But with the current behavior it only works when just the correct string for this item is sent.
If a string with the wrong ``id`` is sent, my item will be updated with with the complete string (if it is a text item is) or we get an error message.


In my eyes the main question is whether the term ``REGEX`` must always belong to transform service?

Case NO: In this case the binding writes wrong data to items if we use more than one item with RegEx on the same interface

Case YES: ``REGEX`` is reserved for the RegEx transformation service. We should use another parameter like ``EXTRACT`` to extract data via RegEx or skip if RegEx does not match.

watou:
In my opinion, "transform" and "extract" mean the same thing.  The serial binding should support any transformation service, and the point of the transformation services is to transform an incoming string into another string.  The RegEx transformation service does this by capturing matched text and returning it.  But unlike the other transformation services, in OH1 RegEx transformation returns the input string instead of throwing a TransformationException.  In ESH/OH2 it instead returns null, which might be an improvement, but really it should instead throw a TransformationException if it can't capture.

So your PR is trying to work around what I believe is a flaw in the RegEx transformation service in OH1, but 1) this binding will work differently under ESH/OS2 with your PR, 2) your workaround misses the valid case where the entire string is captured, and 3) I believe ESH/OS2 is still wrong not to throw a TransformationException when it cannot capture the string.

My advice is to close this PR and open new issues against OH1 and ESH to finally fix the RegEx transformation service to make it behave consistently with the other transformation services, where a TransformationException is thrown when it fails to transform.

I would like @teichsta to chime in on this.

csowada:
Okay, I agree with you. I had not OH2 and the other translation services on the radar.
So it's better to fix it there. But maybe this change could break other bindings.

watou:
> But maybe this change could break other bindings.

Oh, almost certainly, unless every place that transforms are used is closely examined.  But the RegEx transform has to be fixed to work like all the others, at least in ESH (and maybe in 1.8.* too).

teichsta:
> My advice is to close this PR and open new issues against OH1 and ESH to finally fix the RegEx transformation service to make it behave consistently with the other transformation services, where a TransformationException is thrown when it fails to transform.

Agree with @watou! So would you be so kind to create the according PRs @csowada?

PR #3789, opened 2016-01-13
Added Popp Smoke Detector with Siren to zwave database

svenschaefer74:


cdjackson:
Why are there so many commits associated with this PR? Given you've only changed 2 lines, I think this should be fixed.

Issue #2486, opened 2015-04-20
serial binding in 1.6.2 baudrate - not work


roblad:
Hi,

I tryied to use serial binding with

{ serial="/dev/ttyUSB0@57600" }

it generates errors

015-04-20 21:45:28.507 [ERROR] [i.internal.GenericItemProvider] - Binding configuration of type 'serial' of item \u2018TtyUSB0_receiving\u2018 could not be parsed correctly.
org.openhab.model.item.binding.BindingConfigParseException: Could not open serial port /dev/ttyUSB0@57600: Serial port '/dev/ttyUSB0@57600' could not be found. Available ports are:
/dev/ttyUSB0

```
at org.openhab.binding.serial.internal.SerialBinding.processBindingConfiguration(SerialBinding.java:118) ~[na:na]
```
##

steve-bate:
@roblad Do you know if this is still an issue with 1.7.*?

csowada:
I use such an Item 
``Number Temperature_FF_Bath		"Temperature [%.1f °C]"					<__temperature>		(Temperature, FF_Bath)		{serial="/dev/ttyUSB_HN@38400,REGEX(.*T:([0-9.]*),.*D:5.*)"}``
 and for me it works perfect.

Maybe an issue with the configuration encoding?

hmerk:
Seems to be solved

PR #3793, opened 2016-01-14
Replaced unicode single quotes with standard single quotes

9037568:
The error message in #2486 displays badly due to the use of unicode single quotes.  This patch replaces them with standard single quotes.


kaikreuzer:
Please note that this class does not exist anymore in 1.9 (soon to be deleted).
I have ported the fix to https://github.com/eclipse/smarthome/pull/851

PR #3800, opened 2016-01-15
removed old core/runtime bundles now being available via ESH, also ch…

teichsta:
…anged execution environment to 1.7.

To make the build work again i had to exclude some bundles from the build that we'll have to fix the coming days. The bundles are:

* org.openhab.action.xpl
* org.openhab.binding.mqtt.test
* org.openhab.binding.systeminfo.test
* org.openhab.binding.fritzaha
* org.openhab.binding.xpl
* org.openhab.binding.sagercaster
* org.openhab.io.cv
* org.openhab.io.console
* org.openhab.io.multimedia
* org.openhab.io.myopenhab
* org.openhab.io.transport.xpl
* org.openhab.persistence.rrd4j

Signed-off-by: Thomas Eichstädt-Engelen <thomas@openhab.org>

PR #3801, opened 2016-01-15
applied the ESH code-cleanup rules

teichsta:
please apply after #3800!


kaikreuzer:
You should use the opportunity and also update the license headers to 2016 in this PR!

robnielsen:
I assume this also might break most if not all of the open pull requests?

teichsta:
> I assume this also might break most if not all of the open pull requests?

yes, true … but one day we would have to finally bite the bullets!

PR #3804, opened 2016-01-16
Rename duplicate influxdb project Eclipse

idserda:
There were two projects named 'influxdb', this renames the second one to 'influxdb08'.


teichsta:
Thanks @idserda, also cherry-picked into 1.8

PR #3807, opened 2016-01-16
Adding missing dependency libraries

lewie:


teichsta:
Thanks @lewie! Shouldn't the pom.xml be adjusted as well (remove copying the artifacts)?

lewie:
You're right! Changed pom too now.

lewie:
@teichsta, 
JDBC Service works well in OH2 karaf.
After this PR, JDBC could be a integrated in OH2 too.

Why it would be better to rewrite it as real OH2 bundle:
For running in OH1 while Installation, the needed database driver has to be copied too to addons folder. Copying by hand for OH1 is OK, but is not a good idea when using in OH2 I think.

OH2 can load dependencies dynamically via maven.
This seems to be the ideal way to do with the driver bundles for JDBC Persistence bundle, isn't it?

The other thought is, rewriting for OH2 would cause code duplication...

What do you think about?

watou:
@lewie, does this fix a problem in OH 1.8.0 or is it strictly for OH2 compatibility?

lewie:
@watou,
previously I used a maven plugin to load Binding dependencies to the Project /lib folder.
But as usual in openHAB conventions, the dependency jar's (not the Drivers) should be copied statically to /lib folder.

watou:
@lewie, so this PR would result in the JDBC persistence service working under openHAB 2 (per [this comment](https://github.com/openhab/openhab/pull/3807#issuecomment-172902492)), and without this PR, the JDBC persistence service does not work under openHAB 2?  I'm just trying to figure out which milestone it should be tagged with.

PR #3809, opened 2016-01-16
Fix for defect 3808, remove extra Unicode character 0x0xfeff.

robnielsen:
This fixes https://github.com/openhab/openhab/issues/3808

### DIFF
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
@@ -6,7 +6,7 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
-﻿/**
+/**
  * Copyright (c) 2010-2015, openHAB.org and others.
  *
  * All rights reserved. This program and the accompanying materials


robnielsen:
The change isn't showing up when looking at the changes. This is what git diff produces:
> rob@rob-ubuntu:~/openhab/bundles/binding/org.openhab.binding.myq$ git diff .diff --git a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
index a839abe..4a89e2f 100644
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
@@ -6,7 +6,7 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
-<U+FEFF>/**
+/**
  * Copyright (c) 2010-2015, openHAB.org and others.
  *
  * All rights reserved. This program and the accompanying materials

9037568:
Well, since it's a ZERO-WIDTH space, it wouldn't show up ever would it?

But it appears that this file now has two nearly identical headers at the
top. Shouldn't one of them (eg. lines 9-16) be removed entirely?

robnielsen:
Maybe so, but javac doesn't like it and the build fails.

teichsta:
Thanks @robnielsen! I will also remove the second javadoc block in the license header soon.

PR #3812, opened 2016-01-16
MariaDB compatibility update

lewie:
See: https://github.com/openhab/openhab/issues/3788

MariaDB-driver (tested with mariadb-java-client-1.3.4.jar) now works correct with MySql-Server and MariaDB-Server.


teichsta:
has been cherry-picked onto 1.8 already

PR #3813, opened 2016-01-16
[weather] Add name examples to default weather binding config

watou:
Add 
```
#weather:location.<locationId1>.name=
#weather:location.<locationId2>.name=
```
to weather binding defaults in openhab_default.cfg.

From [forum request](https://community.openhab.org/t/openhab-1-8-default-cfg-weather-location-locationid1-name-missing/6333).

PR #3814, opened 2016-01-16
[weather] Added name examples to default weather.cfg (like #3813)

watou:
Same as #3813, but for master branch (OH2).


teichsta:
Thanks @watou!

watou:
Thanks, @teichsta.  Could you also cherry-pick this onto the 1.8 branch, just so weather.cfg isn't left un-updated there?

PR #3815, opened 2016-01-17
Netatmo Welcome Binding

wep4you:
I implemented a binding for the Netatmo Welcome camera. I used some of the classes from Netatmo Binding, so they are now dublicate, would be better to get them out of both bindings to be used over both bindings. On Netatmo Developer Utility you should create a new Application, to don't get in conflict with between the Netatmo weather station and the camera if you have both.

I have implemented all from the GetHomeData API, just events are not final, but its not possible to generate a dynamic list in OpenHab UI so the information is at the moment not really needed and could be implemented later.

For Presence detection for known and unknown People, and Information  about the Home and the camera the binding works, maybe ill try to implement the Face pictures and the movie stream in the future.

### DIFF
--- a/bundles/binding/org.openhab.binding.netatmowelcome/.classpath
+++ b/bundles/binding/org.openhab.binding.netatmowelcome/.classpath
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
+	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="output" path="target/classes"/>
+</classpath>
--- a/bundles/binding/org.openhab.binding.netatmowelcome/.project
+++ b/bundles/binding/org.openhab.binding.netatmowelcome/.project
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.openhab.binding.netatmowelcome</name>
+	<comment>This is the NetatmoWelcome binding of the open Home Automation Bus (openHAB)</comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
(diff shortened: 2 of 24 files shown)


watou:
Thank you, @wep4you, for your contribution.  What are the reasons for creating a binding that is separate from the existing Netatmo binding?  I think there are reasons to have this capability included in the existing binding, such as avoiding duplicating code, and using the same API and credentials as for other Netatmo products.  This was the approach with the Nest binding (one binding for all Nest products, instead of separate bindings depending on which Nest products you want to access).

I would like to understand your thinking, to be certain we don't head down a suboptimal path.  Also note that there is a new Netatmo binding for openHAB 2.  Thanks again.

wep4you:
At the beginning i tried to implement this in the Netamo binding, but i couldn't handle the code side by side without to much refactoring. To clean different the two devices i decided to implement an own binding because the item definition in in some cases are really identical, like devices and cameras, so to handle the item config also should get the info wicht device it is for. With different bindings its clear, because the binding name is different. Also the refresh token has to be refreshed before every call of the gethomedata to get the new state. Also there are new different scopes in the Netamo API, so it has to be handled by configuration, to not have have to use all scopes, for all devices.
I already had problems if i use the same Refresh key for both Api calls, sometimes one of the both bindings get an error and didn't work, with different Apps i didn't get interferences.
I think it would better to implement the classes witch are used in both bindings outside and use them in both projects.

This was my really first try with git and share a project so im not so experienced, and didn't want to complete reprogram a code wich was not from me ;-)

Thanks, i didn't have the OpenHab2 at the moment, ill try to set up the development environment for it and have a look at it. For the moment the binding did what it should and works in my home.

watou:
Thank you, Peter, for taking the time to explain your objectives and reasonable thinking.  I will leave this subject for the project owners, @kaikreuzer and @teichsta, to chart a path for Netatmo support in openHAB 1.x and 2.x, so there is minimal confusion and code maintenance.  And thank you for your contribution to the openHAB community!

robnielsen:
@wep4you, @watou I think that having two netatmo bindings could be confusing for end users. Netatmo has 3 different types of devices: weather, thermostat and cameras (welcome and presence). @clinique has added support for both weather and thermostat to the OH 2 binding, I'm thinking that this might be a better place for camera support.

wep4you:
@robnielsen @watou I think it should be possible to integrate it in one binding, but for backward compatibility in the configuration we need to differentiate it in the item definition, and have some settings put in the configuration. I haven't had a look for OH2, but would be good to implement it in the same binding from @clinique

wep4you:
@watou @robnielsen 
I managed to implement the welcome binding in the netatmo binding, ill pull the netatmo binding and close this request.

watou:
That's great, Peter! Much appreciated.

PR #3819, opened 2016-01-17
Modbus TCP connections closed after every transaction

### LINKED ISSUE #3818: Modbus binding should close TCP connection after every transaction
During 1.8.0-SNAPSHOT development there was regression caused by my pull request [#3512](https://github.com/openhab/openhab/pull/3512) that broke down installation for some users (for details consult [this community discussion](https://community.openhab.org/t/modbus-connection-problem/6108)). In the PR I had removed the code that closes TCP connection after every modbus transaction, i.e. after every read or write operation, ([original pull request](https://github.com/openhab/openhab/pull/3228) from @teichsta to [issue #658](https://github.com/openhab/openhab/issues/658)). As result of this the binding keeps the connections open as long as possible, and reconnecting only in failure situations.

However, as reported by @wtrucci, this is problematic behaviour when the modbus server is only able to serve one client at a time, and the user wishes to access the modbus server at the same time openhab is connected to it (as in the case of @wtrucci). Furthermore, if there are many slaves defined in openhab.cfg that point to the same modbus server (e.g. with different valuetype), many connections would be established to the same server.

ssalonen:
We want to close connection after every transaction since some modbus
tcp servers can accept only single connection at a time. If we
would keep the connection open, the server would be blocked for other
users.

Fixes #3818

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/io/ModbusTCPTransaction.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/io/ModbusTCPTransaction.java
@@ -212,12 +212,7 @@ public class ModbusTCPTransaction implements ModbusTransaction {
                 throw new ModbusSlaveException(((ExceptionResponse) m_Response).getExceptionCode());
             }
 
-            // 6. close connection if reconnecting
-            if (isReconnecting()) {
-                m_Connection.close();
-            }
-
-            // 7. Check transaction validity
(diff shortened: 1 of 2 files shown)


wtrucci:
I can confirm that now it's working again. Thanks for the fix.

Walter

watou:
This PR lgtm, but I will leave for @teichsta to merge and cherry-pick onto 1.8.  Thanks!

teichsta:
Thanks @ssalonen! Also cherry-picked onto 1.8

PR #3820, opened 2016-01-17
Doc update

dvanherbergen:
- Remove links to cloudbees snapshots
- Updated IDE and contributing links to openhab-distro

Signed-off-by: Davy Vanherbergen <davy.vanherbergen@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -243,7 +243,7 @@
         <!-- openHAB dependencies p2 repository -->
         <repository>
             <id>p2-openhab-deps-repo</id>
-            <url>http://repository-openhab.forge.cloudbees.com/release/openhab-deps-repo/${ohdr.version}</url>
+            <url>https://dl.bintray.com/openhab/p2/openhab-deps-repo/${ohdr.version}</url>
             <layout>p2</layout>
         </repository>
 


watou:
So there won't be 1.8 CI builds for future point releases?  I worry (generally) about eroding the ability to document, support, and maintain 1.x (as it's the only production release at the moment).

dvanherbergen:
> So there won't be 1.8 CI builds for future point releases? 

At the moment, there is indeed no CI build for potential 1.8.x runtime releases.
The CI is currently building the 1.9.0-SNAPSHOT, which will have addons compatible with the 1.8 runtime, but no runtime. 
If a point release is necessary, a new build from a 1.8.x branch can be setup.

watou:
Thanks.  
> If a point release is necessary

A 1.8.1 is already envisioned, and likely further ones as well until 2.x is suitable for all 1.x users (IMHO).

dvanherbergen:
> A 1.8.1 is already envisioned

OK. In that case it would make sense to add a CI build for that. I'm just a little concerned about total build duration vs available build minutes.  I would prefer to hold off on adding a new build until the current ones have been optimised. The current builds are not exactly stable yet in terms of duration.

Would you need to have the designer included in 1.8.x builds too? If we could leave that out, it would save a lot of time/space.

watou:
Completely understand your thinking.  I am not certain, but I think a full 1.x build would make the most sense (weighed against the resource costs).  And whatever is done to decrease build time/space will be very welcome.

kaikreuzer:
Note that we never had CI builds for patch releases (1.7.1, etc.). Maybe @teichsta will provide such a build on his separate Jenkins instance, so that we do not run into build minutes issues on Cloudbees.

dvanherbergen:
It looks like I have overwritten the commit on this PR. It is no longer correct...

PR #3822, opened 2016-01-17
Support for Plex Media Servers with required secure connections

### LINKED ISSUE #3715: Plex binding is hardcoded to use http, fails to connect to server via https
Hi there,

@idserda, after a quick glance at the code it seems to me that the Plex binding is building the URL to connect to the Plex server with a hardcoded "http". I've configured my server to require HTTPS:

7w39o2016mc12n016e95006ab7ii17irgod48

If possible, I'd really prefer to not change that. Do you think it's possible to make an option for using https or not? It might be not as trivial as changing the string to actually allow for HTTPS connections, due to how their SSL certificates work: https://blog.filippo.io/how-plex-is-doing-https-for-all-its-users/
That means, that when connecting via HTTPS from a local machine, the cert will fail:
```
user@test:/opt/openhab$ curl https://192.168.1.111:32400/
curl: (51) SSL: certificate subject name '*.8124bf18d3ae4c7c97712v4a48b1d5d2.plex.direct' does not match target host name '192.168.1.111'
```

With curl I can use the **--insecure** option to allow the connection to happen:
```
user@test:/opt/openhab$ curl --insecure https://192.168.1.111:32400/
<html><head><title>Unauthorized</title></head><body><h1>401 Unauthorized</h1></body></html>
```

So you'd probably have set a similar option in Java.

Thanks in any case, happy to help with anything I might have left unclear, or beta-testing. :)

idserda:
This fixes #3715

### DIFF
--- a/bundles/binding/org.openhab.binding.plex/src/main/java/org/openhab/binding/plex/internal/PlexConnectionProperties.java
+++ b/bundles/binding/org.openhab.binding.plex/src/main/java/org/openhab/binding/plex/internal/PlexConnectionProperties.java
@@ -8,19 +8,30 @@
  */
 package org.openhab.binding.plex.internal;
 
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * Holds the connection properties for a Plex server
- * 
+ *
  * @author Jeroen Idserda
  * @since 1.7.0
  */
 public class PlexConnectionProperties {
+
+    private static final Logger logger = LoggerFactory.getLogger(PlexConnectionProperties.class);
 
     private String host;
 
     private int port = 32400;
 
     private String token;
+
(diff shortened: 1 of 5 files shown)


teichsta:
Thanks @idserda! Would you consider this to be fixed in 1.8.1 as well? I'll cherry pick then …

idserda:
Yes that sounds like a good idea

teichsta:
Thanks @idserda!

halkeye:
So I have 1.9.0.b5, and I can't seem to figure out how to get it to https mode.
I assume this isn't yet in a build?
Is there somewhere I can get a nighly? or should I just compile it myself?

idserda:
There is a nightly build here: https://openhab.ci.cloudbees.com/job/openHAB1-Addons/1418/org.openhab.binding$org.openhab.binding.plex/artifact/org.openhab.binding/org.openhab.binding.plex/1.10.0-SNAPSHOT/org.openhab.binding.plex-1.10.0-SNAPSHOT.jar

Issue #3821, opened 2016-01-17
ExecUtil command execution raises exception when exitCode != 1


steve-bate:
The ```ExecUtil.executeCommandLineAndWaitResponse``` function specifies that the only valid exitCode is 1. Other exit codes cause an exception from the ```org.apache.commons.exec.Executor``` and results in a warning log message. I recommend we remove the code that is setting exit code == 1 as the only valid exit code.

@watou @kevin

watou:
[Link to forum report](https://community.openhab.org/t/openhab-1-8-and-2-0-beta1-release-is-out/6025/27)

kaikreuzer:
> the same is submitted to ESH. 

+1

steve-bate:
@watou has suggested we roll back this [commit](https://github.com/openhab/openhab/commit/49d0905581cec101a506a40a9dcdde3417ffbfcf). However, I'm not convinced yet that this is the best solution.

There are three contexts where this issue might arise. The exec binding, the exec action and the exec transformation. Only the action and the transformation use ```ExecUtil```, which was touched by the commit in question. Previously, ```ExecUtil``` silently ignored exceptions during command execution and logged a debug log message like:

    2016-01-17 08:49:57.096 [DEBUG] [g.openhab.io.net.exec.ExecUtil] - exit code '-559038737', result ''

Even if a user has enabled debug logging, this seems unnecessarily cryptic to me. The -559038737 exit code (0xdeadbeef in hex) is provided by the Apache Commons Exec library and not the actual return code of the command that was intended to be executed. After the change in question, there's a warning log message like:

    2016-01-17 08:54:03.041 [WARN ] [g.openhab.io.net.exec.ExecUtil] - Execution failed (Exit value: -559038737. Caused by java.io.IOException: Cannot run program "/tmp/testx.sh" (in directory "."): error=2, No such file or directory)

for a missing executable, or 

    2016-01-17 09:14:42.979 [WARN ] [g.openhab.io.net.exec.ExecUtil] - Execution failed (Exit value: -559038737. Caused by java.io.IOException: Cannot run program "/tmp/test.sh" (in directory "."): error=13, Permission denied)

for invalid permissions.

I believe these are useful log messages when a user is debugging a exec issue.

[...]

PR #3825, opened 2016-01-17
ExecUtil: fixed incorrect exitValue configuration.

steve-bate:
Issue #3821. 
Forum [topic](https://community.openhab.org/t/openhab-1-8-and-2-0-beta1-release-is-out/6025/27).

### DIFF
--- a/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/exec/ExecUtil.java
+++ b/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/exec/ExecUtil.java
@@ -112,7 +112,7 @@ public class ExecUtil {
         ByteArrayOutputStream stdout = new ByteArrayOutputStream();
         PumpStreamHandler streamHandler = new PumpStreamHandler(stdout);
 
-        executor.setExitValue(1);
+        executor.setExitValues(null);
         executor.setStreamHandler(streamHandler);
         executor.setWatchdog(watchdog);
 


watou:
The [JavaDoc for DefaultExecutor](https://commons.apache.org/proper/commons-exec/apidocs/org/apache/commons/exec/DefaultExecutor.html#setExitValues(int[])) does not state what the default behavior is for exit value handling, so I think the line you removed might instead be replaced with
```java
        executor.setExitValues(null);
```
to explicitly establish that checking of exit values is not to be done.

steve-bate:
> The JavaDoc for DefaultExecutor does not state what the default behavior is for exit value handling,

That's interesting. The ```isFailure``` documentation has more information.

> If no exit values are set than the default conventions of the OS is
> used. e.g. most OS regard an exit code of '0' as successful
> execution and everything else as failure.

So while I agree that we should use ```setExitValues(null)``` for openhab1, we may want to reconsider that decision for ESH. The downside of this technique is that it ignores error codes that indicate an error based on the exit code conventions of the host operating system.

watou:
> The downside of this technique is that it ignores error codes that indicate an error based on the exit code conventions of the host operating system.

I don't think default logging a commandLine's exit code (when it indicates an error value in the host OS's convention) is ultimately a good thing to do, because this forces the user to live with log messages from commands that may return non-zero (or non-1, as in openHAB 1.8.0) exit codes.  The users may have no control of those programs they run or their exit codes, thereby forcing those users to suppress logging in logback.xml or live with the spam.

I think the best course of action is to update the PR so that it 1) ignores exit codes with setExitValues(null), and 2) changes your WARN log message to DEBUG level.

steve-bate:
John, thanks for the discussion and thoughtful feedback.

> I don't think default logging a commandLine's exit code (when it indicates an error value 
> in the host OS's convention) is ultimately a good thing to do, because this forces the user 
> to live with log messages from commands that may return non-zero (or non-1, as 
> in openHAB 1.8.0) exit codes. The users may have no control of those programs they run 
> or their exit codes, thereby forcing those users to suppress logging in logback.xml or 
> live with the spam.

I think there are tradeoffs either way. I assume that most programs do return an proper exit code. For programs that don't, it's usually easy to adapt them to do so (e.g., wrap them in a script that returns a correct exit code). OTOH, it's more difficult to compensate for openHAB ignoring legitimate error codes. Debug logging is not an ideal solution in situations where a program may fail occasionally because of some temporary issue. All that said, I agree to ignoring the exit codes in openhab1 since that's the least change to the existing behavior. (To be clear, openhab1 was not ignoring the exit codes but it was ignoring exceptions, including exit code-related exceptions.)

> I think the best course of action is to update the PR so that it 1) ignores exit codes 
> with setExitValues(null), and 2) changes your WARN log message to DEBUG level.

The first change is in the PR. I don't think the second change is a good idea.

watou:
> The first change is in the PR. I don't think the second change is a good idea.

OK.  What kinds of exceptions will be logged at WARN level with this change?  I assume cases like timeouts, but which ones do you see that are going to be valuable?  Just wondering.

steve-bate:
The examples I gave in #3821 are two cases. Any Exception related to process pipes, working directory issues, thread interrupts, watchdog timeouts and so on appear to result in an exception that would be logged with this message (based on reviewing the Apache Commons Exec code).

watou:
Those should be very useful to know.  lgtm if you're ready to merge.  I will leave it for @teichsta to merge and cherry-pick as he sees fit after your go-ahead.

steve-bate:
Yes, I think it's ready to merge. Thanks for the review.

watou:
Could you make sure that ESH has a sync PR?  I think it's the rule to make sure ESH is merged first....

watou:
@steve-bate: Should I create the ESH sync PR, and have you look it over to make sure it matches 1.8.* changes?

steve-bate:
@watou, yes thanks. Otherwise, I'll do it in the next day or two.

watou:
Will do, Steve.

watou:
@teichsta, please cherry-pick this change onto the 1.8 branch (or just point me at how to do it in SourceTree).

steve-bate:
Thanks, @watou!

teichsta:
done!

Cherry picking in Sourcetree is easy:

* make sure you checked out master and 1.8 branches
* make sure all branches are up to date
* checkout 1.8
* right click on the given commit (not the merge on master)
* choose "cherry-pick"
* push to 1.8

Thanks, Thomas E.-E.

watou:
> not the merge on master

that's probably what I did wrong last time I tried.  Thanks.

PR #3831, opened 2016-01-18
fix imports

maggu2810:


teichsta:
anything to pull into 1.8 as well?

maggu2810:
The version / backporting decision should be done by you or @kaikreuzer. I don't know...

Issue #3823, opened 2016-01-17
[exec] Allow any item type to be updated, not just hardcoded list


watou:
The exec binding can only update Number, Contact, Switch, Rollershutter and String items.  If it's any other item type, it attempts to post a StringType state to the item, regardless of whether the receiving item accepts them.

The exec ought to be able to update any item type without containing any hardcoded list.  Very similar in outcome to #3504 which produces no side effects or compatibility issues.  I will submit a PR.

[Forum discussion](https://community.openhab.org/t/display-time-only-from-datetime/6301/6)

watou:
Closed via #3832 and confirmed externally by Rob B [here](https://community.openhab.org/t/display-time-only-from-datetime/6301/16?u=watou).

PR #3832, opened 2016-01-18
[exec binding] Allow all item types to be updated

watou:
The existing exec binding contains a hardcoded list of item types that it can update: Number, Contact, Switch, Rollershutter, String (default case).  However openHAB 1.x and ESH have more item types than that, and there is no reason that the output of a commandLine (optionally transformed) should not be suitable for any type of item.  The goal and outcome is similar to #3504.

This PR addresses issue #3823.

This PR removes the binding's knowledge of specific item types, but instead uses the item's acceptedDataTypes list and the TypeParser.  I tested this change with the following items:
```
Group Test (All)
String ExecString "ExecString [%s]" (Test) { exec="<[echo I am a String:60000:REGEX((.*?))]" }
Number ExecNumber "ExecNumber [%.1f]" (Test) { exec="<[echo -12345.432145663:50000:REGEX((.*?))]" }
Location ExecLocation "ExecLocation [%s]" (Test) { exec="<[echo 52.6406981,-9.503447:40000:REGEX((.*?))]" }
Switch ExecSwitch "ExecSwitch [%s]" (Test) { exec="<[echo ON:30000:REGEX((.*?))]" }
Contact ExecContact "ExecContact [%s]" (Test) { exec="<[echo OPEN:20000:REGEX((.*?))]" }
Rollershutter ExecRollershutter "ExecRollershutter [%d %%]" (Test) { exec="<[echo 60:10000:REGEX((.*?))]" }
Dimmer ExecDimmer "ExecDimmer [%d %%]" (Test) { exec="<[echo 30:70000:REGEX((.*?))]" }
DateTime ExecDateTime "ExecDateTime [%1$tH:%1$tM:%1$tS]" (Test) { exec="<[/tmp/printdate.sh:80000:REGEX((.*?))]" }
```

### DIFF
--- a/bundles/binding/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/ExecBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/ExecBindingProvider.java
@@ -11,8 +11,8 @@ package org.openhab.binding.exec;
 import java.util.List;
 
 import org.openhab.core.binding.BindingProvider;
-import org.openhab.core.items.Item;
 import org.openhab.core.types.Command;
+import org.openhab.core.types.State;
 
 /**
  * This interface is implemented by classes that can provide mapping information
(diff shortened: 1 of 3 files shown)


teichsta:
lgtm … please go ahead and merge.

PR #3833, opened 2016-01-18
add maven-bundle-plugin configuration

maggu2810:
The configuration has been added, so we could a manual manifest
generation. The generated manifest could be used by developers to
identify missing imports.
The manifest generation could be triggered by:
mvn bundle:manifest -DniceManifest=true
The "nice manifest" option will e.g. add line breaks after every
imported package.
Now you can have a look at the manifest that would be generated by the
bndtool using:
cat target/classes/META-INF/MANIFEST.MF

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,65 +1,65 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    
+
     <prerequisites>
         <maven>3.0</maven>
     </prerequisites>
-    
+
     <groupId>org.openhab</groupId>
     <artifactId>openhab</artifactId>
     <version>1.9.0-SNAPSHOT</version>
     <name>openHAB</name>
-    
+
     <organization>
         <name>openHAB.org</name>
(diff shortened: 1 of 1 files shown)


watou:
This tool helped me identify some unneeded imports in the Nest binding:
```diff
Import-Package:
- com.google.common.base,
- com.google.common.io,
- org.apache.commons.httpclient,
  org.apache.commons.httpclient.methods,
  org.apache.commons.httpclient.params,
  org.apache.commons.httpclient.util,
  org.apache.commons.io,
  org.apache.commons.lang,
  org.apache.commons.lang.builder,
- org.apache.commons.logging,
  org.codehaus.jackson,
  org.codehaus.jackson.annotate,
  org.codehaus.jackson.map,
  org.openhab.core.binding,
  org.openhab.core.events,
  org.openhab.core.items,
- org.openhab.core.library.items,
  org.openhab.core.library.types,
  org.openhab.core.types,
  org.openhab.io.net.http,
  org.openhab.model.item.binding,
  org.osgi.framework,
  org.osgi.service.cm,
- org.osgi.service.component,
- org.osgi.service.event,
  org.slf4j
```

maggu2810:
@watou It needs to be said, that the created manifest should be used as a indication what is needed. It is not always correct, that missing packages are not needed (e.g. if the code is using "get class for name" etc.).

hakan42:
Looks extremely nice to me, could be a part of the review process

teichsta:
> if the code is using "get class for name" etc.

which shouldn't be the case … but +1 for @maggu2810 i've often observed wrong results of this import-calculation tool.

teichsta:
Thanks @maggu2810!

PR #3836, opened 2016-01-18
Added PAT02-A/B/C

schmidmuc:


cdjackson:
Thanks.

antimork:
Unfortunately, this does not work with my PAT02-**B**:s (variant with temp. and humidity only). I suspect it has been added with the ID's from a PAT02-A only. (They are as far as I can tell the same PCB with different components & firmware - the included manual documents them all so I think the existing pat02.xml should work fine once the right ID is set.)

Node XML says:
`<manufacturer>0x13c</manufacturer>`
`<deviceId>0x20</deviceId>`
`<deviceType>0x2</deviceType>`

So it should just require adding this, right?
`<Reference>`
`<Type>0002</Type>`
`<Id>0020</Id>`
`</Reference>`

cdjackson:
If this is the type/id of your device, then yes - if you can create a PR it would be appreciated.

antimork:
Well, after a bit of fiddling, it seems I can do that. :-) Hopefully correctly. Thanks for the quick response!
#5050

cdjackson:
Thanks.

PR #3840, opened 2016-01-18
starting all components immediately after bundle start to prevent iss…

kaikreuzer:
…ues with Felix SCR

### DIFF
--- a/bundles/binding/org.openhab.binding.alarmdecoder/OSGI-INF/binding.xml
+++ b/bundles/binding/org.openhab.binding.alarmdecoder/OSGI-INF/binding.xml
@@ -9,7 +9,7 @@
     http://www.eclipse.org/legal/epl-v10.html
 
 -->
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" activate="activate" deactivate="deactivate" name="org.openhab.binding.alarmdecoder.refreshservice">
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" activate="activate" deactivate="deactivate" immediate="true" name="org.openhab.binding.alarmdecoder.refreshservice">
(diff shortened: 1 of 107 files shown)


kaikreuzer:
To be discussed whether this should be merged or not.
Does anybody know of a reason, why any existing component should NOT be started?

Would be good if some openHAB 1 users could test this first and see if it has any impact on the openHAB 1 runtime.

@maggu2810 Any comments from your side?

watou:
Would this also apply to action service bundles?
If this is merged, OH1 testers could have a Cloudbees build to extract from more easily for testing on 1.8.0.  If it's a problem, the commit could be reverted.

kaikreuzer:
> Would this also apply to action service bundles?

It seems as if this pattern is already applied on our actions.

robnielsen:
Please also apply to the netatmo binding also

kaikreuzer:
> Please also apply to the netatmo binding also

Did you test that it still works with that change?
In https://github.com/openhab/openhab/commit/4581431c3db83ffcd75cdf01d5110eb0b5d2189c @clinique set immediate specifically to false - I assume that had some specific reason?

maggu2810:
I think it is ok, to set immediate for all bindings. The DS of the binding should be started after registration.
I assume the Eclipse IDE added the immediate="false", but let's wait for @clinique

robnielsen:
I  changed immediate=true for Netatmo, and it now works with both 1.8 and the latest build of 2.0. I'll create a PR that includes the change, along with the changes to gets picked up by the distribution build

kaikreuzer:
Cool! No need for a separate PR. As this one needs to be rebased anyhow, I will include Netatmo in it.

kaikreuzer:
Done. As nobody vetoed against this PR, I'll merge it now.

PR #3848, opened 2016-01-19
Telegram action: added username/password authorization on image http get

paolodenti:
method sendTelegramPhoto accepts two additional optional parameters (username and password) used to get image on locked servers


watou:
[Discussion on forum](https://community.openhab.org/t/new-telegram-action/4459/30?u=watou)

I see your MANIFEST.MF already has an Import-Package for `org.apache.commons.httpclient.auth`, so this PR lgtm.

@paolodenti, could you add the new usage to the [wiki doc](https://github.com/openhab/openhab/wiki/Actions#telegram-action), and also make note therein of which kinds of URLs are permitted to be used in the `photoURL` parameter to the `sendTelegramPhoto` methods?

paolodenti:
wiki modified accordingly

watou:
Thanks!  Could you also note that the new usage is available as of 1.9 (so users don't report it broken before then)?

paolodenti:
oops, yes, just done!

PR #3841, opened 2016-01-19
Netatmo binding for Netatmo Weather and Netatmo Welcome

wep4you:
I have successfull implemented the Netatmo welcome binding #3815 into the Netatmo binding, and did some refactoring, and also add new keyword weather for the configuration. For backward compatibility i left the config without the keywords welcome or wether default as weather config.

With this implementation there is also issue #3225 closed, this is now the implementation. For correct use the wiki has to be updated after the code was merged.

@watou @robnielsen @clinique Response about the changes would be nice

@napoleonmm83 @steve-bate Info because of the issue #3225

Issue #3225, opened 2015-10-01
Netatmo Welcome Binding


napoleonmm83:
Netatmo Welcome has now a api 
https://github.com/Netatmo/Netatmo-API-PHP#netatmo-welcome-api-client-implementation---php-sdk

Maybe some one can make a Binding?

wep4you:
Yes would be great to use the presence Information for the Home Automation!

wep4you:
Based on the Netatmo Binding i implemented a Binding for Netatmo Welcome. At the moment it will get all Informations from the Netatmo API GetHomeData. Till now i haven't pushed something to GIT so if somebody else is also interested i will try it.

skycryer:
I am interested, how hard is it to add new faces that have been detected?

wep4you:
@skycryer What do you mean, in the netatmo app or in the API? If you mean the app, than its really easy, just tap the unknown face on the phone, click identify. Then you get all unknown faces, and could check them and say to witch user they belong or create a new one ...
Also have a look at the state of he implementation in the netatmo binding #3841

RosaEinhorn:
Hi,
I just got a Netatmo Welcome and installed it (already had a weather station).
I followed all the steps of netatmo pre-setup and requested access to the welcome camera, the acces token seems to work, I can receive updated for the weather station.
Unfortunately, when I try to request information of the camera, I am getting the following error:

`Binding configuration of type 'netatmo' of item ‘Welcome_Carina_AtHome‘ could not be parsed correctly.
java.lang.IllegalArgumentException: Invalid measure: OutOfSight
	at org.openhab.binding.netatmo.internal.NetatmoMeasureType.fromString(NetatmoMeasureType.java:65) ~[na:na]
	at org.openhab.binding.netatmo.internal.NetatmoGenericBindingProvider.processBindingConfiguration(NetatmoGenericBindingProvider.java:190) ~[na:na]
	at org.openhab.model.item.internal.GenericItemProvider.internalDispatchBindings(GenericItemProvider.java:348) [org.openhab.model.item_1.8.3.jar:na]
	at org.openhab.model.item.internal.GenericItemProvider.dispatchBindingsPerType(GenericItemProvider.java:309) [org.openhab.model.item_1.8.3.jar:na]
	at org.openhab.model.item.internal.GenericItemProvider.addBindingConfigReader(GenericItemProvider.java:106) [org.openhab.model.item_1.8.3.jar:na]
[...]

PR #3859, opened 2016-01-21
Implemented Netatmo Welcome into Netatmo binding.

wep4you:
@robnielsen 
I managed to get my development environment working, seems the whole set has changed the last days, so i have to start over with Eclipse installer, but now its running fine ;-)

I grabbed all commits from head, and implemented my changes on top of the latest sources, so merge should be possible now, don't know if this goes in my last pull request so here are the references to the other: #3225 #3841 #3815 

just for Info: @watou @clinique @napoleonmm83 @steve-bate


wep4you:
@robnielsen 
All changes are merged into 1 commit, no conflicts, Testcases are ok and all features are in, so you could merge this back to head.

robnielsen:
@wep4you, I will take a look this weekend.

wep4you:
@robnielsen Thanks i also take a look on the new openhab2 ill try to move this to the new eclipse smart home development, looks really cool and i like the paper ui

robnielsen:
I haven't looked much, but I see you changed the format to require weather and welcome. Can't this be added without changing the way it works currently? How are the welcome devices addressed?

robnielsen:
I also see the Netatmo API uses the generic term camera, we should use the same sytax since they also have external cameras called presence.

wep4you:
@robnielsen Default is the weather binding, so if you don't use the keyword weather it also will work, for backward compatibility. To clarify for new user for which device it is i added the keyword weather.

wep4you:
I implemented the netatmo api with the api names, i didn't changed them, to easily map the binding to the api.

robnielsen:
Yes, but let's find a way to work with the current format, there should be no need to prefix with weather or welcome. What is the address format of the cameras?

wep4you:
same format then for devices so you can't say if its a device or a camera. its the mac address ...

robnielsen:
How about using something like
[weather|camera,]00:00:00:00:00:...
camera,00:00:00:00:00:...

where weather and camera are optional, and defaults to weather. Then the current format doesn't change.

wep4you:
Ok, but i didn't see a difference to the implementation, maybe im missing something. 

[weather|welcome#]00:00:00:00:00:...
welcome#00:00:00:00:00:...

the only difference is a , than a # and camera instead of welcome

wep4you:
We shouldn't use , because as i see in the code this is something for backward compatibility for scale,
and the keyword welcome should describe the product better then just camera, because netatmo also have this as product name (camera is just in the api for us developers)

robnielsen:
There should be some sort of a way to further split the string, the first split is with "#", the second split would be with ',' or something similar. Its just a way to further break up configParts[0], and it will not change how the binding currently works.

As for camera, Netamo has two different camera devices, Welcome and Presence. Let's use camera as a generic way to refer to both devices

wep4you:
i'll have a look at the api tomorrow, i think the functions are different so it would be better to use different keys, but i just have a welcome cam and a weather station so i can't try the apis just with unit tests

wep4you:
@robnielsen At the moment there is no API documentation for the presence camera, but i think the device id will also be the mac address like the weather station and welcome Ids. To use the same name then netatmo i suggest to also use the same names: Weather Station, Thermostat and Welcome.

What do you think to use '=' for further splitting the first part, so Config would look like:

[weatherstation=|welcome=|thermostat=|othernewdevicename=]00:00:00:00:00:...
and weather station is default.

robnielsen:
using = sounds fine, but let's use the generic term camera instead of
welcome, just like Netatmo does.

wep4you:
ad1) Using '=' is ok for me ill recode this and commit my changes after im done
ad 2) Don't know where you read that netatmo tells the welcome cam generic camera. Look at the API (https://dev.netatmo.com/doc/methods), just 1 function with camera in the name, all other are events ans person data. 
From the API:
* Welcome specific methods
    * gethomedata
    * getnextevents
    * getlasteventof
    * geteventsuntil
    * getcamerapicture

If you insist in the generic name camera, i could change it, but i think if the presence camera will get an API the 2 different camera devices couldn't be differentiated and this has to be renamed in the future.

robnielsen:
Presence is an outdoor camera, welcome is an indoor one, camera also
matches what is used by Netatmo, so lets go with it.

wep4you:
Ok ill change it to camera, and the separator to =
is the name weather station ok, or do you have other suggestions?

In the meantime i managed to implement the binding in Openhab2. Think the authentication will be ready today

wep4you:
@robnielsen All discussed changes are now in my last commit

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -52,732 +34,159 @@
...
+    /**
+     * The refresh interval which is used to poll values from the Netatmo server
+     * (optional, defaults to 300000ms)
+     */
+    private long refreshInterval = 300000;
+
+    private static Map<String, OAuthCredentials> credentialsCache = new HashMap<String, OAuthCredentials>();
+
+    private final NetatmoWelcomeBinding welcomeBinding = new NetatmoWelcomeBinding();
+    private final NetatmoWeatherBinding weatherBinding = new NetatmoWeatherBinding();
```
robnielsen: rename welcomeBinding with cameraBinding.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -52,732 +34,159 @@
...
+
+            OAuthCredentials oauthCredentials = getOAuthCredentials(userid);
+            if (oauthCredentials.noAccessToken()) {
+                // initial run after a restart, so get an access token first
+                oauthCredentials.refreshAccessToken();
+            }
+
+            // Start the execution of the weather station and welcome binding
+            weatherBinding.execute(oauthCredentials, this.providers, this.eventPublisher);
+            welcomeBinding.execute(oauthCredentials, this.providers, this.eventPublisher);
```
robnielsen: Can we make this smarter so we only call the weather binding if they have weather items configured and vice versa for camera items?

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
```diff
@@ -59,16 +96,86 @@
  * </li>
  * </ul>
  *
+ * <b>Netatmo welcome:</b>
+ * Valid bindings for a netatmo welcome home are:
+ * <ul>
+ * <li>
+ * <code>{ netatmo="camera=&lt;home_id&gt;#attribute" }</code></li>
+ * <ul>
+ * <li><code>{ netatmo="camera=1234567890abcdefghijklmn#Name" }</code></li>
```
robnielsen: fix 1234... to be similar to the weather one

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
```diff
@@ -24,7 +28,24 @@
  * This class is responsible for parsing the binding configuration.
  *
  * <p>
- * Valid bindings for the main device are:
+ * Valid bindings for the main device are - since 1.9
+ * <ul>
+ * <li><code>{ netatmo="weatherstation=&lt;device_id&gt;#Measurement" }</code></li>
```
robnielsen: Lets leave this weather as discussed, not weatherstation

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
```diff
@@ -59,16 +96,86 @@
...
  * @author Gaël L'hopital
  * @author Rob Nielsen
+ * @author Ing. Peter Weiss
  * @since 1.4.0
  */
 public class NetatmoGenericBindingProvider extends AbstractGenericBindingProvider implements NetatmoBindingProvider {
 
     private static Logger logger = LoggerFactory.getLogger(NetatmoGenericBindingProvider.class);
 
+    private static String NETATMO_WEATHER = "weatherstation";
```
robnielsen: weather

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
```diff
@@ -42,7 +63,23 @@
  * </li>
  * </ul>
  * <p>
- * Valid bindings for a module are:
+ * * Valid bindings for a module are - since 1.9:
```
robnielsen: what's new since 1.9? Also need to indicate what's optional.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/weather/NetatmoWeatherBinding.java
```diff
@@ -0,0 +1,557 @@
...
+     * @param oauthCredentials
+     * @param providers
+     * @param eventPublisher
+     */
+    public void execute(OAuthCredentials oauthCredentials, Collection<NetatmoBindingProvider> providers,
+            EventPublisher eventPublisher) {
+        logger.debug("Querying Netatmo Weather API");
+
+        try {
+            if (oauthCredentials.isFirstExecution()) {
```
robnielsen: oauthCredentials is used by both weather and camera, this check needs to use a private global variable in this binding to indicate that is the first time used.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/welcome/GetHomeDataRequest.java
```diff
@@ -0,0 +1,82 @@
+/**
+ * Copyright (c) 2010-2015, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.netatmo.internal.welcome;
```
robnielsen: change package to org.openhab.binding.netatmo.internal.camera

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/welcome/NetatmoWelcomeBinding.java
```diff
@@ -0,0 +1,487 @@
...
+
+        logger.debug("Querying Netatmo Welcome API");
+        try {
+            oauthCredentials.refreshAccessToken();
+            configuredHomeKnownPersons.clear();
+            configuredHomeUnKnownPersons.clear();
+            configuredHomeMap.clear();
+            configuredhomeCameras.clear();
+
+            processGetHomeData(oauthCredentials, providers, eventPublisher);
```
robnielsen: Should this happen every time? The weather binding only does this the first time.

robnielsen:
@wep4you, in general looks pretty good. Please rename the package to org.openhab.binding.netatmo.internal.camera instead of org.openhab.binding.netatmo.internal.welcome. Also please change the variables and classes to use [cC]amera instead of [wWelcome]. 

@clinique already has a PR for OH 2 (https://github.com/openhab/openhab2-addons/pull/586).

If you could, could you also add unit tests for the camera binding?

@jannegpriv, you might be interested in the PR since it could impact some enhancements you are interested in adding to the binding.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
```diff
@@ -203,10 +393,18 @@ public void processBindingConfiguration(final String context, final Item item, f
...
+        String cameraId;
+        NetatmoWelcomeAttributes attribute;
+
         @Override
         public String toString() {
             return "NetatmoBindingConfig [userid=" + this.userid + ", deviceId=" + this.deviceId + ", moduleId="
-                    + this.moduleId + ", measure=" + this.measureType.getMeasure() + "]";
+                    + this.moduleId + ", measure=" + (this.measureType != null ? this.measureType.getMeasure() : null)
+                    + ", homeId=" + this.homeId + ", personId=" + this.personId + ", cameraId=" + this.cameraId
+                    + ", attribute=" + this.attribute + "]";
```
robnielsen: Can we make toString a little smarter and print out what's appropriate for weather or camera?

wep4you:
@robnielsen I changed the things and did a new commit. Welcome changed to camera, added a Unit Test for the camera binding, and also add information of wether and camera in the toString Method.

@clinique Haven't seen your commit in the meantime i also implemented netatmo Binding in OH2 in my fork. It works with discovery for the camera device, and i already get the home information in the channels, so the implementation of the other informations are ongoing. How about merge our code to get 1 binding?

@watou @napoleonmm83 @steve-bate Information about the state of the welcome camera implementation in OH1 and also OH2

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
wep4you: done

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
wep4you: done, i check if wether a wether or camera device ist configured as item. If the items are not read at the first run, the execution is delayed till the next refresh interval

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
wep4you: Changed to 00000...

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
wep4you: changed to weather

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
wep4you: changed to weather

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
wep4you: nothing, just the keyword weather so removed this and add weather = as optional

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/weather/NetatmoWeatherBinding.java
wep4you: Removed firstRun from OAuth and check this as u suggested

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/welcome/GetHomeDataRequest.java
wep4you: done

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/welcome/NetatmoWelcomeBinding.java
wep4you: Yes the response didn't change in this API call when the refreshToken is not send again.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoGenericBindingProvider.java
wep4you: Check if its a weather or a camera item and just write the items which belongs to the type

teichsta:
PR lgtm … awaiting your indication when to merge into master. Thanks guys!

robnielsen:
@teichsta, lgtm also. @wep4you thanks for your contribution!

teichsta:
ftr: the OH2 PR wich discusses the changes for the OH2 binding https://github.com/openhab/openhab2-addons/pull/586

teichsta:
Thanks again, guys!

robnielsen:
@wep4you , can you update the wiki? Thanks!

PR #3861, opened 2016-01-21
added type exact add/remove methods.

kaikreuzer:
resolves https://github.com/openhab/openhab-distro/issues/101

### DIFF
--- a/bundles/binding/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/HttpBinding.java
+++ b/bundles/binding/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/HttpBinding.java
@@ -107,6 +107,14 @@ public class HttpBinding extends AbstractActiveBinding<HttpBindingProvider>imple
         setProperlyConfigured(true);
     }
 
+    protected void addBindingProvider(HttpBindingProvider bindingProvider) {
+        super.addBindingProvider(bindingProvider);
+    }
+
+    protected void removeBindingProvider(HttpBindingProvider bindingProvider) {
+        super.removeBindingProvider(bindingProvider);
+    }
+
     /**
      * @{inheritDoc}
      */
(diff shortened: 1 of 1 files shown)


maggu2810:
Do you like to add the "override" annotation?

kaikreuzer:
No, because it does not override but overload the method. In this case, this annotation isn't used, right?

maggu2810:
Okay, I have not known, that this is handled different for overloaded methods.
My settings in of the Eclipse IDE save actions / formatter will add "@Override" for overwritten and overloaded functions.
Also the annotation documentation does not different here.

__java.lang.Override__
@Target(value={METHOD})
@Retention(value=SOURCE)
Indicates that a method declaration is intended to override a method declaration in a supertype. If a method is annotated with this annotation type compilers are required to generate an error message unless at least one of the following conditions hold:
* The method does override or implement a method declared in a supertype.
* The method has a signature that is override-equivalent to that of any public method declared in Object.

kaikreuzer:
> My settings in of the Eclipse IDE save actions / formatter will add "@Override" for overwritten and overloaded functions.

Hm, I thought I am using the same formatter as you? Mine doesn't add it. 

> The method has a signature that is override-equivalent

So does this apply here? What does override-equivalent exactly mean...?

maggu2810:
> Hm, I thought I am using the same formatter as you?

For ESH and OH2, but there are other projects and other formatters I am using. :wink: 
But I checked the save actions and "Add missing '@Override' annotations", "Add missing '@Override' annotations to implementations of interface methods" and "Add missing '@Deprecated' annotations" is selected for the ESH and OH2 formatter, too.
Perhaps I am missing something (I tested it with two classes where the one defines a function foo and a child implements foo by calling super.foo()...).
Let's try me it with the Eclipse IDE and investigate what is the difference.

Two method signatures m1 and m2 are override-equivalent iff either m1 is a subsignature of m2 or m2 is a subsignature of m1. (https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.4.2)

kaikreuzer:
Which means that they are not override-equivalent, so not having the annotation is correct.

note that we are talking about the two methods `foo(Class1 x)`and `foo(Class2 y)`, where `Class2 extends Class1` - so their erasure is no the same.

maggu2810:
AbstractActiveBinding is using BindingProvider, HttpBinding is using HttpBindingProvider...
So all fine, nothing to complain :wink:

maggu2810:
I will shut up and you can merge :)

PR #3871, opened 2016-01-22
MiOS: Work-around issue where update() called before config loaded (openHAB 2.0)

mrguessed:
Under openHAB 2.0, the `update()` method is called before the corresponding `config.cfg/mios.cfg` file is loaded.

We're called again, after these are correctly loaded, so we'll skip over the call if the properties are passed as `null`.

Also removed an unused variable that Eclipse highlighted.


kaikreuzer:
Cool, thanks!

PR #3880, opened 2016-01-23
Several Z-Wave database fixes

JensErat:
These two commits fix several issues in the Z-Wave database:

1. Vision Security ZD2012 missed configuration command class
2. Vision Security ZD2012 configuration was badly aligned
3. Several Z-Wave.Me products had non-well-formed XML, at the same time removed empty help strings

Furthermore, I found about half a dozen other products not having the configuration command class, but having configuration items in their XML file. Is this an issue? If so, I'd fix them. I found this using a very basic automated check over all XML files, and I could easily add some further verification (and even automatic cleanup) if interested.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/vision/zd2102.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/vision/zd2102.xml
@@ -6,6 +6,7 @@
 	<CommandClasses>
 		<Class><id>0x20</id></Class>
 		<Class><id>0x30</id></Class>
+		<Class><id>0x70</id></Class>
 		<Class><id>0x71</id></Class>
 		<Class><id>0x72</id></Class>
 		<Class><id>0x80</id></Class>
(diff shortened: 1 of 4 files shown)


cdjackson:
The command class list in the XML files isn't used (except in special cases), so the changes won't matter. I'll merge this, but it would be great if you can consider using the online database editor I'm working on so that we can also keep changes between OH1 and OH2 consistent, otherwise the database is going to be a nightmare (ok, an even bigger nightmare :) ) to maintain.

You can find the database [here](http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-database-guide).

JensErat:
Good to know this editor exists!

Anyway: With this very specific product and Z-Wave-binding from git HEAD, I was not able to configure the door switch. After adding the command class, it worked fine. Is the configuration command class really not needed?

cdjackson:
The online editor is still work in progress, but it’s in reasonable shape now…

Actually, you might well be right that it’s used (badly!) in the config UI - if so, we should change the binding to remove this dependancy. I’ll take a look at this later as I need to do some work on the binding anyway.

cdjackson:
Yes - I just checked, the UI won’t list parameters unless the command class is in the database. I’ll check the binding to change the way this works and remove this dependancy.

Thanks.

PR #3884, opened 2016-01-23
Update project name for OH1 ZWave binding

cdjackson:
I would like to propose to change the project name of the OH1 ZWave binding to avoid workspace conflicts with the OH2 binding. 

@kaikreuzer Is there any downside to this? If not, it seems it would avoid problems...


kaikreuzer:
Should be ok.

cdjackson:
Thanks.

PR #3792, opened 2016-01-14
Fix for tcpbinding base config regex

steve-bate:
This fixes a regression in the tcpbinding base config regex that causes issues when there are multiple binding config elements. The current expression matches space-separated config elements and it should match comma-separated elements.

PR #3890, opened 2016-01-25
[tcp] Same change as #3792 on 1.8

watou:


steve-bate:
@watou, I've tested the change and my tests pass. Thanks.

watou:
Thanks, @steve-bate!

PR #3901, opened 2016-01-26
Added add and remove binding methods for OH2 compatibility.

sciamachy:
Added addBindingProvider and removeBindingProvider to NetworkHealth Binding for OpenHAB2 compatibility.

### DIFF
--- a/bundles/binding/org.openhab.binding.networkhealth/src/main/java/org/openhab/binding/networkhealth/internal/NetworkHealthBinding.java
+++ b/bundles/binding/org.openhab.binding.networkhealth/src/main/java/org/openhab/binding/networkhealth/internal/NetworkHealthBinding.java
@@ -61,6 +61,15 @@ public class NetworkHealthBinding extends AbstractActiveBinding<NetworkHealthBin
         return refreshInterval;
     }
 
+    protected void addBindingProvider(NetworkHealthBindingProvider bindingProvider) {
+        super.addBindingProvider(bindingProvider);
+    }
+ 
+    protected void removeBindingProvider(NetworkHealthBindingProvider bindingProvider) {
+        super.removeBindingProvider(bindingProvider);
+    }
+
+
     /**
      * @{inheritDoc}
      */


kaikreuzer:
Thanks!

teichsta:
this hasn't been cherry-picked into 1.8, right?

kaikreuzer:
No, it hasn't. Should it? It is only relevant for use on OH2. We could make the beta2 use 1.8.1 add-ons then we should pick all required changes (I guess there are many more already that would need to be picked).

watou:
I worry that picking that many master changes into the 1.8 branch threatens 1.8.1 stability, and also the new dev environment will not be based on what is distributed, since it depends on master changes (removal of runtime, code reformatting, etc.).  There are already several new challenges supporting 1.8!

kaikreuzer:
The question is whether we want to pin the version of included 1.x add-ons within 2.0beta2 - otherwise, the online distro will e.g. download different (snapshot) versions, depending on when you install it, so things are not reproducible.

watou:
I suggest that 1.8.1 distro serves a different purpose than 2.0beta2, the code has diverged, and 1.x addons distributed with OH2 need to carry on with new numbers.  Perhaps just tag openhab/master with 2.0beta2-addons when you release 2.0beta2, and sync the repo branches/tags when 2.0 is released?  There are doubtless many more changes on openhab/master before then.

teichsta:
> (I guess there are many more already that would need to be picked)

that would be my guess as well but we could handle these changes one by one.

> I worry that picking that many master changes into the 1.8 branch threatens 1.8.1 stability

we could agree to only cherry-pick binding dependent bugfixes (except the Classic-UI fix)

> The question is whether we want to pin the version of included 1.x add-ons within 2.0beta2

to be honest i didn't dive into the new distro system right now … so i'd appreciate your suggestions.

kaikreuzer:
> Perhaps just tag openhab/master with 2.0beta2-addons when you release 2.0beta2, and sync the repo branches/tags when 2.0 is released? 

Tagging on the repo isn't enough, the version would have to be uploaded to Bintray as well (and which version should it carry then? 1.9.0.b2 maybe?

watou:
1.9 anything is confusing to me, since there is to 1.9 distro.  Maybe branch&tag openhab/master the same way you branch&tag the other repos for the 2.0beta2 release?

kaikreuzer:
> since there is no 1.9 distro

But there are 1.9 add-on releases. And 1.9.0 final will be released with openHAB 2.0 final. So to me it makes sense to release a 1.9.0.b2 together with a 2.0.b2.

PR #3906, opened 2016-01-28
[xbmc] Missing cfg file on Install

### LINKED ISSUE #3904: XBMC Missing cfg file on Install
When installing the current binding, there is no cfg file being included which is necessary to configure the location of your xmbc setup.

The existing configuration from OH1 does work minus the xbmc: at the beginning of each requirement.  I have tested and can also see my items in the sitemap once the cfg file is configured.

watou:
Closes #3904

[Discussion on forum](https://community.openhab.org/t/xbmc-binding/6747).

Renaming to Kodi is for another day, and so as to not break existing user configurations.


teichsta:
lgtm, thanks @watou!

PR #3908, opened 2016-01-28
[koubachi] Added v3 API Device members; changed default URLs to v3.

watou:
Thanks to input [here](https://community.openhab.org/t/koubachi-not-supporting-new-values/4833/13?u=watou), it now makes sense to update this binding to v3 API.  I will update wiki after merge.


watou:
Confirmed working as built from master and running on 1.7 runtime [here](https://community.openhab.org/t/koubachi-not-supporting-new-values/4833/15?u=watou).

teichsta:
lgtm!

PR #3909, opened 2016-01-28
Update zw056.xml

arraylabs:
Fix incorrect parameters, fix incorrect command classes.

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/aeon/zw056.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/aeon/zw056.xml
@@ -1,119 +1,123 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Product>
-    <Model>ZW056-A</Model>
-    <Label lang="en">Doorbell</Label>
-    <CommandClasses>
-        <Class>
-            <id>0x20</id>
-        </Class>
-        <Class>
-            <id>0x70</id>
-        </Class>
-        <Class>
-            <id>0x72</id>
-        </Class>
-        <Class>
-            <id>0x80</id>
-        </Class>
-        <Class>
-            <id>0x85</id>
-        </Class>
-        <Class>
-            <id>0x86</id>
-        </Class>
-     </CommandClasses>
-
-     <Configuration>
-                <Parameter>
-                        <Index>0</Index>
-                        <Type>long</Type>
-                        <Default>2</Default>
-                        <Size>1</Size>
-                        <Minimum>0</Minimum>
-                        <Maximum>100</Maximum>
-                        <Label lang="en">Repetitions of ringtone</Label>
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.zwave/database/aeon/zw056.xml
```diff
@@ -1,119 +1,123 @@
...
+		<Class><id>0x70</id></Class>
+		<Class><id>0x72</id></Class>
+		<Class><id>0x73</id></Class>
+		<Class><id>0x82</id></Class>
+		<Class><id>0x85</id></Class>
+		<Class><id>0x86</id></Class>
+	</CommandClasses>
+	<Associations>
+		<Group>
+			<Index>1</Index>
```
cdjackson: Is it correct that neither of these report to the controller? If so, they should have the SetToController tag.

bundles/binding/org.openhab.binding.zwave/database/aeon/zw056.xml
```diff
@@ -1,119 +1,123 @@
...
+	                    	<Value>1</Value>
+	                    	<Label lang="en">Hail CC</Label>
+	                </Item>
+	                <Item>
+	                    	<Value>2</Value>
+	                    	<Label lang="en">Basic CC Report</Label>
+	                </Item>
+			<Label lang="en">Enable Notifications</Label>
+			<Help lang="en">Enable to send notifications to associated devices (Group1) when the state of Siren changed (0=nothing,1=hailCC,2=basic CC report).</Help>
+		</Parameter>
```
cdjackson: Looking at the manual, there are some parameters missing?

cdjackson:
Thanks. Just a couple of comments - can you also squash your commits into a single commit once ready.

Thanks

bundles/binding/org.openhab.binding.zwave/database/aeon/zw056.xml
arraylabs: I'm not sure, the manual is kinda useless when trying to determine the command classes. I found a document that showed the hex codes with the english class names. But as far as what reports to the controller i don't see anything that says that anywhere, or I don't know what I am looking for if something does say that. Guessing may 0x82 as that is hail CC, but thats simply a guess.

bundles/binding/org.openhab.binding.zwave/database/aeon/zw056.xml
arraylabs: Yes, mostly the stuff for remapping the buttons on the front of the unit. Thought I would leave them alone as the user manual does have instructions for setting volume and ringtone based on the type of button press, switching them via configuration settings would make those portions of the manual incorrect, thought that could be confusing for someone that changed the parameters, then pulled out the manual a few months later and it didn't work right. Was trying to get the config items that can't be changed via the front buttons.

PR #3920, opened 2016-01-30
Reverted timing back to fix reliability.

DJCool1:
From my commit 6f8fd06 on May 9, 2015 I had decreased the latency. Unfortunately, over time it's become clear to me with my lights that this decreased the reliability. The lights do need 100 msec, so I have rolled them back to the 100 msec timing to undo what I did then.


watou:
@DJCool1, could you give a description of the problem this PR fixes and in which release the problem exists?  Is there a discussion thread link or other context?  Thank you!

DJCool1:
Updated description.

watou:
Thanks!  Do you think this fix should be in the 1.8.1 maintenance release?  Any chance other milight binding users could test your change for that extra bit of certainty, or is that unnecessary?

DJCool1:
I think it should be in the maintenance release to improve reliability. As for testing, it's a very minor change and a revert... not much to test. Likely unnecessary to wait.

teichsta:
Thanks @DJCool1! Also cherry-picked onto 1.8

PR #3930, opened 2016-01-30
Fix UI image and chart refresh per #1584

### LINKED ISSUE #1584: Image refresh isn't refreshing
When you have an Image set to a dynamic url (ip camera) with the refresh parameter, the image is not refreshed at the refresh period, but only once.

Here's a very basic sitemap : 
```
sitemap test label="Test"
{
        Frame {
                Image url="http://10.8.1.2:8080/snapshot.cgi?user=user&pwd=password" refresh=2000
        }
}
```

The imageRefresh variable is always set to 0, which prevents the execution of the reloadImage function located here : https://github.com/openhab/openhab/blob/master/bundles/ui/org.openhab.ui.webapp/snippets/main.html#L67

Also, looking at the code, the imagesToRefreshOnPage variable seems to be ignored.

Thanks

watou:
Users in issue #1584 recommend these changes to fix #1584.  I would like to see image and chart refresh in the 1.8 Classic UI as well, but not before verifying that this fix has no regressions.  Please comment here if you are willing to test this change and report results.  It's possible this won't be the fix.

### DIFF
--- a/bundles/ui/org.openhab.ui.webapp/src/main/java/org/openhab/ui/webapp/internal/render/ChartRenderer.java
+++ b/bundles/ui/org.openhab.ui.webapp/src/main/java/org/openhab/ui/webapp/internal/render/ChartRenderer.java
@@ -61,8 +61,8 @@ public class ChartRenderer extends AbstractWidgetRenderer {
 			String snippet = getSnippet("image");			
 
 			if(chart.getRefresh()>0) {
-				snippet = StringUtils.replace(snippet, "%setrefresh%", "<script type=\"text/javascript\">imagesToRefreshOnPage=1</script>");
-				snippet = StringUtils.replace(snippet, "%refresh%", "id=\"%id%\" onload=\"setTimeout('reloadImage(\\'%url%\\', \\'%id%\\')', " + chart.getRefresh() + ")\"");
+				snippet = StringUtils.replace(snippet, "%setrefresh%", "<script type=\"text/javascript\">OH.imagesToRefreshOnPage=1</script>");
+				snippet = StringUtils.replace(snippet, "%refresh%", "id=\"%id%\" onload=\"setTimeout('OH.reloadImage(\\'%url%\\', \\'%id%\\')', " + chart.getRefresh() + ")\"");
 			} else {
 				snippet = StringUtils.replace(snippet, "%setrefresh%", "");
 				snippet = StringUtils.replace(snippet, "%refresh%", "");
--- a/bundles/ui/org.openhab.ui.webapp/snippets/main.html
+++ b/bundles/ui/org.openhab.ui.webapp/snippets/main.html
@@ -50,8 +50,8 @@
         search = search + "sitemap=%sitemap%&w=" + hash.substring(2) + "&poll=true";
         url = path + search;
         WA.Request(url, null, -1, true, null); 
-        OH.images.refresh = OH.images.countOnPage;
-        OH.images.countOnPage = 0;
+        OH.imageRefresh = OH.imagesToRefreshOnPage;
+        OH.imagesToRefreshOnPage = 0;
       });
 
       WA.AddEventListener("load", function () {
@@ -106,13 +106,11 @@
       };
 
       // Functions for image refresh
-      OH.images = {
-        refresh: 0,
-        countOnPage: 0
-      };
+      OH.imageRefresh = 0;
+      OH.imagesToRefreshOnPage = 0;
 
       OH.reloadImage = function reloadImage(url, id) {
-        if (OH.images.refresh) {
+        if (OH.imageRefresh) {
           document.getElementById(id).src = url + new Date().getMilliseconds();
         }
       };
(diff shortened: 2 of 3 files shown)


bundles/ui/org.openhab.ui.webapp/src/main/java/org/openhab/ui/webapp/internal/render/ChartRenderer.java
```diff
@@ -62,7 +62,7 @@ public boolean canRender(Widget w) {
 
 			if(chart.getRefresh()>0) {
 				snippet = StringUtils.replace(snippet, "%setrefresh%", "<script type=\"text/javascript\">imagesToRefreshOnPage=1</script>");
-				snippet = StringUtils.replace(snippet, "%refresh%", "id=\"%id%\" onload=\"setTimeout('reloadImage(\\'%url%\\', \\'%id%\\')', " + chart.getRefresh() + ")\"");
```
yozik04: When I last time looked in the code **reloadImage** function was meant to be global. But due to syntax error it is not (It is in some browsers). I would suggest to fix this **reloadImage** function's globalness instead of modifying template.
I may be wrong...

teichsta:
@watou are you going to contribute these changes to ESH as well (would be great)? Once it is merged we could carry on with this PR as well. Thanks!

watou:
@teichsta I will make sure I understand the history of the changes that led to this, and proceed as you suggest, this week.

watou:
Don't merge yet!  I am still working on it.

watou:
This PR reverts the variable names `imageRefresh` and `imagesToRefresh` to be the same as the ESH version, but the addition of the `OH` namespace here in main.html had broken image and chart refresh, so the namespace is added to the image and chart renderers.

The problem fixed by this PR is not present in ESH, so both ESH and OH1 will be the same in this respect.  

There is a different problem, present in both ESH and OH1, where images and charts are not refreshed on the initial sitemap page, but only after the next "endslide" event (user goes to another page), image refreshes are triggered on that page.  I will open a separate issue for that.

lolodomo:
No chance to have a snapshot to test this fix ?

watou:
I can share a .zip file to you for the runtime (based on the current 1.8 branch) if you want to give it a try?  

But yes, ideally, we would have CI builds on the 1.8 branch, since we lost that feature after 1.8.0.

lolodomo:
Yes please. I would like to check that issue #4037 is fixed too.

watou:
OK, I've uploaded the 37.4MB [distribution-1.8.2-SNAPSHOT-runtime.zip](https://dl.dropboxusercontent.com/u/4286376/fix-issue-1584/distribution-1.8.2-SNAPSHOT-runtime.zip) file.

`MD5 (distribution-1.8.2-SNAPSHOT-runtime.zip) = 89ba33174f3fbf75a742d55488991eea`

Please also note https://github.com/openhab/openhab/issues/4104.

lolodomo:
@watou : unfortunately, your change does not fix my issue #4037.

Here is how is defined one chart in my sitemap:
```
			Switch item=TempChart1Periode mappings=[0="4 heures", 1="12 heures", 2="Jour", 3="Semaine"]
			Chart item=GTemperatureInt period=4h refresh=5000 visibility=[TempChart1Periode==0, TempChart1Periode=="Uninitialized"]
			Chart item=GTemperatureInt period=12h refresh=10000 visibility=[TempChart1Periode==1]
			Chart item=GTemperatureInt period=D refresh=30000 visibility=[TempChart1Periode==2]
			Chart item=GTemperatureInt period=W refresh=30000 visibility=[TempChart1Periode==3]
```
Could you confirm that there were no change and refresh value is still expected to be a number in milliseconds ?

watou:
`refresh=milliseconds` continues to be valid.  I believe the issue with your example is similar to #4104, in that these are two cases where a needed JavaScript variable `imageRefresh` is not being set to a non-false value at the right times.  It is currently only being set when a page slides into view, which doesn't happen either on the initial page, or when the widgets on the current page are changed.  This PR is only intended to fix the regression from around a year ago, where no chart or image refresh was happening in any situation.  

I will assign your #4037 to myself.

lolodomo:
Do you mean I could replace refresh by something else and it would work ? Please explain in this case.
PS: my charts are not in the main page.

watou:
No, it still won't work if you are using `visibility=` in the sitemap, because the initial kickoff of the refreshes doesn't happen after you've switched to a different chart on the page.  I believe the first time that page appears, the chart that's showing should be refreshing at the rate you specified, but as soon as you switch to a different chart on that page, it won't refresh.  I am looking at #4104 and #4037 now.

lolodomo:
So if I suppress visibility= refresh should work as expected ?

watou:
> So if I suppress visibility= refresh should work as expected ?

Yes, as long as it's not the initial page.  I'm researching how to guarantee that the onload/setTimeout cycle happens in all circumstances, which should fix #4104 and #4037, but this PR ought to allow image and chart refresh on all non-initial pages in the meantime.

watou:
@lolodomo please have a look [here](https://community.openhab.org/t/please-help-test-fix-for-image-and-chart-refresh-in-classic-web-ui/8106) for a new JAR to test.  I have a similar sitemap with charts and `visibility=` and the new version works for me for all charts and images.

Issue #3929, opened 2016-01-30
MiOS Action - sendMiosAction() API non-functional under openHAB 2 compat layer


mrguessed:
The MiOS Action binding uses a Rule API with an Item as the first parameter.  This breaks under openHAB 2.x/ESH because they won't resolve to the ESH _Item_ equivalent.

```
sendMiosAction(<Item>, <String>)
sendMiosAction(<Item>, <String>, List<Pairs>)
```

See the community discussion for more details.

ref: https://community.openhab.org/t/problem-with-sendmiosaction-in-oh2/6772

lolodomo:
I tested the jar delivered by @mrguessed for testing and it works well.
You can merge.

PR #3931, opened 2016-01-30
MiOS Action - sendMiosAction() API non-functional under openHAB 2 compat

mrguessed:
The MiOS Action binding uses a Rule API with an Item as the first parameter. This breaks under openHAB 2.x/ESH because they won't resolve to the ESH Item equivalent.

This change moves to declaring the first parameter as Object, and then dynamically calling getName() on this parameter, which resolves correctly in both OH 1.x and ESH.

This PR is to fix issue #3929 

### DIFF
--- a/bundles/action/org.openhab.action.mios/src/main/java/org/openhab/action/mios/internal/MiosAction.java
+++ b/bundles/action/org.openhab.action.mios/src/main/java/org/openhab/action/mios/internal/MiosAction.java
@@ -8,6 +8,7 @@
  */
 package org.openhab.action.mios.internal;
 
+import java.lang.reflect.Method;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


watou:
Would it be possible to only have the `<Object>, ...` method signatures for sendMiosAction and sendMiosScene (not the `<String>, ...` signatures), but then your `getName` method would be:
```
    private static String getName(Object obj) throws Exception {
        if (obj instanceOf String) {
            return (String)obj;
        }
        Method method = item.getClass().getMethod("getName");
        return (String) method.invoke(item);
    }
```
This could also let you remove the overloaded versions of the `sendMios*Internal` methods.

mrguessed:
If I were using an untyped language, I'd likely do that, but I'm trying to play by the Java rules as much as possible....  and localizing my "untyped-ness" to a single chain of the code, and leaving the other parts typed, clean, and not crossing wires too often.

ie. If it's a _String_, go the fastest route, otherwise use the OH1/ESH Hack codeline ;)

Originally I was also going to put in a bunch of validations on the _Object_ path, and then decided against it.  I dislike using the Object-parameter, but it was the best of a bunch of bad options...

lolodomo:
I tested the jar delivered by @mrguessed for testing and it is working well with my rules.

watou:
Please merge this @mrguessed when you're ready.

PR #3950, opened 2016-02-01
[nest] Allow setting setpoints with decimals

### LINKED ISSUE #3944: [Nest] Bug: unable to set temperature to .5 °C values.
Setting temperature to say 19.5°C results in temperature set to 19.0°C. The half degree Celsius is always stripped away.

This is my item (as of binding example):
`Number living_room_target_temperature_c { nest="=[thermostats(Living Room).target_temperature_c]" }`

Even if sending the update command from console, after a short while it re-reads the set value as 19.0°C:
`osgi> openhab update living_room_target_temperature_c 19.5
Update has been sent successfully.
osgi> openhab status living_room_target_temperature_c     
19.5
osgi> openhab status living_room_target_temperature_c     
19.0`

NB: if temperature is set from Nest App it is reported correctly everywhere, so the problem seems to be with sending the value to Nest, not handling it.

(openHAB 1.8.0 on debian RPi B+, org.openhab.binding.nest-1.8.0.jar)

watou:
Fixes #3944.  Reported working https://github.com/openhab/openhab/issues/3944#issuecomment-177942784.

### DIFF
--- a/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/messages/DataModel.java
+++ b/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/messages/DataModel.java
@@ -10,6 +10,7 @@ package org.openhab.binding.nest.internal.messages;
 
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.InvocationTargetException;
+import java.math.BigDecimal;
 import java.net.URLDecoder;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


watou:
@teichsta, this PR could be added to 1.8.1 and 1.9.0, unless it's too late or you would rather not.

teichsta:
also cherry-picked onto 1.8

watou:
Thank you @teichsta and @kaikreuzer!

PR #3956, opened 2016-02-01
RFXCOM Add binding support for 10 sensor types

Speederc:
Add binding support for the following sensor types :
- CurrentEnergy (tested in #3854)
- DateTime (*)
- Fan (*)
- Lighting3 (*)
- Power (*)
- TemperatureHumidityBarometric (tested in #2557)
- TemperatureRain (tested in #3863)
- Thermostat2 (*)
- Thermostat3 (*)
- Weight (*)

 (*) Untested sensor. Can be buggy ! But at least it gives a first draft for future fix.

PR #3957, opened 2016-02-01
RFXCOM Fix message data lengths

Speederc:
Fix message data lengths for the following sensor types : Current, Energy, Rain, UV, Wind.
Not really serious as it is mainly used as info for this kind of sensor, but it will be clearer.

### DIFF
--- a/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComCurrentMessage.java
+++ b/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComCurrentMessage.java
@@ -128,7 +128,7 @@ public class RFXComCurrentMessage extends RFXComBaseMessage {
 
     @Override
     public byte[] decodeMessage() {
-        byte[] data = new byte[13];
+        byte[] data = new byte[14];
 
         data[0] = 0x0D;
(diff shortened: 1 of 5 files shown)


ivanfmartinez:
The new data lengths are correct.

But I found in the RFXComEnergyMessage the enum is wrong, it should have only ELEC2(1) and ELEC3(2), this message does not have ELEC1 according to RFXTRX documentation.

I think this should be commited, because the data lengths are wrong and then the enum can be corrected after.

ivanfmartinez:
Now I have llooked at other PR. This can only be commited after #3956

ivanfmartinez:
I have tried to make a test binding for rfxcom based on the mqtt.test
Can you take a look to see if is correct ?
[https://github.com/ivanfmartinez/openhab/pull/2/files]

Executing here it found many problems in the rfxcom binding, some of them will be solved by @Speederc PRs.

Now after this basic tests we can also create per message specific tests.

Speederc:
Hi @ivanfmartinez 
I agree with you for ELEC2(1) and ELEC3(2) types for RFXComEnergyMessage and not ELEC1(0). I don't know why it was included at the beginning. Can you create a PR to fix this ?

ivanfmartinez:
I have created the PR #4022 for the RFXComEnergyMessage update

ivanfmartinez:
I have created the PR #4034  with a test binding for rfxcom. the test binding already can detect the problem that are being corrected by this PR.

teichsta:
Thanks guys!

PR #3981, opened 2016-02-06
Fix openHAB2 compatibility for RFXCOM 1.x binding

lolodomo:
This patch is a fix for the RFXCOM 1.x binding fixing compatibility with openHAB2.


kaikreuzer:
Thanks!

Issue #3877, opened 2016-01-23
fritzboxtr064 cannot be built


mknet:
It is (still) not possible to build fritzbox064.

Here's my maven error output:

```
[INFO] Compiling 12 source files to /Users/marcel/Development/playground/openhab/bundles/binding/org.openhab.binding.fritzboxtr064/target/classes
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 55.564 s
[INFO] Finished at: 2016-01-23T10:27:59+01:00
[INFO] Final Memory: 36M/285M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:0.24.0:compile (default-compile) on project org.openhab.binding.fritzboxtr064: Compilation failure: Compilation failure:
[ERROR] /Users/marcel/Development/playground/openhab/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java:[22]
[ERROR] import org.apache.xml.serialize.OutputFormat;
[ERROR] ^^^^^^^^^^^^^^
[ERROR] The import org.apache.xml cannot be resolved
[ERROR] /Users/marcel/Development/playground/openhab/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java:[23]
[ERROR] import org.apache.xml.serialize.XMLSerializer;
[ERROR] ^^^^^^^^^^^^^^
[ERROR] The import org.apache.xml cannot be resolved
[ERROR] /Users/marcel/Development/playground/openhab/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java:[43]
[ERROR] OutputFormat format = new OutputFormat(doc);
[ERROR] ^^^^^^^^^^^^
[ERROR] OutputFormat cannot be resolved to a type
[ERROR] /Users/marcel/Development/playground/openhab/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java:[43]
[ERROR] OutputFormat format = new OutputFormat(doc);
[ERROR] ^^^^^^^^^^^^
[ERROR] OutputFormat cannot be resolved to a type
[...]

PR #3989, opened 2016-02-07
Fixing openhab/openhab #3877

mknet:
I just changed two entries of the build.properties. Now, the build runs successfully


bundles/binding/org.openhab.binding.fritzboxtr064/build.properties
```diff
@@ -24,4 +24,5 @@ bin.includes = META-INF/,\
                lib/jna-4.1.0.jar,\
                lib/jna-platform-4.1.0.jar
 output.. = target/classes/
+jars.extra.classpath = lib/xercesImpl.jar
 jars.compile.order = .
```
teichsta: are this both jars.xxx entries necessary?

teichsta:
Thanks @mknet. Could then please also add this binding back to the binding/pom.xml?

teichsta:
also fixed by #4161 which chanhed the implementation …

PR #3994, opened 2016-02-07
Avoid unchanged items to be logged on every modbus poll

Wolfgang1966:
Items were reported changed on every modbus poll although their value didn't change. This caused extensive logging in events.log and heavy cpu load. 

I got the changes from there:
https://groups.google.com/forum/#!topic/openhab/wKLAf5SbMCY

they worked, have it running for several days now


watou:
This PR contains a number of commits that don't seem related to the description. 

Also, a binding should be able to update item states as much as it thinks it needs to; the fact that default logging in events.log logs every update, whether or not the state changed, is possibly what is unwanted, not necessarily the frequent updates themselves. You can change your logging in logback.xml.

Wolfgang1966:
Sorry, I messed up the commits :-((

Forget about this request, I delete it.

PR #3996, opened 2016-02-07
Fixes to tacmi binding

Wolfgang1966:
tacmi-binding:
- rounding values to a meiningful number of digits, cutting off errors from floating point math
- increase timeout on udp socket to avoid returning from read too early
- logging stacktrace of exception to get more helpful output in case of errors
- handle negative temperature values correctly

fritzboxtr064-binding
- adapted version number to make it build

(sorry, I should have done this in different pull requests ... trying to be more careful next time!)


bundles/binding/org.openhab.binding.fritzboxtr064/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.openhab.bundles</groupId>
 		<artifactId>binding</artifactId>
-		<version>1.8.0-SNAPSHOT</version>
+		<version>1.9.0-SNAPSHOT</version>
```
teichsta: remove this change from this PR in order to keep the PRs isolated

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiBinding.java
```diff
@@ -205,8 +208,13 @@ protected void execute() {
                 }
             }
 
+        } catch (SocketTimeoutException e) {
+            logger.info("Receive Timeout on CoE socket");
         } catch (Exception e) {
+            StringWriter w = new StringWriter();
+            e.printStackTrace(new PrintWriter(w));
```
teichsta: remove printStackTrace()

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiBinding.java
```diff
@@ -205,8 +208,13 @@ protected void execute() {
...
             logger.warn("Error in execute: {}, Message: {}", e.getClass().getName(), e.getMessage());
+            logger.warn(w.toString());
```
teichsta: remove this additional log statement and rather enhance the existing statement if that would make sense

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
```diff
@@ -23,6 +25,7 @@
 public final class AnalogValue {
     public BigDecimal value;
     public TACmiMeasureType measureType;
+    public final MathContext mc = new MathContext(6, RoundingMode.HALF_UP);
```
teichsta: i would assume rounding to more than 2 digits would imply false accurance, right?

teichsta:
Hi @Wolfgang1966, thanks for this contribution. Please find some review comments inline. Best, Thomas E.-E.

bundles/binding/org.openhab.binding.fritzboxtr064/pom.xml
Wolfgang1966: Done

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiBinding.java
Wolfgang1966: Changed logging. As message and class are part of the stacktrace nevertheless, I just log the stacktrace. I also changed the severity to "error", because if an exception arrives here it is very likely an error and not just a warning.

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
Wolfgang1966: You are right, my first fix was quite dirty. The new version now really does what it should and does not introduce any precision that doesn't exist in reality.

Wolfgang1966:
Hello, thanks for your comments. I reworked the changes, please recheck and give feedback.

Sorry for the poor quality of my first fixes!

Yours

Wolfgang

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiBinding.java
```diff
@@ -205,8 +208,12 @@ protected void execute() {
                 }
             }
 
+        } catch (SocketTimeoutException e) {
+            logger.info("Receive Timeout on CoE socket");
         } catch (Exception e) {
-            logger.warn("Error in execute: {}, Message: {}", e.getClass().getName(), e.getMessage());
+            StringWriter w = new StringWriter();
+            e.printStackTrace(new PrintWriter(w));
```
teichsta: please remove this printStackTrace()

teichsta:
Thanks for working on this @Wolfgang1966! However the printstackTrace() survived somehow. Please also squash your commits into one after incorporating the feedback.

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiBinding.java
Wolfgang1966: Sorry, I don't understand the problem here. Do you in general want no stack traces in the log?

The reason why I put it there is that I got several log entries that said:

2016-02-07 14:36:57.348 [WARN ] [.b.tacmi.internal.TACmiBinding] - Error in execute: java.lang.NullPointerException, Message: null

At that point, it is not possible to find out what went wrong inside the try-block without having a stacktrace (at least for me). Thats why I want one here. Unfortunately the NPE did not happen again since then, so I was not able to dig deeper into the problem.

So if there is a guideline not to put stack traces to the log, thats fine for me. But where should they go then?

In general I'm not a friend of silently eating up exceptions, because it makes it difficult to correct the problems that they would show.

watou: The correct way to log exceptions would be like:
```
logger.warn("Error in execute", e);
```
The default logback configuration will put most of the stack trace in the log this way.

Wolfgang1966: Thanks for the hint, changed it!

Yours

Wolfgang

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
```diff
@@ -33,7 +33,7 @@ public AnalogValue(int rawValue, int type) {
         switch (type) {
             case 1:
                 measureType = TACmiMeasureType.TEMPERATURE;
-                value = new BigDecimal(rawValue / 10.0);
+                value = new BigDecimal(rawValue).divide(new BigDecimal(10));
```
watou: Minor point: you can save creating one temporary BigDecimal here:
```
value = new BigDecimal(rawValue).movePointLeft(1);
```

teichsta:
PR lgtm except for the squashing …

Wolfgang1966:
Hi, I will ask a colleague for help on squashing. I tried a few things but I'm afraid I will mess up everything now, so please be patient ...

Wolfgang1966:
Sorry, I messed up the commits. I will open a new one with the changes in a few days.

Issue #3992, opened 2016-02-07
[ JDBC Persistence] extend parsing jdbc url


lewie:
@watou,
a little fix for parsing given URL Strings with space char.
from: https://community.openhab.org/t/jdbc-persistence-with-sqlite-not-working-in-openhab-1-8-0/6929/4

````
diff --git a/bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/StringUtilsExt.java b/bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/StringUtilsExt.java
index d524c83..6a06a3a 100644
--- a/bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/StringUtilsExt.java
+++ b/bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/StringUtilsExt.java
@@ -126,10 +126,10 @@
 
         URI dbURI = null;
         try {
-            dbURI = new URI(stringAfterSubstr(url, ":", 1));
+            dbURI = new URI(stringAfterSubstr(url, ":", 1).replaceFirst(" ", ""));
             if (dbURI.getScheme() != null) {
                 props.put("scheme", dbURI.getScheme());
-                dbURI = new URI(stringAfterSubstr(url, ":", 2));
+                dbURI = new URI(stringAfterSubstr(url, ":", 2).replaceFirst(" ", ""));
             }
         } catch (URISyntaxException e) {
             logger.error("parseJdbcURL: URI '{}' is not well formated URISyntaxException: {}", url, e);

````

watou:
Thank you, @lewie!

watou:
Hi @lewie, please double check that #3998 is what you intended and works! :smile:   Thanks a lot.

lewie:
Hi @watou,
Is what I intended it works as expected!

Thank You!

lewie:
Done!

PR #3998, opened 2016-02-08
[jdbc persistence] fix for parsing given URL Strings with space char

watou:
Matches @lewie's diff in #3992.


teichsta:
Thanks @watou … also cherry-picked onto 1.8

PR #4003, opened 2016-02-08
Pebble Timeline action

idserda:
This PR adds an action that enables openHAB to send pins and notifications directly to a Pebble smartwatch.

Documentation: https://github.com/idserda/openhab/wiki/Actions#pebble

Source for the Pebble library: https://github.com/idserda/java-pebble-timeline

Tested on both 2.0 and 1.8. 

I'm still getting used to the 2.0 IDE setup, let me know if I missed any 2.0-specific configuration.


kaikreuzer:
Wow, this sounds great - thanks @idserda!

> let me know if I missed any 2.0-specific configuration.

Yes, it would be nice if you could add similar files as in https://github.com/openhab/openhab/pull/3720/files - this will make the action be packaged with the OH2 distro and have it directly listed as an extension in the UI.

idserda:
Added the config files for OH2.

kaikreuzer:
Looks all good to me, thanks!

PR #4004, opened 2016-02-08
Plex binding: add play/pause property

idserda:
This adds a play/pause property the to Plex binding to easily use one item for toggling between play and pause, depending on the state of the player.


idserda:
Jar with these changes: https://www.dropbox.com/s/ru6lrouzemku1tp/org.openhab.binding.plex_1.9.0.201602082125.jar?dl=0

teichsta:
Thanks @idserda!

PR #4005, opened 2016-02-08
Plex binding: add cover image

idserda:
This adds a property to the Plex binding that holds the cover image for the media that is currently playing.


teichsta:
Thanks @idserda!

PR #4012, opened 2016-02-09
Add suppport for Security2/Keeloq protocol on rfxcom binding

ivanfmartinez:
recognized Security2 messages from rfxcom device


teichsta:
@Speederc could you a look at this one please? @ivanfmartinez could you in turn have a look into the PR series by @Speederc?

Speederc:
Hi @ivanfmartinez 
Are Security2 messages available for all types of RFXCom Transceiver ? According to their user manual it is restricted to RFXtrx433E. In that case, your modification of RFXComInterfaceMessage.java with the new data length would be buggy with the old RFXtrx433.

Can you please squash your commits in a single one as described in : https://github.com/openhab/openhab-distro/blob/master/CONTRIBUTING.md ?
Be careful it seems that the 2 last commits contains various changes not linked to RFXCom (action.Peeble, Zwave...)

ivanfmartinez:
I have one RFXtrk SDK documentarion(9.06a), that talks about RFXtrx433E as different only on Somfy programming, all other text appear to be valid for all RFXtrx transceivers. I have not changed the data length, I have just corrected the indexes for hardware versions and included the checking for two protocols and outputpower.

If the older documentation have different positions, we should ask Bert to known when we can use one position or other. 

This code is giving me the same information as RFxMgr , the old version does not show the correct hardware version. Do you have an older  RFXtrx433 for testing ?

I have made by mistake an upstream  merge, this 2 last commits (peeble and zwave) are already from openhab/master . The git rebase -i does not show the commits to squash.

Speederc:
Hi @ivanfmartinez 
I tested with my RFXtrx433 (with 78 firmware and then 95 firmware).
The hardwareVersion1 and hardwareVersion2 fields moved between the 2 versions to insert the new HomeConfort and KeeLoq capabilities (last one is unavailable with RFXtrx433). They also added the output power field.
Looking at the release notes, it is difficult to find which version change (perhaps version 89_189_243 ?). That could be interesting to add a version test to map the information correctly.
The link between output power field and the value displayed in the RFXMngr is uncleared. On my RFXtrx433, "1C" shows +10dBm.

I'm not an expert on Github. I use "git rebase -i HEAD~nn" with nn the number of commit to squash. But I don't know how to remove the upstream merge...

ivanfmartinez:
I will send and email to Bert asking about how we can work about this changes in messages and about dBm mapping.

Also I have looked again in docs and this InterfaceMessage class should have more work done, because currently  supports only one subtype, and there are more 7 subtypes, with different fields...

I have tried the rebase with ~nn in other patch and become worst.  If necessary I will  create a new branch only with this changes to make sure its correct.

teichsta:
> I have tried the rebase with ~nn in other patch and become worst. If necessary I will create a new branch only with this changes to make sure its correct.

yes, please. The PR should be clean and contain just one commit.

ivanfmartinez:
@techista for the keeloq message support the new #4075 is ok, single commit and no conflicts with current master.

When I got responses from RFXCOM I will make a new PR with a better solution for the RFXComInterfaceMessage.

teichsta:
great, thanks @ivanfmartinez!

This PR is superseded by #4075

Issue #4015, opened 2016-02-09
Netatmo binding is reporting wrong values for battery percentages


robnielsen:
Both of my Netatmo devices that have battery's are currently reporting 100%, when they should be be reporting lower values. The devices have a range of 6000 ... 3600, and my devices have values of 5586 and 5782. Using a formula of (value - min) / (max - min) yields percentages of 83 and 91 respectively.

teichsta:
should have been fixed through #4016

PR #4016, opened 2016-02-09
Netatmo: Fix for defect #4015, added support for BatteryStatus and new BatteryPercentage parameter

robnielsen:
Fix defect https://github.com/openhab/openhab/issues/4015, battery percentages are wrong.

Added support for BatteryStatus, which is similar to WifiStatus and RfStatus.

Added parameter BatteryPercentage which provides the same value as Batteryvp, but is more descriptive of what the parameter actually is used for.

### DIFF
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/weather/NetatmoWeatherBinding.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/weather/NetatmoWeatherBinding.java
@@ -180,13 +180,19 @@ public class NetatmoWeatherBinding {
                                 }
                             }
                                 break;
+                            case BATTERYPERCENT:
+                            case BATTERYSTATUS:
                             case BATTERYVP:
                             case RFSTATUS:
                                 for (Device device : stationsDataResponse.getDevices()) {
                                     for (Module module : device.getModules()) {
                                         if (module.getId().equals(moduleId)) {
                                             switch (measureType) {
+                                                case BATTERYPERCENT:
                                                 case BATTERYVP:
+                                                    state = new DecimalType(module.getBatteryPercentage());
+                                                    break;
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/weather/NetatmoWeatherBinding.java
```diff
@@ -180,13 +180,19 @@ public void execute(OAuthCredentials oauthCredentials, Collection<NetatmoBinding
...
+                            case BATTERYPERCENT:
+                            case BATTERYSTATUS:
                             case BATTERYVP:
                             case RFSTATUS:
                                 for (Device device : stationsDataResponse.getDevices()) {
                                     for (Module module : device.getModules()) {
                                         if (module.getId().equals(moduleId)) {
                                             switch (measureType) {
+                                                case BATTERYPERCENT:
                                                 case BATTERYVP:
```
watou: Does this mean that binding users can't retrieve the battery_vp value because it's returned as a percentage instead of its actual value?  Other the pr lgtm.

robnielsen: yes that is correct. This is from the Netatmo documentation:

    Number   Netatmo_Outdoor_Batteryvp "Outdoor battery status [%d %%]"  (Netatmo {netatmo="00:00:00:00:00:00#02:00:00:00:00:00#Batteryvp"}

watou: Thanks Rob.  Well a percentage is probably a little more useful anyway....

teichsta:
Thanks guys!

lolodomo:
Please include this patch in 1.8.2.

PR #4017, opened 2016-02-10
Rfxcom reset buffer

ivanfmartinez:
This PR clear receiving buffer after sending reset command.
This solves problem when openhab was restated and messages are received from rfxcom during startup and the protocol becomes unsynchronized .
As the RFXCOM protocol does not have any king of start byte, any byte that is != 0 is considered as start.

Now everytime it starts in correct state.


teichsta:
Thanks @ivanfmartinez could you please finally squash your commits into one, please?

ivanfmartinez:
sorry @teichsta  please check the PR #4032 for a correct single commit.

PR #4022, opened 2016-02-11
RFXCOM - define the correct subtypes for EnergyMessage

ivanfmartinez:


teichsta:
Thanks @ivanfmartinez!

PR #4024, opened 2016-02-11
Update of Featured Persistence Services

msfunty:
Added:
- InfluxDB (v. 1.0)
- MapDB

Deleted:
- Logging

### DIFF
--- a/features/openhab-addons/src/main/feature/feature.xml
+++ b/features/openhab-addons/src/main/feature/feature.xml
@@ -467,6 +467,20 @@
         <configfile finalname="${openhab.conf}/services/rrd4j.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/rrd4j</configfile>
     </feature>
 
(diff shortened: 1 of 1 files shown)


watou:
The new .cfg files would also have to be provided (identical to same sections of openhab_default.cfg except for mapdb: and influxdb: components).  Also, why remove logging persistence?

kaikreuzer:
> Also, why remove logging persistence?

Good, question - I would actually assume that the documentation is wrong and it should work.

teichsta:
@msfunty any updates on this? I guess the open question is why you removed the logging service …

kaikreuzer:
Well, testing the logging persistence it shows that it does not work. But the reason is not that logback is missing, since this is included in the feature definition. In order to not lose this information, I would suggest to only put it in comments instead of deleting it.
@teichsta Any chance you could analyze what adaptions the logging persistence service might need to make it work as expected?

teichsta:
> @teichsta Any chance you could analyze what adaptions the logging persistence service might need to make it work as expected?

yes, but i won't be able to work on this before end of next week

kaikreuzer:
This is superseded by https://github.com/openhab/openhab/pull/4144

PR #4028, opened 2016-02-12
Netatmo: Poll The method getstationsdata instead of only calling it at startup.

robnielsen:
The results from the method contains dynamic info such as battery level, wifi strenght, rf strenght, etc.


watou:
Nice change! LGTM

jannegpriv:
Gr8 work!

Now it is easy to add support for fetching temperature/pressure trends :-)

teichsta:
Thanks @robnielsen!

PR #4040, opened 2016-02-13
Changes to have 1.x astro binding work in OH2

watou:
Reported and this change reported working: https://github.com/openhab/openhab-distro/issues/129

Cc: @Marty56


kaikreuzer:
Thanks, @watou!

PR #4044, opened 2016-02-14
Added support for IEC 62056-21 Protocol Mode B and support Landis+Gyr…

sedstef:
… ULTRAHEAT T550 / Siemens UH50

	* added binding parameter initMessage to execute device specific pre request-message codes
	* added support for IEC 62056-21 Protocol Mode B (baudrate change)
	* change data-message buffer dynamicly if data-message bigger than 1MB
	* detect and handle ETX right if device sends some additional data after ETX


teichsta:
Thanks @sedstef!

PR #4045, opened 2016-02-14
OneWireBinding: PushButton Feature

Dennis650:
The new option makes it possible to use a relay like a pushbutton. It
is possible to do that with a rule and timer, but then you have no
control of the real time a button is pushed on slow systems.

Example:
`onewire="deviceId=29.66C30E000000;propertyName=sensed.0;refreshinterval=10";pushbutton=500;invert",autoupdate="false"`


teichsta:
it seems to formatter kills the diff. Could you please rebase this PR again the latest master? The ESH should have been applied to the whole codebase meanwhile …

Dennis650:
Hi Thomas, i made a new PullRequest: #4063

Greetings
Dennis

PR #4047, opened 2016-02-15
[ecobee] Include alerts in the payload when any item is bound to an alerts property

watou:
See [this report](http://developer.ecobee.com/api/topics/need-to-mass-clear-alerts-on-a-thermostat#reply_16614731).

This PR is for the 1.8 branch.  I will make a separate PR for master.

PR #4048, opened 2016-02-15
[ecobee] Include alerts in the payload when any item is bound to an alerts property

watou:
Same as #4047 but for the master branch.


teichsta:
great thanks @watou! No need to split into two separate PRs next time. However a small sidenote "please also merge onto 1.8" would be great however.

watou:
> No need to split into two separate PRs next time.

I only did so because I introduced a breaking change on the master branch and I didn't want to complicate your cherry-pick :smile:, but I will use a single commit in the future.  Thanks, @teichsta.

PR #4049, opened 2016-02-15
Add jdbc Persistence to Paperui and features

lewie:
Based on this [hints ](https://community.openhab.org/t/jdbc-persistence-heave-from-oh1-to-oh2-esh-choosing-a-namespace/7487/5)added JDBC to OH2 features and PaperUI service configuration.

The JDBC Persistence can be fully installed and configured by the PaperUI now.

### DIFF
--- a/features/openhab-addons/src/main/feature/feature.xml
+++ b/features/openhab-addons/src/main/feature/feature.xml
@@ -491,7 +491,64 @@
         <bundle start-level="80">mvn:org.openhab.persistence/org.openhab.persistence.logging/${project.version}</bundle>
         <configfile finalname="${openhab.conf}/services/logging.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/logging</configfile>
     </feature>
-
+    
+    <!-- JDBC Persistence for: Apache Derby, H2, HSQLDB, MariaDB, MySQL, PostgreSQL, SQLite -->
+    <feature name="openhab-persistence-jdbc-derby" description="JDBC Persistence Apache Derby" version="${project.version}">
+        <configfile finalname="${openhab.conf}/services/jdbc.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jdbc</configfile>
+        <feature prerequisite="false" dependency="false">openhab-runtime-base</feature>
+        <feature prerequisite="false" dependency="false">openhab-runtime-compat1x</feature>
+        <bundle start-level="80">mvn:org.apache.derby/derbyclient/${derby.version}</bundle>
+        <bundle start-level="80">mvn:org.openhab.persistence/org.openhab.persistence.jdbc/${project.version}</bundle>
+    </feature>
+    
+    <feature name="openhab-persistence-jdbc-h2" description="JDBC Persistence H2" version="${project.version}">
+        <configfile finalname="${openhab.conf}/services/jdbc.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jdbc</configfile>
+        <feature prerequisite="false" dependency="false">openhab-runtime-base</feature>
+        <feature prerequisite="false" dependency="false">openhab-runtime-compat1x</feature>
+        <bundle start-level="80">mvn:com.h2database/h2/${h2.version}</bundle>
+        <bundle start-level="80">mvn:org.openhab.persistence/org.openhab.persistence.jdbc/${project.version}</bundle>
+    </feature>
+    
+    <feature name="openhab-persistence-jdbc-hsqldb" description="JDBC Persistence HSQLDB" version="${project.version}">
+        <configfile finalname="${openhab.conf}/services/jdbc.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jdbc</configfile>
+        <feature prerequisite="false" dependency="false">openhab-runtime-base</feature>
+        <feature prerequisite="false" dependency="false">openhab-runtime-compat1x</feature>
+        <bundle start-level="80">mvn:org.hsqldb/hsqldb/${hsqldb.version}</bundle>
+        <bundle start-level="80">mvn:org.openhab.persistence/org.openhab.persistence.jdbc/${project.version}</bundle>
+    </feature>
+    
+    <feature name="openhab-persistence-jdbc-mariadb" description="JDBC Persistence MariaDB" version="${project.version}">
+        <configfile finalname="${openhab.conf}/services/jdbc.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jdbc</configfile>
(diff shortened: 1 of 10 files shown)


watou:
Very nice!  I see you changed a number of warns to infos; any chance a number of them could be lowered even further to debug or trace?  Generally, I don't think users will want to see those log messages every time.

lewie:
@watou,
this info-changes based on questions from the Forum to the installation and connectivity of the JDBC Persistence. These messages inform that the database could be contacted and appear only once at the start of the framework.
 If you insist, I'll put them back willingly to debug. ;-)

lewie:
@watou,
````
 -            logger.warn("JDBC::store: ignore Item because it is UnDefType");
+            logger.info("JDBC::store: ignore Item '{}' because it is UnDefType", item.getName());
````
Hmm, this is not only at startup. If a Item is UnDefType it can not be persisted, so this is a unusable state.
I want to know about that, without to  switch to debug.

**Comment by watou** [comment]
I would never insist! :-)  But my preference is to only have a single info message at startup, warn messages on problems, and as many debug messages as would help troubleshoot.

**Comment by lewie** [comment]
Ok, a compromise, I will gather the startup-states in a variable and send them as a single info on  framework start. "UnDefType" tells about a problem, so will set back to warn.

File: features/openhab-addons/src/main/feature/feature.xml [hunk 1]
```diff
@@ -491,7 +491,93 @@
...
+    
+    <feature name="openhab-persistence-jdbc-sqlite" description="JDBC Persistence SQLite" version="${project.version}">
+        <configfile finalname="${openhab.conf}/services/jdbc.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jdbc</configfile>
+        <feature prerequisite="false" dependency="false">jdbc-driver-sqlite</feature>
+        <feature prerequisite="false" dependency="false">openhab-runtime-base</feature>
+        <feature prerequisite="false" dependency="false">openhab-runtime-compat1x</feature>
+        <bundle start-level="80">mvn:org.openhab.persistence/org.openhab.persistence.jdbc/${project.version}</bundle>
+    </feature>
+    
+    <feature name="jdbc-driver-derby" description="JDBC Driver derby-10.12.1.1" version="${project.version}">
```
kaikreuzer: Why did you define features for all the drivers as well? Wouldn't it be good enough to add the bundle to the persistence service features above?

lewie: JPA Persistence, will use the drivers later too. So I externalized them (the user doesn`t see them, or is it?). If we load them all by default, this would cause over 15 MB Traffic.

kaikreuzer: > JPA Persistence, will use the drivers later too

Ok, but as long as the feature only includes a single bundle, you can also reference the single bundle in the JPA Persistence and do not necessarily require a feature. A feature mainly makes sense, if you need to package multiple bundles and/or additional resources or configurations with it.

> the user doesn`t see them

Correct, the will only be visible in the shell when listing available features. But that's ok.

> If we load them all by default, this would cause over 15 MB Traffic.

This was anyhow not my suggestion. I simply meant you could also do a
```
    <feature name="openhab-persistence-jdbc-derby" description="JDBC Persistence Apache Derby" version="${project.version}">
         <configfile finalname="${openhab.conf}/services/jdbc.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jdbc</configfile>
         <feature prerequisite="false" dependency="false">openhab-runtime-base</feature>
         <feature prerequisite="false" dependency="false">openhab-runtime-compat1x</feature>
         <bundle start-level="80">mvn:org.apache.derby/derbyclient/10.12.1.1</bundle>
         <bundle start-level="80">mvn:org.openhab.persistence/org.openhab.persistence.jdbc/${project.version}</bundle>
     </feature>
 ```

lewie: I understand, will change it.

kaikreuzer: If you plan to reuse the bundles, it might also be a good idea to put the version of the drivers as a property into the pom.xml and only refer to it here - then you can change it later at a single place and it will be picked up by JDBC and JPA persistence likewise.

lewie: Excellent idea, will do that! 
in */org.openhab.persistence.jdbc/pom.xml*
````
<bundle start-level="80">mvn:org.apache.derby/derbyclient/${project.properties.derby.version}</bundle>
````
Is that the varpath right?

kaikreuzer: It should probably go best in here (all of them): https://github.com/openhab/openhab/blob/master/features/openhab-addons/pom.xml
And I would call the property simply `${derby.version}`.

kaikreuzer: Alternatively, you could put it in the parent pom, this is how it looks for the openhab2-addons: https://github.com/openhab/openhab2-addons/blob/master/pom.xml#L50

lewie: OK, then all bundles in openHAB can use the same version of the driver later. 
This has advantages for compatibility and stability?!

Can this variable from parent pom used too in local https://github.com/lewie/openhab/blob/add_jdbc_to_paperui_and_features/bundles/persistence/org.openhab.persistence.jdbc/pom.xml as ````${derby.version}````? If yes, I can delete it there.

kaikreuzer: Yes, exactly! I wasn't aware that you need it there anyhow already - then the root pom is indeed the best place as it is shared between the bundles and the features.

lewie: Thank You!

Anything went wrong with my comment wrong document, don' know...
That I can go on, please can you give me a short state about that too:
https://github.com/lewie/openhab/commit/7aee28def126134c7632b75c9b347a6ed079510f#commitcomment-16098127

lewie:
@watou, now I have a single info message at startup and warn messages on problems for "UnDefType".
@kaikreuzer, review round #1 implemented.

kaikreuzer:
Perfect, thanks!

lewie:
@kaikreuzer, it was not as perfect as we thought... ;-)

JDBC Persistence couldn't start without Database mandatory *url*, *user* and *password*. This works with this PR.

But after starting, I do still not get my superduper form, to [config JDBC from ESH-INF](https://github.com/lewie/openhab/blob/add_jdbc_to_paperui_and_features_2/bundles/persistence/org.openhab.persistence.jdbc/ESH-INF/config/config.xml). ;-))

What is missing?

kaikreuzer:
@lewie I saw many exceptions in my log when the service tried to start without any proper configuration set. I have tried to fix it in https://github.com/openhab/openhab/pull/4266 - with this, the service is now correctly activated and I see the configuration screen as expected:

bildschirmfoto 2016-04-08 um 10 48 32

lewie:
Thank You very much @kaikreuzer!

PR #4053, opened 2016-02-16
Updated JodaTime to 2.8.2.

### LINKED ISSUE #3311: Wrong time at rules by using now.getHourOfDay
Hi! 
When I use now.getHourOfDay I take wrong result (+1 hour).
Time at linux (raspbian) is correct. 
Time at Java Machine is correct. 
What's wrong? May it depend of daylight saving? How I can check it? 
TimeZone is Europe/Moscow

Now I use something like this: now.minusMinutes(60).getHourOfDay but it is horrible.

9037568:
This fixes #3311.


teichsta:
ftr: joda has been removed in ESH and is not going to be introduced again (see https://github.com/eclipse/smarthome/issues/594)

kaikreuzer:
> joda has been removed in ESH and is not going to be introduced again

This is not correct, see https://github.com/eclipse/smarthome/tree/master/bundles/model/org.eclipse.smarthome.model.persistence/lib

We therefore also need to update ESH.

kaikreuzer:
Creating the CQ for ESH, I noticed that version 2.8.2 is meanwhile already outdated and 2.9.2 is the latest release.
I would suggest to directly go to 2.9.2 then as it seems to contain many further timezone fixes.
I will do this for ESH, @teichsta, could you please update 1.8.2 accordingly as well?

teichsta:
created PR #4142 … once the (local) build is successful i'll go ahead an merge.

PR #4062, opened 2016-02-17
Denon binding: use different http request method

idserda:
The Denon binding was spamming the log with error level messages when the receiver was not reachable over http (when it was turned off for example). This issue is discussed here: https://community.openhab.org/t/denon-binding-spams-noroutetohostexception/7292

Since HttpUtil catches this error and logs it to level 'error' the binding now uses its own implementation for doing a http request.


teichsta:
Thanks @idserda! Also cherry-picked onto 1.8.

idserda:
Thanks!

PR #4065, opened 2016-02-18
[garadget] initial commit of the Garadget binding

watou:
See wiki [here](https://github.com/openhab/openhab/wiki/Garadget-Binding).


teichsta:
great work! no comment, thanks a lot @watou!

PR #4073, opened 2016-02-20
Trailing whitespace fixes per new IDE setup

dbadia:
These changes are made automatically by the "save as" function of the new IDE if a file is edited.  
Would like to commit all of these white space changes independently of any functional changes to avoid mixing white space and functional changes in the same diff.


teichsta:
Thanks @dbadia!

PR #4077, opened 2016-02-20
[caldav] - add charset option, presence-simulation, filter option

querdenker2k:
- add charset option
- presence-simulation posibility
- filter for categories and event name
- small fixes


teichsta:
nice enhancements, thanks @querdenker2k!

querdenker2k:
Thx @teichsta for merging, Wiki is updated.

PR #4079, opened 2016-02-20
Initial commit for new Powermax alarm binding

lolodomo:
Here is the finished 1.x version of my Powermax/Powermaster binding.

Feedback from beta testers can be found [here](https://community.openhab.org/t/powermax-binding/2824)


bundles/binding/org.openhab.binding.powermax/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+	<parent>
+		<groupId>org.openhab.bundles</groupId>
+		<artifactId>binding</artifactId>
+		<version>1.9.0-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<bundle.symbolicName>org.openhab.binding.powermax</bundle.symbolicName>
+		<bundle.namespace>org.openhab.binding.powermax</bundle.namespace>
+		<deb.name>openhab-addon-binding-PowerMax</deb.name>
```
watou: please use lowercase powermax here.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @author lolodomo
+ * @since 1.9.0
+ */
+
+public class PowerMaxBindingConfig implements BindingConfig {
+
+    private PowerMaxSelectorType selectorType;
+    private String selectorParam;
+    private Item item;
```
watou: Do not hold onto an Item object!  They have an independent lifecycle from BindingConfig objects.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
```diff
@@ -0,0 +1,71 @@
...
+        return selectorParam;
+    }
+
+    /**
+     * @return the item
+     */
+    public Item getItem() {
+        return item;
+    }
+
```
watou: remove

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxActivator.java
```diff
@@ -0,0 +1,56 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Extension of the default OSGi bundle activator
+ *
+ * @author lolodomo
+ * @since 1.9.0
+ */
+
```
watou: remove blank line please

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+    private boolean autoSyncTime = false;
+
+    /** The object to store the current state of the PowerMax alarm system */
+    private PowerMaxState currentState = null;
+
+    /** Number of items managed by the binding */
+    private int itemCount = 0;
+
+    /** Boolean indicating whether or not the update of items is temporarily disabled by the binding */
+    private boolean itemUpdateDisabled = true;
```
watou: Is this left over from debugging, or an intentional feature?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+        logger.debug("Deactivate PowerMax alarm binding");
+        closeConnection();
+    }
+
+    /**
+     * @{inheritDoc
+     */
+    @Override
+    protected long getRefreshInterval() {
+        return 20000; // 20 seconds
```
watou: Any reason to make this configurable with a `refresh` property in the .cfg file?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+     */
+    @Override
+    protected void execute() {
+        logger.debug("PowerMax alarm Execute");
+
+        if (!isProperlyConfigured()) {
+            logger.debug("execute(): not yet properly configured");
+            return;
+        }
+
```
watou: I think the AbstractActiveBinding already manages this situation above?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+                        && ((now - currentState.getSensorLastTripped(i)) > motionOffDelay)) {
+                    if (updateState == null) {
+                        updateState = new PowerMaxState();
+                    }
+                    updateState.setSensorTripped(i, false);
+                }
+            }
+        }
+        if (updateState != null) {
+            updateItemsFronAlarmState(PowerMaxSelectorType.ZONE_STATUS, updateState);
```
watou: Fron typo?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+            currentState.merge(updateState);
+        }
+
+        if (PowerMaxCommDriver.getTheCommDriver() != null) {
+            connected = PowerMaxCommDriver.getTheCommDriver().isConnected();
+        }
+
+        if (connected && (currentState.isPowerlinkMode() != null) && currentState.isPowerlinkMode().equals(Boolean.TRUE)
+                && (currentState.getLastKeepAlive() != null) && ((now - currentState.getLastKeepAlive()) > 60000)) {
```
watou: Make 60000 a constant for better documentation of its meaning?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+                logger.debug("execute(): reconnection failed");
+            }
+        } else if (triggerItemsUpdate) {
+            logger.debug("execute(): update all items");
+            itemUpdateDisabled = false;
+            triggerItemsUpdate = false;
+            // Adjust all the items to the current alarm state
+            updateItemsFronAlarmState(currentState);
+            // Adjust all the items to the current alarm settings
+            updateItemsFronPanelSettings();
```
watou: "Fron" also.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void bindingChanged(BindingProvider provider, String itemName) {
+        logger.debug("bindingChanged(): Item Name: {}", itemName);
+        logger.debug("Provider items count: {}", provider.getItemNames().size());
```
watou: Possible to log only one debug line here?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: I think I added that for openHAB2 which starts the service several times during the init process. Doing that,

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+
+        itemUpdateDisabled = true;
```
watou: I understand now how you are using this instance variable.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+        if (provider.getItemNames().size() > 0) {
+            triggerItemsUpdate = true;
+        }
+        itemCount = provider.getItemNames().size();
+
+        super.allBindingsChanged(provider);
+    }
```
watou: Nice feature -- refresh all after .items file changed, right?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
lolodomo: There is nothing different from the freebox binding for example: https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.freebox/src/main/java/org/openhab/binding/freebox/FreeboxBindingConfig.java

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxActivator.java
lolodomo: Which one ? The line just before "public class PowerMaxActivator ..." ?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+                    }
+                    break;
+                case PARTITION_ARMED:
+                    if (command instanceof OnOffType) {
+                        armMode = command.equals(OnOffType.ON) ? "Armed" : "Disarmed";
+                    }
+                    break;
+                case PGM_STATUS:
+                    if (command instanceof OnOffType) {
+                        actionPGMX10 = command.equals(OnOffType.ON) ? "ON" : "OFF";
```
watou: Could be 
```
actionPGMX10 = command.toString();
```

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+                    }
+                    break;
+                case X10_STATUS:
+                    if (command instanceof OnOffType) {
+                        actionPGMX10 = command.equals(OnOffType.ON) ? "ON" : "OFF";
```
watou: same

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+                    try {
+                        devicePGMX10 = Byte.parseByte(config.getSelectorParam());
+                    } catch (NumberFormatException e) {
+                        logger.info("PowerMax alarm binding: invalid X10 device id: {}", config.getSelectorParam());
+                        actionPGMX10 = null;
+                    }
+                    break;
+                case ZONE_BYPASSED:
+                    if (command instanceof OnOffType) {
+                        bypass = command.equals(OnOffType.ON) ? Boolean.TRUE : Boolean.FALSE;
```
watou: same

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Yes, it is intentional. Certain things have to be done regularly. I cannot let the user fill any value. If I do, certain features will no more work as expected.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Yes, I will fix that.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+                        if (currentState.getLastKeepAlive() != null) {
+                            currentState.setLastKeepAlive(System.currentTimeMillis());
+                        }
+                    }
+                } else if (commandStr.equalsIgnoreCase("log_setup")) {
+                    settings.log();
+                } else if (commandStr.equalsIgnoreCase("help_items")) {
+                    settings.helpItems();
+                } else {
+                    logger.info("PowerMax alarm binding: rejected command {}", commandStr);
```
watou: Could this be a WARN level, or it INFO level better?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+    /**
+     * @{inheritDoc
+     */
+    @Override
+    protected void internalReceiveUpdate(String itemName, State newState) {
+        // the code being executed when a state was sent on the openHAB
+        // event bus goes here. This method is only called if one of the
+        // BindingProviders provide a binding for the given 'itemName'.
+        logger.debug("internalReceiveUpdate({},{}) is called", itemName, newState);
+        logger.info("Item {}: state can be updated only by the PowerMax alarm binding.", itemName);
```
watou: same comment about WARN vs. INFO level.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Do you mean I should rather set them in the constructor ?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: OK
The idea is to check a particular message is received at least once every minute.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Yes and that works well ;)
The refresh of items is done at the next refresh cycle (max 20 seconds later).

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,881 @@
...
+        if (StringUtils.isNotBlank((String) config.get("autoSyncTime"))) {
+            autoSyncTime = Boolean.valueOf((String) config.get("autoSyncTime"));
+        }
+
+        if (StringUtils.isNotBlank((String) config.get("pinCode"))) {
+            pinCode = (String) config.get("pinCode");
+        }
+
+        this.setProperlyConfigured(true);
+    }
```
watou: instead of the form

1. setProperlyConfigured(false)
2. parse and early return if problem
3. setProperlyConfigured(true)

I suggest the form
```java
try {
 parse
 if problem, throw new ConfigurationException("reason why")
 ...
 setProperlyConfigured(true);
} catch (ConfigurationException ce) {
 setProperlyConfigured(false);
 throw ce;
}
```

lolodomo: Am I sure in this case that ProperlyConfigured will be set to false in case of exception ?

bundles/binding/org.openhab.binding.powermax/pom.xml
lolodomo: Ok

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok I set the attributes in the constructor.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Here, I want a Boolean, not a String.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
watou: It probably doesn't matter too much, but be aware that setProperlyConfigure(false) may [stop the thread that calls execute()](https://github.com/openhab/openhab/blob/1.8/bundles/core/org.openhab.core/src/main/java/org/openhab/core/service/AbstractActiveService.java#L119), so it's slightly better to only call that method once with the final answer.  What about
```java
boolean properlyConfigured = false;
try {
 parse and throw new ConfigurationException as needed
 ...
 properlyConfigured = true;
} finally {
 setProperlyConfigured(properlyConfigured);
}
```

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Yes, I changed few logger.info to logger<;warn

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok, fixed.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
lolodomo: Checking my code, I only call getItem in my code to know the kind of the item (StringItem or SwitchItem or ...).
If you think that is better, I will store here the "kind of item" (class name ?) rather than the item itself ?

lolodomo:
I committed locally my changes and then squashed my two commits.
After that, I pushed in Git to a new branch named powermax_upd1.
Finally, how can I update my PR ?

watou:
If you push to the same branch that you used to submit the pull request, then the existing pull request is updated.  If you squash your commits on that branch, you will have to force push the branch.  Only use a single branch for a single pull request.

That being said, I will probably have more review comments (had to take a break earlier).

lolodomo:
Ok, PR updated.

Thank you for your quick feedback, even if not yet finished. I was afraid it could take several months before I got a first feedback !

For your first two comments relative to the Item stored in class PowerMaxBindingConfig, I am waiting for your final advice. As said, I can a class name (kind of item) rather than an Item if it is important.

lolodomo:
I see that there are conflicts for the few files I updated, probably because my fork has been done 1 or 2 weeks ago.
What the best way to solve that ?

watou:
> What the best way to solve that ?

Something like:
```bash
$ git fetch origin
$ git merge-base powermax master
<commit_id>
$ git rebase -i <commit_id>
change second and subsequent `pick` to `squash` and save file
$ git rebase origin/master
fix conflicts and follow instructions printed to finish the rebase
$ git push -f origin powermax
```

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
watou: I like to save the acceptedDataTypes and acceptedCommandTypes from the item when constructing the binding config object:
```java
this.acceptedDataTypes = new ArrayList<Class<? extends State>>(item.getAcceptedDataTypes());
this.acceptedCommandTypes = new ArrayList<Class<? extends Command>>(item.getAcceptedCommandTypes());
```
and then the binding doesn't need to know anything at all about item classes—it only needs to convert to a type or command on the list when it's time to postUpdate or sendCommand.  This removes hard-coded assumptions about which items accept which types and commands, knowledge that lives outside the binding.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
watou: Any chance that was related to the `setProperlyConfigured(false)` call before, that would stop the `AbstractActiveService` and then start it again?  Are there other 1.x bindings that have this problem in OH2?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+     * @{inheritDoc
+     */
+    @Override
+    protected void internalReceiveUpdate(String itemName, State newState) {
+        // the code being executed when a state was sent on the openHAB
+        // event bus goes here. This method is only called if one of the
+        // BindingProviders provide a binding for the given 'itemName'.
+        logger.debug("internalReceiveUpdate({},{}) is called", itemName, newState);
+        logger.warn("Item {}: state can be updated only by the PowerMax alarm binding.", itemName);
+    }
```
watou: Could you just remove this override altogether, since the binding is not interested in others updating the state of an item it's bound to?  Also, the WARN message isn't accurate "state can be updated only by the PowerMax alarm binding" because it *can* be updated by a rule or another binding -- it's just that your binding doesn't care.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+
+    /**
+     * @{inheritDoc
```
watou: {@inheritDoc}

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+
+            if (StringUtils.isNotBlank((String) config.get("pinCode"))) {
+                pinCode = (String) config.get("pinCode");
+            }
+
+            properlyConfigured = true;
+
+        } catch (ConfigurationException e) {
+
+            logger.warn("PowerMax alarm binding: {}", e.getReason());
```
watou: 1. you would want to rethrow the exception if you caught it, and 
2. perhaps no reason to catch it because it would be logged outside this method anyway?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+     * @param event
+     */
+    @Override
+    public void powerMaxEventReceived(EventObject event) {
+        PowerMaxEvent powerMaxEvent = (PowerMaxEvent) event;
+        PowerMaxBaseMessage message = powerMaxEvent.getPowerMaxMessage();
+
+        logger.debug("powerMaxEventReceived(): received message {}", (message.getReceiveType() != null)
+                ? message.getReceiveType().getDescription() : String.format("%02X", message.getCode()));
+
```
watou: perhaps wrap above in `if (logger.isDebugEnabled()) { ... }` to avoid the `String.format` overhead on each call of this method under normal operation?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+        }
+
+        PowerMaxState updateState = message.handleMessage();
+        if (updateState != null) {
+
+            if ((currentState.isPowerlinkMode() != null) && currentState.isPowerlinkMode().equals(Boolean.TRUE)
+                    && (updateState.isDownloadSetupRequired() != null)
+                    && updateState.isDownloadSetupRequired().equals(Boolean.TRUE)) {
+                // After Enrolling Powerlink or if a reset is required
+                logger.info("PowerMax alarm binding: Reset");
```
watou: info message not the same as comment above, correct?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+                PowerMaxCommDriver.getTheCommDriver().startDownload();
+                if (currentState.getLastKeepAlive() != null) {
+                    currentState.setLastKeepAlive(System.currentTimeMillis());
+                }
+            } else if ((currentState.isPowerlinkMode() != null) && currentState.isPowerlinkMode().equals(Boolean.FALSE)
+                    && (updateState.getLastKeepAlive() != null)) {
+                // Were are in standard mode but received a keep alive message
+                // so we switch in PowerLink mode
+                logger.info("PowerMax alarm binding: Switching to Powerlink mode");
+                PowerMaxCommDriver.getTheCommDriver().startDownload();
```
watou: debug messages and method names dissimilar: intentional?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/state/PowerMaxState.java
```diff
@@ -0,0 +1,911 @@
...
+        return result;
+    }
+
+    /**
+     * Keep only data that are different from another state and reset all others data to undefined
+     *
+     * @param otherState
+     *            the other state
+     */
+    public void keepOnlyDiifWith(PowerMaxState otherState) {
```
watou: typo "Diif" should be "Diff" or "Differences"?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+                updateItemsFromAlarmState(prov, name, selector, state);
+            }
+        } else {
+            for (String itemName : provider.getItemNames()) {
+                if ((name == null) || itemName.equals(name)) {
+                    State itemState = null;
+
+                    PowerMaxBindingConfig config = provider.getConfig(itemName);
+                    Integer num = config.getSelectorIntParam();
+                    Item item = config.getItem();
```
watou: This is where you would be concerned with acceptedDataTypes saved in the config instead of the item, and rework the following code to simply satisfy a type on the list instead of being concerned with item types, keeping of the knowledge of item->type mapping out of the binding.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+            PowerMaxSelectorType selector) {
+
+        if (itemUpdateDisabled) {
+            logger.debug("updateItemsFromPanelSettings(): items update disabled");
+            return;
+        }
+
+        if (provider == null) {
+            for (PowerMaxBindingProvider prov : providers) {
+                updateItemsFromPanelSettings(prov, name, selector);
```
watou: there should never be a `null` on the `providers` list, but if there were, stack overflow from infinite recursion! :-/

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
```diff
@@ -0,0 +1,907 @@
...
+            }
+        } else {
+            PowerMaxPanelSettings settings = PowerMaxPanelSettings.getThePanelSettings();
+
+            for (String itemName : provider.getItemNames()) {
+                if ((name == null) || itemName.equals(name)) {
+                    State itemState = null;
+
+                    PowerMaxBindingConfig config = provider.getConfig(itemName);
+                    Item item = config.getItem();
```
watou: Another case where TypeParser could be used and you could support any kind of item by just using its list of accepted data types.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxGenericBindingProvider.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getBindingType() {
+        return "powermax";
+    }
+
+    /**
+     * @{inheritDoc
```
watou: {@inheritDoc}  -- these got mangled by an autoformatter a long time ago :)

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxGenericBindingProvider.java
```diff
@@ -0,0 +1,140 @@
...
+                    }
+                    break;
+                default:
+                    break;
+            }
+        } catch (IllegalArgumentException e) {
+            throw new BindingConfigParseException(
+                    "item '" + item.getName() + "': parse error of string " + bindingConfig);
+        }
+    }
```
watou: some of these fail-fast checks could be unnecessarily restrictive.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxConnector.java
```diff
@@ -0,0 +1,213 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * An abstract class for the communication with the Visonic alarm panel that
+ * handles stuff common to all communication types
+ *
+ * @author lolodomo
+ * @since 1.9.0
+ */
+
```
watou: normally there is no blank line between class comment and beginning of class definition.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxConnector.java
```diff
@@ -0,0 +1,213 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void sendMessage(byte[] data) {
+        try {
+            output.write(data);
+            output.flush();
+        } catch (IOException ioException) {
+            logger.debug("sendMessage(): Writing error: {}", ioException.getMessage());
+            connected = false;
```
watou: `setConnected(false)` instead here and below?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
```diff
@@ -0,0 +1,182 @@
...
+            int len = -1;
+
+            while (((len = in.read(tmpData)) > 0) && !interrupted) {
+
+                for (int i = 0; i < len; i++) {
+
+                    if (index >= dataBufferMaxLen) {
+                        // too many bytes received, try to find new start
+                        logData = Arrays.copyOf(dataBuffer, index);
+                        logger.debug("Truncating message {}", DatatypeConverter.printHexBinary(logData));
```
watou: Note the overhead of `DatatypeConverter.printHexBinary(logData)` even when not in DEBUG logging level.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
```diff
@@ -0,0 +1,182 @@
...
+
+                                // find new preamble
+                                index = 0;
+                            } else if (msgLen == 0) {
+                                // CRC check failed for a message with an
+                                // unknown length
+                                logger.debug("Message length is now {} but message is apparently not complete",
+                                        index + 1);
+                                logData = Arrays.copyOf(dataBuffer, index);
+                                logger.debug("Partial message {}", DatatypeConverter.printHexBinary(logData));
```
watou: same comment as above

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
```diff
@@ -0,0 +1,182 @@
...
+                            } else {
+                                // CRC check failed for a message with a known
+                                // lenght
+                                logData = Arrays.copyOf(dataBuffer, index);
+                                logger.debug("Message with CRC error {}", DatatypeConverter.printHexBinary(logData));
```
watou: same comment as above

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
```diff
@@ -0,0 +1,182 @@
...
+     *            the size of the message in the buffer
+     *
+     * @return true if the CRC is valid or false if not
+     */
+    private boolean checkCRC(byte[] data, int len) {
+        byte checksum = PowerMaxCommDriver.computeCRC(data, len);
+        byte expected = data[len - 2];
+        if (checksum != expected) {
+            logger.debug("CRC check failed (expected {}, got {})", String.format("%02X", expected),
+                    String.format("%02X", checksum));
```
watou: Is this rare and interesting enough to be at WARN level?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxSerialConnector.java
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * @author lolodomo
+ * @since 1.9.0
+ */
+
+public class PowerMaxSerialConnector extends PowerMaxConnector {
+
+    private static final Logger logger = LoggerFactory.getLogger(PowerMaxSerialConnector.class);
+
+    private String serialPortName = "";
```
watou: no need to initialize to a value if only constructor sets it.  Also, if you change it nowhere but in the constructor, you could even make it final.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxSerialConnector.java
```diff
@@ -0,0 +1,164 @@
...
+            if (getInput().markSupported()) {
+                getInput().reset();
+            }
+
+            setReaderThread(new SerialReaderThread(getInput(), this));
+            getReaderThread().start();
+
+            setConnected(true);
+        } catch (NoSuchPortException noSuchPortException) {
+            logger.debug("open(): No Such Port Exception: {}", noSuchPortException.getMessage());
```
watou: These would be important enough to log as WARN, no?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxTcpConnector.java
```diff
@@ -0,0 +1,112 @@
...
+ * @since 1.9.0
+ */
+
+public class PowerMaxTcpConnector extends PowerMaxConnector {
+
+    private static final Logger logger = LoggerFactory.getLogger(PowerMaxTcpConnector.class);
+
+    private String ipAddress = "";
+    private int tcpPort;
+    private int connectTimeout;
```
watou: these could also be marked final if only set in the constructor, and don't need to be initialized here.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxTcpConnector.java
```diff
@@ -0,0 +1,112 @@
...
+
+            setInput(tcpSocket.getInputStream());
+            setOutput(tcpSocket.getOutputStream());
+
+            setReaderThread(new PowerMaxReaderThread(getInput(), this));
+            getReaderThread().start();
+
+            setConnected(true);
+        } catch (UnknownHostException exception) {
+            logger.debug("open(): Unknown Host Exception: {}", exception.getMessage());
```
watou: might be better at WARN level (and below)?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxTcpConnector.java
```diff
@@ -0,0 +1,112 @@
...
+        logger.debug("close(): Closing TCP Connection");
+
+        super.cleanup();
+
+        try {
+            if (tcpSocket != null) {
+                tcpSocket.close();
+            }
+        } catch (IOException ioException) {
+            logger.debug("close(): Closing connection error: {}", ioException.getMessage());
```
watou: Only need the one catch block, since IOException subclass of Exception and messages are the same.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxBaseMessage.java
```diff
@@ -0,0 +1,192 @@
...
+     *
+     * @return a new state containing all changes driven by the message
+     */
+    public PowerMaxState handleMessage() {
+        // Send an ACK if needed
+        if (isAckRequired()) {
+            PowerMaxCommDriver.getTheCommDriver().sendAck(this, (byte) 0x02);
+        }
+
+        logger.debug("{}message handled by class {}: {}", (receiveType == null) ? "Unsupported " : "",
```
watou: missing space between `{}` and message

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxBaseMessage.java
```diff
@@ -0,0 +1,192 @@
...
+            return (PowerMaxBaseMessage) cl.getConstructor(byte[].class).newInstance(message);
+        } catch (InstantiationException e) {
+            return new PowerMaxBaseMessage(message);
+        } catch (IllegalAccessException e) {
+            return new PowerMaxBaseMessage(message);
+        } catch (NoSuchMethodException e) {
+            return new PowerMaxBaseMessage(message);
+        } catch (InvocationTargetException e) {
+            return new PowerMaxBaseMessage(message);
+        }
```
watou: would any logging in this method help diagnose a problem?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxCommDriver.java
```diff
@@ -0,0 +1,621 @@
...
+     * @param armMode
+     *            the arm mode. Allowed values are: Disarmed, Stay, Armed,
+     *            StayInstant, ArmedInstant, Night, NightInstant
+     * @param pinCode
+     *            the PIN code. A string of 4 characters is expected
+     *
+     * @return true if the message was sent or false if not
+     */
+    public boolean requestArmMode(String armMode, String pinCode) {
+        logger.debug("requestArmMode(): armMode = {}", armMode == null ? "null" : armMode);
```
watou: I thought that if `armMode` is `null`, the logger class would do the same as your ternary operator?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxCommDriver.java
```diff
@@ -0,0 +1,621 @@
...
+     * @param action
+     *            the requested action. Allowed values are: OFF, ON, DIM, BRIGHT
+     * @param device
+     *            the X10 device number. null is expected for PGM
+     *
+     * @return true if the message was sent or false if not
+     */
+    public boolean sendPGMX10(String action, Byte device) {
+        logger.debug("sendPGMX10(): action = {}, device = {}", action == null ? "null" : action,
+                device == null ? "null" : device);
```
watou: same comment as above

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxCommDriver.java
```diff
@@ -0,0 +1,621 @@
...
+     *
+     * @return true if the message was sent or false if not
+     */
+    public boolean sendSetTime() {
+        logger.debug("sendSetTime()");
+
+        boolean done = false;
+
+        GregorianCalendar cal = new GregorianCalendar();
+        if (cal.get(Calendar.YEAR) >= 2000) {
```
watou: is this sufficient to meet the "time synchronized" test you intend?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/state/PowerMaxPanelType.java
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.powermax.internal.state;
+
+/**
+ * Used to store main characteristics of each Visonic alarm panel type in an ENUM
```
watou: Could you add links to source reference documentation in comments throughout the binding so other contributors and reviewers could understand the correlation between your code and the documented protocols involved?

watou:
I finished my full pass through the code.  Impressive amount of work, @lolodomo!  I suspect another maintainer may want to have a look, too.  Please find my comments inline.  Interested to have links to reference manuals accompany code as much as practical so others can understand correlation of reference to implementation. Also, probably no need to squash again until final "lgtm".

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxBaseMessage.java
lolodomo: No.Check the first argument.

lolodomo:
The protocol is not officially documented but was reverse-engineered.
I will put somewhere in the code and in the wiki page where to find information about the protocol.
My implementation code is largely inspired from the Vera plugin (lua code).

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: I need to know the item type in oder to send on the bus ON  / OFF for a switch item and OPEN / CLOSE for a contact item. The state object to be used is not the same.
Can you point me to a binding that inplements that in a clean way please ?

watou: > I need to know the item type in oder to send on the bus ON / OFF for a switch item and OPEN / CLOSE for a contact item. The state object to be used is not the same.

You don't need to know the item type, but you need to know the list of types that an item can receive as state and commands.  So keep the lists, and use your own loops over the types, or use the TypeParser if it's a String object.

Examples of other bindings that do something similar:

* o.o.b.myq.internal.MyqBinding.java#L213
* o.o.b.garadget.internal.AbstractDevice.java#L134
* o.o.b.http.internal.HttpGenericBindingProvider.java#L299
* o.o.b.mqtt.internal.MqttMessageSubscriber.java#L65

lolodomo: Ok thank you, I understand now perfectly how to implement that and will update my code.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxConnector.java
lolodomo: Fixed in all files

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: provider and providers
provider is a parameter of the method. When null, I consider all available providers.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxGenericBindingProvider.java
lolodomo: Othet bindings are doing that.
I think it is a good way to inform the user of inappropriate item types.
I could suppress these checks and the user will have items never set without any hints.

watou: My point was that, if using `acceptedDataTypes` instead of a hardcoded mapping of item types to data types, then your binding would be supporting more kinds of items than it was previously.  If you have a reliable way of supporting the fail-fast purpose of this method without unnecessarily restricting which item types can be used, then please do!

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
watou: My point was that, if iterating through providers gave a `null` prov, then this would lead to infinite recursion, stack overflow, etc.  If there are never any `null`s on the providers list, then this won't happen.  Just pointing out the potential for risk of infinite recursion without an explicit `null` check before recursing.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxGenericBindingProvider.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Yes, that is expected.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: All my logs at INFO or WARN level start with "PowerMax alarm binding" while for DEBUG level, I start the message with the name of the method. The name of the method is not talking to the final user.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/state/PowerMaxState.java
lolodomo: Ok => renamed keepOnlyDifferencesWith

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxConnector.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
lolodomo: I added a test if (logger.isDebugEnabled()) {}

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
lolodomo: if (logger.isDebugEnabled()) {

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
lolodomo: if (logger.isDebugEnabled()) {

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxSerialConnector.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxTcpConnector.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxTcpConnector.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok, suppressed

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Changed to
* suppress the try / catch
* trigger an exception
* keep the log
* keep a call to setProperlyConfigured

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: Ok, I added a test if (prov != null)

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxGenericBindingProvider.java
lolodomo: OK, I suppressed the control on item type.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxReaderThread.java
lolodomo: Yes, I changed to WARN level and I added the display of the message content in this log entry.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxSerialConnector.java
lolodomo: Log as INFO level is done in PowerMaxBinding.openConnection to inform the user if connection succeeded or failed.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/connector/PowerMaxTcpConnector.java
lolodomo: Log as INFO level is done in PowerMaxBinding.openConnection to inform the user if connection succeeded or failed.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxBaseMessage.java
lolodomo: My intention here is just to consider as a standard message any message that I don't handle in the binding.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxCommDriver.java
lolodomo: I need to provide a year from 2000 (16 for 2016). If current date is before year 2000, I just avoid synchronizing date/time.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxCommDriver.java
lolodomo: I checked and you're right.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/message/PowerMaxCommDriver.java
lolodomo: OK

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
lolodomo: OK, suppressed

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/PowerMaxBindingConfig.java
lolodomo: OK, removed.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: @watou : what string value shall I provide for a date/time ?

watou: If you want to provide a StringType state from a Date or Calendar (but you don't have to of course), you could construct a DateTimeType and return its toString() result.

lolodomo:
I just updated my PR (squashing my commits).
DON'T MERGE the PR, there is still one thing to be fix.

By the way, it looks like there are still conflicts on the few files I updated. I don't understand because I took the files less than one hour ago from the repo after a call to "git pull" to update the files and then update them to add my new entries. As it concerns mainly the adding of feature for OH2, I could suppress these files from my PR. The only remaining problem would be the file bundles/binding/pom.xml I need to update to add my new binding...

watou:
Even though you have fetched from the remote, your branch is still based on an earlier commit on master.  You would probably want to rebase your commit, something like:
```
git checkout powermax
git rebase origin/master
```
and then fix the merge conflicts (denoted by `<<<<<` and `>>>>>` marks) and follow the instructions that the git command gives you to complete the process.  I suggest making this a complete PR for your new binding.

lolodomo:
To be honest, I think I am a little lost with Git and all the repo !
As soon as I fork openHAB in GitHub to have my own openHAB repo with write access, I can no more update it from the official repo ?

lolodomo:
The easiest could be to fork again openHAB repo ?

watou:
If you type `git remote -v`, do you get this?
```
origin	https://github.com/openhab/openhab.git (fetch)
origin	https://github.com/lolodomo/openhab.git (push)
```
This means that when you `git fetch`, it's coming directly from the main repo, but when you `git push`, you are pushing to your fork.  The above remote is set up automatically when you install the new IDE.  Git used to not allow a separate pushurl, but the modern way as above makes so much more sense.

When you rebase your branch, you usually have to "force push" it, like this:
```
git push -f origin powermax
```
This should automatically update your open PR.

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: I found in DateTimeType what format is expected but you're right I will build a DateTimeType and then call toString to get my string value to be used in TypeParser.

lolodomo:
I realize I could have totally misunderstood how to work with the new IDE !!
After forking in Github, I cloned my fork in a new location. So I have one openhab repo installed with the IDE I am using only to pull data from official openhab repo and I have a second repo linked to my gothub fork that I use to develop my code and push requests.
Your message let me think I should have only one !

lolodomo:
@watou please help.
First I added a new remote: git add remote official https://github.com/openhab/openhab.git
Then I try to create a tracking branch: git checkout -b refmaster official/master
But I get this error:
```
Fatal: Cannot update paths and switch to branch 'refmaster' at the same time
Did you intend to checkout 'official/master' which can not be resolved as commit ?
```
Why official/master cannot be resolved ?
As soon as I have this tracking branch, I should be able to merge my local branch powermax from this new branch I am trying to create..

watou:
I strongly recommend that you use the single `origin` remote for your local repo and remove all the other remotes:
```
git remote remove official
git remote remove origin
git remote remove "whatever other remotes you've defined"
git remote add origin https://github.com/openhab/openhab.git
git remote set-url --push origin https://github.com/lolodomo/openhab.git
```

Then you should be able to push your branch to your fork at github, which will update the PR:
```
git checkout powermax
git push -f origin powermax
```

lolodomo:
@watou: a big thank you for all your help.
Now I was able to merge my powermax branch and fix what git identified as conflicts.
A lot of commits have been done in my powermax branch.
Shall I squash every commits including commits from others before pushing again my branch ?
I should just replace "pick" by "squash" on all the lines except the first one ?

bundles/binding/org.openhab.binding.powermax/src/main/java/org/openhab/binding/powermax/internal/PowerMaxBinding.java
lolodomo: I kept this test by security. It does not introduce any problem.

lolodomo:
I pushed without squashing.
Will see at least if jenkins jobs are ok...

lolodomo:
@watou what's the next step before a final merge ?

I don't know if I have to squash with the commits resulting from the merge of other's changes with my branch ?

watou:
@lolodomo if you think this binding is ready to merge, I have no further comments.  Please squash your 3 commits into 1 and then it should be good to merge (if you agree).  Thanks for working through all the comments!

lolodomo:
@watou: rebase failed with the commit "Adding feature for intertechno" !

The problem is probably due the fact I am not familiar with git rebasing.

When I launch "git rebase -i HEAD~3", I have a long list of commits, the first two ones are mine, all the others are commits from other contributors resulting of the merge of the master branch to my powermax branch.

My idea would be to change the order of commits, putting all the other contributors commits first keeping a "pick" for each of them, and then put my 2 personal commits, the first with "pick" and the second with "squash". Is it the good solution ? Is there a risk that I corrupt my branch and all my work ?!

watou:
I would try to squash like this
```
git merge-base powermax master
<prints common commit>
git rebase -i <that number>
```
That usually works for me.  I hope it doesn't make it worse...

lolodomo:
@watou: is it expected to have such a big list including commits from other contributors ?!

By the way, if I move my commit at the bottom (last commit) when rebasing, it seems I inject again the files conflicts. That makes sense to me.
If I fix the conflict (what I did and you can see the last commit) and then try to rebase/squash, my commit is again at first in the list.
That is probably trivial but I am out of idea !

lolodomo:
@watou: that would be easier to create a new branch from an up-to-date master branch, copy all my files inside and provide a new PR ?

watou:
> create a new branch from an up-to-date master branch, copy all my files inside and provide a new PR ?

Certainly!

lolodomo:
I close this PR and will submit a new one later.

PR #4092, opened 2016-02-23
Allow String commands in MyQ binding

beowulfe:
The MyQ binding can bind to String items. This change allows it to accept String commands as well.

Signed-off-by: Andy Lintner <dev@beowulfe.com> (github: beowulfe)

### DIFF
--- a/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
+++ b/bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
@@ -316,6 +316,19 @@ public class MyqBinding extends AbstractBinding<MyqBindingProvider> {
                 } else if (command.equals(OnOffType.OFF) || command.equals(UpDownType.DOWN)) {
                     myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 0);
                     beginRapidPoll(true);
+                } else if (command instanceof StringType) {
+                    String stringValue = ((StringType) command).toString();
+                    if (stringValue.equalsIgnoreCase(GarageDoorStatus.OPEN.getLabel())
+                            || stringValue.equalsIgnoreCase(GarageDoorStatus.OPENING.getLabel())) {
+                        myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 1);
+                        beginRapidPoll(true);
+                    } else if (stringValue.equalsIgnoreCase(GarageDoorStatus.CLOSED.getLabel())
+                            || stringValue.equalsIgnoreCase(GarageDoorStatus.CLOSING.getLabel())) {
+                        myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 0);
+                        beginRapidPoll(true);
+                    } else {
+                        logger.warn("Unknown string command {}", stringValue);
+                    }
                 } else {
                     logger.warn("Unknown command {}", command);
                 }


beowulfe:
FYI @computergeek1507

bundles/binding/org.openhab.binding.myq/src/main/java/org/openhab/binding/myq/internal/MyqBinding.java
```diff
@@ -316,6 +316,19 @@ private void computeCommandForItem(Command command, String itemName) {
...
+                            || stringValue.equalsIgnoreCase(GarageDoorStatus.OPENING.getLabel())) {
+                        myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 1);
+                        beginRapidPoll(true);
+                    } else if (stringValue.equalsIgnoreCase(GarageDoorStatus.CLOSED.getLabel())
+                            || stringValue.equalsIgnoreCase(GarageDoorStatus.CLOSING.getLabel())) {
+                        myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 0);
+                        beginRapidPoll(true);
+                    } else {
+                        logger.warn("Unknown string command {}", stringValue);
+                    }
```
watou: I agree that accepting string commands is a good thing, but I am concerned that using choices from `GarageDoorStatus` may not be the most appropriate because that enum is about the current state of the garage door, not the commands you can send to the garage door.  Perfectly reasonable string commands could be things like "open", "ON", "1", "close", "OFF", "0", or perhaps a new enum called GarageDoorCommand that produces the 0 or 1 that's needed?

beowulfe: It makes sense to me to have the commands with the same values that a String item is set to when bound to the door. Doing it this way, you can use a Selection in the sitemap that is used to both display the current state and set the target state.

watou: To command a door to "Closing" or "Closed" or "Opening" doesn't make sense as they all describe the current state, not the verb to "close" or "open" the door.  There must be other ways to create a good sitemap, since the button labels can be set to whatever you want.  

But since there is no harm in having several strings all mean the "open" verb, and several strings all mean the "close" verb, perhaps just add PercentType.ZERO, "Close", "OFF" and "0" to your list of "close" synonyms, and add PercentType.HUNDRED, "ON" and "1" to your list of "open" synonyms?  ("Open" is already there because it's both an adjective and a verb in English).

watou: One idea to just make it table-driven:
```java
public class MyqBinding extends AbstractBinding<MyqBindingProvider> {
    private static final Logger logger = LoggerFactory.getLogger(MyqBinding.class);

    private static Command[] onCommands = new Command[] { OnOffType.ON, UpDownType.UP, PercentType.HUNDRED,
            new StringType(GarageDoorStatus.OPEN.getLabel()), new StringType(GarageDoorStatus.OPENING.getLabel()) };
    private static Command[] offCommands = new Command[] { OnOffType.OFF, UpDownType.DOWN, PercentType.ZERO,
            new StringType(GarageDoorStatus.CLOSED.getLabel()), new StringType(GarageDoorStatus.CLOSING.getLabel()),
            new StringType("Close") };
```
...
```java
    private void computeCommandForItem(Command command, String itemName) {

        MyqBindingConfig deviceConfig = getConfigForItemName(itemName);

        if (invalidCredentials || deviceConfig == null) {
            return;
        }

        try {
            GarageDoorData garageStatus = myqOnlineData.getGarageData();
            GarageDoorDevice garageopener = garageStatus.getDevice(deviceConfig.deviceIndex);
            if (garageopener != null) {
                String commandString = command.toString();
                boolean open = false;
                for (int i = 0; i < onCommands.length; i++) {
                    if (commandString.equalsIgnoreCase(onCommands[i].toString())) {
                        open = true;
                        break;
                    }
                }
                if (open) {
                    myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 1);
                    beginRapidPoll(true);
                } else {
                    boolean close = false;
                    for (int i = 0; i < offCommands.length; i++) {
                        if (commandString.equalsIgnoreCase(offCommands[i].toString())) {
                            close = true;
                            break;
                        }
                    }
                    if (close) {
                        myqOnlineData.executeGarageDoorCommand(garageopener.getDeviceId(), 0);
                        beginRapidPoll(true);
                    } else {
                        logger.warn("Unknown command {}", command);
                    }
                }
```
...

teichsta:
any updates @beowulfe?

watou:
On re-reading this PR, I see my earlier suggestions are optional and the PR adds value as is.  So I think it's OK to merge unless @beowulfe has anything to add?

teichsta:
Thanks @beowulfe and @watou for your efforts …

PR #4094, opened 2016-02-23
Fixes #4093. Use thread factories producing named threads

### LINKED ISSUE #4093: Use named threads in ebus binding
EBus binding uses couple thread pools which can not be identified due to default names. This is an easy fix which will make easier to identify OH state via thread dump.

splatch:


kaikreuzer:
lgtm, thanks.

PR #4096, opened 2016-02-23
Fix eBus binding issues for OH2

csowada:
- Remove dependency to logback
- Fix optional dependency on serial
- Remove unnecessary Activator class
- Remove bindingChanged helper thread, also removed configurationAdminService


kaikreuzer:
lgtm, thanks!

PR #4106, opened 2016-02-26
eBUS - Fix NullPointException if telegram value is invalid/out of range

csowada:
If the value is out of min/max range the returned value can be null. This is okay, but need to be catched.


teichsta:
Thanks @csowada, also cherry-picked onto 1.8

PR #4114, opened 2016-02-28
Was a Problem creating main Itemstable

lewie:
SEE: https://github.com/openhab/openhab/issues/4111


lewie:
Found another fault...

Issue #3282, opened 2015-10-12
Getting Error when trying to bind more than one Datapoint to ntp


udo1toni:
I think this is associated with https://github.com/openhab/openhab/pull/3251 ...
I use an Item to update the Clocks in knx System:

    DateTime Date "Zeit [%1$tA, %1$td. %1$tB %1$tY %1$tT]" <calendar> (All) {ntp="Europe/Berlin:de_DE",knx="11.001:15/7/11, 10.001:15/7/10" } 

This has worked as expected until I updated to 1.8.0-1034 (my last nightly Version was 1.8.0-1009, so it's uncertain when this Bug exactly rised)

Now I'm getting an Error:

```
2015-10-12 07:04:04.464 [ERROR] [i.internal.GenericItemProvider] - Binding configuration of type 'knx' of item  ^ ^ Date ^ ^  could not be parsed correctly.
org.openhab.model.item.binding.BindingConfigParseException: No more than 1 datapoint definitions are allowed for this item.
        at org.openhab.binding.knx.internal.config.KNXGenericBindingProvider.parseBindingConfigString(KNXGenericBindingProvider.java:429) ~[na:na]
        at org.openhab.binding.knx.internal.config.KNXGenericBindingProvider.processBindingConfiguration(KNXGenericBindingProvider.java:106) ~[na:na]
        at org.openhab.model.item.internal.GenericItemProvider.internalDispatchBindings(GenericItemProvider.java:348) [org.openhab.model.item_1.8.0.201510100138.jar:na]
        at org.openhab.model.item.internal.GenericItemProvider.internalDispatchBindings(GenericItemProvider.java:324) [org.openhab.model.item_1.8.0.201510100138.jar:na]
        at org.openhab.model.item.internal.GenericItemProvider.processBindingConfigsFromModel(GenericItemProvider.java:171) [org.openhab.model.item_1.8.0.201510100138.jar:na]
        at org.openhab.model.item.internal.GenericItemProvider.modelChanged(GenericItemProvider.java:390) [org.openhab.model.item_1.8.0.201510100138.jar:na]
        at org.openhab.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:159) [org.openhab.model.core_1.8.0.201510100138.jar:na]
        at org.openhab.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:100) [org.openhab.model.core_1.8.0.201510100138.jar:na]
[...]

PR #4127, opened 2016-03-03
Attempt to fix KNX index out of bounds exception.

9037568:
Attempts to fix the index out of bounds exception being thrown by
the KNXGenericBindingProvider when there are multiple strings in
the KNX configuration.

See #3282.


bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProvider.java
```diff
@@ -380,10 +386,18 @@ protected KNXBindingConfig parseBindingConfigString(Item item, String bindingCon
...
+                  typeClass = item.getAcceptedCommandTypes().get(i);
+                }
+                else {
+                  if( item.getAcceptedDataTypes().size() > i ) {
+                    typeClass = item.getAcceptedDataTypes().get(i);
+                  }
+                  else {
+                    typeClass = item.getAcceptedDataTypes().get(0);
+                  }
+                }
```
watou: I wonder why this code is indexing into the list of the item's accepted data and command types?  I know this existed before this PR, but doesn't it seem wrong that the KNX binding would expect to have knowledge of which States and Commands exist at certain indices in the lists?  This would be a great opportunity to figure out how to disconnect this binding from any expectations of which State and Command classes exist at specific indices in the list.

9037568:
Agreed, @watou. I'm digging into this code now to see if I can glean enough understanding to make that disconnection.

bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProvider.java
9037568: From what I can tell, the reason the binding is so incredibly tightly tied to the underlying implementation is that the configuration was designed to allow the DataPoint ID (dptId) to be optional instead of being required:

     knx="[<][<dptId>:]<mainGA>[[+[<]<listeningGA>]+[<]<listeningGA>..], [<][<dptId>:]<mainGA>[[+[<]<listeningGA>]+[<]<listeningGA>..]"

When the Datapoint ID is omitted, as seems to be the common case, the binding will likely fail to function properly if all of the datapoints are not configured in the same order as the types were defined in the Item source code (in [org.openhab.core.library.items.*] (https://github.com/openhab/openhab/tree/master/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items)).

The only way I see around this at present is to change the rules to make the Datapoint ID required in the config string. Otherwise it looks like we're stuck with the current system of guessing at which command type the user is trying to configure.

Unless you can spot something I've overlooked?

watou: I will look into it later today.  Thanks for your analysis, Chris.

watou: >  wrong that the KNX binding would expect to have knowledge of which States and Commands exist at certain indices in the lists?

After finally reading the wiki, I see that the binding authors direct users to the source code to determine which Types exist at which indices and to write their binding config strings to match.  Maybe not the friendliest approach, but it obviously works. :smile:   

Regarding your fix for index out of bounds, maybe the existing implementation is correct, except that the potential `ArrayIndexOutOfBoundsException` should be caught and a `BindingConfigParseException` should be thrown that explains the mistake in the binding config string that led to the bad list index?  I worry that your fix for the indexing might actually alter how the typeClasses are mapped.

9037568: Well, just catching and throwing is the safer fix. I'll make those changes.

Should I close this PR and open a new one?

9037568: And after more reworking, I've just realized nothing we're discussing fixes the original poster's problem.

He wants to bind more than one datapoint to a DateTime item, which only supports one command type (DateTime).

watou: Is there a way to compatibly "relax" the current binding config string syntax so that the proper `Type` is chosen in cases where an `ArrayIndexOutOfBounds` would otherwise occur?  It's tricky, because this moves into the "do what I'm thinking" territory.

9037568: >Is there a way to compatibly "relax" the current binding config string
syntax so that the proper Type is chosen in cases where an
ArrayIndexOutOfBounds would otherwise occur? It's tricky, because this
moves into the "do what I'm thinking" territory.

I may see a way. I'll poke at it some more...

teichsta:
Thanks @9037568, fix lgtm. Could you please finally squash your commits into one? Thanks …

9037568:
Lovely. it appears my attempts to squash commits into one have now completely destroyed the code that was to have fixed the problem.

If anyone knows of a way to recover them, I'd be very grateful to hear it.

watou:
Aside from the incorrect commits, this PR still shows only [2 files changed](https://github.com/openhab/openhab/pull/4127/files).  You could take those two raw changed files from [here](https://github.com/openhab/openhab/pull/4127/files), make a new branch off latest master, copy those two saved files over, `git diff` until they look like you want them to, stage them for commit, commit them, and then submit a new PR with the new branch?  And a good opportunity to name the PR more directly as adding the requested enhancement. :smile:

9037568:
I will close this PR and open a new one. Hope to have that done in the next day or two.

PR #4134, opened 2016-03-05
[lgtv] Removed required bundle javax from MANIFEST file

kungfoolfighting:
to make plugin compatible with openHAB 2.


watou:
Thanks @kungfoolfighting!

PR #4139, opened 2016-03-06
[mail] Upgrade commons email to 1.4; add new ssl parameter

### LINKED ISSUE #4130: SMTP not sending messages
Openhab is failing to send a message on a rule to a non-TLS SMTP server.  Openhab sends EHLO, then HELO, then QUIT.  At that point the remote side hangs up, since Openhab requested QUIT, and no message is sent.  Seems as though there's either a missing payload between HELO and QUIT, or sending QUIT is premature.

mail settings I'm using:

mail:hostname=mail.thekorn.net
mail:port=26
mail:username=(redacted)@thekorn.net
mail:password=(redacted #2)
mail:from=(redacted)@thekorn.net
mail:tls=false
#mail:popbeforesmtp=false
#mail:charset=

dump generated on openhab host (assuming it only has one network port) by running:

sudo tcpdump host (your mail server) -vvXS -w tcpdump_capture_file.cap

then dragging tcpdump_capture_file.cap into wireshark to examine.

I'm no SMTP expert, but wireshark says that openhab is sending three commands in packet 7:

EHLO
HELO
QUIT

tcpdump capture file available here, since github insists this isn't a zip file (it is): http://www.thekorn.net/temp/tcpdump_capture_file.zip

watou:
This is a partial fix for #4130.  This PR only fixes the case where the target SMTP server does not handle STARTTLS negotiation (when using `mail:tls=true`), and therefore the mail action has to negotiate TLS on socket connection by instead setting the new `mail:ssl=true`.

The other half of #4130 is caused by a [bug in the org.apache.geronimo implementation of javax.mail](https://issues.apache.org/jira/browse/GERONIMO-3427), in javax.mail_1.4.0.v201005080615.jar.  This bug does not exist in [this different javax.mail implementation](http://mvnrepository.com/artifact/javax.mail/mail/1.4.7).  Therefore, it makes sense to replace the very old bundle javax.mail_1.4.0.v201005080615.jar with mail-1.4.7.jar.

In OH2, I can `bundle:install file:///path/to/mail-1.4.7.jar`, and this org.openhab.action.mail service works and fixes #4130.  This tells me that a new openhab-deps-repo release with [this changed](https://github.com/openhab/openhab-deps-repo/blob/master/features/org.openhab.deps.runtime/feature.xml#L522) would fix this bug in OH2.  However, I don't know if or how this can be fixed in openHAB 1.8, short of building the mail action bundle with mail-1.4.7.jar embedded (which would be a shame to add more bloat).

@teichsta and @kaikreuzer, what do you think?


teichsta:
> ever, I don't know if or how this can be fixed in openHAB 1.8, short of building the mail action bundle with mail-1.4.7.jar embedded (which would be a shame to add more bloat).

mail-x.jar dependency is resolved via orbit drop. Since even the latest [orbit build](http://download.eclipse.org/tools/orbit/downloads/drops/R20160221192158/) does only contain the mail-1.4.jar i don't see any other chance to add it as binary. Which is absolutely ugly … true!

watou:
> the latest orbit build does only contain the mail-1.4.jar

That may indeed be perfectly workable.  mail-1.4.7.jar is my preferred javax.mail implementation, but mail-1.4.jar might work just as well to fix #4130, since it's a totally different implementation.  Is there any chance that could be switched to to replace the ancient javax.mail_1.4.0.v201005080615.jar?

watou:
On closer inspection, mail-1.4.jar is not suitable, as it's not a proper OSGi bundle and it's 10 years old.  In any case, this PR is independent of which javax.mail implementation is provided in a separate bundle (as I removed the `version="1.4.0"` from the manifest).  This could at least be fixed in openHAB 2 without embedding the mail JAR in the mail action, correct?  Do you have any recommendation?

teichsta:
Thanks @watou! I assume we are not going to cherry-pick this onto 1.8, right?

watou:
Since this PR does not fully fix the user-reported problem, cherry-picking it onto 1.8 doesn't yet serve a purpose. If we can also replace the javax.mail implementation, then together they would fix a 1.8 bug.

PR #4161, opened 2016-03-12
fritzboxtr064 binding: fix compilation error

stefanroellin:
Use org.w3c.dom.ls.LSSerializer instead of org.apache.xml.serialize.XMLSerializer, which is deprecated.

Since this binding did not compile for me on the command line, I replaced some deprecated symbols.

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java
@@ -8,9 +8,8 @@
  */
 package org.openhab.binding.fritzboxtr064.internal;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Helper.java
```diff
@@ -34,20 +35,26 @@
...
+            lsSerializer.getDomConfig().setParameter("format-pretty-print", true);
+
+            Writer stringWriter = new StringWriter();
+            LSOutput lsOutput = domImplLS.createLSOutput();
+            lsOutput.setEncoding("UTF-8");
+            lsOutput.setCharacterStream(stringWriter);
+            lsSerializer.write(doc, lsOutput);
+            strMsg = stringWriter.toString();
+        } catch (Exception e) {
             e.printStackTrace();
```
watou: Could you replace this and any occurrences of printStackTrace with logger.warn in the code (since you are making other fixes)? Bindings should not print to stdout or stderr but instead use log4j.

Issue #3776, opened 2016-01-12
setpermissions.sh should not change the permissions for all files and folders in ~openhab


rkoshak:
The forum discussion is here: https://community.openhab.org/t/a-restart-of-openhab-causes-file-permissions-to-change-in-openhab-ssh/6121/7

Currently setpermissions.sh changes the permissions of all files and subfolders of ~openhab during an openhab restart. There are two problems with this.

1) It appears the intent is to set the permissions of the folders in /var/lib/openhab which include the Designer workspace, persistence, and zwave folders. However, if the user creates the openhab user themselves and maps the home directory to somewhere else, these folders get missed by setpermissions.sh

2) If the user wants to run ssh as the openhab (Exec binding) using ssh keys as opposed to a password (a very good security practice) they need to put those keys in ~openhab/.ssh. And the permissions on those ssh key files must be user rw only or else some (all?) ssh servers will reject the connection because the permissions on the private key are too open. Because the setpermissions.sh script blindly changes the permissions of all files and folders to 755 a restart of openhab will break ssh for the openhab user.

theoweiss:
setpermissions is needed because the openhab execution user can be configured with /etc/default/openhab and it must be granted that this user has rw access to certain files.

I would suggest the following changes:
1) the OPENHAB_HOME_DIR value should be configurable in /etc/default/openhab
2) add a new variable to /etc/default/openhab SET_PERMISSIONS=yes|no with the default value "yes". If SET_PERMISSIONS is set to "no" no permission changes will happen.

rkoshak:
If I'm understanding the changes correctly the changes are an all or nothing sort of change, correct? 1) lets you deal with issue 1) above where ~openhab isn't at /var/lib/openhab. 2) addresses the problem of it changing the .ssh directory permissions by essentially turning off the script entirely.

So my questions is if running setpermissions.sh every time OH starts is important, won't turning it off like this cause problems? If it isn't important, why is it run every time?

How hard would be it be to update the installer to put ~openhab somewhere else and change the script to operate on /var/lib/openhab instead of ${OPENHAB_HOME_DIR}.

Or make the script a little smarter to only change the permissions on those files or folders that are needed instead of blanket recursively update whole folders (at least for the ~openhab folder).

Or add a new variable to /etc/default/openhab to add exclusion patterns so certain folders like .ssh can be excluded.

watou:
[...]

Issue #3601, opened 2015-12-27
setpermission.sh missing some directories


slaboure:
For an apt-get installation to work, setpermission.sh must be modified to also "chown" /var/log/openhab (for logs) and /usr/share/openhab (for Dropbox timestamp files for example). This is not being done on 1.7.1

teichsta:
could you have a look please @theoweiss?

theoweiss:
Ok, I will check that.

teichsta:
any chance to have a quick at this one today @theoweiss ?

theoweiss:
Sorry no chance: a serious cold is visiting my family. I have to get these things managed.

theoweiss:
@slaboure can you describe the issue a bit more in detail. 
Normally set permissions is only needed when the openhab user in /etc/default/openhab was reconfigured or  the openhab service has been started in between manually as root.  It's meant for handling these special cases.

teichsta:
> Sorry no chance: a serious cold is visiting my family. I have to get these things managed.
no worries … get well soon!

teichsta:
any update on this @slaboure @theoweiss ?

theoweiss:
I will provide a fix till the weekend.

theoweiss:
setpermissions.sh already changes permissions of /var/log/openhab .  /usr/share/openhab has to be readonly for the openhab user, for security reasons. Files should only be changed in the $userdata directory and the log dir. The dropbox is already fixed with #3451. Open issues of setpermissions.sh should be fixed with #4166 .

teichsta:
ftr: #4166 has been merged meanwhile …

slaboure:
perfect, thanks!

PR #4166, opened 2016-03-12
don't change file mode of files in OPENHAB_HOME_DIR

theoweiss:
The mode of files and directories in $OPENHAB_HOME_DIR ( /var/lib/openhab) are not changed anymore, only the ownership is changed.
Fix for #3776 and #3601 .


teichsta:
great, thanks @theoweiss!

PR #4168, opened 2016-03-13
Maven warnings on build start

slawekjaranowski:
Because of missing version of plugin org.vafer:jdeb maven generates many warnings on each build start.


teichsta:
Thanks @slawekjaranowski! Also cherry-picked onto 1.8 …

PR #4172, opened 2016-03-13
Fritzbox TR064 - wrong parent version

slawekjaranowski:


teichsta:
oh yes, thanks @slawekjaranowski!

PR #4174, opened 2016-03-14
Now JDBC Persistence can start whithout config

lewie:
Forget this! based on too old code!


lewie:
Forget this! based on too old code!

PR #4179, opened 2016-03-16
Allow multiple KNX datapoints without type guessing.

9037568:
Don't do type guessing until it's needed. This allows binding multiple
datapoints for the DateTimeItem, resolving #3282.

### DIFF
--- a/bundles/binding/org.openhab.binding.knx.test/src/test/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProviderTest.java
+++ b/bundles/binding/org.openhab.binding.knx.test/src/test/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProviderTest.java
@@ -20,6 +20,7 @@ import org.openhab.binding.knx.internal.config.KNXGenericBindingProvider.KNXBind
 import org.openhab.binding.knx.internal.config.KNXGenericBindingProvider.KNXBindingConfigItem;
 import org.openhab.core.items.GenericItem;
 import org.openhab.core.items.Item;
+import org.openhab.core.library.items.DateTimeItem;
 import org.openhab.core.library.types.PercentType;
 import org.openhab.core.library.types.StopMoveType;
 import org.openhab.core.library.types.StringType;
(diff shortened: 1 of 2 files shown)


watou:
Thanks for this, Chris.  Has @udo1toni tested this change, is it compatible with all KNX binding config strings, and will it still throw an IndexArrayOutOfBounds exception (and not a BindingConfigParseException) when the binding config string is out of spec?

udo1toni:
I would like to test it, but I don't have a build system, so I need a link to the .jar :)

9037568:
Don't think a new jar exists yet.  The last cloudbees build failed...

https://openhab.ci.cloudbees.com/job/PR-openHAB1-Addons/378/console

watou:
@9037568 would you be able to build one based off your PR and put in on Dropbox, Google Drive, etc.?

9037568:
I've placed a locally built jar here:

https://drive.google.com/file/d/0B8vYheWfObP9N01RUFlBZTlBZU0/view?usp=sharing

9037568:
Answers to @watou's earlier questions:

>is it compatible with all KNX binding config strings

I don't know of any way to ascertain the answer to that one, as I'm unaware of any catalog containing all possible KNX binding config strings. I can say, however, that this did not break any of the tests in the [knx.test](https://github.com/openhab/openhab/tree/master/bundles/binding/org.openhab.binding.knx.test) package.

>and will it still throw an IndexArrayOutOfBounds exception (and not a BindingConfigParseException) when the binding config string is out of spec?

The behaviour should be unchanged for all items where the datapoint ID is **NOT** specified. The change is that we skip the magic type guessing routine for items where the datapoint ID **IS** specified.

udo1toni:
I can confirm, that the error described in #3282 seems to be resolved with this update. I will use this version the next days to see if the binding works as expected :)

udo1toni:
So far no errors occured :)

9037568:
Awesome!

udo1toni:
Still no errors occured, so I think the patch is ok?

9037568:
Thank you, John! I can finally reclaim my master branch. ;-)

watou:
Sorry for the delay in merging the PR, and thanks for your care in creating it!

PR #4183, opened 2016-03-17
[systeminfo] Use Bundle-NativeCode to load embedded native libraries

watou:
The current systeminfo binding requires the user to download and install the appropriate native sigar library in their openhab/lib directory, and the sigar Java library uses its own arch/osname/osversion logic to load the correct library with a specific arch/osname/osversion file name from the `java.library.path`.  This approach is error prone and relies on the user downloaing native library files from sourceforge.net or the old openHAB Google Group.

This PR embeds all supported native libraries in the binding jar, and uses OSGi's `Bundle-NativeCode` instruction to extract the correct version.  The binding uses a [slightly modified version of sigar.jar](https://github.com/watou/openhab-sigar/commit/4399fb57fadb01fb3e1157c4bec37b5aa282b057) that just performs a `System.loadLibrary(lib)` to load the native library.

The only new configuration choice is an optional `systeminfo:variant` property in openhab.cfg for ARM systems running Linux, since `Bundle-NativeCode` does not account for those differences.  The 3 currently supported values for `systeminfo:variant` are `cubian`, `odroid-u3` and `raspbian`.

The old approach of placing the correct native library in openhab/lib still works, but this PR should make that obsolete and unnecessary.  The native library in openhab/lib would have to be called `[lib]sigar[-variant][.so|.sl|.dll|.dylib]` in that case.

The only platforms tested so far are x86_64/macosx, arm/linux-raspbian and x86_64/Windows10.  Testing would be nice to have for all supported platforms (even though it seems to be working on at least 3).  To test the JAR:

* shut down the openHAB runtime
* move any sigar native libraries out of your openhab/lib directory (and any other directories where your OS would look for native libraries)
* replace the systeminfo binding JAR with [this one](https://dl.dropboxusercontent.com/u/4286376/systeminfo-bundle-nativecode/org.openhab.binding.systeminfo-1.9.0-SNAPSHOT.jar) (MD5 = 60b6009742d6f44cb4ca9a6f540b100a) **UPDATED 22 March 2016 for Windows 8 and 10**
* If you are running openHAB on an ARM processor running Linux, specify one of the following in your openhab.cfg: `systeminfo:variant=cubian`, `systeminfo:variant=odroid-u3`, or `systeminfo:variant=raspbian`.
* restart the openHAB runtime.

[Forum discussion here](https://community.openhab.org/t/unknown-error-on-startup/8684) and [here](https://community.openhab.org/t/systeminfo-binding-requesting-help-to-test-simpler-packaging/8794).


maggu2810:
Sounds good

watou:
Could another maintainer review this PR, so it can be merged or closed?  My opinion is that this PR improves installation of the binding on both openHAB 1.x and 2.0, and his been tested and works on 3 different platforms.

It might also make sense to move my fork of sigar from watou/openhab-sigar to openhab/openhab-sigar, since a small modification was required for this PR.  A later improvement to that repo could be to rework the build process so it's more in line with the openHAB repos.

kaikreuzer:
> Could another maintainer review this PR

Hm, there do not seem to be any maintainers left besides yourself :-/
So I am willing to jump in, but allow me a few days time.

Regarding "improved installation on OH2": Did you already have a look at https://github.com/openhab/openhab2-addons/pull/785? I plan to review this as well in the next days - if it gets merged, do you see any features that the 1.x might bring on top so that it is still relevant for OH2 as well?

kaikreuzer:
@watou The code looks all good to me - I didn't test it on any further platform, but from the feedback on the forum it seems to do its job.

> It might also make sense to move my fork of sigar from watou/openhab-sigar to openhab/openhab-sigar

If we do not really expect any required maintenance I think we don't need this step necessarily.

FYI: I also reviewed the OH2 systeminfo binding today, which looks pretty good. I didn't compare all the features though to be able to say whether it is (or could become) a full replacement of the sigar-based binding.

watou:
@kaikreuzer Thank you.  Regarding the OH2 systeminfo binding: it may not have all the features of the OH1 binding yet (I think process monitoring might be incomplete), but the fact that it doesn't require native libraries is a huge improvement for wider arch/OS support, and especially for maintainability.

kgoderis:
@watou John, is this now in 2.0.0.201607030102 of the systeminfo binding? I am under the impression this is not the case yet. I just installed the 1.9 SNAPSHOT one you provided earlier on. It is unclear to me what should be the preferred version to use on OH2

watou:
@kgoderis, I had just made this change to the 1.x binding, and shortly thereafter a completely new 2.x binding was merged.  I think the 2.x binding would generally be the preferred one to use under openHAB 2, but I think there was some different native code issue there after all (sorry I don't have the link).  If there are any issues with either one when running under OH2, please open new issues or PRs.

PR #4194, opened 2016-03-19
Z-wave: New device Everspring AN181 Mini-plug with meter

FredrikFornstad:
Please add support for the Everspring AN181 Mini-plug with meter function (Gen5). 

My pull request is slightly different from the database at www.cd-jackson.com:
- My xml-file is named an181.xml since "AN181" is the device name, not "AN1812" (AN181-2 means it is the german type plug, AN181-6 the French and so on)
- My file has maximum 4 allowed associations for group 2 (tested and also according to the manual)

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/everspring/an181.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/everspring/an181.xml
@@ -1,0 +1,119 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product>
+	<Model>AN181</Model>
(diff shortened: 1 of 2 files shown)


cdjackson:
> My pull request is slightly different from the database at www.cd-jackson.com:

Do you mind updating this database to avoid them getting out of sync? (I assume of course that your definition is correct).

FredrikFornstad:
Database updated as per request, awaiting review.

cdjackson:
Thanks.

PR #4197, opened 2016-03-20
bugfix in the way house codes and unit codes are packed into a single…

renescherer:
… byte. The 4 bit house code has to be packed into the 4 most significant bits of the address byte. The previous version of the code only works for devices in house code A.


teichsta:
Thanks @renescherer! Also cherry-picked onto 1.8 …

Issue #3537, opened 2015-12-12
Better synchronization for openHAB and industerial controllers via Modbus


masoud-omidvar:
I need to integrate openHAB and some other controllers in a project. They provide some Modbus registers. To prevent inconsistent data set, they reject any Modbus request while running their control logic in each cycle.

I think if we can adjust a configurable small delay (e.g. 20 ms) after only first execution-try, then it is more probable to have a successful set of I/O requests in each pooling interval, otherwise we miss some part of our data set in some pooling actions.

![image](https://cloud.githubusercontent.com/assets/8285663/11760948/25853ed8-a0c5-11e5-82bb-35e0600462c3.png)
##

masoud-omidvar:
Two more issues:
1. The order of putting read register in ByteBuffer is incorrect (in ModbusBinding.java) 
2. Modbus users are accustomed to start Addressing from '1'  instead of '0'

diverseg:
1. If this Is this regarding a floating point data type, Likely a Big Endian Vs Little Endian issue. 
The endianness of the master and slave must be considered. Byte swap, word swap options for the 4 possible byte orders would solve any issue. 
ABCD
BADC
CDAB
DCBA

2.  Not all Modbus users. 
  This issue has been around a long time.
he Modbus.org standard documents refer to register addresses using "one-based" numbering. However, the addresses that are actually sent in a Modbus command message are "zero-based". For example, to read register 1001, the address that is actually sent in the command message is 1000.
Modbus specification says "Registers are addressed starting at zero. Therefore input registers 1-16 are addressed as 0-15."
Some software deals with it by enabling/disabling a shift data addresses by 1.

masoud-omidvar:
As you can see in the following screenshot from debugger, ByteBuffer class expect to put higher value word (register) first !
![image](https://cloud.githubusercontent.com/assets/8285663/11835478/1be1688a-a3e9-11e5-9daf-b9cc9fc25c85.png)

So, I change the code as:
![image](https://cloud.githubusercontent.com/assets/8285663/11836080/6c662ba2-a3ed-11e5-8288-ae3031967ce6.png)

ssalonen:
So if I understand correctly, would this issue be resolved if a configurable retry-timeout is introduced for each slave?

masoud-omidvar:
yes, dear Sami. It will be will be helpful.

ssalonen:
@masoud-omidvar are you willing to make a PR with a fix? :smile:

[...]

PR #4226, opened 2016-03-26
Modbus: shared connections and fixes for serial slaves.

ssalonen:
Modbus binding: shared connections and serial fixes

Main issues with the modbus binding
- many slaves referring to same endpoint were problematic since there was no protection against executing a command while polling. With serial slaves this naturally messes up the communication. With TCP slaves we have got extensive feedback that many slaves are able to process single connection at a time. Issue [#3818](https://github.com/openhab/openhab/issues/3818) improved the sitatuation a little bit
- serial slaves had [static connection parameters](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSerialSlave.java#L32) making it impossible to configure the binding to read two different serial ports

With a lot of [testing help from the community](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246) and quite much coding, I'm finally happy to introduce pull request to resolve these issues. Based on the community feedback, the modbus serial binding seems to work without issues or errors now. Previously it was usually noted that modbus is unusable (see links to discussions below). Furthermore, even the modbus TCP had its issues, especially when the slave has some connection limitations.

Changes in this huge pull request has these critical fixes
- configuration parameters between serial slaves are now isolated
- **connections are pooled** using Apache Pool. We allow only one transaction per endpoint to be active at a time. This means that binding waits for polling to finish before starting to process command.
-- endpoint refers to protocol+hostname+ip in case ip slaves and port name in case of serial slaves
- **serial slaves: blocking read** by using `CommPort.enableReceiveThreshold`. Previously serial connections were really error prone because this was not set
- **configurable delay between transactions**. Turned out this was critical as well for maximum robustness. The binding bombarded the serial slaves too fast, and I believe the it either mixes up the RTU modbus protocol (where communication silence is used to separate modbus frames), or it is just too much for the slave

Other fixes:
- read timeout now applies for all serial protocols, not just RTU
- wait between failed modbus transactions, otherwise retries might be bound to fail (see [#3818](https://github.com/openhab/openhab/issues/3818), and sensitivity of the slave to delays as [tested by mbs38](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/93?u=ssalonen))
- logging error situations improved
- obsolete documentation updated
- connection retry for ip slaves (related [#3537](https://github.com/openhab/openhab/issues/3537))
- configurable time to wait before trying connect closed connection for ip slaves (related [#3537](https://github.com/openhab/openhab/issues/3537))
- configurable connection max tries for ip slaves
- configurable time-to-live setting for ip slave connections. Instead of closing connection after every transaction (see [#3818](https://github.com/openhab/openhab/issues/3818)), we let user to optionally configure how long to let the connections live. For example, the user can specify that connection can remain open for 10 minutes, but not more. The default (TTL=0ms) remains the same. This parameter was introduced for high-performance situations where we want to avoid spending time with connections -- but also to make the system more robust with some modbus slaves (see [#3537](https://github.com/openhab/openhab/issues/3537)) which might refuse connection now and and then.
- test bundle introduced with large set of end-to-end tests to verify the operation and act as regression tests(actually this identified couple of bugs I have already reported and resolved separately). Note that the tests start up a java modbus tcp server with randomly assigned free port.
- [setProperlyConfigured](https://github.com/openhab/openhab/commit/60d49bdff1e6054d2f733b16115b6563545e41ec#diff-dc828f03d20f3c7032d4ab0c86ea782cR353) reseted on every configuration change
- [configuration parsing streamlined](https://github.com/openhab/openhab/commit/60d49bdff1e6054d2f733b16115b6563545e41ec#diff-dc828f03d20f3c7032d4ab0c86ea782cR415)
- configurable receiveTimeout for serial slaves (time before we give up waiting for a byte when reading response)
- configurable flow control for serial slaves

Relevant discussions:
- https://community.openhab.org/t/current-modbus-changes-from-ssalonen/5356/
- https://community.openhab.org/t/modbus-connection-problem/6108/
- https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/
- https://community.openhab.org/t/modbus-serial-failed-to-read/3430/
- https://community.openhab.org/t/modbus-binding-multiple-serial-masters/3383

[This thread](https://community.openhab.org/t/modbus-connection-problem/6108/) is especially important since it contains test reports from [Ferryb4](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/96?u=ssalonen) (serial), [mbs38](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/93?u=ssalonen) (serial) and [Ymedox](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/63?u=ssalonen) (tcp). Ferryb4 and mbs38 used [relatively new version, tag modbus-pooling-test-version-6, rev fb3f7da](https://github.com/ssalonen/openhab/tree/modbus-pooling-test-version-6) show it should be comparable with this PR. 

Serial performance is improved greatly: [mbs38 was able to confirm](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/93?u=ssalonen) performance of 1732 polls per minute

Should likely fix these issues:
- https://github.com/openhab/openhab/issues/626 (stale issue?)
- https://github.com/openhab/openhab/issues/3537 (positive feedback @masoud-omidvar)
- https://github.com/openhab/openhab/issues/2043 (stale issue?)
- https://github.com/openhab/openhab/issues/1865 (stale issue?)

Possibly explains
- https://community.openhab.org/t/modbus-binding-unstable/3928
- https://github.com/openhab/openhab/issues/2843


ssalonen:
As the pull request is huge, I'm open to suggestions how it would be best to review this.

ssalonen:
Compled version available [here](https://github.com/ssalonen/openhab/releases/tag/pr-v1)

ssalonen:
Configuration is backward compatible but extended for "power-users" looking for extreme performance. With some slaves the default delays might need to be increased but I've tried to keep the defaults conservative.

Please advice how proceed with editing `openhab.cfg` configuration templates

mbs38:
> Compiled version available here
I can confirm that this version works perfectly with serial devices.

ssalonen:
Configuration is extended as follows:

**Modbus tcp connection string**

As before, it still consists of colon separated parameters. Only host is mandatory.

Expecting at most 6 parameters: hostname (mandatory) and optionally (in this order): port number, interTransactionDelayMillis, reconnectAfterMillis, interConnectDelayMillis, connectMaxTries.



**Modbus serial connection string**
As before, it still consists of colon separated parameters. Only port name is mandatory.

Optional parameters are as follows 
- baud 
- data bits 
- parity 
- stop bits
- encoding 
- inter transaction delay (New)
- receive timeout (new) 
- flow control in (new)
- flow control out (new)

The inter transaction delay has a default value of 35ms with serial devices. 60ms is the default for tcp connections. For RTU serial this plays a big part as silence between two modbus frames is part of the protocol. For tcp there is no such restriction but experiments show that slaves might refuse the connection if the delay between transactions is too short. 60ms was the minimum that works with my siemens plc.

Receive time out default has been increased from 500ms to 1500ms. Previously this was not configurable. For long requests (many registers) and slow baud rates the original default might be too small.

Flow control is now made configurable also, it was the last serial parameter that was not exposed to user configuration.


-----

Inter transaction delay means time to wait between modbus transactions that would interact with the same modbus endpoint (same host+port or same serial port)

Reconnect after millis  means time to keep connections alive before disconnecting. Default with tcp connections is to disconnect after every transaction (same as before).

Inter connect delay is the time to wait between connection attempts to the tcp slave. Default 0ms.

Connect max tries tells how many times to try to connect to the slave if there are issues. Default 1 emulates old behavior.

-----

The delay and reconnect parameters should be equal for all slaves connecting to the same endpoint. If that is not the case, warning is logged when the configuration is parsed.

ssalonen:
Update regarding testing. sCythe in the community forums is testing the new binding with windows.

It seems that the serial timeout handling is different with windows (read() returns immediately EOF=-1) than linux. Naturally this causes issues in this binding. See this [post](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/136?u=ssalonen) for details.

I do find these issues in the nrjavaserial which I believe is the library used by the openhab (?)

https://github.com/NeuronRobotics/nrjavaserial/issues/27
https://github.com/NeuronRobotics/nrjavaserial/issues/53


Is this known issue or am I using the serial somehow "wrong"?

DCandE:
This version works for my Modbus/TCP devices.

ssalonen:
It's now 1 month since the PR submission. Anything I can do to progress this?

masoud-omidvar:
Thx Sami, for this version. It works for my Modbus/TCP device too.

ssalonen:
Mbs38 tested on windows with positive results: https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/148?u=ssalonen

ssalonen:
Rebased against master -- now should PR #4339

Wolfgang1966:
Tested the latest jar on my Banana Pi, 4 slaves, >200 items in total, with good results. I strongly recommend to merge this PR for 1.9.0!

watou:
@ssalonen @mbs38 @DCandE @masoud-omidvar @Wolfgang1966 @teichsta 

Does anyone see a reason **not** to merge this now?

ssalonen:
I'm happy to do any changes required, or any additional tests required. So far the feedback has been positive. Only exception is the windows+serial modbus users (some [negative](https://github.com/openhab/openhab/pull/4226#issuecomment-204046492) and some [positive](https://github.com/openhab/openhab/pull/4226#issuecomment-217621180) experience). 

Per my understanding it seems that our java library _nrjavaserial_ (or the underlying OS details -- goes beyond my understanding) behaves differently on nix and windows. As discussed above, I believe right way to tackle this is to engage upstream to the discussion and try see if that's the place to fix it. In my opinion this shouldn't be a blocker for this PR. Compared to the previous status of the serial modbus support (as summarized in the initial post), I hope this will improve the situation significantly.

Nevertheless the change list is huge and I understand the risks involved here. It's always something slips in eventhough we have [pretty extensive testing conducted by our good community](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/147), with both tcp and serial devices.

One technical issue I just received from @Wolfgang1966  this week: apparently configuration changes to `openhab.cfg` will disable the binding instead of refreshing it. I wonder if it is because of [setProperlyConfigured(false);](https://github.com/ssalonen/openhab/blob/77afc4564d5be3236cd95fe447cd156bb8aaea2f/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java#L384) call in the beginning of `ModbusBinding.updated(.)`? I tried to look reference what is the correct way to mark the configuration as "stale" but could not find anything. Perhaps you can comment on this @watou?

One more thing: I will have limited internet connection during the next three weeks (traveling starting tomorrow), but happy to work on this after that if required.

watou:
Thank you for the detailed response, @ssalonen.  Very helpful!  If remaining issues are restricted to Windows + nrjavaserial and this PR did not make the situation worse, then I would agree it's worth merging.

> apparently configuration changes to openhab.cfg will disable the binding instead of refreshing it. I wonder if it is because of setProperlyConfigured(false); call in the beginning of ModbusBinding.updated(.)? I tried to look reference what is the correct way to mark the configuration as "stale" but could not find anything.

I think you are correct that bindings should not call `setProperlyConfigured(false)` unless it is true that the configuration is incorrect.  That method does more than just set a flag that can be set again to `true`. There are more elegant fixes than the following suggestion, but if you removed the `setProperlyConfigured` calls from `updated`, but then put the body of the `updated` method in this construct:
```java
try {
 // put whole body of method here
 setProperlyConfigured(true);
} catch (ConfigurationException ce) {
 setProperlyConfigured(false);
 throw ce;
}
```
Perhaps @Wolfgang1966 could test if that resolves what he reported this week?

Safe travels and thanks for your contributions.

Wolfgang1966:
As the current version of the binding throws several exceptions and stops working correctly after a config change, I already was used to restart the server afterwards anyways :-)

@ssalonen Please let me know when you have an updated version, I will test it as soon as possible. 

But even with this problem left I would strongly suggest to merge because the binding works fine and e.g. fixes some transient exceptions that I saw in the logs before, so for me it is sure an improvement. Also I have the impression that the CPU load on my system went down a little since I installed the new version.

@ssalonen Thanks for the large amount of work you put into this.

watou:
Pulled this PR locally and did a full build with no errors.  I will open a PR that tries to improve the `updated` method as mentioned above; @Wolfgang1966 and anyone else willing to test, I will provide a link to a JAR with that change in the hope that the binding adapts to the updated configuration and continues running.

watou:
Please see https://github.com/openhab/openhab/pull/4398 for a test JAR that will hopefully keep running after the config is updated.  Please test [it](https://github.com/openhab/openhab/pull/4398)!

bundles/binding/org.openhab.binding.modbus.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,13 @@
...
+Bundle-SymbolicName: org.openhab.binding.modbus.test
+Bundle-Version: 1.8.0.qualifier
+Fragment-Host: org.openhab.binding.modbus
+Bundle-RequiredExecutionEnvironment: JavaSE-1.6
+Require-Bundle: org.junit
+Bundle-Vendor: openHAB.org
+Bundle-ClassPath: lib/mockito-all-1.10.19.jar,
+    lib/slf4j-simple-1.7.5.jar,
+    ../org.openhab.binding.modbus/lib/commons-pool2-2.4.2.jar
+    .
```
martinvw: @ssalonen in my IDE setup this gives some nasty errors, it also looks strange?

It has been some time but can you recall what should have been here?

ssalonen: Are you referring to the current directory (.)?

I'm sorry, cannot recall why that would need to be there

Issue #1922, opened 2015-01-02
RFXCOM binding not supporting Oregon WGR800 WindGust


paalkr:
The RFXCOM binding does not expose the WindGust ValueSelector for the Oregon WGR800 sensor. Can support for WindGust be added?

Reported by RFXmngr
1056022DD40000E10024001D0000000059
Packettype    = WIND
subtype       = WIND2 - WGR800
Sequence nbr  = 45
ID            = 54272
Direction     = 225 degrees  SW
Average speed = 3,6 mtr/sec = 12,96 km/hr = 8,05 mph
Wind gust     = 2,9 mtr/sec = 10,44 km/hr = 6,49 mph
Signal level  = 5
Battery       = OK

Reported by OpenHAB debug log
Raw data = 105602C4D40000B40019000D0000000059
 - Packet type = WIND
 - Seq number = 196
 - Sub type = WGR800
 - Id = 54272
 - Wind direction = 180.0
 - Wind speed = 1.3
 - Signal level = 5
 - Battery level = 9

Alpoy:
Indeed would be nice.. anyone up for the task? ;-)

Speederc:
Already included in the binding, "wind speed" and "wind gust" are two different names (for RFXmngr and Openhab) but they describe the same field in the RFXCOM data.
Can be closed.

paalkr:
Hi

Sorry for the late comment ;) Unfortunately you are wrong. The sensor reports Wind Speed as the average of the last minutes, while Wind Gust are the current readout from the sensor. So they are two distinct values, and unfortunately In OpenHAB Wind Speed is implemented as Average speed. Please look at my sensor dump data from RFXmngr in the OP.

Average speed = 3,6 mtr/sec = 12,96 km/hr = 8,05 mph
Wind gust = 2,9 mtr/sec = 10,44 km/hr = 6,49 mph

The binding has implemented
    WIND_DIRECTION("WindDirection", NumberItem.class),
    WIND_SPEED("WindSpeed", NumberItem.class),
    WIND_AVSPEED("WindAvSpeed", NumberItem.class),
    ....
    GUST("Gust", NumberItem.class),

I think that OpenHAB WindAvSpeed should point to RFXcom Wind Speed, and that OpenHAB WindSpeed should point to RFXcom Wind Gust.

Regards,
PK

Speederc:
To clarify what is currentl implemented in Openhab binding for this type of message :
- RAW_DATA : raw received message
[...]

PR #4237, opened 2016-03-28
RFXCOM Add Average speed for all wind sensors

Speederc:
Add the average speed value for all wind sensors : cf #1922


teichsta:
Thanks @Speederc! Also cherry-picked onto 1.8 …

PR #4256, opened 2016-04-04
[weather] Quiesce repeatedly logged stacktraces

watou:
The weather binding can spam openhab.log with stack traces at ERROR level, when it would be sufficient to instead log the exception message at WARN level.


watou:
@gerrieg, any problem with this PR?  Thanks.

PR #4258, opened 2016-04-04
Onkyo Binding multiple devices status update fix

### LINKED ISSUE #4240: Onkyo binding status update processing problem
Please refer to [this thread](https://community.openhab.org/t/onkyo-binding-netusb-list-handling/9099) for background.

The problem is with the way status update events from the receiver are handled. For advanced commands (those starting with '#') only the first three characters of the command string are compared. See line 355 in [OnkyoBinding.java](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoBinding.java). This makes it impossible to display status updates to the onkyo menu list as they begin with strings like:

"NLSU0-"
"NLSU1-"
"NLSU2-"
"NLSU3-"
...
"NLSU9-"

The current code maps all of these to the same item regardless of how the items are defined. For example, all of these items would get every "NLS" update.

Number onkyoNETListSel	"List Select"	(Onkyo1)	{onkyo="*:onkyo1:#NLSL%01X"}
String onkyoNETList	"List [%s]"	(Onkyo1)	{onkyo="*:onkyo1:#NLS"}
String onkyoNETList0	"1 [%s]"	(Onkyo1)	{onkyo="*:onkyo1:#NLSU0"}
String onkyoNETList1	"2 [%s]"	(Onkyo1)	{onkyo="*:onkyo1:#NLSU1"}
String onkyoNETList2	"3 [%s]"	(Onkyo1)	{onkyo="*:onkyo1:#NLSU2"}

Notice that the first item (straight from the wiki example on the onkyo binding page) isn't even intended to receive status updates. It's intended for selecting one of the list items and causes type conversion errors when string status updates are applied to it. I think two things would clean this up and really improve the effectiveness of the binding:

1) Do not automatically apply status updates based on command string. Commands are for sending, not receiving.
2) Add a receive/input specification like the HTTP binding supports and compare against the whole specification, even for advanced commands.

With these two changes the list items could be re-written as follows:

String onkyoNETList0	"1 [%s]"	(Onkyo1)	{onkyo=">[*:onkyo1:#NLSL0] <[onkyo1:#NLSU0:REGEX(.(.*))"}

String onkyoNETList1	"2 [%s]"	(Onkyo1)	{onkyo=">[*:onkyo1:#NLSL1]<[onkyo1:#NLSU0:REGEX(.(.*))"}

String onkyoNETList2	"3 [%s]"	(Onkyo1)	{onkyo=">[*:onkyo1:#NLSL2]<[onkyo1:#NLSU0:REGEX(.(.*))"}

This would allow each string item to capture one line of the list and sending any command to a string item would select it's line from the list. I'm hoping that using a Text element in the site map for these items will send them some command when touched. If not, some type of Text element that sends commands would be required for the intended functionality.

Regards,

~petev

### LINKED ISSUE #4255: Onkyo binding multiple receiver status update problem
When there are more than one Onkyo receiver configured, status updates can get routed to the wrong item. In OnkyoBinding.java:statusUpdateReceived(), there doesn't appear to be a check against the host name, only the device command string is compared to that in the item configuration. After the first match, line 393 breaks out of the loop. So only the first item with a matching device command is updated, which may not be the correct device.

`
public void statusUpdateReceived(EventObject event, String ip, String data) {
        // find correct device from device cache
        DeviceConfig deviceConfig = findDevice(ip);

        if (deviceConfig != null) {
            logger.debug("Received status update '{}' from device {}", data, deviceConfig.host);

            // KLUDGE: Clear menu list when first line is received.
            if ("NLSU0".equals(data.substring(0, 4))) {
                clearMenuList(deviceConfig);
            }

            for (OnkyoBindingProvider provider : providers) {
                for (String itemName : provider.getItemNames()) {
                    // Update all items which refer to command
                    HashMap<String, String> values = provider.getDeviceCommands(itemName);

                    for (String cmd : values.keySet()) {
                        String[] commandParts = values.get(cmd).split(":");
                        String deviceCmd = commandParts[1];

                        boolean match = false;
                        if (deviceCmd.startsWith(ADVANCED_COMMAND_KEY)) {
                            // skip advanced command key and compare remaining characters
                            if (data.startsWith(deviceCmd.substring(1))) {
                                match = true;
                            }
                        } else {
                            try {
                                String eiscpCmd = EiscpCommand.valueOf(deviceCmd).getCommand();

                                // compare 3 first character
                                if (data.startsWith(eiscpCmd.substring(0, 3))) {
                                    match = true;
                                }

                            } catch (Exception e) {
                                logger.error("Unrecognized command '" + deviceCmd + "'", e);
                            }
                        }

                        if (match) {
                            Class<? extends Item> itemType = provider.getItemType(itemName);
                            State v = convertDeviceValueToOpenHabState(itemType, data);
                            eventPublisher.postUpdate(itemName, v);
                            break;
                        }
                    }
                }
            }
        }
    }
`
~petev

vavaroutsos:
Fixes #4255 regarding problems with status updates when multiple devices are configured and fixes #4240 regarding advanced command status updates. There was no check to match the device ID against the device ID in the binding specification, so status updates from devices were being posted to items bound to different devices. A full comparison (instead of the first three characters) was also added for advanced command keys as three characters is not enough to distinguish advanced commands. This caused status updates to be written to the wrong item.

Signed-off-by: Pete Vavaroutsos<pete@vavaroutsos.com> (github: vavaroutsos)

### DIFF
--- a/bundles/binding/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoBinding.java
+++ b/bundles/binding/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoBinding.java
@@ -347,12 +347,17 @@ public class OnkyoBinding extends AbstractBinding<OnkyoBindingProvider>
 
                     for (String cmd : values.keySet()) {
                         String[] commandParts = values.get(cmd).split(":");
+                        String deviceId = commandParts[0];
                         String deviceCmd = commandParts[1];
+
+                        if (!deviceConfig.deviceId.equals(deviceId)) {
+                            continue;
+                        }
 
                         boolean match = false;
                         if (deviceCmd.startsWith(ADVANCED_COMMAND_KEY)) {
-                            // skip advanced command key and compare 3 first character
-                            if (data.startsWith(deviceCmd.substring(1, 4))) {
+                            // skip advanced command key and compare remaining characters
+                            if (data.startsWith(deviceCmd.substring(1))) {
                                 match = true;
                             }
                         } else {
@@ -363,7 +368,6 @@ public class OnkyoBinding extends AbstractBinding<OnkyoBindingProvider>
                                 if (data.startsWith(eiscpCmd.substring(0, 3))) {
                                     match = true;
                                 }
-
                             } catch (Exception e) {
                                 logger.error("Unrecognized command '" + deviceCmd + "'", e);
                             }
@@ -380,7 +384,6 @@ public class OnkyoBinding extends AbstractBinding<OnkyoBindingProvider>
             }
         }
     }
-
     /**
      * Convert receiver value to OpenHAB state.
      *


teichsta:
Thanks @vavaroutsos! Also cherry-picked onto 1.8 …

Issue #4257, opened 2016-04-04
[owntracks] account for accuracy


travisghansen:
I'd like to see an option similar to the Home Assistant option described here:

https://home-assistant.io/components/device_tracker.owntracks/

> max_gps_accuracy (Optional): Sometimes Owntracks can report GPS location with a very low accuracy (few kilometers). That can trigger false zoning in your Home Assistant installation. With the parameter, you can filter these GPS reports. The number has to be in meter. For example, if you put 200 only GPS report with an accuracy under 200 will be take in account.

steve-bate:
+1

PR #4272, opened 2016-04-10
Owntracks: filter events based on accuracy

steve-bate:
See request #4257.

This is relatively straightforward. However, I needed to change the JSON Map to Map<String,Object> since the payload accuracy value is an Integer and not a String.

### DIFF
--- a/bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeBinding.java
+++ b/bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeBinding.java
@@ -49,6 +49,8 @@ public class MqttitudeBinding extends AbstractBinding<MqttitudeBindingProvider> 
     private Location homeLocation = null;
     private float geoFence = 0;
     
+    private int maxGpsAccuracy;
+    
     // list of consumers (grouped by broker)
     private Map<String, List<MqttitudeConsumer>> consumers = new HashMap<String, List<MqttitudeConsumer>>();
     
@@ -118,6 +120,15 @@ public class MqttitudeBinding extends AbstractBinding<MqttitudeBindingProvider> 
 			logger.debug("Mqttitude binding configuration updated, 'home' location specified ({}) with a geofence of {}m.", homeLocation.toString(), geoFence);
         }
 		
+		maxGpsAccuracy = Integer.parseInt(getOptionalProperty(properties, "max_gps_accuracy", "-1"));
+		if (maxGpsAccuracy < 0)
+		{
+			logger.debug("Mqttitude binding configuration: no limit on GPS accuracy");
+		}
+		else {
+			logger.debug("Mqttitude binding configuration: GPS accuracy limit is {}", maxGpsAccuracy);
+		}
+		
 		// need to re-register all the consumers/topics if the home location has changed
 		unregisterAll();
 		registerAll();
@@ -170,7 +181,7 @@ public class MqttitudeBinding extends AbstractBinding<MqttitudeBindingProvider> 
 		// 		 have multiple item bindings - i.e. monitoring multiple regions
 		if (consumer == null) {
 			// create a new consumer for this topic
-			consumer = new MqttitudeConsumer(homeLocation, geoFence);
+			consumer = new MqttitudeConsumer(homeLocation, geoFence, maxGpsAccuracy);
 			consumer.setTopic(topic);
 			
 			// register the new consumer
--- a/bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeBinding.java
+++ b/bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeBinding.java
@@ -121,11 +121,10 @@ public class MqttitudeBinding extends AbstractBinding<MqttitudeBindingProvider> 
         }
 		
 		maxGpsAccuracy = Integer.parseInt(getOptionalProperty(properties, "max_gps_accuracy", "-1"));
-		if (maxGpsAccuracy < 0)
-		{
+		
+		if (maxGpsAccuracy < 0) {
 			logger.debug("Mqttitude binding configuration: no limit on GPS accuracy");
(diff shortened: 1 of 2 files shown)


steve-bate:
I tested this change manually by injecting an opentrack transition payload into the associated MQTT topic  and stepping through the code in the Mqttitude binding in a debugger as the message was processed. I tested both acceptable and excessive GPS (in)accuracies. I'm also running the addon in my openHAB 1 installation for additional testing. After the PR is merged, I'll update the wiki with information about the new feature.

steve-bate:
@kai or @watou. Have you seen the following 'openHAB 2 Feature Verification' failure when building an openhab1 PR?

> shaded.org.eclipse.aether.resolution.ArtifactResolutionException: Could not find artifact org.openhab.addons:openhab-addons:xml:features:1.8.3-SNAPSHOT in cloudbees-private-snapshot-repository (https://repository-openhab.forge.cloudbees.com/snapshot/)

kaikreuzer:
Weird, I do not see any references to a 1.8.3-SNAPSHOT in the master branch, so I have no clue why it is looking for this...

bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeBinding.java
```diff
@@ -118,6 +120,15 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
 		
+		maxGpsAccuracy = Integer.parseInt(getOptionalProperty(properties, "max_gps_accuracy", "-1"));
+		if (maxGpsAccuracy < 0)
+		{
+			logger.debug("Mqttitude binding configuration: no limit on GPS accuracy");
+		}
+		else {
+			logger.debug("Mqttitude binding configuration: GPS accuracy limit is {}", maxGpsAccuracy);
+		}
+
```
teichsta: reformat using ESH formatter

bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeConsumer.java
```diff
@@ -109,12 +111,12 @@ public void processMessage(String topic, byte[] payload) {
...
 		
 		// read the payload into a JSON param/value map
-		Map<String, String> jsonPayload = readJsonPayload(decoded);
+		Map<String, Object> jsonPayload = readJsonPayload(decoded);
 		if (jsonPayload == null)
 			return;
 
 		// only interested in 'location' or 'transition' publishes
-		String type = jsonPayload.get("_type");
+		String type = jsonPayload.get("_type").toString();
```
teichsta: better use `(String)` rather than `toString()` in order to prevent NPEs

bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeConsumer.java
```diff
@@ -166,22 +168,30 @@ public void processMessage(String topic, byte[] payload) {
 				}
 			} else {
 				// we are only interested in location updates with an 'event' (i.e. enter/leave)
-				String event = jsonPayload.get("event");
+				String event = jsonPayload.get("event").toString();
```
teichsta: see above

bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeConsumer.java
```diff
@@ -166,22 +168,30 @@ public void processMessage(String topic, byte[] payload) {
...
 				if (StringUtils.isEmpty(event)) {
 					logger.trace("Not a location enter/leave event, ignoring");
 					continue;
 				}
 
 				// check this event is for the region we are monitoring
-				String desc = jsonPayload.get("desc");
+				String desc = jsonPayload.get("desc").toString();
```
teichsta: see above

bundles/binding/org.openhab.binding.mqttitude/src/main/java/org/openhab/binding/mqttitude/internal/MqttitudeBinding.java
steve-bate: @teichsta Sorry about the MS style brackets which have become a habit from .NET programming. I can fix that. However, are we requiring ESH formatting for openHAB 1 code? If so, where is the formatter configuration in the openhab repository? I've searched for it but haven't found it, at least in the version 1.8 branch where this code lives.

cdjackson: > However, are we requiring ESH formatting for openHAB 1 code? If so, where is the formatter configuration in the openhab repository? I've searched for it but haven't found it, at least in the version 1.8 branch where this code lives.

If you use the new IDE, then it enforces it - I guess people using older IDEs don’t have the ESH formatter installed...

teichsta: > However, are we requiring ESH formatting for openHAB 1 code?

no, we don't … simply moving the brackets would also be possible :-)

steve-bate:
Please hold off on merging this commit. I'm seen some errors in my log file that indicate that the OwnTracks JSON accuracy field is sometimes parsed as an Integer and sometimes as a Double. I haven't seen anything other than integer values in the raw JSON but I'm investigating.

steve-bate:
The accuracy is a floating point number although apparently GPS and wifi locations from iOS only arrive as integers. However, iBeacons have a floating point accuracy. The other issue is that the location and accuracy is determined by GPS or wifi when entering or leaving beacon regions. Given the beacons are often inside a building, this location may not very accurate although the physical location of the beacon is well known (for non-moving beacons). I've modified the code to not check for accuracy on beacon updates. This may not be the ideal solution for mobile beacons but it seems like that's a different use case than openHAB's presence detection support.

PR #4117, opened 2016-02-29
ClassicUI: full image and chart refresh

watou:
Support the `refresh=` parameter to Image and Chart widgets in all cases in the classic web UI.  Testing feedback should go in [this topic](https://community.openhab.org/t/please-help-test-fix-for-image-and-chart-refresh-in-classic-web-ui/8106).

Same change as in https://github.com/eclipse/smarthome/pull/1106.

PR #4278, opened 2016-04-11
fix Classic-UI

### LINKED ISSUE #4214: No updates in Classic UI after 1.8.2 upgrade
Setpoint and Switch widgets do not update in 1.8.2 while they did in 1.8.1.

Discussion [here](https://community.openhab.org/t/switch-setpoint-broken-after-1-8-2-upgrade/8916).

teichsta:
@watou @sja could you please have a look if i reapplied your changes correctly?


sja:
Nope, that won't work. Line 36 relies on the return value of WA.request(). But that was an extension for Webapp.NET.
I suggest to follow Kai and revert the complete ClassicUI to an early stage and let the contributors (just me and @watou ?) Re-PR this in.

teichsta:
ok @sja and @watou i've now reverted ALL changes to the state before commit 1025a73cd2fa84dc4a5de72ae7ddd5e80af3e067. How should be now proceed. Should first start @sja and than @watou with reapplying?

watou:
I am happy to prepare a new PR with the same changes that were in #4117.  (#4117 superceded #3930 so #3930 doesn't need to be considered).  Just let me know when my starting point is merged into the 1.8 branch.  Thanks.

teichsta:
great, thanks @watou! @sja so you would be the one to start, right?

teichsta:
@watou reapplied #4117 would you agree it was successful?

watou:
@teichsta I don't see the changes from #4117 reapplied.  Where did you re-apply them?

teichsta:
it's all squashed into the single commit …

watou:
OK, I now see my changes are *not* in this diff, so that makes sense.  But I want to do a build and test the bundle on my live 1.8.2 system, because I'm still not sure the "OH." namespace is consistent.  I will do actual tests to make sure #4117 is present in this PR, and let you know my tomorrow evening?

teichsta:
>  I will do actual tests to make sure #4117 is present in this PR, and let you know my tomorrow evening?
Appreciate your help, thanks a lot!

watou:
Testing fails (sorry):

1. Clean install of OH 1.8.2 on RPi2 Jessie using apt-get
1. Removed `org.openhab.ui_1.8.2.jar` from `server/plugins`.
1. Full `mvn clean install` build of origin/pr/4278
1. Put resulting `bundles/ui/org.openhab.ui.webapp/target/org.openhab.ui.webapp-1.8.3-SNAPSHOT.jar` into `addons` folder on RPi2.
1. `sudo systemctl start openhab`
1. Current state of the PR leads to 404s:

teichsta:
so i assume this is the part @sja could PR in again, right? @sja could you please help with that soon?

teichsta:
it boils down to two ".." dots to much :S

vavaroutsos:
Personally I have not had a problem with the 1.8.0 slider elements. The situation seems worse now in 1.8.3. The slider is replaced with a setpoint element, but when attached to a dimmer item, the behavior is less than ideal. Using chrome from a windows PC, pressing the up/down setpoint buttons results in values of 100 and 0 respectively. This is fine, although it really doesn't work for dimming. This means all items need to have explicit entries in the site map to specify a step. This totally breaks old site maps and is very tedious compared to using item groups. To make matters worse, the up/down setpoint buttons don't work at all using chrome from an android phone or safari from an ipad. Pressing the buttons does nothing from these devices.

Regards,

~petev

teichsta:
@vavaroutsos yes the slider is represented with Up/Down Buttons in Classic-UI. Pressing and holding the button should do the trick for dimming. At least the current demo shows it's working (see http://demo.openhab.org:8080/openhab.app?sitemap=demo#_0301).

vavaroutsos:
@teichsta pressing and holding the buttons does nothing when tested with Chrome on a windows 7 PC. Now my Android devices are having trouble connecting to my openhab site via Chrome as well as IOS devices via Safari. There was never a problem before I upgrades to 1.8.3. Android devices tell me the site is unreachable, yet they can ping the openhab server. Not sure what's going on.

Regards,

~petev

sja:
IMHO this is not related to this Bug. 

Von meinem Toaster gesendet

teichsta:
@vavaroutsos can you please direct your clients (Android/iOS) to the demo server and see if that works? You will find a Slider under "Widget Overview"

vavaroutsos:
@sja, it has something to do with the changes made after 1.8.1. I reverted only the openhab-runtime to version 1.8.1 and everything works again. Android and IOS clients can connect to the site, sliders are back, set points work from mobile clients, etc.

~petev

vavaroutsos:
@teichsta, from Android via chrome:

1) the dimmer at ground floor ->living room->table pretty much doesn't work at all. One time I got the light icon to turn off. All other attempts of long and short presses did nothing.

2) the dimmer at widget overview->dimmer behaves erratically. I can see the percentage change during long presses, however it doesn't always stop on release. Short presses sometimes do nothing, cause a 5% change, or start a sequence of 5% changes which continue after release until 100% or 0%.

I don't have access to an IOS client right now.

~petev

Issue #4239, opened 2016-03-29
IEC-62056-21 Binding: Support for ISKRA MT174 OBIS Codes


Puncher4444:
Hello,

no values are shown in Openhab even data comes in. I guess its because of the OBIS codes of the ISRKA MT174 meter.
When starting OH in debug mode, the binding gets the data (from my point of view) correct from the meter. The log shows the following:

```
19:26:42.579 [DEBUG] [o.b.i.i.Iec6205621MeterBinding:119  ] - Read meter: meter1; /dev/ttyUSB2
19:26:46.660 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: MT174-0001;;
19:26:46.661 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 0-0:F.F.0*255;0000000;
19:26:46.662 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.0.1*255;  1ISK00;
19:26:46.663 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.0.0*255;64153041;
19:26:46.663 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.0.2;4M012531;
19:26:46.664 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.9.1*255;192607;
19:26:46.665 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.9.2*255;1160329;
19:26:46.666 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.1.0*255;8;
19:26:46.666 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.1.2*01;1603010000;
19:26:46.667 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.1.2*02;1602010000;
19:26:46.668 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.1.2*03;1601010000;
19:26:46.669 [DEBUG] [iec6205621meter.internal.Meter:83   ] - DataSet: 1-0:0.1.2*04;1512010000;
. . .
```

I´ve created a Item Number in my Items files:

[...]

PR #4281, opened 2016-04-12
Fix issue reported in #4239

kreutpet:
IEC-62056-21 Binding: Support for ISKRA MT174 OBIS Codes 
OBIS code of the format A-B:C.D.E*F conflict with item configuration in which conflicts with separation string in item configuration


kreutpet:
will change the wiki as soon as the pull request in released
/kreutpet

teichsta:
Thanks @kreutpet! Also cherry-picked onto 1.8 …

PR #4286, opened 2016-04-14
New binding for chinese RGB led controller

theater:
Please review README.MD inside the plugin folder and provide feedback. 
My first binding development so be merciful :)

### DIFF
--- a/bundles/binding/org.openhab.binding.chinesergbledcontroller/README.MD
+++ b/bundles/binding/org.openhab.binding.chinesergbledcontroller/README.MD
@@ -4,3 +4,5 @@ TCP packets with specific bytes as a payload and based on the prefix/sufix and p
 Binding is very simple and has been tested with RGB UI control and regular switch (ON/OFF).
 Because there is no model or any other way to label it it's currently named chinesergbledcontroller. :)
 
+The type is this one: http://www.aliexpress.com/item/16-millions-color-RGB-WIFI-LED-controller-with-remote-Group-or-timing-control-music-mode-for/32256711379.html 
+Not sure how else people can identify it. :)


hakan42:
Is there really a product / company that creates this "Chinese RGB controller" or is it a clone of some other product?

theater:
Hi,
on the sites - no. Looking at the SDK documents that they provided me they're refering to it as an "HF-A11-SMT module" but documentation is very chinesish. No idea if there is any simmilar comercial solution by western companies.

Cheers,
Konstantin

watou:
Adding to @hakan42's comment, could you perhaps rename the binding and Java packages to something like `hfa11smt` from `chinesergbledcontroller`?

9037568:
Where's the actual led controller this is for? When I search for hf-a11-smt, I get a bunch of references to usb wifi modules like [this one](http://www.hi-flying.com/products_detail/&productId=60326fef-3398-422b-9350-62f5f99a8dbc&comp_stats=comp-FrontProducts_list01-111.html) ...

theater:
Hi guys,
I've purchased one of these: http://www.aliexpress.com/item/16-millions-color-RGB-WIFI-LED-controller-with-remote-Group-or-timing-control-music-mode-for/32256711379.html
Of course I can rename it to anything you would like but since there is no strict naming I don't know what to use for it.

Cheers,
K.

watou:
The AliExpress link talks about brand "HI-POWER" and core of product "WF200 Controller" -- are either/both combined a more suitable name for the binding/packages?

hakan42:
Honestly, the Alixepress pictures and description look like yet another milight clone.

@theater , maybe you could enhance that binding instead of creating a fresh one?

theater:
Hi guys, I've looked at the milight addon code. It's not the same but I would propose... let me have a deeper look and to see if I can fit my code + the chinese module inside and get back to you. If this can be done fairly easy I will integrate it insight. If there are architectural differences I will rename this one and we may push it further if you agree. What do you think?

Cheers,
Konstantin

PR #4288, opened 2016-04-15
Mystrom Binding: Reduces http call due to complaint of mystrom.

coolweb:
Improve performance of Mystrom binding by reducing http calls to mystrom server.
It's an important fix because mystrom had performance issue on their servers without this improvement.


coolweb:
@teichsta I do a completly new fork with one commit from a feature branch :-) .
Could you review it?
Thx.

PR #4289, opened 2016-04-15
[MyStrom] Reduces http call due to complaint of mystrom.

coolweb:
Improve performance of Mystrom binding by reducing http calls to mystrom server.
It's an important fix because mystrom had performance issue on their servers without this improvement.

@teichsta Could you review it?
Thx.

### DIFF
--- a/bundles/binding/org.openhab.binding.mystromecopower/src/main/java/org/openhab/binding/mystromecopower/MyStromEcoPowerBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.mystromecopower/src/main/java/org/openhab/binding/mystromecopower/MyStromEcoPowerBindingProvider.java
@@ -12,43 +12,43 @@ import org.openhab.core.binding.BindingProvider;
 
 /**
  * @author Jordens Christophe
- * @since 1.8.0-SNAPSHOT
+ * @since 1.9.0-SNAPSHOT
  */
 public interface MyStromEcoPowerBindingProvider extends BindingProvider {
-    /**
-     * Get the name in the configuration file which must correspond to the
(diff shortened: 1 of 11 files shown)


watou:
@coolweb, what led to all the reformatting?  Ideally, the reformatting had already been done by @teichsta after the 1.8 release, and your PR would only have the code changes without reformatting, so it will be clearer what has changed for the purpose of review.  Was this PR based on the latest master branch?  If somehow the IDE formatting wasn't already on master, could you make a separate PR that only reformats?

Also, for all `@since` JavaDoc tags, could you fix them to be `1.8.0` for those files you edited and didn't create (and also remove the `-SNAPSHOT` parts).  Only new files you're introducing here should be marked `1.9.0`.

coolweb:
@watou yes it is based on latest version of master. So the only solution is to delete my fork and this PR and do a new one for the formatting and after it is accepted do a new one for the code fix? That's right?

watou:
@coolweb maybe not the only solution, but a good one that will deal with the two issues separately (1. Formatting only changes, since not correct w.r.t. current IDE setup, 2. Your binding fixes.).  Would that be OK?

coolweb:
@watou my ide setup is the setup from openhab eclipse installer...

watou:
> my ide setup is the setup from openhab eclipse installer...

Perfect.

coolweb:
@watou  I do a new pull request, I don't know why but from a fresh install, the code formatting seems to be good in eclipse. The new pull request contains removing the keyword `snapshot` and the fix to reduce http call, the new pull request is: #4295 

Thx.

watou:
Closing in preference to #4295.

PR #4291, opened 2016-04-15
Corrected handling of changed modbus values

Wolfgang1966:
The current modbus implementation reports items as updated at every poll cycle even if the value actually did not change. For setups with many modbus items (mine has about 200) and poll cycles of 500ms this simply overloads the CPU of my system. The code already tries to post updates only on changed values, but there is an error as only the very first value is stored but then never updated. This fix corrects this behaviour.

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus/OSGI-INF/modbusbinding.xml
+++ b/bundles/binding/org.openhab.binding.modbus/OSGI-INF/modbusbinding.xml
@@ -26,4 +26,7 @@
    <reference bind="addBindingProvider"  cardinality="1..n" 
       interface="org.openhab.binding.modbus.ModbusBindingProvider" name="ModbusBindingProvider"
       policy="dynamic" unbind="removeBindingProvider"/>
+    <reference bind="setItemRegistry" cardinality="0..1" 
+      interface="org.openhab.core.items.ItemRegistry" name="ItemRegistry" 
+      policy="dynamic" unbind="unsetItemRegistry"/>
 </scr:component>
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
@@ -25,6 +25,9 @@ import org.openhab.binding.modbus.ModbusBindingProvider;
 import org.openhab.binding.modbus.internal.ModbusGenericBindingProvider.ModbusBindingConfig;
 import org.openhab.core.binding.AbstractActiveBinding;
 import org.openhab.core.binding.BindingProvider;
+import org.openhab.core.items.GenericItem;
+import org.openhab.core.items.ItemNotFoundException;
+import org.openhab.core.items.ItemRegistry;
 import org.openhab.core.library.items.NumberItem;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.types.Command;
@@ -53,7 +56,7 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
     private static final String TCP_PREFIX = "tcp";
     private static final String SERIAL_PREFIX = "serial";
 
-    private static final String VALID_COFIG_KEYS = "connection|id|start|length|type|valuetype|rawdatamultiplier|writemultipleregisters";
+    private static final String VALID_COFIG_KEYS = "connection|id|start|length|type|valuetype|rawdatamultiplier|writemultipleregisters|updateunchangeditems";
     private static final Pattern EXTRACT_MODBUS_CONFIG_PATTERN = Pattern.compile(
             "^(" + TCP_PREFIX + "|" + UDP_PREFIX + "|" + SERIAL_PREFIX + "|)\\.(.*?)\\.(" + VALID_COFIG_KEYS + ")$");
 
@@ -62,6 +65,24 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
 
     /** slaves update interval in milliseconds, defaults to 200ms */
     public static int pollInterval = 200;
+
+    /**
+     * ModbusBinding uses the ItemRegistry to check for the last Item state and, depending on the slave
+     * configuration, avoid pushing updates of unchanged items.
+     */
+    private ItemRegistry itemRegistry = null;
+
+    protected void setItemRegistry(ItemRegistry itemRegistry) {
+        this.itemRegistry = itemRegistry;
+    }
+
+    protected void unsetItemRegistry(ItemRegistry itemRegistry) {
+        this.itemRegistry = null;
+    }
+
+    public ItemRegistry getItemRegistry() {
+        return itemRegistry;
+    }
 
     @Override
     public void activate() {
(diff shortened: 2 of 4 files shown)


ssalonen:
Looks like the correct way to implement fix for this...really nasty bug! However, I use this opportunity to discuss the feature in general.

There has been [discussion in the community forums about this](https://community.openhab.org/t/modbus-tcp-polling-generates-events-even-if-there-is-no-state-change/6708/5) and similar pull request has been posted previously, see  [#2939](https://github.com/openhab/openhab/pull/2939) (at the time I could not see the actual bug but now realizing why you would have multiple state updates). Especially note openhab maintainer [@watou's comment](https://community.openhab.org/t/modbus-tcp-polling-generates-events-even-if-there-is-no-state-change/6708/2?u=ssalonen), quoting here for future reference:

> It is normal for a binding to generate update events for items even when those items have the same state as before. There may be unwanted side effects, however, like the default logging of events in events.log, which in my opinion is not a good default. That's why I set event logging to WARN in my logback.xml, so no events are logged.

> A separate subject is, is polling 5 times a second a sensible thing to do? Obviously the developer thought so, but I don't know anything about Modbus to independently evaluate that decision. Absent anything special about Modbus, polling 5 times a second seems way too frequent.

> About MQTT publishing updates too often, I just provided a workaround to another user that uses a rule to only publish to MQTT when an item changes, instead of being updated, that uses openHAB 1.8 and the new MQTT action bundle:

````
rule MQTTPublishOnChange
when
  Item YourModBusItem changed
then
  publish(yourBroker, yourTopic, YourModbusItem.state.toString)
end
````

(as discussed the thread, there are perfectly valid reasons for having short polling period)

Note the **changed** keyword in the rule. So you would still get the event, but the rules would not do any calculations. Would this lower the CPU load you are experience? What kind of rules you have configured?

To put this in more practical context: In my use case I'm actually interested having explicit events whenever read is successfull. This allows rules to react to broken connections (e.g. "if no modbus value received in 5 minutes -> send sms") and communication faults. I'm using this for z-wave wireless sensors but why not have the same functionality here also?

If the CPU load would be issue even with the rules reacting only _changed_ items, could we make this feature configurable by the user?

Finally, if we decide to go with this "event protection", should we compare against most current item state instead of binding keeping the state? Is it possible that item state is updated in e.g. user rules and thus becomes out-of-sync with binding state (`config.getItem()`)?

Best regards,
Sami

Wolfgang1966:
Hello Sami,

thanks for your reply!

In my case I use modbus to poll my SPS which controls nearly everything in my house, e.g. all lights. When I turn on a light using a switch at the wall (yeah, such ones still exist :-) ), I want openHAB to detect this immediately, hence the polling frequency.

I think the best option is to make this configurable. Is there already a name established for this kind of configuration value? I would suggest to use something like

modbus:tcp.slave1.updateunchangeditems=(true|false}

so that the value can be set individually for each slave.

Regarding changes triggered by user rules: Thanks for the hint, I will test this.

I will add the necessary changes to this pull request the next days.

Wolfgang1966:
It is now configurable whether unchanged items are causing updates or not. Configuration can be set per slave; default is not to push updates. Please check if this looks ok for you.

Wolfgang1966:
By the way: I just read the "How to Contribute" article, which states:

Bundles must not require any substantial CPU time. Test this e.g. using "top" or VisualVM and compare CPU utilization with your bundle stopped vs. started.

To me this reads like a strong hint not to push unnecessary updates :-)

watou:
> To me this reads like a strong hint not to push unnecessary updates :-)

Yes, but the question is, what is an unnecessary update? Knowing most recently to "now" what the state of an item is valuable, and is accomplished by posting updates as they are known. Suppressing duplicate updates loses this knowledge.  All that said, if the volume of updates creates too high of a load, that would have to be mitigated somehow in the binding.  Complicating things further is that the meaning of "too high of a load" isn't clearly definable.

Wolfgang1966:
Of course openHAB should have an up to date view of the state of the configured item. With bindings that use polling this means that you have to poll frequently so that you get to know when things change. Polling frequently also means that you may receive the same result over and over as things did not change for a while. In that case, publishing the same result over and over to the bus simply causes a lot of overhead. This is why there should be at least a chance to tell the binding to only send events on changes.

To answer your question: I consider an update unnecessary when the item shows no state change compared to the last posted update. As soon as there is a change, how small it might be, it is of course a necessary update. Also when I get a changed value with every polling cycle then of course an update has to be posted. In fact, the latter case is an indication that the polling frequency might be to low.

In my case (modbus binding on bananapi) CPU usage of openHAB went from 14% to 190% when using the unchanged modbus binding (it's a two core machine, and the remaining CPU capacity was needed by the other processes like mysql a.s.o.), so I consider this a definitive case of "too high".

Now with the latest additions everyone who uses the binding is able to configure whether updates should be posted only on changed values or on every polling, as Sami suggested.

ssalonen:
From code perspective the configuration looks good to me.

Just for the record, can you confirm what kind of rules you had when experiencing high cpu usage? Did you have rules reacting to changes or something else?

Another point about how to detect changes: should we use ItemRegistry instead of storing the state in the binding? See for example how [mongo persistence uses ItemRegistry in getItem](https://github.com/openhab/openhab/blob/master/bundles/persistence/org.openhab.persistence.mongodb/java/org/openhab/persistence/mongodb/internal/MongoDBPersistenceService.java). Would the same work here? (I think you need some declare dependency to item registry in some osgi config file)

Wolfgang1966:
Hi Sami,

regarding the rules: I have rules that convert the integer values received from modbus to decimal values. They look like this:

rule "Temp Wohnzimmer"
when
	System started or
	Item Rawtemp_Wohnzimmer changed
then
	var Number realtemp = Rawtemp_Wohnzimmer.state as DecimalType
	realtemp = (realtemp) / 10
	Temp_Wohnzimmer.postUpdate(realtemp)
end

(Rawtemp_Wohnzimmer is a modbus item) 

Furthermore I store many of the modbus values in a database using openhab persistence.

I also think using ItemRegistry is the better option compared to the local cache currently implemented. I already found ItemRegistry in the openhab core, I just had no clue on how to retrieve the object, so thanks for the hint!

Update will follow!

ssalonen:
Check this snippet how to configure dependency. I think that would be enough 

https://github.com/openhab/openhab/blob/master/bundles/persistence/org.openhab.persistence.mongodb/OSGI-INF/mongodb.xml

Wolfgang1966:
Hello Sami,

got it working, updated pull request.

Thanks

Wolfgang

--
Wolfgang@Klimt.de - http://www.klimt.de/ 

It is difficult to get a man to understand something, when his salary depends upon his not understanding it! -- Upton Sinclair

Wolfgang1966:
Hi,

obviously there were conflicting changes to the head release in between. I reapply my changes on top of them and create a new pull request afterwards.

slawekjaranowski:
If you interested I also have similar issue with another binding, I compared some system utilization depends on publishing all event or only changed ones. 

https://github.com/openhab/openhab/pull/2784#issuecomment-214891831

ssalonen:
@Wolfgang1966 apparently it is not "allowed" to use item registry in the binding, see [this comment](https://github.com/openhab/openhab/pull/2784#issuecomment-215119903)

Wolfgang1966:
Ah, interesting.

Then how does a good solution look like? Back to internal caching inside the binding?

From the discussion going on in the past and also in #2784 I think the way of dealing with Item updates is a general topic to think about. To bring in some numbers: On my system (banana pi, original model), it looks like this (modbus:poll set to 1000ms):

updateunchangeditems false, no client:
Java CPU usage: 10-15% load 0,1-0,2

updateunchangeditems false, one client:
Java CPU usage: 15%, short peaks at 150% (rendering chart) load 0,5

updateunchangeditems true on one slave, no client:
Java CPU usage: 90% load 1,3

updateunchangeditems true on one slave, 1 client:
Java CPU usage: 90%, peaks at 160% load 1,3

updateunchangeditems true on two slaves, no client:
Java CPU usage: 180% load 2,2

updateunchangeditems true on two slaves, 1 client:
Java CPU usage: 180%, load 2,2
(system is cpu-limited)

So for modbus I think it is a "must" to take care of this. For bindings that only provide few items or allow poll cycles of several seconds or even minutes, that might not be an issue at all. But in general I think we want a smart home platform to reflect a quite up to date view of the home, while running smooth and energy-efficient on a system with limited resources, so we should allow frequent polling without putting heavy load on the system.

PR #4296, opened 2016-04-17
Fix #4294 Add call to improved method

### LINKED ISSUE #4294: [Mystrom] Fix code formatting
Reduces http call due to complaint of mystrom.

coolweb:


bundles/binding/org.openhab.binding.mystromecopower/src/main/java/org/openhab/binding/mystromecopower/internal/MyStromEcoPowerBinding.java
```diff
@@ -172,7 +180,7 @@ protected void execute() {
                         }
                     }
                 } else {
-                    logger.warn("The device itemName '{}' not found on discovery. Verify device is not offline",
+                    logger.warn("The device itemName '{}' not found on discovery verify device is not offline",
```
watou: If I merge this as is, could you then open a new PR to revert this change in the message?  Otherwise, could you revert the change with a new commit on this PR branch, and then squash so there is one commit?

coolweb:
@watou Done, revert the message and squash my commits into one ;-)

PR #4301, opened 2016-04-19
Only send Intertechno settings if specified

### LINKED ISSUE #3987: Do not send intertechno sending parameters per default
Currently the intertechno sends repetitions and wavelenght parameters always on startup.
This causes issues with some sticks (e.g. nanocul)

See (German discussion) http://forum.fhem.de/index.php?topic=35064.msg404579#msg404579

I think it would make much more sense not to send the parameters unless they are configured in the settings. Normally sticks are already configured with default settings and it does not make sense to change the default settings from the stick to "openhab default" settings. If someone did configure the settings he probably knows what he is doing and we would still send them.

I have a fix for this but it's currently based on top of #3885 which needs to get merged first.

tarioch:
Fixes #3987

Signed-off-by: Patrick Ruckstuhl patrick@ch.tario.org (github: tarioch)

### DIFF
--- a/bundles/binding/org.openhab.binding.intertechno/src/main/java/org/openhab/binding/intertechno/internal/CULIntertechnoBinding.java
+++ b/bundles/binding/org.openhab.binding.intertechno/src/main/java/org/openhab/binding/intertechno/internal/CULIntertechnoBinding.java
@@ -42,7 +42,7 @@ public class CULIntertechnoBinding extends AbstractBinding<CULIntertechnoBinding
      * href="http://culfw.de/commandref.html">Culfw Command Ref</a> for more
      * details.
      */
-    private final static String KEY_REPITIONS = "repetitions";
+    private final static String KEY_REPETITIONS = "repetitions";
     /**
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.intertechno/src/main/java/org/openhab/binding/intertechno/internal/CULIntertechnoBinding.java
```diff
@@ -42,7 +42,7 @@
      * href="http://culfw.de/commandref.html">Culfw Command Ref</a> for more
      * details.
      */
-    private final static String KEY_REPITIONS = "repetitions";
+    private final static String KEY_REPETIONS = "repetitions";
```
kaikreuzer: Spelling not fully fixed. Please change REPETIONS to REPETITIONS.

tarioch:
thanks, should be fixed

kaikreuzer:
Thanks!

PR #4308, opened 2016-04-22
adds esh addons to the verification

kaikreuzer:
fixes https://github.com/openhab/openhab-distro/issues/163

### DIFF
--- a/features/openhab-addons-verify/pom.xml
+++ b/features/openhab-addons-verify/pom.xml
@@ -45,6 +45,13 @@
             <type>xml</type>
         </dependency>
         <dependency>
+            <groupId>org.openhab.core</groupId>
+            <artifactId>openhab-esh-addons</artifactId>
+            <version>${ohc.version}</version>
+            <classifier>features</classifier>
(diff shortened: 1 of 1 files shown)


features/openhab-addons-verify/pom.xml
```diff
@@ -45,6 +45,13 @@
             <type>xml</type>
         </dependency>
         <dependency>
+            <groupId>org.openhab.core</groupId>
+            <artifactId>openhab-esh-addons</artifactId>
```
mrguessed: To get this to be enabled, I had to manually add:
```
   <descriptor>mvn:org.openhab.core/openhab-esh-addons/${ohc.version}/xml/features</descriptor>
```

to the `descriptors` block below.  Without it, the repo entry was missing and the MiOS Binding wouldn't resolve the `openhab-transformation-map` feature dependency.

Not sure if that's correct or not...

kaikreuzer: Good point, I had missed that!

PR #4309, opened 2016-04-22
FritzBoxTR064 Binding: Update for network stability

gitbock:
if network failure occurrs remediate from error properly. Close connections.

- Release connection after communicating SOAP to FritzBox
- Decrease "no TR064 service found" to warning
- Set Timeouts for SOAP response to 4 seconds


teichsta:
also cherry-picked onto 1.8

PR #4312, opened 2016-04-24
PR for consideration : Oppo bluray player against OH 1.9.0

netwolfuk:
Hi All.

First PR for OpenHAB. I welcome your review.
Adds support for Oppo Bluray player over RS232.

Please let me if there is anything I've done wrong and should fixup.


netwolfuk:
Hmm. I'm not sure what I did wrong here, but my binding didn't appear in the build above.

ssalonen:
You probably need to add it to [binding/pom.xml](https://github.com/openhab/openhab/blob/master/bundles/binding/pom.xml)

netwolfuk:
Oh. Yes. Good point. Thanks. I'll fix that shortly.

9037568:
There are a number of errors in this code set that make it unbuildable in either Maven or Eclipse. You'll need to fix those first.

Once those are fixed you should consider cleaning up the large number of warnings for unused code.

Are you able to make these changes, @netwolfuk ?

Thanks!

TheIndustriousCoder:
What is the best way to move this PR forward? Should the changes be made to this branch?

netwolfuk:
I bought a RPi3 this week, and will look to moving my stuff to OH2.0
Maybe we should abandon this PR, and I'll rewrite it for OH2.
Or is OH2 support for OH1.x bindings still an acceptable option?

netwolfuk:
@9037568 Sorry, your original message came in when I was travelling and it fell to the back of my mind.

> Maybe we should abandon this PR, and I'll rewrite it for OH2.
> Or is OH2 support for OH1.x bindings still an acceptable option?

9037568:
OH2 can certainly still use OH1 bindings. This would be advantageous to those who can't move to OH2 or who choose not to.

Your call, though, if you want to migrate this binding to a full OH2 addon. If so, just close this ticket at your convenience. Thanks!

netwolfuk:
I'll have another look at what it takes to migrate the binding and will either update this PR or close it and raise another against 2.0

netwolfuk:
This is probably not the right place to ask, but...

This binding uses RS232 over serial. I'm not sure how I would do auto discovery with that. I don't think just randomly probing serial ports is a good idea. ;-)

9037568:
Agreed, poking serial ports at random is probably not a good idea.

TheIndustriousCoder:
@netwolfuk you should be able to do something similar to what the z-wave binding does. It asks for the serial port when installing the binding and uses that port for auto discovery. I just got started with OH2 and am looking forward to using your Oppo binding and hopefully contributing bindings for some of the other gear that I use.

netwolfuk:
Ah. That's good to know. I'll have a look into how that works. Thanks.

PR #4321, opened 2016-04-26
[mios] Install MAP and REGEX transformation dependencies for OH2

mrguessed:
The MiOS Binding internally uses the MAP and REGEX transformation
services.  This change ensures that OH2 will install those deps
when the MiOS Binding is deployed.

Also tweaks #4308 to add the other component for complete feature
verification.  See discussion in openhab/openhab-distro#163

Signed-off-by: Mark Clark <mr.guessed@gmail.com> (github: mrguessed)


kaikreuzer:
Many thanks!

PR #4326, opened 2016-04-27
Fix #4320 by restoring the htmlescape.

### LINKED ISSUE #4320: Ampersand "&" in item label text breaks item in classic UI
I put an ampersand in an item label text as shown below:

> Group:Contact:OR(OPEN, CLOSED) gContacts "Windows & Doors [(%d)]" <contact> (gAll)

This caused the group page to not open when clicked. The URL in the browser would change to that with the fragment of the group page, but the page would not open.

This was the case with the classic UI. The page opened properly when tested from habdroid.

~petev

9037568:
This should allow the classic UI to show items containing the ampersand character again.


9037568:
If this change is approved, please also cherry-pick into 1.8.3...

kaikreuzer:
It cannot be approved, because the Classic UI is not part of the master branch at all - so best first discuss on https://github.com/openhab/openhab/issues/4320#issuecomment-215505228 whether it is critical enough to create a PR against the 1.8 branch.

PR #4328, opened 2016-04-27
Add Openhab 1.x WOL binding to be included in Openhab2 distribution

avdleeuw:


watou:
Hi @avdleeuw, looking at [feature.xml](https://github.com/avdleeuw/openhab/blob/add-wol-binding/features/openhab-addons/src/main/feature/feature.xml), I see [examples](https://github.com/avdleeuw/openhab/blob/add-wol-binding/features/openhab-addons/src/main/feature/feature.xml#L144) of [bindings](https://github.com/avdleeuw/openhab/blob/add-wol-binding/features/openhab-addons/src/main/feature/feature.xml#L173) that do not have config files and don't need dummy config files.  Is there a problem with this approach here?  If not, please remove the dummy config file and references to it so this config-less binding is like other config-less bindings.  Thank you!

kaikreuzer:
@watou is right, if there are no config options, you do not need to add an empty file.
It would be nice though, if you could also add a binding.xml as described in (2) at https://github.com/openhab/openhab-distro/blob/master/docs/sources/features/compatibilitylayer.md#how-to-add-a-successfully-tested-1x-add-on-to-the-distribution.

avdleeuw:
No problem! 

To be honest, I really only worked from the first bit of the instructions, which made me believe it was the actual config file that caused the binding to be included. 

I will try and get this corrected today or tomorrow, hopefully including the changes to binding.xml as well.

kaikreuzer:
thanks!

avdleeuw:
I've added the binding.xml and removed the dummy config file.

watou:
Thanks!

kaikreuzer:
Perfect!

kaikreuzer:
@watou Could you please also apply the new merge strategy on all PRs (see https://community.openhab.org/t/merging-prs/10008/)? It seems you missed it here.

watou:
OK, sorry Kai!

kaikreuzer:
No worries!

PR #4331, opened 2016-04-29
Default item added to caldav command calendar.

paphko:
Optional option added for command calendar: `caldavCommand:defaultItemOnBegin=<item>`
If the command-calendar contains items whose contents does not follow the `BEGIN/END:<item>:<value>` pattern AND a default item is specified, then the respective lines are interpreted as `BEGIN:<defaultItemOnBegin>:<line>`

This is really nice to use in combination with a rule such as described here: https://github.com/openhab/openhab/wiki/Controlling-openHAB-with-your-voice
Set the default item to: `caldavCommand:defaultItemOnBegin=VoiceCommand`

Then you can use the very same logic of your voice commands also in your calendar events (which should also improve the [WAF factor](https://en.wikipedia.org/wiki/Wife_acceptance_factor) ;-)).
Depending on your rule implementation, it is possible to use event entries like these:
- Swtich on light in kitchen
- Swtich off light in bedroom
- Switch on radio
- Close roller shutter in living room
- Close roller shutter on first floor
etc.

I also changed the parser of the event contents to use regular expressions.

### DIFF
--- a/bundles/binding/org.openhab.binding.caldav-command/src/main/java/org/openhab/binding/caldav_command/internal/CalDavBinding.java
+++ b/bundles/binding/org.openhab.binding.caldav-command/src/main/java/org/openhab/binding/caldav_command/internal/CalDavBinding.java
@@ -68,6 +68,9 @@ public class CalDavBinding extends AbstractBinding<CalDavBindingProvider>impleme
 
     private static final String KEY_READ_CALENDARS = "readCalendars";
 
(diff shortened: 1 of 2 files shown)


paphko:
I will document that feature in the [wiki](https://github.com/openhab/openhab/wiki/CalDAV) as soon as this PR is merged.

kaikreuzer:
Thanks @paphko!

PR #4334, opened 2016-04-30
Z-Wave: adding new device Enerwave ZWN-RSM2

lukaskuzmiak:
Updating XML definition as discussed in issue
https://github.com/openhab/openhab/issues/2595

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -2456,9 +2456,10 @@
                 <Type>101</Type>
                 <Id>0603</Id>
             </Reference>
-            <Model>ZW20R</Model>
-            <Label lang="en">20A Tamper Resistant Duplex Receptacle</Label>
-            <ConfigFile>enerwave/zw20r.xml</ConfigFile>
+            <Model>ZW20R/ZWN-RSM2</Model>
+            <Label lang="en">20A Tamper Resistant Duplex Receptacle / Smart Dual Relay Switch Module</Label>
+            <ConfigFile VersionMax="3.9">enerwave/zw20r.xml</ConfigFile>
+            <ConfigFile VersionMin="4">enerwave/zwn-rsm2.xml</ConfigFile>
         </Product>
         <Product>
             <Reference>
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -2458,7 +2458,7 @@
             </Reference>
             <Model>ZW20R/ZWN-RSM2</Model>
             <Label lang="en">20A Tamper Resistant Duplex Receptacle / Smart Dual Relay Switch Module</Label>
-            <ConfigFile VersionMax="3.9">enerwave/zw20r.xml</ConfigFile>
+            <ConfigFile VersionMax="3.255">enerwave/zw20r.xml</ConfigFile>
             <ConfigFile VersionMin="4">enerwave/zwn-rsm2.xml</ConfigFile>
         </Product>
         <Product>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveConverterHandler.java
```diff
@@ -365,6 +365,11 @@ public void handleEvent(ZWaveBindingProvider provider, String itemName, ZWaveCom
         if (commandClassName != null
                 && !commandClassName.equalsIgnoreCase(event.getCommandClass().getLabel().toLowerCase())
                 && !(respondToBasic && event.getCommandClass() == CommandClass.BASIC)) {
+            logger.warn("Not matching command class for item = {}, command class = {}, item command class = {}, ignoring event.",
```
cdjackson: Please update these log entries to the standard with NODE number

bundles/binding/org.openhab.binding.zwave/database/products.xml
```diff
@@ -2456,9 +2456,10 @@
                 <Type>101</Type>
                 <Id>0603</Id>
             </Reference>
-            <Model>ZW20R</Model>
-            <Label lang="en">20A Tamper Resistant Duplex Receptacle</Label>
-            <ConfigFile>enerwave/zw20r.xml</ConfigFile>
+            <Model>ZW20R/ZWN-RSM2</Model>
+            <Label lang="en">20A Tamper Resistant Duplex Receptacle / Smart Dual Relay Switch Module</Label>
+            <ConfigFile VersionMax="3.9">enerwave/zw20r.xml</ConfigFile>
```
cdjackson: Are you sure about this 3.9? Maybe you should use a higher number as we discussed?

lukaskuzmiak:
Version updated. There is no Node object in the function which I updated - the other logger msgs also don't have the NODE number .. I can try to pull the node from the item but is that worth it just for logging?

cdjackson:
Isn't this debug with the BASIC command class just going to cause a lot of confusion?  If we have multiple command classes in a node, then only 1 will be linked to basic, and the others will give an error. I'd prefer actually to remove this message as I think it will cause more confusion.

For example, for the FGMS001, which has 4 user command classes, we'll get a lot of errors.

lukaskuzmiak:
Honestly, I'm not sure - I've updated it so that the DEBUG msg includes itemName so in case someone wants to debug the same problem I was debugging they will see at least a debug message in case their device sends BASIC command class but this IF() ignores it.. better? :-)

cdjackson:
I'm reasonably sure this will cause multiple of your new statements to be printed when there's nothing wrong. The point here is that this code will be called multiple times for each device - ie for each item attached to the device. Only one (max!) can be attached to BASIC, but if there are many items attached to a device, then they will all print this message. So, we will cause a lot of these messages to be needlessly printed.

There should already be a message printed in the calling method that states that the event wasn't handled. This gets around the issue by checking that the event is handled. I really don't see a need to add this message, but maybe I'm missing something?

lukaskuzmiak:
Perhaps you're right this would be printed more times, I'm sorry, I don't the the project well, I just wanted to help users like me.

Look at the issue where we discussed this - https://github.com/openhab/openhab/issues/2595 - in my 2nd comment there are two DEBUG logs - both of this exact case .. there (IMHO) is nothing that would hint this BASIC command class message was ignored and not handled .. I only found it out by adding my own debug, recompiling, lots of trial and error until I realized this IF() where I added it was silently eating the event.

I think the more user-friendly the project is more people are gonna use it, this to me seems like impossible to configure for a regular user of openhab.

Perhaps you can suggest what's best in this case, I really lack the knowledge of the project to be able to tell where I can put this so it doesn't cause trouble in other cases.

lukaskuzmiak:
I reverted the commits that contained the change for the confusing if to at least get a merge on the product database..

please?

PR #4339, opened 2016-05-01
Improved modbus item update handling

Wolfgang1966:
This is a new version of pull request https://github.com/openhab/openhab/pull/4291 . Please refer to the discussion there.

It is now configurable per slave whether an item change should be published on every update or only when the item value really changes.

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
@@ -53,9 +53,9 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
     private static final String TCP_PREFIX = "tcp";
     private static final String SERIAL_PREFIX = "serial";
 
-    private static final String VALID_COFIG_KEYS = "connection|id|start|length|type|valuetype|rawdatamultiplier|writemultipleregisters";
+    private static final String VALID_CONFIG_KEYS = "connection|id|start|length|type|valuetype|rawdatamultiplier|writemultipleregisters|updateunchangeditems";
     private static final Pattern EXTRACT_MODBUS_CONFIG_PATTERN = Pattern.compile(
-            "^(" + TCP_PREFIX + "|" + UDP_PREFIX + "|" + SERIAL_PREFIX + "|)\\.(.*?)\\.(" + VALID_COFIG_KEYS + ")$");
+            "^(" + TCP_PREFIX + "|" + UDP_PREFIX + "|" + SERIAL_PREFIX + "|)\\.(.*?)\\.(" + VALID_CONFIG_KEYS + ")$");
 
     /** Stores instances of all the slaves defined in cfg file */
     private static Map<String, ModbusSlave> modbusSlaves = new ConcurrentHashMap<String, ModbusSlave>();
@@ -113,8 +113,9 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
                 continue;
             }
 
-            String slaveValueType = modbusSlaves.get(slaveName).getValueType();
-            double rawDataMultiplier = modbusSlaves.get(slaveName).getRawDataMultiplier();
+            ModbusSlave slave = modbusSlaves.get(slaveName);
+            String slaveValueType = slave.getValueType();
+            double rawDataMultiplier = slave.getRawDataMultiplier();
 
             State newState = extractStateFromRegisters(registers, config.readIndex, slaveValueType);
             /* receive data manipulation */
@@ -122,14 +123,14 @@ public class ModbusBinding extends AbstractActiveBinding<ModbusBindingProvider>i
                     .translateBoolean2State(!newState.equals(DecimalType.ZERO));
             if (!UnDefType.UNDEF.equals(newStateBoolean)) {
                 newState = newStateBoolean;
-            } else if ((rawDataMultiplier != 1) && (config.getItem() instanceof NumberItem)) {
+            } else if ((rawDataMultiplier != 1) && (config.getItemClass().isAssignableFrom(NumberItem.class))) {
                 double tmpValue = ((DecimalType) newState).doubleValue() * rawDataMultiplier;
                 newState = new DecimalType(String.valueOf(tmpValue));
             }
 
-            State currentState = config.getItemState();
-            if (!newState.equals(currentState)) {
+            if (slave.isUpdateUnchangedItems() || !newState.equals(config.getState())) {
                 eventPublisher.postUpdate(itemName, newState);
+                config.setState(newState);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


ssalonen:
Looks good to me.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusGenericBindingProvider.java
```diff
@@ -152,14 +153,19 @@ State getItemState() {
             return item.getState();
         }
 
+        public void setItemState(State state) {
+            ((GenericItem) item).setState(state);
+            ;
```
watou: Bindings should not directly set the state of items, or even hold onto references to items.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
```diff
@@ -53,7 +53,7 @@
     private static final String TCP_PREFIX = "tcp";
     private static final String SERIAL_PREFIX = "serial";
 
-    private static final String VALID_COFIG_KEYS = "connection|id|start|length|type|valuetype|rawdatamultiplier|writemultipleregisters";
+    private static final String VALID_COFIG_KEYS = "connection|id|start|length|type|valuetype|rawdatamultiplier|writemultipleregisters|updateunchangeditems";
```
watou: I know this PR doesn't introduce it, but could you fix the spelling to `VALID_CONFIG_KEYS`?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusGenericBindingProvider.java
Wolfgang1966: Please check the discussion for PR https://github.com/openhab/openhab/pull/4291 (the predecessor of this one). Basically the binding has to cache values somehow to identify changed values. Sami's first idea was to use the ItemRegistry to avoid local caching, but Kai Kreuzer commented in another binding that ItemRegistry should also not be used, so I reworked it to do it the way you see here.

There is another way to do it: Internally store the source value that comes from the modbus server (16 bit or 1 bit), before they are converted to items. If this is the way to go, I can rework the pull request (again) to avoid effects that might happen using the Item for this (although I have this configuration running for many weeks now without problems). But I would really like to clarify this before starting to write the third version of a PR to solve this issue. Sami, John: Would that be ok?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
Wolfgang1966: Yep! (I saw this line several times but did not realize the spelling error :-) ). I just wait for the result of the discussion about caching Items before working on this PR again ...

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusGenericBindingProvider.java
watou: Could the binding config just hold onto a `private State currentState` member, and do something like:
```
            if (slave.isUpdateUnchangedItems() || !newState.equals(config.getCurrentState())) {
                eventPublisher.postUpdate(itemName, newState);
                config.setCurrentState(newState);
            }
```

Wolfgang1966: Of course, that would be an easy solution. Just to be sure: As I learned that keeping a reference to an Item is a bad idea, is keeping a reference to a State really save, or could that also potentially cause problems?

watou: There is no problem holding onto a State object in your binding (in a special case like this where you can't be stateless).

Wolfgang1966: Done

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
Wolfgang1966: Done. Found one more spelling error (responce --> response) and corrected it also.

Wolfgang1966:
Changes done, please review

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
```diff
@@ -113,23 +112,24 @@ protected void internalUpdateItem(String slaveName, InputRegister[] registers, S
...
+            double rawDataMultiplier = slave.getRawDataMultiplier();
 
             State newState = extractStateFromRegisters(registers, config.readIndex, slaveValueType);
             /* receive data manipulation */
             State newStateBoolean = provider.getConfig(itemName)
                     .translateBoolean2State(!newState.equals(DecimalType.ZERO));
             if (!UnDefType.UNDEF.equals(newStateBoolean)) {
                 newState = newStateBoolean;
-            } else if ((rawDataMultiplier != 1) && (config.getItem() instanceof NumberItem)) {
+            } else if ((rawDataMultiplier != 1) && (config.getState() instanceof DecimalType)) {
```
watou: Is there any chance `config.getState()` could return `null` or an instance of `UnDefType`, causing this `if` block to be skipped, and thus `newState` not getting set?  Or at least, is there any chance of different behavior than when the item was being checked before this PR?  Might be OK; just wanting to confirm.

watou:
@Wolfgang1966 thanks!  Aside from the question above and of course the results of your testing 😄 , then the PR lgtm.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
watou: Would `config.getItemClass() == NumberItem.class` be more consistent with the previous logic than `config.getState() instanceof DecimalType`, or is that not a concern?

Wolfgang1966: Changed this to use Class.isAssignableFrom, to match the previous logic.

watou: Great!  Thanks @Wolfgang1966.  If your own testing confirms all good, then I think this PR is good to merge.

Wolfgang1966:
Regarding testing: The changes run fine on my system with three different pollers. Are there any testcases you want me to verify?

watou:
> Are there any testcases you want me to verify?

Nothing in particular!  Just want to avoid regressions is all.  😉   Any reason you can see to have others test a JAR before merge?

Wolfgang1966:
I don't expect any problems with this changes. However, having it tested by somebody else could always be helpful.

Sami, as you also provided changes to this binding: If you contact me directly under wolfgang [at] klimt [dot] de, I can provide the JAR file to you for testing.

ssalonen:
Tested the [attached jar](https://github.com/openhab/openhab/pull/4339#issuecomment-217122653). 

Works as expected with/without `updateunchangeditems`.

Test config (items): http://pastebin.com/gDuPJLZK
openhab.cfg: http://pastebin.com/iF8UM0BL
rule I used to test: http://pastebin.com/7Y7WxkVv

teichsta:
great, thanks guys!

PR #4347, opened 2016-05-02
Added configuration meta-data for powermax binding

lolodomo:
Signed-off-by: lolodomo lg.hc@free.fr

### DIFF
--- a/bundles/binding/org.openhab.binding.powermax/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.powermax/ESH-INF/binding/binding.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="powermax" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+	<name>Visonic PowerMax Binding</name>
+	<description>The Visonic PowerMax binding interfaces with Visonic PowerMax and PowerMaster alarm panel series.</description>
+	<author></author>
+
+	<config-description>
+		<parameter name="serialPort" type="text" required="false">
+			<label>Serial port</label>
+			<description>The serial port to use for connecting to the serial interface of the PowerMax alarm system e.g. COM1 for Windows and /dev/ttyS0 or /dev/ttyUSB0 for Linux.</description>
+		</parameter>
+		<parameter name="ip" type="text" required="false">
+			<context>network_address</context>
+			<label>IP address</label>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.powermax/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="powermax" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+	<name>Visonic PowerMax Binding</name>
+	<description>The Visonic PowerMax binding interfaces with Visonic PowerMax and PowerMaster alarm panel series.</description>
+	<author>lolodomo</author>
```
kaikreuzer: Would you be ok to put your real name here? If not, rather leave it as an empty string. Thanks!

lolodomo: Why not just keeping the same name as inside my source code and the name I use to sign-off my PR ?

kaikreuzer: Because we have the real names in all other bindings as well. And it is an info shown in the UI and not just in the source code.

lolodomo:
Author suppressed !

kaikreuzer:
No, you suppressed the whole XML tag - which is mandatory, so that the XML is now invalid.
Simply leave it empty (but don't remove it). Thanks!

lolodomo:
"author" tag restored.

lolodomo:
@Kai, testing with distro 284, I discovered that my file is rejected because the author is empty !
```
2016-05-06 12:21:08.800 [WARN ] [ig.xml.osgi.XmlDocumentBundleTracker] - The XML document '/ESH-INF/binding/binding.xml' in module 'org.openhab.binding.powermax' could not be parsed: The author must neither be null nor empty! : The author must neither be null nor empty!
---- Debugging information ----
message             : The author must neither be null nor empty!
cause-exception     : java.lang.IllegalArgumentException
cause-message       : The author must neither be null nor empty!
class               : org.eclipse.smarthome.core.binding.xml.internal.BindingInfoXmlResult
required-type       : org.eclipse.smarthome.core.binding.xml.internal.BindingInfoXmlResult
converter-type      : org.eclipse.smarthome.core.binding.xml.internal.BindingInfoConverter
path                : /binding
line number         : 78
version             : 1.4.7
-------------------------------
```
So, can this constraint be changed or can I just put "-" if lolodomo is not accepted ?

watou:
Just a thought:  come up with a generic author like "Contributor(s)" that can stand in for the author tag value when it can't be otherwise a proper name.

lolodomo:
I will probably put my proper name...

kaikreuzer:
I think we should make it optional. Could you please enter an issue for ESH?

PR #4351, opened 2016-05-03
Bugfix at binding EM. only Datapoint.CUMULATED items could be updated…

svenschreier:
… and multiple jitter updates for CUMULATED.


teichsta:
Thanks @svenschreier! Also cherry-picked onto 1.8

PR #4374, opened 2016-05-16
FHT: Implement support for valve FHT8v

tobiaswegner:
This patch allows openhab to control FHT8v valves directly without the need of a FHT80b controller. See also http://www.fhemwiki.de/wiki/FHT_8v_direkt_ansprechen (German page that describes doing so for/with the help of the FHEM project).


teichsta:
Thanks @tobiaswegner!

PR #3888, opened 2016-01-24
[weather] workaround for #3860

watou:
The Yahoo weather provider now requires a `woeid` location config parameter in openhab.cfg, since the [geo.placefinder service was discontinued by Yahoo](https://developer.yahoo.com/forums/#/discussion/7838/tahoo-geo-placefinder-does-not-work%3D%3D%3D%3D) in January, 2016.

@gerrieg, could you comment on this PR?

This PR addresses #3860.

PR #4397, opened 2016-05-22
Updated notes and attributes for the weather binding.

9037568:
Updated the notes for the weather binding and added the woeid configuration attribute.


watou:
I forgot to update openhab_default.cfg in #3888, oops!  Thanks for this, Chris.

9037568:
I would definitely want to see this in 1.8.3.

As to 1.9.0, it still supports the weather binding, right? Anyway, the changes here are only in comments, so can't cause a harmful effect...

teichsta:
Thanks guys! Also cherry-picked onto 1.8 …

PR #4403, opened 2016-05-25
[Weather] Add a default of 240 minutes for the updateInterval.

9037568:
Also added an early short circuit in the validation logic.

See [discussion](https://community.openhab.org/t/weather-binding-updateinterval-should-be-optional/10859/2).

### DIFF
--- a/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/LocationConfig.java
+++ b/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/LocationConfig.java
@@ -19,12 +19,14 @@ import org.openhab.binding.weather.internal.model.ProviderName;
  * @since 1.6.0
  */
 public class LocationConfig {
+    private static final int DEFAULT_UPDATE_INTERVAL = 240;
+
     private ProviderName providerName;
     private String language = "en";
     private Double latitude;
     private Double longitude;
     private String woeid;
-    private Integer updateInterval = 240;
+    private Integer updateInterval = DEFAULT_UPDATE_INTERVAL;
     private String locationId;
     private String name;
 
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/LocationConfig.java
```diff
@@ -145,12 +145,15 @@ public void setName(String name) {
      */
     public boolean isValid() {
         boolean valid = providerName != null && language != null && updateInterval != null && locationId != null;
+        if( !valid ) {
```
watou: Please use the formatter in the Eclipse IDE to comply, and also so this line doesn't get reformatted on the next commit.

watou:
Thanks Chris!

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/LocationConfig.java
```diff
@@ -24,7 +24,7 @@
     private Double latitude;
     private Double longitude;
     private String woeid;
-    private Integer updateInterval;
+    private Integer updateInterval = 240;
```
watou: could you add a `private static final` constant for documentation purposes?

9037568: something like this?

    private static final int DEFAULT_UPDATE_INTERVAL = 240;

Not sure I understand what this is for, but it's easy enough...

watou: Yes; that is what I mean.  Doing it this way has value to future developers who want to understand where the magic number 240 came from.  By explicitly describing it the way you propose, any doubt is removed as to what the significance of 240 is.  [This stackexchange](https://stackoverflow.com/questions/66066/what-is-the-best-way-to-implement-constants-in-java) is a bit all over the place, but the "consensus" there is that constants expressed this way help explain the intention.  Also, the compiler will effectively produce the same bytecode anyway.  You could make it `public` instead of `private` if it was needed from another class, but since it's `static` and `final`, it would be just as safe.    In general, it's better to write verbose code like this so it's easier for others to understand your intentions.

9037568: Ok, I think that should do it.

watou:
Thanks!

PR #4411, opened 2016-05-27
[knx] Allow the use of Network Address Translation for connections.

### LINKED ISSUE #3629: KNX binding via NAT
Hi,

I have just migrated my openhab server at home from a centos server to a docker container. Everything just works fine except the KNX binding. I found the reason for this. I see in the code (KNXConnection.java) that the KNXNetworLink method return a KNXNetworkLinkIP with by default the useNat param set to false (line 263). However, when you use openhab in a docker you have to use NAT (managed by docker itself).

I was able to trace the packet on the router and indeed the Siemens N148 IP interface return the packets to the internal docker ip address.

23:27:54.113919 IP HOST_RUNNING_DOCKER.44292 > SIEMENTS_N148_IP_INT.3671: UDP, length 26
23:27:54.114251 IP SIEMENTS_N148_IP_INT.3671 > INTERNAL_DOCKER_IP.44292: UDP, length 20

Would it be possible to add a setting to be able to set the useNat setting to true in the openhab configuration file ?

In the meanwhile I go through a eibd to bypass the issue. I am not a java developer but I assume it would not be too complicated to rebuild it with the param set to true.

9037568:
Set up a configuration setting for Network Address Translation.

Should resolve #3629

### DIFF
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
@@ -118,6 +118,13 @@ public class KNXConnection implements ManagedService {
      */
     private static int sMaxRefreshQueueEntries = 10000;
 
+    /**
+     * Determines whether Network Address Translation (NAT) will be used for IP connections.
+     *
+     * Default value is <code>false</code>.
+     */
+    private static boolean sUseNAT = false;
+
     /** listeners for connection/re-connection events */
     private static Set<KNXConnectionListener> sConnectionListeners = new HashSet<KNXConnectionListener>();
 
@@ -282,7 +289,7 @@ public class KNXConnection implements ManagedService {
             }
         }
 
-        return new KNXNetworkLinkIP(ipConnectionType, localEndPoint, new InetSocketAddress(ip, port), false,
+        return new KNXNetworkLinkIP(ipConnectionType, localEndPoint, new InetSocketAddress(ip, port), sUseNAT,
                 new TPSettings(new IndividualAddress(sLocalSourceAddr), true));
     }
 
@@ -431,6 +438,14 @@ public class KNXConnection implements ManagedService {
                             "Error when trying to read parameter 'scheduledExecutorServiceShutdownTimeout' from configuration. '{}' is not a number: using default.",
                             scheduledExecutorServiceShutdownTimeoutString);
                 }
+            }
+
+            String shouldUseNAT = (String) config.get("useNAT");
+            if (StringUtils.isNotBlank(shouldUseNAT)) {
+                sUseNAT = shouldUseNAT.equalsIgnoreCase("true");
+            }
+            else if (StringUtils.isBlank(shouldUseNAT)) {
+                sLogger.warn("Error when trying to read parameter 'useNAT' from configuration. Parameter was missing, so using default.");
             }
 
             if (sPC == null) {
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
@@ -441,12 +441,7 @@ public class KNXConnection implements ManagedService {
             }
 
             String shouldUseNAT = (String) config.get("useNAT");
-            if (StringUtils.isNotBlank(shouldUseNAT)) {
-                sUseNAT = shouldUseNAT.equalsIgnoreCase("true");
-            }
-            else if (StringUtils.isBlank(shouldUseNAT)) {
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
```diff
@@ -433,6 +440,14 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
                 }
             }
 
+            String shouldUseNAT = (String) config.get("useNAT");
+            if (StringUtils.isNotBlank(shouldUseNAT)) {
+                sUseNAT = shouldUseNAT.equalsIgnoreCase("true");
+            }
+            else if (StringUtils.isBlank(shouldUseNAT)) {
```
watou: if isNotBlank is false, isn't isBlank true?  If so, the `if` above would be unnecessary?

bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
```diff
@@ -433,6 +440,14 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
...
+                sLogger.warn("Error when trying to read parameter 'useNAT' from configuration. Parameter was missing, so using default.");
```
watou: If `useNat` is optional, this should be silent unless DEBUG level so the log doesn't have a WARN every time.

bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
9037568: I didn't grasp that isBlank is also true if the passed value is null. I'll rework this.

bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
9037568: This is the case I was trying to hadlie in the 'else if':
>    knx:useNAT=

Since the line is not commented out, this would mean the user intended to define a value for the 'useNAT' setting, but for whatever reason, forgot/failed to do so. Even though the setting is optional, it seems like there should be a warning issued for that.

Also, what do you think should be done for the case where the value specified for the setting is not as expected? 
eg:
>    useNAT=falsee

watou: The only part that matters, IMHO, is:
```java
String shouldUseNAT = (String) config.get("useNAT");
sUseNAT = StringUtils.isNotBlank(shouldUseNAT) && shouldUseNAT.equalsIgnoreCase("true");
```
and a proper commented section in openhab_default.cfg and knx.cfg to that effect.

9037568: Thanks for the pointer to knx.cfg. I'd have missed that one...

distribution/openhabhome/configurations/openhab_default.cfg
```diff
@@ -619,6 +619,10 @@ logging:pattern=%date{ISO8601} - %-25logger: %msg%n
 # ScheduledExecutorService. (optional, defaults to 5)
 #knx:scheduledExecutorServiceShutdownTimeoutString=
 
+# Use NAT (Network Address Translation)
+#  (optional; defaults to false)
+#knx:useNAT=
```
watou: At the risk of seeming too picky, if this showed the commented-out `true` setting, then it would not be left to assume that `true` was the precise thing to enter here when the user uncomments the line.

features/openhab-addons-external/src/main/resources/conf/knx.cfg
```diff
@@ -60,3 +60,7 @@
 # Seconds to wait for an orderly shutdown of the auto refresher's 
 # ScheduledExecutorService. (optional, defaults to 5)
 #scheduledExecutorServiceShutdownTimeoutString=
+
+# Use NAT (Network Address Translation)
+#  (optional; defaults to false)
+#useNAT=
```
watou: At the risk of seeming too picky, if this showed the commented-out `true` setting, then it would not be left to assume that `true` was the precise thing to enter here when the user uncomments the line.

9037568: Considering that's not done for any of the other boolean entries, and that
it's only done for 1 knx entry at all, you're right, that seemed
unnecessary.

In fact, some of the entries are missing an equals sign.

But I did it anyway.

HTH...

watou:
Thanks @9037568!  You may already be in process of updating [the wiki page](https://github.com/openhab/openhab/wiki/KNX-Binding#binding-configuration) referencing the new configuration parameter available in 1.9, but just to mention it....

PR #4416, opened 2016-05-29
DSC Alarm: Fix Premature Setting of Partition Arm Mode Item

RSStephens:
The binding would prematurely set the item `PARTITION_ARM_MODE` when issuing a partition arm command.  This would sometimes show the partition as being armed when it actually wasn't.  This fixes that situation by only setting the item when a verification message is received from the alarm system.  This issue was brought to light on the openHAB community website [here](https://community.openhab.org/t/dsc-partition-arm-mode/11012).


teichsta:
Thanks @RSStephens!

PR #4443, opened 2016-06-04
Use UPNP Eventing for Wemo switch devices

SilverGhostBS:
Removed the "polling" for state of the wemo switches devices, instead it leverages the UPNP eventing system, which allow us to save ressources by avoiding useless polling and allows us to get INSTANT update of state when an external event changes the state (manually pushing the switch for example).

### DIFF
--- a/bundles/binding/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/WemoBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/WemoBindingProvider.java
@@ -12,27 +12,26 @@ import org.openhab.binding.wemo.internal.WemoGenericBindingProvider.WemoChannelT
 import org.openhab.core.binding.BindingProvider;
 
 /**
- * @author Hans-Jörg Merk
+ * @author Hans-J�rg Merk
  * @since 1.6.0
  */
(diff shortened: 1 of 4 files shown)


hmerk:
Thanks for your contribution.
On first sight I saw that you completely removed the polling for service:basicevent.
This should be kept as a backup if subscribing to events fails.
Why did you not implement event subscription for the insight params ?
Can you implement this as well and reimplement the polling?

hmerk:
@SilverGhostBS any feedback to my comments?

Adrian-Sherwood:
Oh yes sorry, I forgot about this thread, I didn't add for insight params as I didn't have an insight to test the code, so I preferred not to do it.

Regarding the polling, yes it might be a good idea to reimplement, but I don't have time now :(
And I don't really see a scenario where the subscription will fail and not the polling... (I mean either the device is connected or it is not)
If the subscription fail, it should be retried anyway if I remember correctly.

PS: I have been using the module like this for a few months now, and it's working fine, unless there is a power outage and the subscriptions are "severed".

hmerk:
@Adrian-Sherwood Thanks for your feedback. In that case, I would suggest not to merge your PR, as it will break functionallity for the WeMo Insight switches.
Polling should also be available for all devices, cause event subscription does not always work reliable.

PR #4445, opened 2016-06-05
[TA-CMI Binding]: Added support for more data types, rudimentary support for the whole protocol

Wolfgang1966:
The TA CMI protocol supports 21 different analog value types according to the documentation. I found out that the documentation is not completely accurate, but by analyzing the network traffic I was able to figure out how to support at least kilowatt, kilowatthours and megawatthours. The code also adds generic dummy values for all other value types so that it is easy to add them as soon as somebody discovers them with their device.

Coments welcome; binding jar attached for testing
[org.openhab.binding.tacmi-1.9.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab/files/299288/org.openhab.binding.tacmi-1.9.0-SNAPSHOT.jar.zip)

### DIFF
--- a/bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
+++ b/bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
@@ -9,44 +9,85 @@
 package org.openhab.binding.tacmi.internal;
 
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This enum holds all the different measures and states available to be
- * retrieved by the TACmi binding Not the complete list yet.
- * 
+ * retrieved by the TACmi binding, including the scale factors needed to convert the received values to the real
+ * numbers.
+ *
  * @author Timo Wendt
+ * @author Wolfgang Klimt
  * @since 1.8.0
  */
 public enum TACmiMeasureType {
-    NONE(0),
-    TEMPERATURE(1),
-    SECONDS(4),
-    UNSUPPORTED(-1);
+    NONE(0, 0),
+    TEMPERATURE(1, 1),
+    UNKNOWN2(2, 0),
+    UNKNOWN3(3, 0),
+    SECONDS(4, 0),
+    UNKNOWN5(5, 0),
+    UNKNOWN6(6, 0),
+    UNKNOWN7(7, 0),
+    UNKNOWN8(8, 0),
+    UNKNOWN9(9, 0),
+    KILOWATT(10, 1),
+    KILOWATTHOURS(11, 1),
+    MEGAWATTHOURS(12, 0),
+    UNKNOWN13(13, 0),
+    UNKNOWN14(14, 0),
+    UNKNOWN15(15, 0),
+    UNKNOWN16(16, 0),
+    UNKNOWN17(17, 0),
+    UNKNOWN18(18, 0),
+    UNKNOWN19(19, 0),
+    UNKNOWN20(20, 0),
+    UNKNOWN21(21, 0),
 
-    int measure;
+    UNSUPPORTED(-1, 0);
 
-    private TACmiMeasureType(int measure) {
-        this.measure = measure;
+    private int typeval;
+    private int scale;
+
+    private static final Logger logger = LoggerFactory.getLogger(TACmiMeasureType.class);
+
+    private TACmiMeasureType(int typeval, int scale) {
+        this.typeval = typeval;
+        this.scale = scale;
     }
 
-    public int getMeasure() {
-        return measure;
+    public int getTypeValue() {
+        return typeval;
+    }
+
+    public int getScale() {
+        return scale;
     }
 
     /**
      * Return the measure type for the specified name.
-     * 
+     *
      * @param measure
      * @return
      */
     public static TACmiMeasureType fromString(String measure) {
         if (!StringUtils.isEmpty(measure)) {
-            for (TACmiMeasureType measureType : TACmiMeasureType.values()) {
-                if (measureType.toString().equalsIgnoreCase(measure)) {
-                    return measureType;
-                }
-            }
+            return TACmiMeasureType.valueOf(measure.toUpperCase());
         }
         throw new IllegalArgumentException("Invalid measure: " + measure);
+    }
+
+    /**
+     * Return measure type for a specific int value
+     */
+    public static TACmiMeasureType fromInt(int type) {
+        for (TACmiMeasureType mtype : TACmiMeasureType.values()) {
+            if (mtype.getTypeValue() == type) {
+                return mtype;
+            }
+        }
+        logger.warn("Received unexpected measure type {}", type);
+        return TACmiMeasureType.UNSUPPORTED;
     }
 }
--- a/bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
+++ b/bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
@@ -18,6 +18,7 @@ import org.slf4j.LoggerFactory;
  * This class handles analog values as used in the analog message.
  *
  * @author Timo Wendt
+ * @author Wolfgang Klimt
  * @since 1.8.0
  */
 public final class AnalogValue {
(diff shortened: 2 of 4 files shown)


bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
```diff
@@ -35,18 +67,26 @@ public int getMeasure() {
...
+
+    /**
+     * Return measure type for a specific int value
+     */
+    public static TACmiMeasureType fromInt(int type) {
+        if (type >= 0 && type <= 21) {
+            return measureTypes[type];
+        } else {
+            logger.warn("Received unexpected measure type {}", type);
+            return TACmiMeasureType.UNSUPPORTED;
```
watou: Would it not be better to leave this enum ignorant of how it's used, but instead just throw an exception like is done in `fromString` above, for consistency and also keeping concerns separated?  Then the user of this enum will catch the exception and log from that layer.

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
```diff
@@ -26,21 +26,24 @@
...
-                measureType = TACmiMeasureType.TEMPERATURE;
-                value = new BigDecimal(rawValue).movePointLeft(1);
-                break;
-            case 4:
-                measureType = TACmiMeasureType.SECONDS;
-                value = new BigDecimal(rawValue);
-                break;
-            default:
-                measureType = TACmiMeasureType.UNSUPPORTED;
+        if (type < 1 || type > 21) {
```
watou: 1 and 21 should be static final constants from the TACmiMeasureType enum.

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
```diff
@@ -26,21 +26,24 @@
...
+            measureType = TACmiMeasureType.UNSUPPORTED;
+            value = new BigDecimal(rawValue);
+            logger.warn("Got unsupported measure type {}, value is {}", type, rawValue);
+        } else {
+            measureType = TACmiMeasureType.fromInt(type);
+            value = new BigDecimal(rawValue).movePointLeft(factors[type]);
+            logger.debug("Got measure: type {}, raw value {}, converted: {}, {}", type, rawValue, measureType.name(),
+                    value);
         }
```
watou: Better yet, don't range check here at all, just try/catch and let the TACmiMeasureType enum reject unsupported types by throwing the IAE.

watou:
Hi @Wolfgang1966, since the documentation isn't accurate about which numbers go with which types, perhaps remove the UKNOWNx entries from the enum, and just throw an IllegalArgumentException in fromInt for those you aren't sure of, and let the caller log the fact?

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
Wolfgang1966: fromString is used only to convert item definitions, so an unknown value there is a sign of an error in the item definition. fromInt is used for data read from the CMI device, see my general comment below on that.

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/message/AnalogValue.java
Wolfgang1966: Correct, I will change this.

Wolfgang1966:
Hello @watou ,

thanks for your comments. Let me describe my basic thoughts behind the changes:

My idea was: Whatever type comes back, finally there is an integer value that is sent by the CMI module, so one can use it for an item. As the user configures in CMI which value is sent, we may not know about type and scale, but nevertheless it can be used in a rule and, if the scale is not correct, use a rule to adapt it. Thats why I treat them as unknown value type and return the value. If I get reports from users of TA controllers that see such values and can provide the type information, I will be able to further work on the binding using the information I get from them. They may not be Java developers and therefore not able to handle the changes by themselves.

Strictly spoken the type information may safely be ignored  as long as one is only reading values and not writing them. Using this protocol for writing is difficult because the protocol enforces to write 4 values at once, as the original author already stated. So I assume most users of the device will never do this at all. I just assume that for writing a value one has to provide the correct value type, so the binding will allow that.

watou:
Thanks for the explanation, @Wolfgang1966; I understand your point now.  But could you remove the redundant `measureTypes` static member, since it's a duplicate of what the enum does, and also encapsulate the `factors` array into the enum, so the enum itself holds these things (and a new method to handle the factors)?  It would be easier for readability and maintainability if they lived neatly in the enum.

Wolfgang1966:
regarding measureTypes array: I can remove it, but then the fromInt method will become more expensive, because I will have to iterate through the enums. Unfortunately java.lang.Enum does not provide a way to convert an Int to the corresponding Enum. If the costs don't hurt you then I will change this.

Moving the factors to the measure types will be done with the next update.

watou:
> the fromInt method will become more expensive

My opinion is that looping over the enum is worth it, for the sake of brevity in the code and ease of future maintenance with lessened risk of mysterious regressions.  Thanks for your updates!

Wolfgang1966:
Hi @watou, please have a look at my changes.

All users of the TA-CMI binding, please find the new binding here for testing:

[org.openhab.binding.tacmi-1.9.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab/files/310152/org.openhab.binding.tacmi-1.9.0-SNAPSHOT.jar.zip)

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
```diff
@@ -9,44 +9,84 @@
...
         }
         throw new IllegalArgumentException("Invalid measure: " + measure);
     }
+
+    /**
+     * Return measure type for a specific int value
+     */
+    public static TACmiMeasureType fromInt(int type) {
+        for (TACmiMeasureType mtype : TACmiMeasureType.values()) {
+            if (mtype.ordinal() == type) {
```
watou: Would it not be better to use `getMeasure()` instead of `ordinal()` because you don't want to depend on the measure types being the same as the ordinal index of their enum constants?

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
```diff
@@ -9,44 +9,84 @@
...
 
 /**
  * This enum holds all the different measures and states available to be
- * retrieved by the TACmi binding Not the complete list yet.
- * 
- * @author Timo Wendt
+ * retrieved by the TACmi binding, including the scale factors needed to convert the received values to the real
+ * numbers.
+ *
+ * @author Timo Wendt, Wolfgang Klimt
```
watou: It's better to add a new `@author` line right after Timo's.

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
Wolfgang1966: Done

bundles/binding/org.openhab.binding.tacmi/src/main/java/org/openhab/binding/tacmi/internal/TACmiMeasureType.java
Wolfgang1966: Done

Wolfgang1966:
Hello @watou ,

updated the source according to your comments. Made a few more small corrections.

New binding jar is here:
[org.openhab.binding.tacmi-1.9.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab/files/310393/org.openhab.binding.tacmi-1.9.0-SNAPSHOT.jar.zip)

watou:
Thank you, @Wolfgang1966.  Changes look good, but please let me know when you feel this PR is tested well enough to merge.

Wolfgang1966:
Hello @watou 

I did a few more tests today and did not find any problems, so I consider this PR stable.

Thanks a lot for your comments and support!

PR #4446, opened 2016-06-05
[mios] openHAB2 delayed startup/reload improvements under compat layer.

mrguessed:
Under the compat1x layer of OH 2, the MiOS binding was seeing a different startup sequence than when run natively under OH 1.x.

This change adds a delay to the startup processing to allow Item loading to stabilize before attempting to load data from the MiOS Unit, to avoid the sequencing issues.

Community discussion:
* https://community.openhab.org/t/openhab2-new-issue-with-the-startup-process-mios-binding/8320

Signed-off-by: Mark Clark <mr.guessed@gmail.com> (github: mrguessed)


watou:
Thanks @mrguessed!

PR #4447, opened 2016-06-05
Adding feature for Nikobus.

crnjan:
Have been using this binding in OH2 and seems to work fine.


watou:
Thank you @crnjan!

watou:
@kaikreuzer : what causes the documentation site to be rebuilt after a commit like this one?  Also, how can my local build take into account the new 1.x binding?

kaikreuzer:
I am currently doing this manually, but I plan to setup a nightly Jenkins build to do so.
All you need to do locally is to call `update.sh` - this will pull the changes from the submodule repos and re-generate and copy all required files.

kaikreuzer:
Just pushed an update, so that Nikobus now shows up as well.

watou:
Thanks, Kai.

Issue #4337, opened 2016-05-01
[WeatherBinding] Fails silently at startup


9037568:
## Expected Behavior
If the weather binding experiences a failure at startup, it should be logged and have an appropriate severity.

## Current Behavior
When the weather binding fails, it does so silently, producing no log output.

I've poked at this a bit, but haven't been able to determine the cause, although it *looks* like no request is ever being made to the weather service.

Debug log can be found [here](https://drive.google.com/file/d/0B8vYheWfObP9SUp4X0hRX1hTeGc/view?usp=sharing)

Weather settings from openhab.cfg. (the api key is left blank deliberately. The same behaviour occurs when specifying an invalid key):
```
#weather:apikey.ForecastIo=
#weather:apikey.OpenWeatherMap=
#weather:apikey.WorldWeatherOnline=
weather:apikey.Wunderground=
#weather:apikey.Hamweather=
#weather:apikey2.Hamweather=

# location configuration, you can specify multiple locations
weather:location.home.name=Home
weather:location.home.latitude=37.0
weather:location.home.longitude=-84.0
weather:location.home.provider=Wunderground
weather:location.home.language=en
weather:location.home.updateInterval=15
```

## Steps to Reproduce (for bugs)
1. None needed. Just configure as listed above.

## Your Environment

* OH Version: 1.8.2 with 1.8.2 bindings
* Java version: 
```
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
```
* Operating System: Windows 10 64-bit

9037568:
Further digging indicates that the weather binding is throwing an exception in WeatherConfig.java ([line 99 in the 1.8 branch](https://github.com/openhab/openhab/blob/1.8/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java)), but that exception never shows up in the log output.

Should this exception automatically appear in the log output, or does the code have to explicitly call logger.error... ?

```
09:48:43.922 [WARN ] [o.o.b.w.i.common.WeatherConfig:79   ] - Weather api key 'apikey.Wunderground' had no value.
09:48:43.922 [WARN ] [.o.b.w.i.common.ProviderConfig:44   ] - setApiKey(): received null value for provider 'WUNDERGROUND'
[...]

PR #4458, opened 2016-06-11
[WeatherBinding] Add log messages for invalid configurations

9037568:
In order to prevent silent failures of the weather binding (see #4337), record appropriate log messages.

Also prevent duplicate warnings on binding change.

### DIFF
--- a/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/bus/WeatherBinding.java
+++ b/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/bus/WeatherBinding.java
@@ -44,7 +44,7 @@ public class WeatherBinding extends AbstractBinding<WeatherBindingProvider>imple
     }
 
     /**
-     * Set providers in WeatherContext and generates metadata from the weater
+     * Sets providers in WeatherContext and generates metadata from the weather
      * model annotations.
      */
     @Override
@@ -71,15 +71,22 @@ public class WeatherBinding extends AbstractBinding<WeatherBindingProvider>imple
      */
     @Override
     public void updated(Dictionary<String, ?> config) throws ConfigurationException {
-        if (config != null) {
-            context.getJobScheduler().stop();
+        if (config == null) {
+            logger.warn("Unable to find any configuration settings for weather binding. Check openhab.cfg.");
+            throw new ConfigurationException("weather", "Unable to find any configuration settings for weather binding. Check openhab.cfg.");
+        }
 
-            context.getConfig().parse(config);
-            context.getConfig().dump();
+        context.getJobScheduler().stop();
 
-            if (context.getConfig().isValid()) {
-                context.getJobScheduler().restart();
-            }
+        context.getConfig().parse(config);
+        context.getConfig().dump();
+
+        if (context.getConfig().isValid()) {
+            context.getJobScheduler().restart();
+        }
+        else {
+            logger.warn("Unable to restart weather job because weather configuration is not valid. Check openhab.cfg.");
+            throw new ConfigurationException("weather", "Unable to restart weather job because weather configuration is not valid. Check openhab.cfg.");
         }
     }
 
@@ -96,8 +103,15 @@ public class WeatherBinding extends AbstractBinding<WeatherBindingProvider>imple
      */
     @Override
     public void allBindingsChanged(BindingProvider provider) {
+        if (!context.getConfig().finishedParsing() ) {
+            return;
+        }
+
         if (context.getConfig().isValid()) {
             context.getJobScheduler().restart();
+        }
+        else {
+            logger.warn("All bindings changed, but unable to restart weather job because weather configuration is not valid. Check openhab.cfg.");
         }
     }
 
@@ -106,10 +120,18 @@ public class WeatherBinding extends AbstractBinding<WeatherBindingProvider>imple
      */
     @Override
     public void bindingChanged(BindingProvider provider, String itemName) {
+        if (!context.getConfig().finishedParsing() ) {
+            return;
+        }
+
         if (context.getConfig().isValid()) {
             if (provider instanceof WeatherBindingProvider) {
                 context.getJobScheduler().restart();
             }
+        }
+        else {
+            logger.debug("Binding for item '{}' changed, but unable to restart weather job "
+                        +" because weather configuration is not valid. Check openhab.cfg.", itemName);
         }
         super.bindingChanged(provider, itemName);
     }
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/bus/WeatherBinding.java
```diff
@@ -71,15 +71,22 @@ public void deactivate() {
      */
     @Override
     public void updated(Dictionary<String, ?> config) throws ConfigurationException {
-        if (config != null) {
-            context.getJobScheduler().stop();
+        if (config == null) {
+            logger.warn("Unable to find any configuration settings for weather binding. Check openhab.cfg.");
+            throw new ConfigurationException("weather", "Unable to find any configuration settings for weather binding. Check openhab.cfg.");
+        }
```
watou: Are you certain that there aren't expected situations when `updated` is called with a `null` config?  A thorough exercise under OH 1.8.3 and OH 2 might reveal if this is correct; perhaps you've already done that?

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
```diff
@@ -167,8 +186,7 @@ private void parseApiKey(String key, String value) throws ConfigurationException
         } else if (StringUtils.equalsIgnoreCase(keyId, "apikey2")) {
             pConfig.setApiKey2(value);
         } else {
-            throw new ConfigurationException("weather",
-                    "Unknown configuration key '" + key + "', please check your openhab.cfg!");
+            logger.debug("Unknown configuration key '{}'. Check openhab.cfg.", key);
```
watou: why debug and not warn?

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/bus/WeatherBinding.java
9037568: You're thinking we shouldn't throw an exception here, but just issue a warning? I'd be ok with that.

I can't think of any situations when <code>updated</code> would be called with a <code>null</code> config, though. I've run through quite a number of test scenarios without seeing any bad behaviour.

At any rate, if you prefer to not throw an exception here, I'll change it.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
9037568: In this case, someone's added a key that's not recognized by the system, like <code>myInvalidKey=7</code>.

This shouldn't cause any problems for the binding, however, as it only examines known keys. So it's merely a nicety for the user. "Hey, we found this setting we don't recognize. It's no problem for us since we're ignoring it completely. Why's it here again?"

watou: The reason I prefer warn over debug in this case is that a user might mistype the configuration key.  So with debug-only logging of the unrecognized key, the user will think they set the configuration correctly, but the default value is being used instead.  By making it warn-level, I think it's more likely the user will see their mistake.

All that said, it's a very minor issue, so if you're not persuaded, then leave it as is.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/bus/WeatherBinding.java
9037568: Changed debug to warn.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
9037568: Changed debug to warn.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
```diff
@@ -61,62 +61,77 @@
...
-            while (keys.hasMoreElements()) {
-                String key = keys.nextElement();
-
-                String value = StringUtils.trimToNull((String) properties.get(key));
-                if (StringUtils.startsWithIgnoreCase(key, "apikey")) {
-                    parseApiKey(key, value);
-                } else if (StringUtils.startsWithIgnoreCase(key, "location")) {
-                    parseLocation(key, value);
-                }
+        if (properties == null ) {
```
watou: Hi Chris, please format your changes with the Eclipse IDE formatter, so they won't be reformatted on the next commit.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
```diff
@@ -61,62 +61,77 @@
...
             }
+        }
 
-            valid = locationConfigs.size() > 0;
+        // check all ProviderConfigs
+        for (ProviderConfig pc : providerConfigs.values()) {
+            if (!pc.isValid()) {
+                parseCompleted = true;
+                logger.warn("Invalid apikey config for provider '{}'. Check openhab.cfg.", pc.getProviderName());
+                throw new ConfigurationException("weather", "Invalid apikey config for provider '"
```
watou: Is the description in this exception reflective of the `!pc.isValid()` test?

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
```diff
@@ -225,6 +245,13 @@ public boolean isValid() {
     }
 
     /**
+     *
```
watou: Comment?

watou:
Thanks, Chris.  Please see my last comments.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
9037568: As far as I can see, yes. I haven't changed it, though, so it should still be as accurate as it has ever been.

bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/common/WeatherConfig.java
9037568: Odd, I don't recall adding that one. Fixed.

watou:
Thanks Chris!

Issue #4459, opened 2016-06-12
HTTP Binding : Issue when URL has spaces being 20%


TommySharpNZ:
It appears that if a URL has the special characters of 20% to represent a space then the HTTP Binding fails.

The use case is the Google Maps Distance Matrix API that allows you to query the estimated travel time between two addresses.

This is my Item
`Number TTW_Sonay "Sonay TTW [%s]" (GroupOutdoors) { http="<[https://maps.googleapis.com/maps/api/distancematrix/json?origins=OriginAddressHere&destinations=DestinationAddressHere&key=GoogleAPIKeyHere&language=en&departure_time=now:300000:JS(ttw_trafficTimesNumber.js)]"}`

 In my case the origin and distination addresses have spaces in them represented by 20%

This causes the error below.

```
2016-06-11 16:10:37.053 [ERROR] [.service.AbstractActiveService] - Error while executing background thread HTTP Refresh Service
java.util.UnknownFormatConversionException: Conversion = 'R'
at java.util.Formatter$FormatSpecifier.conversion(Unknown Source) ~[na:1.8.0_60]
at java.util.Formatter$FormatSpecifier.(Unknown Source) ~[na:1.8.0_60]
at java.util.Formatter.parse(Unknown Source) ~[na:1.8.0_60]
at java.util.Formatter.format(Unknown Source) ~[na:1.8.0_60]
at java.util.Formatter.format(Unknown Source) ~[na:1.8.0_60]
at java.lang.String.format(Unknown Source) ~[na:1.8.0_60]
at org.openhab.binding.http.internal.HttpBinding.execute(HttpBinding.java:136) ~[na:na]
at org.openhab.core.binding.AbstractActiveBinding$BindingActiveService.execute(AbstractActiveBinding.java:156) ~[na:na]
at org.openhab.core.service.AbstractActiveService$RefreshThread.run(AbstractActiveService.java:173) ~[na:na]
```

With assistance from watou it appears the issue is with this line?
https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/HttpBinding.java#L144
##

9037568:
Should be fixed with #4460.

PR #4460, opened 2016-06-12
[http] allow String.format to be optional to avoid conflicts

watou:
For user-specified URLs that happen to contain sprintf-style formatting strings, the binding would attempt to substitute the current time and/or item state.  But this may not be what the user wants, instead preferring to use the URL as given.  A conflict can easily happen when the URL contains escaped characters, like "http://sample/?This%20string%20is$5%20a%problem," and so the user would want to suppress the sprintf-style formatting.

This PR adds an optional configuration property called `format` that is `true` by default, for backward compatibility.

See #4459 and [the forum](https://community.openhab.org/t/help-with-transforming-google-maps-distance-matrix-api-data/11390).


9037568:
LGTM.

PR #4474, opened 2016-06-17
snmp binding: added remote port support via url and OUT_BINDING_PATTERN_VERSION bugfix

a-glazachev:
[+] snmp binding: added remote port support via url ("ip/port" with specified port or just "ip" with default 161 port). I use "/" symbol instead ":" because do not want to change schemes `(a:b:c:...)`
[-] snmp binding: fixed version/url parameters order in OUT_BINDING_PATTERN_VERSION. Second parameter was version, but the code expects address

For now, you can configure only listening port. Remote port hardcoded as 161.
Current patch allows manual remote port setting.


a-glazachev:
For now, you can configure only listening port. Remote port hardcoded as 161.
Current patch allows manual remote port setting.

watou:
Thank you, @a-glazachev.   Would you please update [the wiki page](https://github.com/openhab/openhab/wiki/Snmp-Binding) to indicate that in 1.9 of the binding, the ability to specify the remote port has been added, much as you described above?

a-glazachev:
Good day, Watou.

I've added the NOTE about "V1.9" and the port description in "Generic Item Binding Configuration" section.

watou:
That's perfect, @a-glazachev. Thank you.

PR #4502, opened 2016-06-27
[mochadx10] Add binding to OH2 distro

smar000:
Add mochadX10 binding to OH2 distro


kaikreuzer:
Many thanks, looks good!

Issue #4495, opened 2016-06-23
[MySQL] Storing of items with MySQL Server time problematic


herby3:
The time value of the items stored by the MySQL binding should accord to the time of the system Openhab is running. Currently on inserting items the MySQL time is stored (MySQL function NOW()). If you store items in a MySQL database on another server and the clocks differ between the MySQL server and the Openhab Server, several persistence querys (for example item.historicState(now.minusMinutes(1)) leads to undesirable results.

My suggestion is to use the Java time in the INSERT statement of items to avoid these problems.

watou:
Closed via #4503

PR #4503, opened 2016-06-28
Changed timestamp of item insert from MySql server time to local Java…

herby3:
see #4495

### DIFF
--- a/bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java
+++ b/bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java
@@ -14,6 +14,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
@@ -102,6 +103,9 @@ public class MysqlPersistenceService implements QueryablePersistenceService {
     private int errReconnectThreshold = 0;
 
     private int waitTimeout = -1;
+    
+    // Time used for persisting items, False: MySQL Server time (default), True: openHAB Server time
+    private boolean localtime = false;
 
     private Connection connection = null;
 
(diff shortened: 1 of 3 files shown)


bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java
```diff
@@ -421,15 +421,20 @@ public void store(Item item, String alias) {
         String sqlCmd = null;
         PreparedStatement statement = null;
         try {
+            // Get current timestamp
+            long timeNow = Calendar.getInstance().getTimeInMillis();
+    	    java.sql.Timestamp timestamp = new java.sql.Timestamp(timeNow);
```
watou: Indentation (consider using the Eclipse IDE to auto-format code).  Also, please import java.sql.Timestamp and use here without package qualifier.

watou:
Thanks for this.  

What are the potential effects on existing users of MySQL persistence?  If MySQL server and openHAB server times are very different, there will be a time leap in persisted item states in rows inserted after this change is running, which may be unwanted.  Perhaps a compatibility flag for openhab.cfg / mysql.cfg, and the code would use the time source the user specified?

9037568:
I like the idea of making this user configurable.  Beneficial for those who
want to continue to use the server time...

watou:
Maybe a `#mysql:localtime=true` setting which defaults to `false` if not specified, so that an upgrade does not surprisingly change how times are stored?

herby3:
I added flag 'localtime' to configuration and it defaults to 'false'.

watou:
Excellent, thanks @herby3.  Could you also update [mysql.cfg](https://github.com/openhab/openhab/blob/master/features/openhab-addons-external/src/main/resources/conf/mysql.cfg) and also [the wiki page](https://github.com/openhab/openhab/wiki/MySQL-Persistence) (noting in the wiki page that this is a 1.9 enhancement)?  With that, and your testing that it works as intended, we should be ready to merge this PR.

features/openhab-addons-external/src/main/resources/conf/mysql.cfg
```diff
@@ -11,4 +11,9 @@
...
 
 # the connection timeout (in seconds)
-#waitTimeout=
\ No newline at end of file
+#waitTimeout=
+
+# Use MySQL Server time to store item values (=false) or use openHAB Server time (=true).
+# For new installations, its recommend to set "mysql:localtime=true".
+# (optional, defaults to false)
+#mysql:localtime=true
```
watou: The `mysql:` part is not used in `mysql.cfg`.

features/openhab-addons-external/src/main/resources/conf/mysql.cfg
```diff
@@ -11,4 +11,9 @@
 #reconnectCnt=
 
 # the connection timeout (in seconds)
-#waitTimeout=
\ No newline at end of file
+#waitTimeout=
+
+# Use MySQL Server time to store item values (=false) or use openHAB Server time (=true).
+# For new installations, its recommend to set "mysql:localtime=true".
```
watou: also remove `mysql:` here as well

herby3:
Sry, done!

watou:
Thanks @herby3!

PR #4516, opened 2016-07-04
RRD4j - remove dependency on TimerTasks

kgoderis:
I can not fix the formatter issue, I give up! The changed lines are (in the updated file):

70, 80, 155-167, 595-623

Signed-off-by: Karel Goderis <karel.goderis@me.com>


kaikreuzer:
Ok, let's merge it then!

PR #4520, opened 2016-07-06
Add a Data value selector that doesn't include RFXCOM headers

Jamstah:
The Undecoded implementation only has one value selector, RawData. This updates the item state with the full message.

I'd like to introduce a "Data" value selector, which drops the header from the message. The bytes of the header are:
1. Length - not required as we get a byte array of the right length anyway.
2. Type (Undecoded) - Not required as we know the item is undecoded from the items file.
3. Subtype - This could potentially be of use to someone, but it wasn't in my case, and RawData is still an option.
4. Sequence number, this changes consistently, unlikely to be required.

I have devices that just spit out their ID as a message and nothing else, by using the Data selector instead of the RawData selector, I get a consistent output (without sequence), which means my rules file can detect the state update really easily. With RawData, the value would change every time so would need more work.

### DIFF
--- a/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/RFXComValueSelector.java
+++ b/bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/RFXComValueSelector.java
@@ -29,6 +29,7 @@ import org.openhab.core.library.items.SwitchItem;
 public enum RFXComValueSelector {
 
     RAW_DATA("RawData", StringItem.class),
+    DATA("Data", StringItem.class),
     SHUTTER("Shutter", RollershutterItem.class),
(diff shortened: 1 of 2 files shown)


Jamstah:
Assuming this is pulled into openhab, I next plan to upgrade my system to openhab2 and make the same changes. The code for undecoded messages hasn't reached openhab2-addons yet, so I'll port that in as well.

watou:
Thank you for this.  Would you also please update the [wiki](https://github.com/openhab/openhab/wiki/RFXCOM-Binding#specifics-for-packettypes-and-subtypes--packettypesubtype) to document the new "Data" value selector, noting that it will be available in 1.9 of the binding?

Jamstah:
Thanks. wiki updated.

watou:
Thank you very much, @Jamstah!

PR #4522, opened 2016-07-06
[samsungac] Allow port to be specified in cfg other than default 2878

watou:
Also replaced a few writes to System.out with logger calls.

Signed-off-by: John Cocula john@cocula.com


watou:
[Forum discussion](https://community.openhab.org/t/samung-ac-port-changed-to-8888/12189)

watou:
Never got complete feedback, but this change fixes known logging problems and provides the option to set a non-default port.

PR #4549, opened 2016-07-20
[MochadX10] Introduce delay before trying to reconnect

smar000:
Wait 20 seconds before attempting to reconnect as otherwise the log gets full with thousands of reconnect attempt messages when mochad server is down.

### DIFF
--- a/bundles/binding/org.openhab.binding.mochadx10/src/main/java/org/openhab/binding/mochadx10/internal/MochadX10Binding.java
+++ b/bundles/binding/org.openhab.binding.mochadx10/src/main/java/org/openhab/binding/mochadx10/internal/MochadX10Binding.java
@@ -49,7 +49,7 @@ import org.slf4j.LoggerFactory;
  * @author Jack Sleuters
  * @since 1.7.0
  */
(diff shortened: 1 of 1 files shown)


watou:
Thank you, @smar000.  @Jakey69 and others who use this binding, do you have any thoughts about adding the 20-second delay in this PR?  Would a dynamic throttle (wait increasingly longer times after consecutive failures to connect up to some threshold) address the problem without having to wait too long, or is this approach likely to be acceptable to everyone?

smar000:
@watou - I thought about dynamically increasing delays but then decided that a fixed delay was probably a better compromise as it would potentially allow a faster recovery overall. 

Mochad server frequently stops responding for no, as yet, known reason. A number of users, myself included, end up having `cron `jobs to restart the Mochad service. Or it may be down for legitimate reasons (e.g. upgrade/migration etc). With a 20 second delay, we get only 3 failure entries a minute in the log file. In an ideal world, once the Mochad server is down, the binding should go into an idle state and wait to be notified that the server is properly back up. However, this seemed much more effort than introducing the delay!

watou:
Thank you @smar000!

Issue #4546, opened 2016-07-17
[MySQL] Frequent state changes are not added to database.


kjasdal:
## Expected Behavior
When using a persistence service, with strategy = everyChange, all changes of state should be stored in the database - regardless of how quickly the state changes. 

## Current Behavior
Frequent changes of state (changes within the time period of a second) are not persisted.

## Possible Solution
As far as I can tell, the problem is related to the fact that the Time column in the database (table) only supports resolution down to seconds, and at the same time it is the PRIMARY KEY (that needs to be unique for each entry). This probably leads to entries on the same second not being accepted.

MySQL 5.7 introduces support for fractional seconds (down to microseconds): 

<http://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html>

Modifying the following line (311/312) in MysqlPersistenceService.java should possibly be enough: 

sqlCmd = new String(
                "CREATE TABLE " + tableName + " (Time DATETIME, Value " + mysqlType + ", PRIMARY KEY(Time));");

DATETIME => DATETIME(3)

An alternative solution (for MySQL version prior to 5.7) would be to change the Time column to some large general number type (e.g. BIGINT) and then simply use milliseconds as time stamp.

## Steps to Reproduce (for bugs)
1. Set up MySQL persistence for an item using strategy = everyChange.
2. Write and execute a test rule that changes the state of the item several times as quickly as possible, e.g.

item.postUpdate(0)
item.postUpdate(2)
item.postUpdate(3)
item.postUpdate(4)

3. Browse the openHAB events.log file to see that all state changes are "registered" by openHAB event bus.
4. Query the MySQL database (e.g. using MySQL Workbench) and verify that some of the state changes are missing from the database.

## Context
Any use of MySQL as persistence service.

## Your Environment
* Version used: openHAB 1.8.3
* Environment name and version (e.g. Chrome 39, node.js 5.4, Java 8, ...): OpenJDK 1.8.0_91
* Operating System and version (desktop or mobile, Windows 10, Raspbian Jessie, ...): Ubuntu 14.04
* MySQL 5.7.13

watou:
[...]

PR #4550, opened 2016-07-22
JDBC now with high precision by fractional seconds in general.

lewie:
mysql can use fractal seconds since 5.6.

added to configure store and read date format: 
- #jdbc:sqltype.TABLEPRIMARYKEY   =   TIMESTAMP
- #jdbc:sqltype.TABLEPRIMARYVALUE =   NOW()

needed to update YANK and HicariCP 

now we can differentiate between database and driver versions, if
needed.

Its for #4546


lewie:
All databases are tested, and working correct!
All (tested) and mysql (since 5.6) and mariadbl (since 5.1) now reads and writes fractal seconds.

bundles/persistence/org.openhab.persistence.jdbc/.classpath
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry exported="true" kind="lib" path="lib/yank-3.1.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/yank-3.2.0.jar" sourcepath="C:/Users/Helmut/.m2/repository/org/knowm/yank/3.2.0/yank-3.2.0-sources.jar"/>
```
watou: please remove sourcepath parameter

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -46,17 +46,27 @@
...
+import org.openhab.persistence.jdbc.utils.DbMetaData;
 import org.openhab.persistence.jdbc.utils.StringUtilsExt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Default Database Configuration class.
+ *
+ * @author Helmut Lehmeyer
+ * @since 1.8.0
```
watou: @since 1.9.0

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -77,6 +87,42 @@ public JdbcBaseDAO() {
         initSqlQueries();
     }
 
+    /**
+     * ## Get high precision by fractal seconds, examples ##
+     *
+     * mysql > 5.5 + mariadb > 5.2:
+     * DROP TABLE FractionalSeconds;
+     * CREATE TABLE FractionalSeconds (time TIMESTAMP(3), value TIMESTAMP(3));
+     * INSERT INTO FractionalSeconds (time, value) VALUES( NOW(3), '1999-01-09 20:11:11.126' );
```
watou: According to [this](https://dev.mysql.com/doc/refman/5.6/en/date-and-time-functions.html#function_now), "As of MySQL 5.6.4, if the fsp argument is given to specify a fractional seconds precision from 0 to 6, the return value includes a fractional seconds part of that many digits. Before 5.6.4, any argument is ignored."  Is it incorrect?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -97,8 +143,8 @@ private void initSqlQueries() {
         SQL_DELETE_ITEMS_ENTRY = "DELETE FROM items WHERE ItemName=#itemname#";
         SQL_GET_ITEMID_TABLE_NAMES = "SELECT itemid, itemname FROM #itemsManageTable#";
         SQL_GET_ITEM_TABLES = "SELECT table_name FROM information_schema.tables WHERE table_type='BASE TABLE' AND table_schema=#jdbcUriDatabaseName# AND NOT table_name=#itemsManageTable#";
-        SQL_CREATE_ITEM_TABLE = "CREATE TABLE IF NOT EXISTS #tableName# (time TIMESTAMP NOT NULL, value #dbType#, PRIMARY KEY(time))";
-        SQL_INSERT_ITEM_VALUE = "INSERT INTO #tableName# (TIME, VALUE) VALUES( NOW(), ? ) ON DUPLICATE KEY UPDATE VALUE= ?";
+        SQL_CREATE_ITEM_TABLE = "CREATE TABLE IF NOT EXISTS #tableName# (time #TABLEPRIMARYKEY# NOT NULL, value #dbType#, PRIMARY KEY(time))";
```
watou: why #TABLEPRIMARYKEY# (all caps) vs. #tablename# lowercase?  (Same question for line below.)

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -109,13 +155,15 @@ private void initSqlTypes() {
         sqlTypes.put("CALLITEM", "VARCHAR(200)");
         sqlTypes.put("COLORITEM", "VARCHAR(70)");
         sqlTypes.put("CONTACTITEM", "VARCHAR(6)");
-        sqlTypes.put("DATETIMEITEM", "DATETIME");
+        sqlTypes.put("DATETIMEITEM", "TIMESTAMP");
```
watou: Why change the default SQL type for DateTime items in this PR?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcDerbyDAO.java
```diff
@@ -50,19 +50,21 @@ public JdbcDerbyDAO() {
...
+        SQL_CREATE_ITEM_TABLE = "CREATE TABLE #tableName# (time #TABLEPRIMARYKEY# NOT NULL, value #dbType#, PRIMARY KEY(time))";
         // Prevent error against duplicate time value (seldom): No powerful Merge found:
         // http://www.codeproject.com/Questions/162627/how-to-insert-new-record-in-my-table-if-not-exists
-        SQL_INSERT_ITEM_VALUE = "INSERT INTO #tableName# (TIME, VALUE) VALUES( CURRENT_TIMESTAMP, CAST( ? as #dbType#) )";
+        SQL_INSERT_ITEM_VALUE = "INSERT INTO #tableName# (TIME, VALUE) VALUES( #TABLEPRIMARYVALUE#, CAST( ? as #dbType#) )";
     }
 
     private void initSqlTypes() {
-        sqlTypes.put("DATETIMEITEM", "DATE");
+        sqlTypes.put("DATETIMEITEM", "TIMESTAMP");
```
watou: See earlier question.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcDerbyDAO.java
```diff
@@ -81,6 +83,13 @@ private void initDbProps() {
 
     }
 
+    @Override
+    public void initAfterFirstDbConnection() {
+        logger.debug("JDBC::initAfterFirstDbConnection: Initializing step, after db is connected.");
+        // Initialize sqlTypes, depending on DB version for example
+        // derby does not like this... dbMeta = new DbMetaData();// get DB information
+    }
+
```
watou: perhaps just not override the method then?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcMariadbDAO.java
```diff
@@ -67,6 +66,18 @@ private void initDbProps() {
         databaseProps.setProperty("minimumIdle", "2");
     }
 
+    @Override
+    public void initAfterFirstDbConnection() {
+        logger.debug("JDBC::initAfterFirstDbConnection: Initializing step, after db is connected.");
+        dbMeta = new DbMetaData();
+        // Initialize sqlTypes, depending on DB version for example
+        if (dbMeta.isDbVersionGreater(5, 1)) {
```
watou: Maybe should check for >= 5.6.4 if [this](https://dev.mysql.com/doc/refman/5.6/en/date-and-time-functions.html#function_now) is correct?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
```diff
@@ -68,39 +70,155 @@ public JdbcConfiguration(Map<Object, Object> configuration) {
...
-            logger.error(
-                    "JDBC::updateConfig: url The SQL database URL is missing - please configure the jdbc:url parameter in openhab.cfg");
+
+        if (user == null || StringUtils.isBlank(user)) {
+            logger.warn(
+                    "JDBC::updateConfig: SQL user is missing - please configure the jdbc:user parameter in openhab.cfg");
+        } else if (password == null || StringUtils.isBlank(password)) {
+            logger.warn("JDBC::updateConfig: SQL password is missing. Attempting to connect without password. "
+                    + "To specify a password configure the jdbc:password parameter in openhab.cfg.");
+        } else if (!(url == null || StringUtils.isBlank(url) || parsedURL.getProperty("parseValid") == "false")) {
```
watou: Don't use == for string equality tests, as it may only accidentally work.  Is `.equalsIgnoreCase("false")` more suitable here?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
```diff
@@ -68,39 +70,155 @@ public JdbcConfiguration(Map<Object, Object> configuration) {
...
+
+            // set password
+            dBDAO.databaseProps.setProperty("dataSource.password", password);
+
+            // set sql-types from external config
+            setSqlTypes();
+
+            String et = (String) configuration.get("reconnectCnt");
+            if (StringUtils.isNotBlank(et)) {
+                errReconnectThreshold = Integer.parseInt(et);
```
watou: Just to confirm that any NumberFormatExceptions are handled as you want in this section....

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
```diff
@@ -68,39 +70,155 @@ public JdbcConfiguration(Map<Object, Object> configuration) {
...
+                logger.debug("JDBC::updateConfig: tableIdDigitCount={}", tableIdDigitCount);
+            }
+
+            String rt = (String) configuration.get("rebuildTableNames");
+            if (StringUtils.isNotBlank(rt)) {
+                rebuildTableNames = "true".equals(rt) ? Boolean.parseBoolean(rt) : false;
+                logger.debug("JDBC::updateConfig: rebuildTableNames={}", rebuildTableNames);
+            }
+
+            // undocumented
```
watou: What is the meaning/value of the `// undocumented` comments here and below?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcMapper.java
```diff
@@ -43,6 +48,8 @@ public boolean pingDB() {
         long timerStart = System.currentTimeMillis();
         if (openConnection()) {
             if (conf.getDbName() == null) {
+                logger.debug(
+                        "JDBC::pingDB asking db for name as absolutelly first db action, after connection is established.");
```
watou: spelling: absolutely

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/DbMetaData.java
```diff
@@ -0,0 +1,117 @@
...
+import org.knowm.yank.Yank;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.zaxxer.hikari.HikariDataSource;
+
+/**
+ *
+ * @author Helmut Lehmeyer
+ * @since 1.8.0
```
watou: @since 1.9.0

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/DbMetaData.java
```diff
@@ -0,0 +1,117 @@
...
+            meta = h.getConnection().getMetaData();
+
+            // Oracle (and some other vendors) do not support
+            // some the following methods; therefore, we need
+            // to use try-catch block.
+            try {
+                dbMajorVersion = meta.getDatabaseMajorVersion();
+                logger.debug("dbMajorVersion = '{}'", dbMajorVersion);
+            } catch (Exception e) {
+                logger.error("Asking for 'dbMajorVersion' is a unsupported: '{}'", e);
```
watou: Do you mean `e.getMessage()` or dump the stack trace, so no `{}` in message string?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/DbMetaData.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            try {
+                dbMinorVersion = meta.getDatabaseMinorVersion();
+                logger.debug("dbMinorVersion = '{}'", dbMinorVersion);
+            } catch (Exception e) {
+                logger.error("Asking for 'dbMajorVersion' is a unsupported: '{}'", e);
```
watou: Same question as above.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/DbMetaData.java
```diff
@@ -0,0 +1,117 @@
...
+            logger.debug("driverMinorVersion = '{}'", driverMinorVersion);
+
+            dbProductName = meta.getDatabaseProductName();
+            logger.debug("dbProductName = '{}'", dbProductName);
+
+            dbProductVersion = meta.getDatabaseProductVersion();
+            logger.debug("dbProductVersion = '{}'", dbProductVersion);
+
+        } catch (SQLException e1) {
+            logger.error("Asking for 'dbMajorVersion' seams to be unsupported: '{}'", e1);
```
watou: Same question as above.

watou: Spelling: seems

Also, could the exception be from a different method call on the `meta` object, so maybe the exception message might have the true answer, so you don't need to guess?

watou:
Thank you @lewie for this PR!  Is there a reason you didn't remove the older Hikari and Yank libraries when adding the new ones?  Please see my other comments inline.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
watou: To be more specific, if the documentation is correct and someone is using MySQL 5.6.2 or 5.6.3, TIMESTAMP(n) and NOW(n) will ignore n, but I suppose that's not too bad.  Maybe a small note in the JDBC wiki page on the subject is warranted?

lewie: No, it will not ignore, it will throw an error, unhappily! As soon as possible I will give you statements back to your thorough review. Thanks watou

lewie: More specific, some Databases will throw an error. Older MySql (mysql <= 5.5) is ignoring the fsp argument only while reading with *TIMESTAMP(3)*. 

If using it while creating a Table, most DB's and MySql <= 5.5 throws an error: 
*CREATE TABLE FractionalSeconds (time TIMESTAMP(3), value TIMESTAMP(3));* 
In older DB's you only can use: 
*CREATE TABLE FractionalSeconds (time TIMESTAMP, value TIMESTAMP);*

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
lewie: I got problems using TIMESTAMP and DATETIME mixed. With TIMESTAMP I get better Results now. For me it was crucial that a TIMESTAMP is used with Time zones conversation.

After some investigation, there are pros and cons:
http://cdn.oreillystatic.com/en/assets/1/event/36/Time%20Zones%20and%20MySQL%20Presentation.pdf

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcDerbyDAO.java
lewie: Do not understand? I have to prevent from original Function:
*// derby does not like this... dbMeta = new DbMetaData();*

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcMariadbDAO.java
lewie: Mariadb is nor MySql?
See above comment.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
lewie: Huuuu, I do too much JavaScript, yet. ;-)
Is .equalsIgnoreCase("false") equivalent to the null value?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
lewie: Do you mean *setSqlTypes*? The user can set what he want at the moment. Do you think it should be restricted?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
lewie: *undocumented* means, this is not documented for the *normal* user outside the code and not in wiki.
For debug/testing and helping when driver problems occur, I would like to stay this at the moment.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/utils/DbMetaData.java
lewie: Yes, *e.getMessage()* is much better!
Removed older Hikari and Yank libraries.
Thanks @watou, next PR should come soon.

lewie:
Thank you @watou, struggled a little bit with git, is all ok here with the new commit for the PR?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcDerbyDAO.java
watou: I was confused by a debug message in an otherwise empty method.  If your override is to prevent default behavior, I understand now.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
watou: I mean Integer.parseInt can throw NFE, and my comment was to verify with you that you want that exception to propagate, instead of catching it and logging a warning for the user.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
watou: The code checks for >5.5, but MySQL 5.6.2 and 5.6.3 do not support fractional seconds.  The documentation says the fsp arguments are ignored in those releases, hence my suggestion to document how the code in this PR does not precisely match MySQL release behavior.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
watou: I am concerned that existing JDBC persistence users, who have existing databases, will upgrade to a release that contains this PR, then the user will add items to their persistence, and their database will contain tables created with DATETIME columns before this PR, and other tables created with TIMESTAMP columns created after their upgrade.  Please let me know what the compatibility issues are for this PR.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
watou: Better to use `"false".equalsIgnoreCase(parsedURL.getProperty("parseValid"))`

watou:
Hi @lewie, I added new comments/questions, specifically about compatibility by changing the default SQL type to TIMESTAMP (resulting in mixed table definitions for existing users), and MySQL 5.6.2 and 5.6.3 releases with fractional seconds.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
lewie: PostgreSql and HsqlDB used "TIMESTAMP" before.
Derby, H2, MySql, Mariadb, Sqlite used "DATETIME".

I did Tests with Tables created with DATETIME and wrote TIMESTAMPs. 
Normally this causes no problems. 
However, side effects may occur with certain configured timezones. From that Time I replace the bundle, it would then save values with the correct timezone offset.

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
lewie: I think to propagate is ok here?! It is well documented, that should be enough.

lewie:
Hi @watou, 
I suggest, I will document the Changes. Explicitly for the mixed table definitions and importance for the various database types and versions. Would that for you too the right way?

bundles/persistence/org.openhab.persistence.jdbc/java/org/openhab/persistence/jdbc/internal/JdbcConfiguration.java
```diff
@@ -85,13 +85,14 @@ public void updateConfig(Map<Object, Object> config) {
         String url = (String) configuration.get("url");
         Properties parsedURL = StringUtilsExt.parseJdbcURL(url);
 
-        if (user == null || StringUtils.isBlank(user)) {
+        if (user.equalsIgnoreCase("false") || StringUtils.isBlank(user)) {
```
watou: So `user`, `password` and `url` can have the string value `"false"`?  That doesn't seem correct.  If they are in fact `null`, the `.equalsIgnoreCase` will cause an NPE.

lewie: @watou, I will change these points and this: https://github.com/openhab/openhab/pull/4550#discussion-diff-71987438 and create a new PR.

lewie:
Hmm, git is sometimes not my friend...

lewie:
@watou, created a new PR [here](https://github.com/openhab/openhab/pull/4605).

PR #4556, opened 2016-07-25
made Onkyo binding 1.x legacy as there is now a 2.0 version

kaikreuzer:


watou:
Thanks @kaikreuzer

PR #4566, opened 2016-08-04
[satel] Thread interruption changes.

druciak:
This PR introduces some minor changes in thread management.

### DIFF
--- a/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
+++ b/bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
@@ -107,7 +107,7 @@ public abstract class SatelModule extends EventDispatcher implements SatelEventL
 
     /**
      * Returns firmware revision of Integra connected to the module.
-     * 
+     *
      * @return version of Integra firmware
      */
     public String getIntegraVersion() {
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
```diff
@@ -271,15 +271,15 @@ private SatelMessage readMessage() {
 
                 // if meanwhile thread has been interrupted, exit the loop
                 if (Thread.interrupted()) {
```
watou: Was there a reason to not change this line to `!Thread.currentThread().isInterrupted()` as below?

watou:
What problem(s) are corrected with this PR?  Good to document that here I think.  Thanks @druciak!

bundles/binding/org.openhab.binding.satel/src/main/java/org/openhab/binding/satel/internal/protocol/SatelModule.java
druciak: I decided it would be better to throw InterrupedException in this case. No specific reason, in my opinion it's slightly better, but I am open to any solutions. :-)

druciak:
No problems actually, just a refactor. Previous code worked well, new one works also very well.

watou:
Thanks @druciak!

PR #4567, opened 2016-08-04
Prevent NPE when neohub is down

carlossg:


bundles/binding/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubBinding.java
```diff
@@ -85,6 +85,11 @@ protected void execute() {
         try {
             // send info request
             final InfoResponse response = createProtocol().info();
+            if (response == null) {
+                logger.error("Failed to connect to neohub [host '{}' port '{}']",
```
watou: Won't this lead to two ERROR messages in the log for every polling cycle where it failed to connect, because of [this line](https://github.com/carlossg/openhab/blob/e557b5a510b6ea12155cdeccaafa148710cd4377/bundles/binding/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubConnector.java#L81)?

bundles/binding/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubBinding.java
```diff
@@ -85,6 +85,11 @@ protected void execute() {
...
+                        new Object[] { hostname, port });
```
watou: The `new Object[] {hostname, port } ` method signature isn't needed, as you can just pass variable arguments as `hostname, port`.

bundles/binding/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubBinding.java
carlossg: yes it did, updated it to just ignore the null

watou:
Thank you @carlossg!

Issue #3685, opened 2016-01-02
Modbus binding: coil commands are not written if the last read data matches the command


ssalonen:

ssalonen:
I believe this can be closed now that PR is merged?

ssalonen:
Closing as pr merged.

PR #4576, opened 2016-08-08
Modbus: enable tests that used to fail

ssalonen:
Some of the failing modbus binding tests were skipped for performance reasons. Now that #3685 is fixed, those tests can be enabled once again. This PR implements just this.


bundles/binding/org.openhab.binding.modbus.test/src/test/java/org/openhab/binding/modbus/internal/WriteCoilsAndDiscreteTestCase.java
```diff
@@ -197,9 +184,6 @@ public void testWriteDigitalsAfterRead() throws Exception {
 
     private void verifyRequests(boolean readRequestExpected) throws Exception {
         // XXX: for test performance reasons, skipping failing tests
```
watou: is this comment still valid?

ssalonen: Good spot! Removed the comment.

watou:
Thanks @ssalonen!

PR #4577, opened 2016-08-08
[Modbus] Read errors now emit UNDEF (configurable)

### LINKED ISSUE #4552: Modbus Binding doesn't set items to undefined once devices have become unresponsive
Hello everyone,
I'm using Modbus to connect window contacts and some other stuff to Openhab. 

## Expected Behavior
A failed bus device should be detectable by looking at the item state.
 
## Current Behavior
When a device is removed from the Bus the item states are never updated and therefore keep their current state.

## Possible Solution
Make the binding set the state of all items associated to a certain Modbus device to UNDEFINED if the device does not respond after three retries.

1. Disconnect your modbus device while the system is running
2. check the state that is never going to change :)

## Context
My problem is that when a Modbus device fails and is not reachable anymore the associated items keep their current state. For me this means that even though a window is open it might very well be marked as closed in the UI. Currently the only way to be sure that all windows are indeed closed is looking for Modbus errors in the log files - which is not very practical... :grin: 

## Your Environment
current online runtime version of openhab2, binding installed using the paper UI

ssalonen:
Resolves #4552. 

This implementation emits Undefined state even with all read errors:
- connection establishment fails
- protocol errors (that is, modbus slave response is error response)
- IO errors
- transaction id mismatch (request's transaction id does not match response's transaction id)

Additionally, this PR will also implement transaction id check for writes, and in case of such mismatch, error is logged with WARN. Previously this case was not checked and not logged.

Pre-compiled jar available [here](https://github.com/ssalonen/openhab/releases/tag/modbus-error-detection-pr-2016-08-08)

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus.test/src/test/java/org/openhab/binding/modbus/internal/ErroringQueriesTestCase.java
+++ b/bundles/binding/org.openhab.binding.modbus.test/src/test/java/org/openhab/binding/modbus/internal/ErroringQueriesTestCase.java
@@ -17,6 +17,7 @@ import org.junit.Test;
 import org.openhab.binding.modbus.ModbusBindingProvider;
 import org.openhab.core.library.items.SwitchItem;
 import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.types.UnDefType;
 import org.openhab.model.item.binding.BindingConfigParseException;
 import org.osgi.service.cm.ConfigurationException;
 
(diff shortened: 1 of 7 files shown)


ssalonen:
@mbs38  Could you please test this version?

mbs38:
@ssalonen I've tried it and there is no change. Maybe my configuration is wrong?

modbus:serial.client15in.connection=/dev/ttyUSB0:115200:8:none:1:rtu:3:500
modbus:serial.client15in.id=15
modbus:serial.client15in.start=0
modbus:serial.client15in.length=8
modbus:serial.client15in.type=discrete
modbus:serial.client15in.postundefinedonreaderror=true

ssalonen:
The configuration looks ok. Are you expecting errors with that config? What does it say in the logs?

mbs38:
@ssalonen  I confounded the files because all the files i had laying around were named org.openhab.binding.modbus-1.9.0-SNAPSHOT.jar.
Sorry about that.

As far as I can tell it works perfectly now. 
Thank you very much!

ssalonen:
Great to hear!

I will fix one technical thing as line commented above. Then let's see if we could get this merged in.

ssalonen:
Fixed (2755861) the accidental try-catch change. Also documented in javadoc why initial state needs to be really `null` (so that we get UNDEF state even if the first connect fails)

watou:
Thank you Sami!

watou:
@ssalonen just a check that the wiki is up to date with this enhancement?

ssalonen:
@watou  Good thing you mentioned. Updated the wiki.

PR #4614, opened 2016-09-03
Update Nest Binding To only update if state has changed

vzorglub:
The binding is clogging up the events log file with item updates. I propose a change so that the binding only updates an item when the value has changed.

### DIFF
--- a/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/NestBinding.java
+++ b/bundles/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/internal/NestBinding.java
@@ -195,6 +195,7 @@ public class NestBinding extends AbstractActiveBinding<NestBindingProvider>imple
         }
 
         DataModel newDataModel = dmres;
+        DataModel olDataModel2 = this.oldDataModel
         this.oldDataModel = newDataModel;
 
         // Iterate through bindings and update all inbound values.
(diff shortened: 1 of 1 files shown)


watou:
Hi @vzorglub, this has been discussed before, and the outcome of previous discussions, as best as I can remember, is that it is a poor default installation to log all state changes in events.log.  Bindings should be unconstrained in asserting the states of items they are bound to, and remain ignorant of previous state as much as possible, so that if a binding knows every minute what the state of an item should be, it should be free to update the item to that state every minute.  This is why in the rules engine there is a distinction between the "received update" trigger and the "changed" trigger.  Instead of blunting the useful information value of repeated updates, I recommend you change your openHAB configuration to not log all state changes, if the size of the events.log file is problematic for you.  I did that in openHAB 1.x by changing INFO to WARN for `runtime.busevents`:

```xml
	<logger name="runtime.busevents" level="WARN" additivity="false">
		<appender-ref ref="EVENTFILE" />
		<appender-ref ref="STDOUT" />
	</logger>
```

vzorglub:
Hi John,
Thank you for this detailed and comprehensive explanation. I understand
completely.
The size of my event log is not a problem. My problem is that the nest
updates are drowning the rest of the data
and it get difficult to find individual or series of events. I just want to
remove the nest updates when the value has not changed.
Is that possible?

Thanks

Vincent

watou:
> the nest updates are drowning the rest of the data and it get difficult to find individual or series of events

Do you mean in the events.log file? A command line tool like grep might make it easier to find specific log lines, if that's what you mean. You could also write a rule that triggers on `changed` and log something to openHAB.log.

vzorglub:
John,

Yes I mean the events log file. I just find it strange that a value is sent
an update when it is not actually updated (value changed). Is there anyway
to prevent the updates to appear in the log?

If I wanted to modify the binding for my own purposes, would the changes
that I proposed have worked? I am not very conversant in Java I am afraid.
And after how do I build a personalised jar file?

Thanks

Vincent

watou:
> Is there anyway to prevent the updates to appear in the log?

See my suggestion earlier for changing logback.xml.

> If I wanted to modify the binding for my own purposes, would the changes that I proposed have worked?

Careful using `==` to compare objects.  You would want to instead compare using `obj1.equals(obj2)`.

There are also syntax errors in the PR, but a compile will find them.

> how do I build a personalised jar file?

In the binding's top-level directory, issue `mvn clean install` and find the resulting JAR in the `target` directory.  Remove the shipped binding JAR from addons and put in your version.

I'm going to close this PR now, and please consider asking any further questions on the forum as there is a much larger potential audience who can offer suggestions.  Thanks!

PR #4622, opened 2016-09-05
some changes to XBMC binding

giig1967g:
I hope this time is fine


bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/internal/XbmcActiveBinding.java
```diff
@@ -46,12 +47,12 @@
...
      * (optional, defaults to 60000ms)
      */
-    private long refreshInterval = 60000;
+    private long refreshInterval = 60000L;
 
     @Override
     public void activate() {
         logger.debug(getName() + " activate()");
-        setProperlyConfigured(true);
+		//setProperlyConfigured(true); // <-Removed as it has been moved to the "updated" call
```
sumnerboy12: Just remove the commented code - the history will be retained in github.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/InputBack.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.ning.http.client.AsyncHttpClient;
+
+/**
+ * Input.Back RPC
+ *
+ * @author Il Plebs
+ * @since 1.5.0
+ */
+public class InputBack extends RpcCall {
```
sumnerboy12: Rather than have all these classes which are exactly the same apart from the command/name, perhaps it would be better to have a single `XxxRpcCall` class with an enumerator for the command?

giig1967g: Hi Ben,
the reason I have added the classes in that way, was to maintain the same
original logic to reduce the changes.

I could probably simplify it removing all the 11 Inputxx classes with a
single INPUT class with several options, like the PlayerGetLabels class.

This means that I will be adding only 2 new classes: InputCommands and
playerGetLabels.

What do you think?

sumnerboy12: I think it is probably worth doing it - much simpler code to follow then.

giig1967g: ok.
I will out of town for few days.
Will fix it next week.

Is it ok?

sumnerboy12: Of course - just whenever you can - thanks for improving the binding!

giig1967g: Hi Ben,
I have just updated the commit.
Can you please check if you can see it correctly?

stefano

sumnerboy12:
Looks ok to me, if you have tested it and are happy it works then I am happy to approve it. The only other thing to do would be to update the WIKI page for this binding with instructions on how to use all the new actions. Thanks very much!

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/XbmcConnector.java
```diff
@@ -581,25 +586,82 @@ private void requestPlayerUpdate(int playerId, boolean updatePolledPropertiesOnl
...
+                item2.setPlayerId(playerId);
+                item2.setProperties(propertiesProperties);
+
+                item2.execute(new Runnable() {
+                    @Override
+                    public void run() {
+                        // now update each of the openHAB items for each property
+                        for (String property : propertiesProperties) {
+                            String value = item2.getPropertyValue(property);
+                            // logger.debug("UPDATEfromGETPROPERTIES: property=" + property + " & value=" + value);
```
watou: Either remove the commented out code, or uncomment the debug line (and switch to parameterized logging).

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/XbmcConnector.java
```diff
@@ -581,25 +586,82 @@ private void requestPlayerUpdate(int playerId, boolean updatePolledPropertiesOnl
...
                     }
-                }
-            });
+                });
+            }
+
+            // make the request for the player item3 details using GETLABELS
+            if (!propertiesLabels.isEmpty()) {
+                final PlayerGetLabels item3 = new PlayerGetLabels(client, httpUri);
+                // item3.setPlayerId(playerId);
```
watou: Remove the commented out code.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/XbmcConnector.java
```diff
@@ -581,25 +586,82 @@ private void requestPlayerUpdate(int playerId, boolean updatePolledPropertiesOnl
...
+                item3.setProperties(propertiesLabels);
+
+                item3.execute(new Runnable() {
+                    @Override
+                    public void run() {
+                        // now update each of the openHAB items for each property
+                        for (String property : propertiesLabels) {
+                            String value = item3.getPropertyValue(property);
+                            // logger.debug("UPDATEfromGETLABELS: property=" + property + " & value=" + value);
```
watou: Either remove the commented out code, or uncomment the debug line (and switch to parameterized logging).

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetItem.java
```diff
@@ -23,6 +23,8 @@
  *
  * @author tlan, Ben Jones, Marcel Erkel
  * @since 1.5.0
+ * @author tlan, Ben Jones, Plebs
+ * @since 1.9.0
```
watou: Just add your name  to line 24 (and add back Marcel Erkel), and leave it at 1.5.0, and remove lines 26 & 27.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetLabels.java
```diff
@@ -0,0 +1,131 @@
...
+import com.ning.http.client.AsyncHttpClient;
+
+/**
+ * Player.GetLabels RPC
+ *
+ * @author Ben Jones
+ * @since 1.5.0
+ * @author tlan, Ben Jones, Plebs
+ * @since 1.9.0
+ */
```
watou: Since you added this file, the comment should just be
```java
 * @author Plebs
 * @since 1.9.0
```

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetLabels.java
```diff
@@ -0,0 +1,131 @@
...
+        List<String> paramProperties = new ArrayList<String>();
+        for (String property : properties) {
+            if (property.startsWith("Label.")) {
+                String paramProperty = getParamProperty(property);
+                paramProperties.add(paramProperty);
+            }
+        }
+
+        Map<String, Object> params = new HashMap<String, Object>();
+        // params.put("playerid", playerId);
```
watou: Remove commented code lines.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetLabels.java
```diff
@@ -0,0 +1,131 @@
...
+        params.put("labels", paramProperties);
+        return params;
+    }
+
+    @Override
+    protected void processResponse(Map<String, Object> response) {
+        Map<String, Object> result = getMap(response, "result");
+        // item = getMap(result, "label");
```
watou: Remove commented code lines.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetProperties.java
```diff
@@ -13,21 +13,27 @@
...
 
 import com.ning.http.client.AsyncHttpClient;
 
 /**
  * Player.GetProperties RPC
  *
- * @author Ben Jones
- * @since 1.5.0
+ * @author Plebs
+ * @since 1.9.0
```
watou: Comment tags should just be
```java
 * @author Ben Jones, Plebs
 * @since 1.5.0
```

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetProperties.java
```diff
@@ -37,32 +43,87 @@ public void setPlayerId(int playerId) {
...
         params.put("playerid", playerId);
-        params.put("properties", properties);
+        params.put("properties", paramProperties);
         return params;
     }
 
     @Override
     protected void processResponse(Map<String, Object> response) {
         Map<String, Object> result = getMap(response, "result");
+        // item = getMap(result, "item");
```
watou: Remove commented code lines.

giig1967g:
Hi,
I sent the files with the changes suggested.
Sorry for the confusion with the tags, but it's the first time I actually
give a contribution to an Open Source app so please apologies.

I have also added a new command in the meantime...
"Refresh"

giig1967g:
Hi all,
did you have a chance to merge my changes?
Please keep me updated and then I will update the wiki.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/RpcCall.java
```diff
@@ -131,11 +131,14 @@ private void postRequest(Map<String, Object> request, Runnable completeHandler)
         try {
             // we fire this request off asynchronously and let the completeHandler
             // process any response as necessary (can be null)
-            ListenableFuture<Response> future = client.preparePost(uri).setBody(writeJson(request))
+            String resultWrite = writeJson(request);
+			logger.debug("Write JSON: {}", resultWrite ); // Stefano: logging the JSON request
```
watou: Please put the code through the Eclipse IDE formatter so future changes don't reformat lines with incorrect formatting.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/RpcCall.java
```diff
@@ -131,11 +131,14 @@ private void postRequest(Map<String, Object> request, Runnable completeHandler)
...
+            ListenableFuture<Response> future = client.preparePost(uri).setBody(resultWrite)
                     .setHeader("content-type", "application/json").setHeader("accept", "application/json")
                     .execute(new AsyncCompletionHandler<Response>() {
                         @Override
                         public Response onCompleted(Response response) throws Exception {
+							logger.debug("Read JSON: {}", response.getResponseBody()); // Stefano: logging the JSON response
```
watou: Please put the code through the Eclipse IDE formatter so future changes don't reformat lines with incorrect formatting.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/XbmcConnector.java
```diff
@@ -276,15 +278,16 @@ public void run() {
...
      * @param itemName
      *            The name of the item which should receive updates
      * @param property
      *            The property of this xbmc instance, which is to be
      *            watched for changes
-     * 
+     *
      */
     public void addItem(String itemName, String property) {
+		logger.debug("Mapping: itemname = {} & property = {}", itemName, property);
```
watou: Please put the code through the Eclipse IDE formatter so future changes don't reformat lines with incorrect formatting.

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/internal/XbmcActiveBinding.java
```diff
@@ -326,6 +328,8 @@ protected void internalReceiveCommand(String itemName, Command command) {
                 connector.playerOpenPVR(command.toString(), 2);
             } else if (property.equals("PVR.OpenRadio")) {
                 connector.playerOpenPVR(command.toString(), 1);
+            } else if (property.equals("Refresh")) {
+                execute();
```
watou: Are you certain that it is threadsafe to do this?  The `execute()` method is called at some refreshInterval in another thread, so it may be possible that this method could be run from each thread overlapping.  You might consider using a technique like in the HTTP or Ecobee binding, where the execute() method is called based on a short `granularity` (five seconds, for example), but checks a future timestamp to see if it should do anything.  The "Refresh" action would then just expire the future timestamp, so the next scheduled `execute` happens on average `granulary/2` into the future.  The `execute()` method would just set the future timestamp to the current timestamp + refreshInterval.  (There are certainly other sound approaches to avoiding thread safety issues.)

bundles/binding/org.openhab.binding.xbmc/src/main/java/org/openhab/binding/xbmc/rpc/calls/PlayerGetProperties.java
```diff
@@ -13,21 +13,27 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.ning.http.client.AsyncHttpClient;
 
 /**
  * Player.GetProperties RPC
  *
- * @author Ben Jones
+ * @author Ben jones, Plebs
```
watou: Jones

giig1967g:
Hi all,
I don't understand what I am supposed to do...

Do I have to reformat the single line in the 3 files and resend to you? Or
you do it?

Regarding the refresh() method, I didn't want to change the original
structure of the binding.
In the original binding the excute() method is already called asyncronously
by a timer and by specific events, and can be overlapping. I just added a
manual call.
The Kodi JSON can handle cuncurrent calls with no issues at all.
I wouldn't know how to change it, so if what I added is a problem, I can
remove the manual refresh command.
Just tell me what do you prefer.

What are the next steps?

thanks
stefano

watou:
> In the original binding the excute() method is already called asyncronously by a timer and by specific events, and can be overlapping. I just added a manual call.

My comment was just to verify that this PR wasn't introducing a new thread safety issue.  If other parts of the binding invoke `execute()` explicitly, then your change is probably OK.

So I will merge this now, and make a new PR myself to only correct the minor formatting issues.

watou:
@giig1967g Please update the wiki to document your new enhancements. Thanks!

giig1967g:
Hi watou,
I need an advice...
I was about to start modifying the wiki but then I realized that the new features will not be available to all, but just to the ones who download the nightly build (SNAPSHOT).

So, I think that the options are three:
1) change the wiki now (but it might confuse the users who have not downloaded the latest version)
2) change the wiki only when the next public version of OpenHAB will become available (1.9)
3) change the wiki specifying that some changes will only be available for the SNAPSHOT users.

What is the best practice?

thanks

watou:
Hi Stefano, it's OK to update the wiki now, but please indicate which features will be available in version 1.9 and later of the binding, which will be the next official release that contains your enhancements.  So the wiki will still be helpful to 1.8 users, but they won't be confused about which features they don't have if they are marked something like "available in 1.9."

giig1967g:
ok. thanks for the tip.

stefano

giig1967g:
Hi,
I have updated the wiki.
Would appreciate if you check if the way I have done it is correct.

thanks

watou:
Thanks @giig1967g!  Comments on your wiki changes:

> -    # Username to connect to XBMC. (optional, defaults to none)
> +    # Username to connect to XBMC. (optional, defaults to xbmc)
>     xbmc:livingRoom.username=xbmc
> -    # Password to connect to XBMC. (optional, defaults to none)
> +    # Password to connect to XBMC. (optional, defaults to xbmc)
>     xbmc:livingRoom.password=xbmc

I am concerned that this change to the binding will break existing users' configurations.  Every PR should not break any existing configurations unless critically necessary.  If this is a breaking change, it and the wiki should be reverted.  What are your thoughts?

> ### Available properties (until V1.8)

I suggest that you only label features that are new in 1.9 as such, and remove any references to 1.8 that you added if nothing about the feature has changed from 1.8 to 1.9.

> +canrepeat    | <         | Returns true if the media can be repeated

When you say `true`, what is actually being returned?  ON? OPEN?

> +Examples: 
> +* xbmc=">[livingRoom|Player.PlayStop]"
> +* xbmc="<[livingRoom|Player.Title]"

I think the examples would be more helpful if they showed a complete Item definition, as one would have in a .items file.  That way, the user knows better what item type is being used, a label string could in some cases add more context, etc.

giig1967g:
Hi John,
Thanks for your comments.
Regarding the username/password change, I didn't change the code.
The wiki and the .cfg were wrong. In the code the default username and
password have always been set to "xbmc"/"xbmc".
I just documented it and updated the cfg file.

You can check in github.com.

You are right about the wiki regarding the expected returned text for the
properties.
Will update it.
When I have done, I will ask a again for your feedback :)

Thanks for your tips.

Il 10/Ott/2016 18:59, "John Cocula" <notifications@github.com> ha scritto:

watou:
> Regarding the username/password change, I didn't change the code. The wiki and the .cfg were wrong. In the code the default username and password have always been set to "xbmc"/"xbmc".  I just documented it and updated the cfg file.

That's the best possible answer.  Thanks for fixing the documentation.

> You are right about the wiki regarding the expected returned text for the properties.
Will update it. When I have done, I will ask a again for your feedback :)

Perfect.  Thank you!

PR #4625, opened 2016-09-06
Add option to send plain literals messages.

ppt000:
The TCP binding does not allow to send plain literal strings straight from the item declaration <transformationrule> (the 4th argument in brackets), but forces to go through transformations.
A simple way to achieve that without changing the code much is to treat an 'empty' transformation to pass the argument directly as the response: instead of "MAP(my.device.map)" or "REGEX((.*))" for example, just type "(1234ABC)" to send the string 1234ABC. The parenthesis are needed to match the pattern tested before, otherwise some more code needs to changed. Also we can assume that no transformation rule will ever be named with an empty string, but it still matches the pattern.

### DIFF
--- a/bundles/binding/org.openhab.binding.tcp/src/main/java/org/openhab/binding/tcp/protocol/internal/TCPBinding.java
+++ b/bundles/binding/org.openhab.binding.tcp/src/main/java/org/openhab/binding/tcp/protocol/internal/TCPBinding.java
@@ -261,32 +261,37 @@ public class TCPBinding extends AbstractSocketChannelBinding<TCPBindingProvider>
 
     protected String transformResponse(String transformation, String response) {
         String transformedResponse;
-
-        try {
-            String[] parts = splitTransformationConfig(transformation);
-            String transformationType = parts[0];
-            String transformationFunction = parts[1];
-
-           if (transformationType == "") {     // test for empty type first to avoid the WARN from getTransformationService
-                transformedResponse = transformationFunction;
-            } else {
+        Matcher matcher = EXTRACT_FUNCTION_PATTERN.matcher(transformation);
+        String transformationServiceName;
+        String transformationServiceParam;
+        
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.tcp/src/main/java/org/openhab/binding/tcp/protocol/internal/TCPBinding.java
```diff
@@ -267,14 +267,18 @@ protected String transformResponse(String transformation, String response) {
             String transformationType = parts[0];
             String transformationFunction = parts[1];
 
-            TransformationService transformationService = TransformationHelper
-                    .getTransformationService(TCPActivator.getContext(), transformationType);
-            if (transformationService != null) {
-                transformedResponse = transformationService.transform(transformationFunction, response);
+           if (transformationType == "") {     // test for empty type first to avoid the WARN from getTransformationService
```
watou: You should always compare strings using the `equals` method in Java.  Testing using the `==` method might only return `true` accidentally.

ppt000:
Well, I would say that a really purist approach would be to have a
'straight through' transformation, called STRING() for example, that just
sends the argument as a string. The parenthesis on their own almost do
that, even if I admit it's unorthodox.
The problem with sending the string 'plainly', is that it forces to nullify
all the checks if a proper transformation exists or not or if it's
arguments are acceptable or not. That means that any typo in the syntax of
a real transformation will just send the string as a result rather than
generate a WARN or an ERROR. It becomes more difficult to debug.
I am fine either way, changing the code in the TCP binding for making it
work like you say is not complicated I agree, and I can propose another PR
for that.
I would prefer you would get the feedback from the community on what route
to take though.
Thanks for your help.

watou:
Hi Paolo, I started a thread here: https://community.openhab.org/t/opinion-on-tcp-binding-enhancement/14176

PR #4627, opened 2016-09-07
[modbus] Connection timeout parameter for TCP slaves.

### LINKED ISSUE #4595: [modbus] Connection timeout with modbus TCP slaves should be configurable
## Expected Behavior
- in the connection string of a slave, one should be able to configure timeout it takes to connect to a TCP slave. 
- Motivation is to fail fast with offline modbus slaves, and try the other slaves instead of hanging up to 20s (windows) or 120s (modern ubuntu).

## Current Behavior
Current behaviour is to wait some undocumented platform specific time when connecting to the TCP slave.

Related discussion in community forums, starting from [this post](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/171?u=ssalonen) by Botura.

For more details, see these links
- [SO "Socket connect timeouts: where is the specification?"](http://stackoverflow.com/questions/13266403/socket-connect-timeouts-where-is-the-specification)
- [SO "Where does the socket timeout of 21000 ms come from?"](http://stackoverflow.com/questions/26896414/where-does-the-socket-timeout-of-21000-ms-come-from)

Debug level logs w/ associated config posted [here](https://community.openhab.org/t/connection-pooling-in-modbus-binding/5246/176?u=ssalonen)

## Possible Solution

Since our forked jamod does not expose ability to set connection timeout in [TCPMasterConnection](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java), the API needs to be changed a little bit:
- **Jamod modification** Instead of [creating new (connected) Socket using (host, port) constructor](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java#L68), one should create unconnected socket [with empty constructor](https://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#Socket()) and then call [connect(endpoint, timeout)](https://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#connect(java.net.SocketAddress,%20int)) method, where socket address can be created using [InetSocketAddress constructor](https://docs.oracle.com/javase/8/docs/api/java/net/InetSocketAddress.html#InetSocketAddress-java.lang.String-int-).
- **Connection string parsing** should take intro consideration the new parameter in [ModbusBinding.updated](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java#L500). 

Current format (>=1.9.0) for the connection string with TCP slaves is:
````
host_ip[:port[:interTransactionDelayMillis[:reconnectAfterMillis[:interConnectDelayMillis[:connectMaxTries]]]]].
````

With <1.9.0, it is much simpler `host_ip[:port]`.

I propose that we would introduce default connection timeout of 30 seconds, and make it configurable with the following format (appended the parameter)
````
host_ip[:port[:interTransactionDelayMillis[:reconnectAfterMillis[:interConnectDelayMillis[:connectMaxTries[:connectTimeout]]]]]].
````
The change would be backwards-compatible with existing 1.9.0-SNAPSHOT users.

Default 30s should be conservative enough in my opinion, but system tools seem to be conservative as well: wget ([no timeout](http://stackoverflow.com/questions/2291524/does-wget-timeout)), curl ([300 s](https://curl.haxx.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html))

Since this is a connection parameter, the change needs to be also implemented for EndpointPoolConfiguration.

- **wiki update** regarding configuration change
- **config template** update (both 1.x and 2.x) since it includes the explanation of the connection parameter
- **integration test**: since this should be pretty easy to test, we could extend integration tests to include a timeout scenario.

Note that connection timeout is different from [TCPMasterConnection.setTimeout](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java#L130) parameter which controls `SO_TIMEOUT` parameter of the socket (default 3s with our binding).

1. slave pointing to ip not responding at all

## Your Environment

Botura reported 21s timeout with windows. I can experience 128s (!) timeout with ubuntu.

All tests with binding version 1.9.0-SNAPSHOT (2016-08)

### PR DESCRIPTION
**Pull request by ssalonen**
Resolves #4595

The default connection timeout is zero which is basically matches the previous behavior (OS default or infinite).

Pre-compiled JAR available at https://github.com/ssalonen/openhab/releases/tag/modbus-timeout-pr-2016-09-07

TODO (once merged): wiki page update

### DIFF
--- a/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java
+++ b/bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java
@@ -18,6 +18,7 @@ package net.wimpi.modbus.net;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.net.Socket;
 
 import org.apache.commons.lang.builder.StandardToStringStyle;
@@ -49,6 +50,8 @@ public class TCPMasterConnection implements ModbusSlaveConnection {
     // private int m_Retries = Modbus.DEFAULT_RETRIES;
     private ModbusTCPTransport m_ModbusTransport;
 
+    private int m_ConnectTimeoutMillis;
+
     private static StandardToStringStyle toStringStyle = new StandardToStringStyle();
 
     static {
@@ -70,6 +73,11 @@ public class TCPMasterConnection implements ModbusSlaveConnection {
         setPort(port);
     }
 
+    public TCPMasterConnection(InetAddress adr, int port, int connectTimeoutMillis) {
+        this(adr, port);
+        setConnectTimeoutMillis(connectTimeoutMillis);
+    }
+
     /**
      * Opens this <tt>TCPMasterConnection</tt>.
      *
@@ -79,7 +87,8 @@ public class TCPMasterConnection implements ModbusSlaveConnection {
     public synchronized boolean connect() throws Exception {
         if (!isConnected()) {
             logger.debug("connect()");
-            m_Socket = new Socket(m_Address, m_Port);
+            m_Socket = new Socket();
+            m_Socket.connect(new InetSocketAddress(m_Address, m_Port), this.m_ConnectTimeoutMillis);
             setTimeout(m_Timeout);
             m_Socket.setReuseAddress(true);
             m_Socket.setSoLinger(true, 1);
@@ -221,6 +230,14 @@ public class TCPMasterConnection implements ModbusSlaveConnection {
     @Override
     public String toString() {
         return new ToStringBuilder(this, toStringStyle).append("socket", m_Socket).toString();
+    }
+
+    public int getConnectTimeoutMillis() {
+        return m_ConnectTimeoutMillis;
+    }
+
+    public void setConnectTimeoutMillis(int m_ConnectTimeoutMillis) {
+        this.m_ConnectTimeoutMillis = m_ConnectTimeoutMillis;
     }
 
 }// class TCPMasterConnection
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by ssalonen** [comment]
@botura please test the version linked here, and share your experiences. The default timeout is OS default ("infinite"). The configuration was documented above.

**Comment by botura** [comment]
@ssalonen, it worked as expected.
Without timeout parameter, timeout depends on the OS (that´s good for legacy application).
And with new parameter, timeout occurs according to the parameter value. Great for systems with more than one slave.

I´ll make new tests during the weekend, but I believe everything is ok!

File: bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java [hunk 1]
```diff
@@ -223,4 +232,12 @@ public String toString() {
         return new ToStringBuilder(this, toStringStyle).append("socket", m_Socket).toString();
     }
 
+    public int getConnectTimeoutMillis() {
+        return m_ConnectTimeoutMillis;
+    }
+
+    public void setConnectTimeoutMillis(int m_ConnectTimeoutMillis) {
```
watou: Is the `m_` prefix for the method parameter make sense? (obv. very minor point)

watou:
@ssalonen @botura please report when you are confident that this PR is ready to merge, thanks!

bundles/binding/org.openhab.binding.modbus/src/main/java/net/wimpi/modbus/net/TCPMasterConnection.java
ssalonen: I agree that it does not really make sense but I'm keeping the "jamod" (net.wimpi.modbus) code base consistent. 

For whatever reason they (=original authors of jamod) are prefixing the private variables with `m_`.

Perhaps at some point we should consider formatting the whole net.wimpi.modbus code to comform to modern java style guide.

ssalonen:
I'm good with the merge if @botura is happy.

botura:
Pretty happy @ssalonen. With this enhancement I can definitely use OH in my system.

ssalonen:
Wiki updated.

Issue #4615, opened 2016-09-03
[rfxcom] RFXComLighting1 Chime not correctly implemented (also in OH2)


martinvw:
## Expected Behavior

The code suggests support for the chime command for Lighting1 but does not because of a small bug.
## Current Behavior

When the chime data is received it's not correctly recognized:
2016-08-30 15:26:44.334 [DEBUG] [.b.r.internal.RFXComConnection] - Data received:
Raw data = 0710015242080780
- Packet type = LIGHTING1
- Seq number = 82
- Sub type = ARC
- Id = B
- Unit code = 8
- Command = UNKNOWN
- Signal level = 8
## Possible Solution

The code should be fixed, which I can do, but this most likely influences other situations and since the test coverage is very low. I need some testers because all my switches and remotes are newer and work via the Lighting2 protocol.

The problems lies in the Commands enum together with how it is accessed. The numbers refer to the bytes which come in however, they are retrieved by their position in the enum. Since there is no element 4. The whole logic breaks. But if the group_on / group_off commando's work now. Then the bytes for them should most likely be 4 & 5. 6 will then be some unsupported command. And 7 is indeed Chime but it's now on position ~~5~~ _6_.
## Steps to Reproduce (for bugs)
1. Push the button of the chime
2. Command is interpreted as UNKNOWN, see log fragment above.
## Context

I cannot use my Chime now :-(
Chime is the CoCo KaKu ACDB-7000BC
## Your Environment

- Version used: 1.9.0-SNAPSHOT
- Operating System and version:  Raspbian Jessie
##

martinvw:
It could also be the case that a chime can be given with 6 or a 7 and that is the reason no else complained.

martinvw:
A similar problem exists in handling responses from the RFXCom after sending a message, see the following log fragment:

2016-09-05 19:27:18.092 [DEBUG] [.b.r.internal.RFXComConnection] - Data received:
Raw data = 0402014300
 - Packet type = TRANSMITTER_MESSAGE
 - Seq number = 67
 - Sub type = UNKNOWN
 - Response = ACK

PR #4629, opened 2016-09-08
Feature/4615 rfxcom lighting1 chime

martinvw:
Besides the original problem (#4615 incorrect support of Lighting1 I fixed some similar problems.

 I did this by using the SDK which I got from the developers of the RFXCom and I got permission to use their exampe messages in tests.

The following changes are included:
- Added additional tests for RFXCom binding and made the old test complete sucessfully
- RFXComWindMessage 
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was working correct because of UNDEF placeholder)
  - add support for an additional subtype
  - fixed the outgoing message (corrected message type from RAIN to WIND)
- RFXComWeightMessage 
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was incorrect here)
- RFXComUndecodedRFMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComUVMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was incorrect here)
- RFXComTransmitterMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was incorrect here)
- RFXComThermostat3Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComThermostat2Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - mapping incoming message gave an ArrayIndexOutOfBoundsException
  - Length definition byte of outgoing messages was incorrect
-  RFXComThermostat1Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Corrected the bitshift for the mode
  - Length definition byte of outgoing messages was incorrect
- RFXComTemperatureRainMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Sending an outgoing message gave an ArrayIndexOutOfBoundsException
  - Length definition byte of outgoing messages was incorrect
- RFXComTemperatureMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComTemperatureHumidityMessage 
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - add support for four additional subtypes
- RFXComTemperatureHumidityBarometricMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Corrected byte of ForecastStatus.RAIN
- RFXComSecurity2Message 
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComSecurity1Message
  - add support for an additional subtype
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComRfyMessage
  - add support for an additional subtype
  - add support for four additional commands
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComRainMessage 
  - add support for an additional subtype
  - outgoing message had in incorrect bitshift
- RFXComPowerMessage 
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was incorrect here)
  - voltage gave some problems because it was converted to a negative number (from the unsigned byte, to a signed int)
- RFXComLighting6Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComLighting5Message
  - add support for three additional commands
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComLighting4Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComLighting3Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Length definition byte of outgoing messages was incorrect
- RFXComLighting2Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComLighting1Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Corrected support for Chime, group_on and group_off commands (factory method)
  - add support for three additional SubTypes
- RFXComInterfaceMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Corrected TransceiverType names
- RFXComHumidityMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - add support for an additional SubType
  - Length definition byte of outgoing messages was incorrect
- RFXComFanMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - add support for an additional SubType
- RFXComEnergyMessage
  - float are not good enough, even for this
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - bit shifts were out-of-range for an int an explicit cast to a long was needed (Sonar & Intellij helped me on this)
  - Encoded message was incorrect, doubleToRawLongBits does not convert the value to a long, it does something else for storing a double in a long value while maintaining the information.
- RFXComDateTimeMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
  - Sending an outgoing message gave an ArrayIndexOutOfBoundsException
- RFXComCurtain1Message
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComCurrentMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was incorrect)
- RFXComCurrentEnergyMessage
  - Encoded message was incorrect, doubleToRawLongBits does not convert the value to a long, it does something else for storing a double in a long value while maintaining the information.
  - added static factory method to SubType/Command/etc to map byteValues of Enums (was incorrect)
  - bit shifts were out-of-range for an int an explicit cast to a long was needed (Sonar & Intellij helped me on this)
  - Sending an outgoing message gave an ArrayIndexOutOfBoundsException
- RFXComChimeMessage
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComBlinds1Message 
  - added static factory method to SubType/Command/etc to map byteValues of Enums (for consistency)
- RFXComBaseMessage
  - added static factory method to PacketType to map byteValues of Enums (for consistency)


lolodomo:
After this PR is merged, could you please check what changes have to be applied to the 2.0 version of the binding?

martinvw:
Thanks for the reminder I already started my efforts to port this to
version 2. I will also try to port the test skeleton because that will
definitely help in keeping the quality good.

Op zondag 11 september 2016 heeft lolodomo <notifications@github.com> het
volgende geschreven:

martinvw:
He guys, anything I could or should do to make sure the pull requests can be reviewed and merged?

martinvw:
The oh2 pull request was approved, see: https://github.com/openhab/openhab2-addons/pull/1247

And I applied those recommended changes here too.

PS: Maybe I did do something wrong in the combination of merging with the master and rebasing my commits to a single one, any suggestions are appreciated.

bundles/binding/org.openhab.binding.rfxcom.test/src/test/java/org/openhab/binding/rfxcom/internal/messages/RFXComMessagesTest.java
```diff
@@ -33,7 +29,7 @@ public void checkNotImplementedPackets() throws Exception {
                 }
             }
             if (errors.length() > 0) {
-                throw new Exception("Packet messages not implemented : " + errors);
+                System.out.println("Packet messages not implemented : " + errors);
```
Jamstah: Is this left over debug output?

Jamstah:
Besides that one small review comment, LGTM.

bundles/binding/org.openhab.binding.rfxcom.test/src/test/java/org/openhab/binding/rfxcom/internal/messages/RFXComMessagesTest.java
watou: If you remove the System.out... and are pretty confident this PR does not introduce regressions, I'm OK to merge.  Are there wiki changes outstanding for this PR?

martinvw: In the original test an exception was thrown, which caused tests always to fail since not all RFXCom protocols are implemented and most likely never (or not in the near future) will be. The tests were also not included in the parent pom.

So I changed the behaviour to at least do a print out of any unsupported protocol, but not fail the test.

 I could change it to logging but I don't know whether a logging framework is already imported there. Any suggestions?

Jamstah:
Can't you tell the test to expect the exception for unimplemented protocols?

martinvw:
I added a list of message for which we know that they are not supported for now and throw an exception if a protocol which is not on that list becomes unsupported.

martinvw:
@Jamstah do you agree with my current solution?

Jamstah:
I'm not sure I fully get how the tests are running, is it looking every defined packet type and trying getMessageInterface on it? If we're now just skipping anything we know is unsupported and only checking those which are, isn't that no longer performing the tests for the method, which is called checkNotImplementedPackets?

We seem to have a test class for every packet type that is implemented, which already calls getMessageInterface in order to run those tests. I would have suggested creating a new test class for each of the unsupported packet types and having a single test along the lines of:

```
@Test(expected = RFXComException.class)
public void checkNotImplemented() throws Exception {
   RFXComMessageFactory.getMessageInterface(PacketType.whichever);
}
```

That way, each packet has a test class, and if someone implements a new packet without updating the tests, then the tests will complain about it, and they will be reminded to add some tests.

I might just be missing the point, I've only really skimmed over the code in the test project so do correct me if I'm wrong about how its working.

https://github.com/junit-team/junit4/wiki/exception-testing

martinvw:
The original test was added while not (yet) succeeding, so the exact long term goal of the test is (also?) not completely clear to me.

With the current solution, I prove that no support for previously supported PacketTypes broke (I did not add tests for all possible PacketType's so this adds value) and as side effect it forces someone to think about testing (or adding it to the ignore list) if he/she adds a new option to the PacketType ENUM (eg. because RFXCom adds support for an additional packet).

What I do not support (yet) is forcing users who introduce support for a previously unsupported PacketType to add a test and/or remove it from the unsupported list.

The last one can be forced by adding 11 of the tests you suggested or with the following test:
```java
for (PacketType p : acceptedToBeUnsupported) {
    try {
        RFXComMessageFactory.getMessageInterface(p);
        fail("This message type is/was marked as not supported"); // or some static import 
    } catch (Exception ignore) {
         // can be ignored, this exception is expected
    }
}
```

I think I prefer above over 11 very simple tests, the good thing is that it loops over the same list which is ignored in the other test. It could also be combined with the existing checkNotImplementedPackets method in one (more complex) test of course.

Jamstah:
Yes, that would have the same technical effect, I just tend to prefer the more declarative approach, to me it seems cleaner.

Saying that, I'd be happy with your approach (either in the same test or a second) and it would make the test actually checkNotImplementedPackets(), which is nice. Having said that, if we don't really see value in that test (I don't think its adding a lot) the other option is to just remove it entirely and forget about it.

I'd be happy with any of those options. If you do leave it in there, can you add a quick javadoc to say what the test(s) are testing and why, maybe we can help future devs avoid having the same discussion. And, one code review comment, it shouldn't just catch Exception, it should catch the RFXComException and perhaps even check the message is a "Not Implemented" message to make sure we don't mask other problems.

martinvw:
While I'm still working on it let's make it good:

For each class which is supported and where no additional tests are available, I'll add this:

```java
@Test
    public void checkForSupportTest() throws RFXComException {
        RFXComMessageFactory.getMessageInterface(PacketType.CHIME);
    }

    @Test
    public void basicBoundaryCheck() throws RFXComException {
        PacketType packetType = PacketType.CHIME;

        RFXComMessageInterface intf = RFXComMessageFactory.getMessageInterface(packetType);

        // This is a place where its easy to make mistakes in coding, and can result in errors, normally
        // array bounds errors
        byte[] message = intf.decodeMessage();
        if (message[0] != (message.length - 1)) {
            fail(intf.getClass().getName() + " wrong packet length");
        }

        if (packetType.toByte() != message[1]) {
            fail(intf.getClass().getName() + " wrong packet type");
        }
    }
```

For the last I can extract a method to some test helper.

And for all unsupported messages I'll create a separate class with the skeleton you suggested above.

I think we then have the best possible solution without changing any additional non-testing code.

martinvw:
@Jamstah Please review my commit (not yet rebased to give you some overview).

Thanks for your time and a pleasure to write great code and tests :-)

Jamstah:
Exactly what I was thinking, lgtm.

Your the one doing all the work, I'm just offering opinions! Great to work with you as well.

martinvw:
@watou anything you need or is it now ready to be merged?

No functional changes should be made so no wiki updates are necessary.

martinvw:
I had to make a new commit because of some markup difference in a change which was just merged with the repo, everything should be fine now.

martinvw:
Build fails on:

org.openhab.binding.enocean.internal.bus.RockerSwitchInDimmerOnOffProfileTest

RFXCom test succeed:

image

martinvw:
@watou is there a chance that this pull request is going to make it to the 1.9.0 milestone. Since I still see no labels added to it? Does it add value if I keep resolving merge conflicts as the appear?

9037568:
I see a lot of the new files are missing their OH copyright/licence header. Similarly also missing are the class doc-comments carrying the @author and @since attributes. Please add these where needed.

Also, if you haven't already, be sure to run all source through the Eclipse formatter.

bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComRainMessage.java
```diff
@@ -1,211 +1,217 @@
...
+    }
+
+    private final static List<RFXComValueSelector> supportedValueSelectors = Arrays.asList(RFXComValueSelector.RAW_DATA,
+            RFXComValueSelector.SIGNAL_LEVEL, RFXComValueSelector.BATTERY_LEVEL, RFXComValueSelector.RAIN_RATE,
+            RFXComValueSelector.RAIN_TOTAL);
+
+    public SubType subType = SubType.UNKNOWN;
+    public int sensorId = 0;
+    public double rainRate = 0;
+    /* public byte rain = 0; */
```
9037568: Removed commented-out code.

bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComWindMessage.java
```diff
@@ -1,253 +1,259 @@
...
+    public RFXComWindMessage() {
+        packetType = PacketType.WIND;
+    }
+
+    public RFXComWindMessage(byte[] data) {
+        encodeMessage(data);
+    }
+
+    @Override
+    public String toString() {
```
9037568: Consider refactoring this to use a StringBuffer for building up the return value.
Absolutely not required for this PR.

martinvw:
@9037568 thanks for your review, I was already planning to do some programming, so I get right onto it.

bundles/binding/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComWindMessage.java
martinvw: Somehow git / github blamed me for this whole file, anyways I fixed it.

martinvw:
All your remarks should now be resolved.

At this moment I'm the only one who tested this, and of course the newly added tests :)

I do agree that it adds a major value if it was tested by someone else from the community, what do you think will work better an invitation for that on your or my behalf?

Thanks for your time!

9037568:
I don't think I have any extra pull with the community. Just drop a post in an appropriate spot, maybe flag a couple users that have been known to use the binding. See what you get. If noone's interested in testing after a few days, we'll go ahead with the merge without them.

9037568:
LGTM.

martinvw:
Sorry, I had the commit with the actual changes still pending ...

9037568:
No worries. Thanks.

martinvw:
For people interested in tested the binding, I have a jar:
http://mtin.nl/org.openhab.binding.rfxcom-1.9.0-SNAPSHOT.jar

martinvw:
@9037568 unfortunately no response 👎

martinvw:
@9037568 Can you play a similar role for the pull request for OH2, its with the similar changes and adding of tests. The main difference is that a lot of messages are not yet supported there.

See: https://github.com/openhab/openhab2-addons/pull/1247

9037568:
As to this PR, I think it's safe to merge unless @watou has objections.

As to OH2, I haven't used it much and I have no permissions in that repo, so I'll have to leave that to others.

martinvw:
Okay, thanks for your help on this one!

watou:
Very impressive effort, @martinvw.  Thank you!  My only, completely optional comment regards

> RFXComEnergyMessage
> * float are not good enough, even for this

Switch from `float` to `double` improved that code, but did you consider using `BigDecimal` for those computations?  Other than that small question, it looks like you really thought through all the implications, so cheers!  LGTM.

martinvw:
Thanks for the review and the compliments.

Given the demands here and (IMHO) the code clutter (especially for untrained developers) which is caused by computations with BigDecimals I would like to keep it on the double's.

watou:
> Given the demands here and (IMHO) the code clutter (especially for untrained developers) which is caused by computations with BigDecimals I would like to keep it on the double's.

OK!  If `double` does the job, then that's all that's needed.

Issue #4565, opened 2016-08-03
[Plugwise] Add support for Scan, Sense, Stealth and Switch


wborn:
Adding support for other Plugwise device types would make migrating to openHAB (and Linux) a more viable option for people with currently unsupported of Plugwise device types. This is a feature I've already been working on myself because it is blocking for me to use openHAB.

The protocol is pretty straightforward. By handling several more message types (and adding some configuration), support can be added for:
* Scan
* Sense
* Stealth
* Switch

### Feature requirements
Below follows the list of requirements I currently have for this feature:
* Configuration remains backwards compatible so people can easily test and upgrade
* Static configuration of non-Circle device types through the configuration (.cfg) file to keep things simple for now
* Binding keeps working with openHAB 1.8.x+ and does not introduce new issues when used with openHAB 2.0
* Battery powered devices (Scan, Sense, Switch) push their information and are not polled for it
* Performance of the binding stays the same or is improved

### Feature limitations
* Adding new devices to the network and updating device settings (Scan, Sense) is not part of this feature and should be done via Source (similar to the current Circle implementation)
* Automatic device type detection is not part of this feature. That is something that should be added when the binding supports the new Inbox feature of openHAB 2.0. Detected devices are assumed to be Circles for backwards compatibility
* Status values are based on messages pushed from online devices. I.e. when messages are missed because a device is offline, no historic lookup is done in the Scan or Sense memory buffers when it rejoins the network.
* The switching groups defined in the Source are unknown to the binding. Devices will still react to them when configured

### Device type configuration in the configuration file
Below follows an example on how to configure device types in your configuration file.
openHAB 1.8: `configurations/openhab.cfg`
openHAB 2.0: `conf/services/plugwise.cfg`

```
############################## Plugwise Binding #######################################
#
# "stick" is reserved plug wise id
plugwise:stick.port=/dev/ttyUSB0
# interval in ms to wait between messages sent on the ZigBee network
#plugwise:stick.interval=150

#plugwise:<plugwise-id-1>.mac=

# you can also name your device circleplus and omit the line specifying the circleplus type (backwards compatibility)
plugwise:lamp.mac=000D6F0000011111
plugwise:lamp.type=circleplus

# the default device type is circle, so in this case the type could be ommitted (backwards compatibility)
plugwise:fan.mac=000D6F0000022222
plugwise:fan.type=circle

plugwise:motionsensor.mac=000D6F0000033333
plugwise:motionsensor.type=scan

[...]

PR #4644, opened 2016-09-15
[Plugwise] Add openHAB 2 feature addon

wborn:
While implementing #4565 I've done a lot of testing with the Plugwise Binding on OH2. So here are the OH2 feature addon and binding definition.


kaikreuzer:
Perfect, thanks a lot!

PR #4648, opened 2016-09-17
[Plugwise] NullPointerException may occur when concurrent threads get PlugwiseDevice from Stick

wborn:
When using a lot of Plugwise devices, occasionally a `NullPointerException` may occur. The root cause is that multiple threads simultaneously sort and search the `plugwiseDeviceCache`. This causes the binary search to be searching in the wrong part of the cache. I've made things a bit simpler by removing the sorting. It did not affect performance for me with tens of configured devices.


watou:
Thanks @wborn.  While the `plugwiseDeviceCache` could be made fancier by keeping two Maps, one keyed by MAC, the other by name, and only modified on add and clear, your approach doing a linear search is probably perfectly fine, as long as the number of devices in the cache is not in the hundreds.  At least the cache is not being sorted on read any more, which wasn't good for performance either. 😄

wborn:
Yes those were exactly my thoughts as well @watou! Also I opted not for creating the fancier solution because it results in more complex code and added risk for new regressions.

Issue #4653, opened 2016-09-19
[modbus] add usage of Rollershutter items 


yellowmm:
## Expected Behavior

The implementation may reflect the rules as shown below:

```
{ modbus:slave1:>readReg:<writeReg }
readReg:       Percent: 0-100 state of rollershutter set by MODBUS controller
writeReg:      UP/DOWN cmd (UP > 0 / DOWN < 0)
writeReg+1:    STOP cmd (value=1, must be reset by MODBUS controller)
```

The sample configuration (openhab.cfg) shall be as follows:

```
modbus:tcp.slave1.connection=192.168.178.1:502
modbus:tcp.slave1.id=1
modbus:tcp.slave1.type=holding
```

Important is to use the type=holding

`Rollershutter myRS "rollershutter" (livingRoom) {modbus="slave1:<0:>1"}
`
-> reg 0: is read register and keeps the actual shutter pos, must be written by MODBUS master
-> reg 1: is write register, value>0 for UP command, value<0 for DOWN command, MODBUS master must reset this value after processing
-> reg 2: is write register, value ==1 is STOP command, MODBUS master must reset this value after processing

Next used Item: 
`Rollershutter myRS2 "rollersutter2" (livingRoom) {modbus="slave1:<3:>4"}
`
## Current Behavior

[ERROR] [i.internal.GenericItemProvider] - Binding configuration of type 'modbus' of item ?myRollerShutter? could not be parsed correctly.
## Possible Solution

This enhancement needs to be implemented (my me).
## Steps to Reproduce (for bugs)

1.
2.
3.
4.
## Your Environment

- Version used: openhab (1.8.x and 1.9.x), org.openhab.binding.modbus
- Environment name and version (e.g. Chrome 39, node.js 5.4, Java 8, ...): n/a
- Operating System and version (desktop or mobile, Windows 10, Raspbian Jessie, ...): Raspbian Jessie
##

ssalonen:
Really great to see improvements from other developers!

I noticed the discussion in the forums and am delighted to see the pull request. I already put some code review comments there.

Generally speaking I'm interested hearing whether the approach of encoding the commands like described is something general or is it a "custom protocol". 

If it is custom, I would perhaps like to see it some how specified in the item binding both of the write registers. As proposed, this would be the first which writes to registers outside what is written to the config.

[...]

PR #4654, opened 2016-09-19
[modbus] enhance modbus binding for rollershutter items

yellowmm:
implement usage of RollerShutter item for modbus binding

address enhancement #4653


yellowmm:
Hi @watou, many thanks for your code review. Just took the mentioned changes to my fork and merged it accordingly.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
```diff
@@ -183,7 +187,20 @@ protected void internalReceiveCommand(String itemName, Command command) {
             if (provider.providesBindingFor(itemName)) {
                 ModbusBindingConfig config = provider.getConfig(itemName);
                 ModbusSlave slave = modbusSlaves.get(config.slaveName);
-                slave.executeCommand(command, config);
+                /* we need to handle rollershutter commands in a different way */
+                if (config.getItemClass() == RollershutterItem.class) {
```
ssalonen: Would it be better if this logic of handling roller shutter rules would be in the slave? Now I think some commands are handled in slave.executeCommand, some here directly

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
```diff
@@ -222,7 +239,10 @@ protected void internalUpdateItem(String slaveName, InputRegister[] registers, S
                 double tmpValue = ((DecimalType) newState).doubleValue() * rawDataMultiplier;
                 newState = new DecimalType(String.valueOf(tmpValue));
             }
-
+            if (config.getItemClass() == RollershutterItem.class) {
```
ssalonen: Could you combine this to the above...seems to be identical handling

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -135,15 +136,17 @@ public ModbusSlave(String slave, KeyedObjectPool<ModbusSlaveEndpoint, ModbusSlav
...
      * works only with types "coil" and "holding"
      *
-     * @param command OpenHAB command received
+     * @param command openHAB command received
      * @param config
+     * @param readRegister
+     * @param writeRegister
      */
-    public void executeCommand(Command command, ModbusBindingConfig config) {
+    public void executeCommand(Command command, ModbusBindingConfig config, int readRegister, int writeRegister) {
```
ssalonen: why the additional parameters `readRegister` and `writeRegister`? Shouldn't config include that?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -135,15 +136,17 @@ public ModbusSlave(String slave, KeyedObjectPool<ModbusSlaveEndpoint, ModbusSlav
...
         try {
             if (ModbusBindingProvider.TYPE_COIL.equals(getType())) {
-                setCoil(command, config);
+                setCoil(command, config, readRegister, writeRegister);
```
ssalonen: why the additional parameters `readRegister` and `writeRegister`? Shouldn't config include that?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -135,15 +136,17 @@ public ModbusSlave(String slave, KeyedObjectPool<ModbusSlaveEndpoint, ModbusSlav
...
             } else if (ModbusBindingProvider.TYPE_HOLDING.equals(getType())) {
-                setRegister(command, config);
+                setRegister(command, config, readRegister, writeRegister);
```
ssalonen: why the additional parameters `readRegister` and `writeRegister`? Shouldn't config include that?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -182,9 +185,8 @@ protected static boolean translateCommand2Boolean(Command command) {
      * @throws ModbusException ModbusIOException on IO errors, ModbusSlaveException with protocol level exceptions
      * @throws ModbusUnexpectedTransactionIdException when response transaction id does not match the request
      */
-    private void setCoil(Command command, ModbusBindingConfig config)
+    private void setCoil(Command command, ModbusBindingConfig config, int readRegister, int writeRegister)
```
ssalonen: why the additional parameters `readRegister` and `writeRegister`? Shouldn't config include that?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -198,10 +200,9 @@ private void setCoil(Command command, ModbusBindingConfig config)
      * @throws ModbusException ModbusIOException on IO errors, ModbusSlaveException with protocol level exceptions
      * @throws ModbusUnexpectedTransactionIdException when response transaction id does not match the request
      */
-    protected void setRegister(Command command, ModbusBindingConfig config)
+    protected void setRegister(Command command, ModbusBindingConfig config, int readIndex, int writeRegister)
```
ssalonen: why the additional parameters `readRegister` and `writeRegister`? Shouldn't config include that?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -243,6 +244,13 @@ protected void setRegister(Command command, ModbusBindingConfig config)
             } else if (command.equals(OpenClosedType.CLOSED)) {
                 newValue.setValue(0);
             }
+        } else if (command instanceof StopMoveType) {
+            newValue = new SimpleRegister();
+            if (command.equals(StopMoveType.MOVE)) {
```
ssalonen: Is this intentional that MOVE=0 and STOP=1. Is the logic inversed?

watou:
Can I suggest that there be a discussion of the objectives along the lines that @ssalonen suggested that can generalize a solution that would apply to all modbus users?  Or in the alternative, an action service for modbus, so rules could call actions like below from a rule or script?
```
modbusExecuteCommand(slave, command, args...)
modbusSetRegister(...)
modbusSetCoil(...)
```
so instead of binding the Rollershutter item to the modbus binding, a rule could be triggered based on STOP, MOVE, UP, DOWN, etc., and the rule would call modbus action calls to perform the user-specific logic?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
yellowmm: from the shutter point of view it makes indeed sense to trigger for a stop command != 0 since the movement (UP/DOWN) is handled by different type.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
yellowmm: You are right, they are. 
But to the RollerShutter item working (and to keep it as simple as possible inside the modbus-server configuration) I decide to use 2 consecutive write registers to handle UP/DOWN and STOP commands. Here it was mandatory to define a offset respectively.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
yellowmm: see comment on setRegister(...)

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
yellowmm: In principle yes. But I decide to implememt it here because there is currently no Item depedency inside the slave.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
yellowmm: Sorry. I do not really understand what you exactly mean here?

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusGenericBindingProvider.java
```diff
@@ -73,10 +74,13 @@ public void validateItemType(Item item, String bindingConfig) throws BindingConf
...
             return;
         }
+		if (item.getClass() == RollershutterItem.class) {
+			return;
+        }
 
         throw new BindingConfigParseException(
                 "item '" + item.getName() + "' is of type '" + item.getClass().getSimpleName()
-                        + "', only Switch, Contact or Number are allowed - please check your *.items configuration");
+                        + "', only Switch, RS, Contact or Number are allowed - please check your *.items configuration");
```
yellowmm: Already fixed

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -198,16 +198,15 @@ private void setCoil(Command command, ModbusBindingConfig config)
...
-        int readIndex = config.readIndex;
-        int writeRegister = getStart() + config.writeIndex;
+        writeRegister += getStart();
 
         Register newValue;
         if (command instanceof IncreaseDecreaseType) {
             newValue = readCachedRegisterValue(readIndex);
             if (newValue == null) {
-                logger.warn("Not polled value for item {}. Cannot process command {}", config.getItemName(), command);
+                logger.warn("Not polled value. Cannot process command {}", command);
```
yellowmm: My 1st approach was w/o the config param in this call to keep the API small. 
Since it is back I can revert this.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java
```diff
@@ -218,7 +217,7 @@ protected void setRegister(Command command, ModbusBindingConfig config)
         } else if (command instanceof UpDownType) {
             newValue = readCachedRegisterValue(readIndex);
             if (newValue == null) {
-                logger.warn("Not polled value for item {}. Cannot process command {}", config.getItemName(), command);
+                logger.warn("Not polled value. Cannot process command {}", command);
```
yellowmm: My 1st approach was w/o the config param in this call to keep the API small. 
Since it is back I can revert this.

yellowmm:
The implementation effort was quite low to add the Rollershutter binding into the modbus binding. All other solutions were too difficult (at least for me) to manage this with rules and so forth. I can say that this enhancement works w/o any issues since more than 1 year. And now I was forced to make this enhancement public due to several demands from other users.

bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java
ssalonen: I understand what you mean. However, there are very similar cases in Slave already, for example:
````
        } else if (command instanceof OnOffType) {
            newValue = new SimpleRegister();
            if (command.equals(OnOffType.ON)) {
                newValue.setValue(1);
            } else if (command.equals(OnOffType.OFF)) {
                newValue.setValue(0);
            }
````

I would keep the command handling in same place, would be easier for developer since the logic would be in one place?

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java (hunk 2 above)
Comment (ssalonen) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusBinding.java]: Sorry, I misread the number handling. This is slightly different case here so this is OK.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 7 above)
Comment (ssalonen) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: I might have misunderstood but could you have the offsets in the body of `setRegister`, depending on `command` class?

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 10 above)
Comment (ssalonen) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Would be nice if reverted

**Comment by ssalonen** [comment]
I wanted to try out how the solution would look like with standard modbus binding, and just using rules. I have something like this:

## default.rules
````
import org.openhab.core.library.types.DecimalType
import org.openhab.core.library.types.PercentType


rule "Process roller shutter commands"
when
	Item myRS received command
then
	logInfo("rollerShutterRules", "Processing RS command")
	if(receivedCommand != null){        
        switch(receivedCommand.toString.upperCase) {        
            case "UP" : {
                sendCommand(myRSUpDown, 1)
            }
            case "DOWN":{
                sendCommand(myRSUpDown, -1)
            }
            case "STOP":{ 
                sendCommand(myRSMoveStop, 1)
            }   
            case "MOVE":{
            	sendCommand(myRSMoveStop, 0)
            }
        }
    }
	
end


rule "Update roller shutter position"
when
	Item myRSMoveStop received update
then
	logInfo("rollerShutterRules", "Processing position update")
	var currentPosition = myRSMoveStop.state as DecimalType
	postUpdate(myRS, new PercentType(currentPosition.toBigDecimal))
end
````

## default.items
````

// Main roller shutter item. 
// - State will be updated from register index 0. 
// - Up (1) & Down (-1) commands will be written to register index 1
// - Move (0) & Stop (1) commands will be written to register index 2
Rollershutter myRS "RollerShutter" (livingRoom)

// Helper items for writing data to MODBUS
Number myRSUpDown "RollerShutter Up/Down Control (writing to index 1, read from index 0)" (livingRoom) {modbus="slave1:<0:>1"} 
Number myRSMoveStop "RollerShutter Move/Stop Control (writing to index 2, read from index 0)" (livingRoom) {modbus="slave1:<0:>2"} 
````

## default.sitemap
````
sitemap demo label="Main Menu" {
    // https://github.com/openhab/openhab/wiki/Rollershutter-Bindings
    Switch item=myRS label="Roller shutter [(%d)]" mappings=[UP="up", STOP="X", DOWN="down"]
    Text item=myRSMoveStop
    Text item=myRSUpDown
}
````

## For local testing

Starting up modbus tcp server (i.e. modbus slave)
````
./diagslave -m tcp -p 55502
````

openhab.cfg

````
modbus:tcp.slave1.connection=localhost:55502:0:0:0:1
modbus:tcp.slave1.type=holding
modbus:tcp.slave1.length=20
modbus:tcp.slave1.postundefinedonreaderror=true
````

As inspiration I used [Roller shutter wiki page](https://github.com/openhab/openhab/wiki/Rollershutter-Bindings)


I'm still thinking different ways to handle the roller shutter in general. Wouldn't some people want to just write the "target position" of the roller shutter? E.g. when you would like to lower the roller shutter by a bit you would input 80% as command?

I see four alternatives here: 
1. we go with your approach. Would like to see that this kind of roller shutter handling is "just" one of the alternatives. This would guarantee future proofness and extensions
2. Extend item binding config syntax such that you could map command(s) -> (register, value(s)). For example, in the example above you could say that 
 - UP will be converted to value=1, and written to register=1
 - DOWN will be converted to value=-1, and written to register=2
 - etc.
3. Document the rule item configuration as above (if it works for you) and make no changes to implementation
4. Provide general purpose actions (as [suggested by @watou](https://github.com/openhab/openhab/pull/4654#issuecomment-248372866)). Basically we could use the above rule/item solution but we would not need to have the extra "helper" items.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 10 above)
Comment (yellowmm) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Reverted.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 11 above)
Comment (yellowmm) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Reverted.

**Comment by watou** [comment]
@yellowmm @ssalonen Any updates for this PR?  Thanks!

**Comment by ssalonen** [comment]
No updates that I know of. Been waiting for discussion to continue...

Btw I am working on transformation support for modbus binding which would allow to implement alternative 2 above. But that's not here yet so it's not an option yet.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 5 above)
Comment (yellowmm) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: see comment on setRegister(...) below

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 4 above)
Comment (yellowmm) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: see comment on setRegister(...) below

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 5 above)
Comment (ssalonen) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Please see comment I also commented below a while ago.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 4 above)
Comment (ssalonen) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Please see comment I also commented below a while ago.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 6 above)
Comment (ssalonen) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Please see comment I also commented below a while ago.

File: bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java (hunk 7 above)
Comment (yellowmm) [inline bundles/binding/org.openhab.binding.modbus/src/main/java/org/openhab/binding/modbus/internal/ModbusSlave.java]: Your idea is to **remove** the offset calculation _RollershutterItem.class_ of _ModbusBinding::internalReceiveCommand_

```
       if (config.getItemClass() == RollershutterItem.class) {
            if (command instanceof StopMoveType) {
                /* the second write register assigns the stop_move condition */
                slave.executeCommand(command, config, config.readIndex, config.writeIndex + 1);
                return;
            }
            if (command instanceof UpDownType) {
                /* wont read from percent read register */
                slave.executeCommand(command, config, config.readIndex + 1, config.writeIndex);
                return;
           }
```
... **remove** all the new read/write index parameters ...
and take care about the required index calculation in _ModbusSlave::setRegister_
```
        } else if (command instanceof StopMoveType) {
	    if (config.getItemClass() == RollershutterItem.class) {
	        /* the second write register assigns the stop_move condition */
		writeRegister++;
            }
            newValue = new SimpleRegister();
            if (command.equals(StopMoveType.MOVE)) {
                newValue.setValue(0);
            } else if (command.equals(StopMoveType.STOP)) {
                newValue.setValue(1);
            }
        } ...
```
and 		
```
        } else if (command instanceof UpDownType) {
            if (config.getItemClass() == RollershutterItem.class) {
	        /* wont read from percent read register */
	        readIndex++;
	    }
            newValue = readCachedRegisterValue(readIndex);
            if (newValue == null) {
                logger.warn("Not polled value for item {}. Cannot process command {}",   config.getItemName(), command);
                return;
            }...
```
 
But please keep in mind that a **new** package needs to be imported in _ModbusSlave.java_:
`import org.openhab.core.library.items.RollershutterItem;`

Please let me know whether this could be acceptable for you. Then I can change it accordingly.

yellowmm: Sure, I can change it as described in the source-code snippeds.

Do you mean by implementing the Rollershutter by rules and leave the code as it is?

ssalonen: These 

> Wouldn't some people want to just write the "target position" of the roller shutter? E.g. when you would like to lower the roller shutter by a bit you would input 80% as command?

And

> I see four alternatives here:...SNIP...

yellowmm:
Some words to the given summary above.

> Wouldn't some people want to just write the "target position" of the roller shutter? E.g. when you would like to lower the roller shutter by a bit you would input 80% as command?

This quite easy to implement when the roller shutter item sets the current position in time. When someone has such requirement it can be done be checking the current position == desired position. If condititon matches, the function sends a STOP command.

> I see four alternatives here:

Since I am not a maintainer it is finally up to you to decide. From my point of view I want to change the code as requested above and leave it as it is. What do you think? Is there anything else to do to finalize this PR?

ssalonen:
Actually I meant in some cases the user does not want openhab to be the low level controller. Analogy in temperature control would be to set thermostat setpoint vs controlling On/Off manually. 

This relates to [my comment some months](https://github.com/openhab/openhab/pull/4654#pullrequestreview-617470) ago that different users have different ways to encode the rolleshutter commands to the modbus registers. At least there was a user with different approach already.

If I understand correctly this implementation makes the choice for the user, and provides no way for future implementations to provide an alternative way to store the data to modbus. I suggested some ways to explicitly to tell the binding what kind of way to store the data to modbus. 

Not sure if you noticed my comment above or perhaps I missed the reply?

Anyway, please note that I'm actually not a maintainer, just an openhab enthuastist contributing to the modbus binding :) so the decision comes from the official maintainers eventually.

watou:
Is there a problem that prevents the more general solution using rules that @ssalonen suggested?  If you implemented a rules-based version and shared that with other users, that could produce the same outcome but leave control in the openHAB users' hands. WDYT?

yellowmm:
@watou Finally there is not a problem to leave the implementation as it is. But I want to emphatize that it is very difficult (or even not possible for beginners) to write all the rules and configs required for the shutter binding. I know what I'm talking about, because as I have started with openhab I had exactly these [problems](https://community.openhab.org/t/modbus-binding-of-rollershutter/3273) and I found absolutely no solution for all my problems. Even not by the community. That's why I was looking for a simple way to integrate shutter binding with modbus. The PR is the result.

yellowmm:
@ssalonen 

> Actually I meant in some cases the user does not want openhab to be the low level controller. Analogy in temperature control would be to set thermostat setpoint vs controlling On/Off manually.

Understood. But this PR shall provide a simple interface to bind shutters by modbus. If someone prefers rules and so forth - still possible. 
By the way, we can easily add a further write register to set a dedicated target position. But all these stuff makes it more complex.

> If I understand correctly this implementation makes the choice for the user...

Exactly. From my point of view this is a benefit and a proven point to start if you are not an expert.

> ... and provides no way for future implementations to provide an alternative way to store the data to modbus. I suggested some ways to explicitly to tell the binding what kind of way to store the data to modbus.

If this is the base understanding to keep everything as flexible as possbile, this PR is completely wrong and obsolete. 
But (I repeat myself) for me it was important to provide a simple way (unfortunately not much flexible) for the outer world to bind shutters w/o usage of many rules to modbus.

watou:
@yellowmm I can absolutely understand your experience with configuration and rules.  Assuming the use of rules and "helper items" like @ssalonen's example can address your use case, it really is the most flexible approach.

I think that a few clear examples, including one that addresses your use of a rollershutter, would be very helpful to lots of Modbus binding users.  Either you, @ssalonen (I'm just asking 😄 ) or myself could create a [Modbus Examples](https://github.com/openhab/openhab/wiki/Modbus-Examples) wiki page that contains a table of contents followed by several examples of items, rules and sitemaps (similar maybe to [this one](https://github.com/openhab/openhab/wiki/Ecobee-Examples)).  Sami's above implementation could be the first entry, but anyone who has a configuration that is likely to help others could be added by that person.  What do you both think?

yellowmm:
When I read between the lines from the last comments it seems that the PR has not the best chance getting approved? 

Sure, this is a great idea and this would help to find a proper solution which finally could work.

watou:
> When I read between the lines from the last comments it seems that the PR has not the best chance getting approved?

All I can say is that I prefer a more flexible solution, but I am just a maintainer with an opinion!  

Rules and helper items I think can work, but it would be great to have a compact way to map the full semantics of openHAB items to Modbus concepts (@ssalonen's option 2) that users of the binding could modify instead of representing it in binding code that users can't adapt easily to their implementations.  I know very little about Modbus myself, and the last thing I want to do is put up unnecessary barriers to solutions — I just want to make sure sufficient discussion has taken place before a decision is made.

@ssalonen, among all options, what do you think should be done going forward?

At this point let me say that I think my earlier suggestion of Modbus action calls from rules (option 4) would advance to goal of simplicity only a small bit, and is not much of an improvement over "helper items" as in the example above.

yellowmm:
Please also keep comments from community like [this one](https://community.openhab.org/t/modbus-binding-of-rollershutter/3273/35) in mind when thinking about possible scenarios...

ssalonen:
Yeah I tend to agree with you @watou. I personally think that the bindings should provide flexible interface between openhab and the external system, in a opinionated way. Of course we should strive for having common use cases fulfilled as easily as possible, and hopefully without complex rules.

However, considering the nature of modbus (low level protocol, no complex datatypes or generally agreed data conventions), I think we should avoid implementing "one-offs" to fulfill certain use cases. In the end, the good intention(s) might show as complex result in the end? As a developer I see the end result harder to maintain as well. In the end, I have seen that openhab does have a quite strict policy on not breaking backwards compatibility, which means that every feature/functionality should be supported in the future as well (without configuration changes). Unfortunately I think that this PR might have the aforementioned characteristics.

It's quite challenging for me to come to this conclusion as I truly value these contributions and the effort required to contribute to an open source project like openhab.

I have kept the option 2 in mind when implementing transformation support for the openhab (public in [my github](https://github.com/ssalonen/openhab/tree/modbus-transformation-support)). I think this would fullfill the original need with a simple and easy way. Naturally we need to open up common use cases still with proper examples.

In the meanwhile (before we do have the transformation), we should resolve the obvious immediate need by the users. I think the idea of the examples page was great idea. @watou would you mind creating the page? I might be quite busy till end of year...

Best regards,
Sami

watou:
@yellowmm @ssalonen I've added [Modbus Examples](https://github.com/openhab/openhab/wiki/Modbus-Examples) to the wiki.  Please correct and clarify the example as needed by editing the wiki page.  @yellowmm, if the example there addresses the same needs as this PR, I think it makes the most sense to close this PR, and with the hope that this course of action makes the most sense to you as well.  Thanks to you both for your continued contributions.  Regards, John

ssalonen:
Thanks, the example looks nice! Great that you added openhab2 examples as well.

Would like hear what  @yellowmm thinks!

Best
Sami

yellowmm:
Now I took the _bitter pill_ and modified the according rules and items I need to control my shutters. 
It works as described above in the sample w/o any problems.

watou:
@yellowmm @ssalonen Are we agreed that this PR can be closed (with big thanks to both of you)?

yellowmm:
From my point of view, yes.

watou:
Thanks very much @yellowmm.

ssalonen:
For the record, 

I now published experimental version implementing transformation support in the [community forums](https://community.openhab.org/t/modbus-support-for-transformations-roller-shutter-items-read-only-items-write-only-items-and-others/21500). Due to the extended item configuration syntax, the use case should be now covered.

I appreciate all the feedback!

PR #4658, opened 2016-09-23
TCP/UDP Binding enhancement 'MQTT style'

ppt000:
The TCP binding currently does not allow straight string literals to be passed as message, it can only done through MAP files or other non direct means. The MQTT binding gives more options: "default", transformation or straight strings.  This PR takes from the MQTT Binding and applies the same rules while maintaining as much as possible the current code. The private method splitTransformationConfig is removed and its functionality incorporated directly into the transformResponse method.

### DIFF
--- a/bundles/binding/org.openhab.binding.tcp/src/main/java/org/openhab/binding/tcp/protocol/internal/TCPBinding.java
+++ b/bundles/binding/org.openhab.binding.tcp/src/main/java/org/openhab/binding/tcp/protocol/internal/TCPBinding.java
@@ -236,53 +236,40 @@ public class TCPBinding extends AbstractSocketChannelBinding<TCPBindingProvider>
     protected void configureChannel(Channel channel) {
     }
 
-    /**
-     * Splits a transformation configuration string into its two parts - the
-     * transformation type and the function/pattern to apply.
-     *
-     * @param transformation the string to split
-     * @return a string array with exactly two entries for the type and the function
-     */
-    protected String[] splitTransformationConfig(String transformation) {
-        Matcher matcher = EXTRACT_FUNCTION_PATTERN.matcher(transformation);
-
-        if (!matcher.matches()) {
-            throw new IllegalArgumentException("given transformation function '" + transformation
-                    + "' does not follow the expected pattern '<function>(<pattern>)'");
-        }
-        matcher.reset();
-
-        matcher.find();
-        String type = matcher.group(1);
-        String pattern = matcher.group(2);
-
-        return new String[] { type, pattern };
-    }
 
     protected String transformResponse(String transformation, String response) {
         String transformedResponse;
-
-        try {
-            String[] parts = splitTransformationConfig(transformation);
-            String transformationType = parts[0];
-            String transformationFunction = parts[1];
-
-            TransformationService transformationService = TransformationHelper
-                    .getTransformationService(TCPActivator.getContext(), transformationType);
-            if (transformationService != null) {
-                transformedResponse = transformationService.transform(transformationFunction, response);
-            } else {
+        Matcher matcher = EXTRACT_FUNCTION_PATTERN.matcher(transformation);
+        String transformationServiceName;
+        String transformationServiceParam;
+        
+        if (StringUtils.isEmpty(transformation) || transformation.equalsIgnoreCase("default")) {
+        	transformedResponse = response;
+        } else if (matcher.matches()) {
+            matcher.reset();
+            matcher.find();
+            transformationServiceName = matcher.group(1);
+            transformationServiceParam = matcher.group(2);
+            try {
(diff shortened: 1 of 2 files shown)


watou:
Hi Paolo, I like the direction you're taking.  Question, though: In the [wiki page](https://github.com/openhab/openhab/wiki/TCP-Binding), it says

> Note: The examples here below are for the TCP protocol. UDP binding configuration are exactly the same, one only has to substitute "tcp=" with "udp="

So will your changes also apply to the UDP binding in the same bundle?

ppt000:
You are right, the same changes should apply to the UDP binding. I have not checked it yet but I assume it is roughly the same code. I'll have a look and propose the changes there also. By the way, I have made a little mess of these pull requests, the latest one is back on my first branch.

ppt000:
This is the latest PR that takes care of both TCP and UDP bindings, removes the unnecessary methods and has similar syntax than the MQTT binding for the transformation field.

watou:
Hi Paolo, looks good, but I have made a suggestion to you here: ae35535e8812a15730ff0422aa2c8fc6fa549fe7 to fix formatting, slight reorganisation, and use parameterized logging instead of string concatenation.  Regardless of my suggestion, if you've tested your change and are able to update the wiki, then I think it's OK to merge.

ppt000:
I have updated the wiki. As far as testing goes, I have not done much frankly, I don't have a functioning full environment yet to compile the whole thing.  Let me know what needs to be done to satisfy the testing requirement. Also, I don't know if your changes are already incorporated or if I have to commit something to make them part of this PR.

watou:
Hi Paolo, please try [this test JAR](https://dl.dropboxusercontent.com/u/4286376/tcp-test/org.openhab.binding.tcp-1.9.0-SNAPSHOT.jar) which incorporates your changes and my changes, to make sure they work as intended.  If they do, I will either fix up your PR or make a new PR (crediting you) with these changes.  Thanks.

ppt000:
Thanks John.

Is the JAR supposed to replace the 1.8.2 binding or do I have to put 
this with the old one in the addons folder?  The new one on its own does 
not seem to start the TCP service at all. Sorry for the basic questions...

watou:
No worries, Paolo.  You would remove the 1.8.2 binding from the addons folder, and put the test JAR in its place.  Possibly restarting openHAB might be needed, but usually that's not required.  I don't know of a reason why the test JAR should not operate in place of the official one under openHAB 1.8.*.

ppt000:
So I had to restart openhab. Maybe removing first the old JAR and 
replacing afterwards with the new JAR would have worked automatically? 
Don't know and didn't test that, anyway it works now.

The new code all works fine for me, but I had limited cases to start 
with to be fair: a switch with a MAP transformation (which I kept) and a 
new switch with straight text. Both work the same, no errors or warns in 
the log.  Is there a way to give it to test by others before release or 
do you want me to do more tests? It will take some time though.

Separately, just a comment: I just realised that the parsing of the 
whole item string is quite advanced (in 
ProtocolGenericBindingProvider.java 
<https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.tcp/src/main/java/org/openhab/binding/tcp/protocol/internal/ProtocolGenericBindingProvider.java>) 
as it is based on regex checks rather than more basic string checks as 
in the MQTT binding.  As a consequence I realised that the command can 
have single quotes or not, only the content is taken anyway, as the 
pattern accommodates that.  I will adapt the wiki just for precision sake.

watou:
> Is there a way to give it to test by others before release or do you want me to do more tests? It will take some time though.

Good idea! If you post on the community site for others to test the test JAR, referring to the new options, hopefully others will give it a go and report their findings.

> As a consequence I realised that the command can
have single quotes or not, only the content is taken anyway, as the
pattern accommodates that.  I will adapt the wiki just for precision sake.

Yes, the part of the pattern `(?:'(.*)'|(.*))` seems to make the single quotes optional before it even gets to your new code, so the net effect of your changes are to allow an empty string or `default` or a literal string that doesn't match the TEXT1(TEXT2) pattern.  And thanks for your precision in updating the wiki -- many will benefit from good documentation!

ppt000:
Started the new topic 
<https://community.openhab.org/t/test-enhancement-in-tcp-udp-binding/14507>as 
suggested. Thanks for your help.

watou:
@ppt000 Hi Paolo, any chance you could make one more push to get tester feedback on this PR?  It's probably just fine, but I'm stuck between having little time myself, and always nervous about some small regression sneaking past us...  There was recently forum discussion on updating the HTTP binding with the same objective, so at least this is worthwhile effort for others. :-)

watou:
Thanks Paolo for your improvement.

ppt000:
You are welcome! and thanks to you for the guidance. See you soon for some other PR...

PR #4685, opened 2016-10-07
Fix #4665 by emitting an error when an HTTP command fails to be created.

### LINKED ISSUE #4665: [HttpBinding] Http commands fail silently for illegal URLs
## Current Behavior

The HTTP binding fails to send an HTTP command when it encounters a bad/malformed/illegal URL, but does not report any error in the log files to indicate that the failure occurred.

(This problem does not occur in OpenHAB 2.x)
## Expected Behavior

The binding should log the error so that the user can tell the command failed.
## Possible Solution

The problem occurs in the `org.openhab.io.net.http.HttpUtil` class.  My proposal is to catch and log the error.  I have the code change almost done...
1. Create an HTTP "out" binding with an illegal URL
   eg.

`{http=">[*:POST:http://www.google.com/home/lights/23873/?status=%2$s&type=\"text\"]"}`
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37992514-httpbinding-http-commands-fail-silently-for-illegal-urls?utm_campaign=plugin&utm_content=tracker%2F482064&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F482064&utm_medium=issues&utm_source=github).
</bountysource-plugin>

9037568:
Hopefully this commit was done properly into my own fork...

### DIFF
--- a/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/http/HttpUtil.java
+++ b/bundles/io/org.openhab.io.net/src/main/java/org/openhab/io/net/http/HttpUtil.java
@@ -41,7 +41,6 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * Some common methods to be used in both HTTP-In-Binding and HTTP-Out-Binding
  * 


watou:
Hi Chris, it looks like this PR is on the 1.8 branch, so it would only ever see the light of day if there were a 1.8.4 release in the future, which seems a bit unlikely to me at present.  

Separately, it looks like your earlier commit cb1eb414c90764ec8b9f4bb2aee74e63c92547ed was made directly onto the 1.8 branch, so if you carefully revert it, we can get the 1.8 branch back to its previous state.

9037568:
True enough, this can only be fixed on the 1.8 branch.

I reverted cb1eb41, so the 1.8 branch should be back to its original state at this point.

PR #4678, opened 2016-10-05
Formatting and add comments

xmor:

PR #4686, opened 2016-10-07
Formatting and add comments

xmor:
Yes for both


watou:
Hi @xmor.  Does this PR replace #4678, and also, is it formatted per the Eclipse IDE settings that result from [this IDE setup](http://docs.openhab.org/developers/development/ide.html)?

9037568:
The formatting is not correct. Would you run the Eclipse formatter and repush, @xmor ?

Issue #4689, opened 2016-10-07
[lcn] missing possiblity to send status variable values


crosenbe:
The binding does not allow to provide variable values sent to group 4, means pushing own values as PCHK to the bus. That kind of values could be fetched by the modules itself.

At the moment it's only possible to send values to variables of the modules with resetting them first and change it relative afterwards. n at community.openhab.org first and include link to topic -->

An alternative could be the X2 command to provide values to the bus.

watou:
The LCN binding appears to be originally authored by @Issendorff .  Would you be able to test some code changes and if successful, submit a pull request?

PR #4690, opened 2016-10-08
Added X2 command to send status messages with variable values (#4689)

crosenbe:


bundles/binding/org.openhab.binding.lcn/src/main/java/org/openhab/binding/lcn/common/PckGenerator.java
```diff
@@ -256,6 +256,29 @@ public static String varAbs(LcnDefs.Var var, int value) throws IllegalArgumentEx
...
+     * @param var the target variable to set
+     * @param value the absolute value to set
+     * @return the PCK command (without address header) as text
+     * @throws IllegalArgumentException
+     */
+    public static String updateStatusVar(LcnDefs.Var var, int value) throws IllegalArgumentException {
+        int id = LcnDefs.Var.toVarId(var);
+        if (id != -1) {
+            // define variable to set, offset 0x01000000
+            int x2cmd = id |= 0x40;
```
watou: It doesn't affect the result, but it would be clearer to use the simple bitwise OR operator (`|`) instead of the bitwise inclusive OR and assignment operator (`|=`).

watou:
Thank you @crosenbe!  Do you know of another user of this binding who could confirm your changes?  Just to be sure!  Please confirm when you are fairly certain this change fixes #4689 and won't lead to regressions for other users of this binding.  Thanks!

bundles/binding/org.openhab.binding.lcn/src/main/java/org/openhab/binding/lcn/common/PckGenerator.java
crosenbe: fixed

crosenbe:
I asked another user for additional testing this feature but there is no feedback until now. I could confirm this patch fix #4689.
Because the if statement "if (this.addr.getId() == 4 && this.addr.isGroup())" only messages to group number 4 are affected. Group number 4 is reserved to send status messages, like described in LcnAddrGrp.java.

```
        // grpId:
        // 3 = Broadcast, 4 = Status messages, 5..254
```

PR #4693, opened 2016-10-09
Documentation to X2 send status variable values (#4689)

crosenbe:


watou:
Thanks @crosenbe!  Would you mind also updating [the wiki](https://github.com/openhab/openhab/wiki/LCN-Binding)?  Also, please note both in this PR and in the wiki that your enhancement pertains to binding version 1.9 and later.

9037568:
I don't see anything here or in the wiki referencing version 1.9 and later, per @watou. Does this doc change apply to all versions?

PR #4696, opened 2016-10-11
Fix dimming bug in MiLight Binding

hmerk:
Signed-off-by: Hans-Jörg Merk hans-joerg.merk@t-online.de


hmerk:
Jenkins failed at XMPP Action, Don't know why

bundles/binding/org.openhab.binding.milight/src/main/java/org/openhab/binding/milight/internal/MilightBinding.java
```diff
@@ -341,6 +335,11 @@ private PercentType sendIncrease(int bulb, int rgbwSteps, String bridgeId) {
             newValue = new PercentType(newPercent);
             logger.debug("milight: Bulb '{}' getting increased to '{}'", bulb, newPercent);
         }
+        try {
+            sendOn(bulb, bridgeId);
+            Thread.sleep(100);
+        } catch (InterruptedException e) {
```
watou: Is it correct to catch and "swallow" the InterruptedException (and below as well)?

watou: I see that is was moved/copied, but I worry that the thread ought to wrap up what's doing, set the interrupted flag again (since it was cleared when the exception was thrown and the rest of the call stack might want to know).

[This](http://www.yegor256.com/2015/10/20/interrupted-exception.html) seems like a good description of the nuances of InterruptedException.  Excerpt:

>  If InterruptedException is thrown, it means someone checked the flag and our thread has to finish what it's doing ASAP.

> The owner of the thread doesn't want to wait any longer. And we must respect the decision of our owner.

> Thus, when you catch InterruptedException, you have to do whatever it takes to wrap up what you're doing and exit.

PR #4704, opened 2016-10-13
[Plugwise] FindBugs issues fix

wborn:
Fixes:
* Locks that are not released when exceptions occur
* Potential NPEs
* Wrong String.format parameters

Signed-off-by: Wouter Born <github@maindrain.net>


watou:
Thanks @wborn!

Issue #4714, opened 2016-10-20
[MPD Binding] Doesn't handle out-of-range volume values correctly


9037568:
## Expected Behavior
If OH receives values for the Volume that are outside of the allowed range (0-100), it should ignore them, instead of attempting to use them.  There should probably be a warning message logged to let the user know.

## Current Behavior
When OH receives values for the Volume that are outside of the allowed range, it attempts to use them. This can happen, for example, when the MPD audio outputs are misconfigured and MPD reports a value of -1.

## Possible Solution
I have created a fix for this that checks for values less than 0 or greater than 100 and ignores them, as well as issuing log warnings about the invalid values.

## Steps to Reproduce (for bugs)
1. Don't configure an audio output in MPD

watou:
The MPDPlayer class setVolume method will force negative numbers to 0 and numbers over 100 to 100, so is the user experience that a -1 is passed to setVolume and is muting the volume, or what exactly?

9037568:
Well, as you noted, most of the changes I made for this weren't needed, since the player class was handling most of the out of range values.

The remaining issue is when OH receives a volumeChanged() event with an invalid value.  During startup, OH receives a value of -1 from the MPD server.  volumeChanged() is subsequently called.  A user watching the live output will see this:
```
20:56:57.545 [DEBUG] [inding.mpd.internal.MpdBinding:485  ] - Volume on emerald changed to -1
Exception in thread "MPD Monitor (player:emerald)" java.lang.IllegalArgumentException: Value must be between 0 and 100
        at org.openhab.core.library.types.PercentType.validateValue(PercentType.java:48)
        at org.openhab.core.library.types.PercentType.<init>(PercentType.java:33)
        at org.openhab.binding.mpd.internal.MpdBinding.volumeChanged(MpdBinding.java:489)
        at org.bff.javampd.monitor.MPDStandAloneMonitor.fireVolumeChangeEvent(MPDStandAloneMonitor.java:272)
        at org.bff.javampd.monitor.MPDStandAloneMonitor.checkVolume(MPDStandAloneMonitor.java:616)
        at org.bff.javampd.monitor.MPDStandAloneMonitor.run(MPDStandAloneMonitor.java:391)
        at java.lang.Thread.run(Unknown Source)
```

So the problems here are:
1) The volume was NOT changed to -1, despite the debug log message claim to the contrary.
2) The exception is NOT recorded in the log files, so is effectively lost.
[...]

PR #4715, opened 2016-10-20
Fix MPD handling of invalid volume levels

9037568:
Ignore invalid volume levels.
Add enhancements to log messages (including eliminating string concatenation).

(For #4714)


bundles/binding/org.openhab.binding.mpd/src/main/java/org/openhab/binding/mpd/internal/MpdBinding.java
```diff
@@ -219,10 +219,20 @@ private void executePlayerCommand(String playerCommandLine, Object commandParams
...
+                        }
                         break;
                     case VOLUME_DECREASE:
-                        player.setVolume(player.getVolume() - VOLUME_CHANGE_SIZE);
+                        result = player.getVolume() - VOLUME_CHANGE_SIZE;
+                        if (volumeInvalid(result)) {
+                            logger.warn("Command ignored because the resulting volume would be invalid: {}", result);
+                        } else {
+                            player.setVolume(result);
+                        }
```
watou: Doesn't the [player.setVolume() method](https://github.com/msteiger/javampd/blob/master/src/main/java/org/bff/javampd/MPDPlayer.java#L445) already do the same work as the volumeInvalid method, except forces it to be between 0 and 100?  Is a WARN log message necessary?

bundles/binding/org.openhab.binding.mpd/src/main/java/org/openhab/binding/mpd/internal/MpdBinding.java
```diff
@@ -219,10 +219,20 @@ private void executePlayerCommand(String playerCommandLine, Object commandParams
                         player.stop();
                         break;
                     case VOLUME_INCREASE:
-                        player.setVolume(player.getVolume() + VOLUME_CHANGE_SIZE);
+                        int result = player.getVolume() + VOLUME_CHANGE_SIZE;
+                        if (volumeInvalid(result)) {
+                            logger.warn("Command ignored because the resulting volume would be invalid: {}", result);
+                        } else {
+                            player.setVolume(result);
+                        }
```
watou: See below.

bundles/binding/org.openhab.binding.mpd/src/main/java/org/openhab/binding/mpd/internal/MpdBinding.java
```diff
@@ -273,20 +283,25 @@ private void executePlayerCommand(String playerCommandLine, Object commandParams
                         break;
                     case VOLUME:
                         logger.debug("Volume adjustment received: '{}' '{}'", pCommand, commandParams);
-                        player.setVolume(((PercentType) commandParams).intValue());
+                        result = ((PercentType) commandParams).intValue();
+                        if (volumeInvalid(result)) {
+                            logger.warn("Command ignored because the volume is invalid: {}", result);
+                        } else {
+                            player.setVolume(result);
+                        }
```
watou: See above.

bundles/binding/org.openhab.binding.mpd/src/main/java/org/openhab/binding/mpd/internal/MpdBinding.java
9037568: Indeed, you're correct. I should've checked the player class to see what it was doing with the values.

I'll undo this change.

bundles/binding/org.openhab.binding.mpd/src/main/java/org/openhab/binding/mpd/internal/MpdBinding.java
9037568: Same.

bundles/binding/org.openhab.binding.mpd/src/main/java/org/openhab/binding/mpd/internal/MpdBinding.java
9037568: Same.

PR #4719, opened 2016-10-21
[DSMR] Add openHAB 2 feature addon

wborn:
The DSMR Binding works just fine with OH2 so here is the OH2 feature addon. This should make using OH2 a bit more easy for the more than 1 million households in The Netherlands that already have a smart meter installed.

### DIFF
--- a/bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="dsmr"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>DSMR Binding</name>
+    <description>The DSMR Binding obtains data from a Dutch smart meter ("Slimme meter" in Dutch) via the P1-port.</description>
+
+    <config-description>
+        <parameter name="port" type="text" required="true">
+            <label>Serial port</label>
+            <description>The DSMR serial port, e.g. "COM1" for Windows or "/dev/ttyUSB0" for Linux.</description>
+            <default>/dev/ttyUSB0</default>
+        </parameter>
+    </config-description>
+</binding:binding>
--- a/bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
@@ -6,6 +6,7 @@
 
     <name>DSMR Binding</name>
     <description>The DSMR Binding obtains data from a Dutch smart meter ("Slimme meter" in Dutch) via the P1-port.</description>
+    <author>Marcel Volaart</author>
 
     <config-description>
         <parameter name="port" type="text" required="true">
--- a/features/openhab-addons-external/src/main/resources/conf/dsmr.cfg
+++ b/features/openhab-addons-external/src/main/resources/conf/dsmr.cfg
@@ -1,0 +1,11 @@
+############################## DSMR Binding ################################
+#
+# Port of the DSMR port (mandatory, e.g. /dev/ttyUSB0)
+#port=/dev/ttyUSB0
+# Configuration of additional meters (channel 0 is used for the main electricity meter)
+#gas.channel=1
+#water.channel=2
+#heating.channel=3
+#cooling.channel=4
+#generic.channel=5
+#slaveelectricity.channel=6
--- a/bundles/binding/org.openhab.binding.dsmr/build.properties
+++ b/bundles/binding/org.openhab.binding.dsmr/build.properties
@@ -2,5 +2,6 @@ source.. = src/main/java/,\
            src/main/resources/
 bin.includes = META-INF/,\
                .,\
-               OSGI-INF/
+               OSGI-INF/,\
+               ESH-INF/
 output.. = target/classes/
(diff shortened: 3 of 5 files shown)


bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,17 @@
...
+    <description>The DSMR Binding obtains data from a Dutch smart meter ("Slimme meter" in Dutch) via the P1-port.</description>
+
+    <config-description>
+        <parameter name="port" type="text" required="true">
+            <label>Serial port</label>
+            <description>The DSMR serial port, e.g. "COM1" for Windows or "/dev/ttyUSB0" for Linux.</description>
+            <default>/dev/ttyUSB0</default>
+        </parameter>
+    </config-description>
+</binding:binding>
```
watou: I  don't think this file needs to be in a binding subdirectory.  I'm also wondering how binding.xml and the similarly purposed dsmr.cfg file interact -- which source of configuration data is used, @kaikreuzer?

features/openhab-addons-external/src/main/resources/conf/dsmr.cfg
```diff
@@ -0,0 +1,11 @@
...
+#
+# Port of the DSMR port (mandatory, e.g. /dev/ttyUSB0)
+#port=/dev/ttyUSB0
+# Configuration of additional meters (channel 0 is used for the main electricity meter)
+#gas.channel=1
+#water.channel=2
+#heating.channel=3
+#cooling.channel=4
+#generic.channel=5
+#slaveelectricity.channel=6
```
watou: See earlier comment about which source of config information -- ESH-INF/bindinng.xml vs. dsmr.cfg -- is used.

bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
wborn: If the file should not be here then following page needs to be updated:

http://docs.openhab.org/developers/development/compatibilitylayer.html#how-to-use-openhab-1x-add-ons-that-are-not-part-of-the-distribution

> How to add a successfully tested 1.x Add-on to the distribution
> ...
> Note that with defining a Karaf feature, bindings are available for installation through the Paper UI in the “Extensions” menu, but they are not listed under “Configuration->Bindings” (although it is fully operational after installation). In order to have bindings listed there as well, you need to add some meta-information to the binding bundle. This information should be put into ESH-INF/binding/binding.xml and its content is described here. Do not forget to add ESH-INF to your build.properties, so that it is packaged in the bundle. See a real life example of such meta-data here.

Also 7 other OH1 bindings that use it this way need to be updated.

The compatibility page could also have some notes on the interactions that @watou asks for as well as do's and don'ts.

The main issue with the page is its discoverability though. I know it exists but even had problems finding it again. It could also help when the broken links in the [opening post](https://community.openhab.org/t/we-need-your-help-on-testing/298) in the community topic are updated.

watou: @wborn Sorry for the bad information!  You are absolutely correct.  I may have found the source of a different issue.  But any doc improvements will be welcome as always. :-)

bundles/binding/org.openhab.binding.dsmr/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="dsmr"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>DSMR Binding</name>
+    <description>The DSMR Binding obtains data from a Dutch smart meter ("Slimme meter" in Dutch) via the P1-port.</description>
+
```
watou: The [schema](https://eclipse.org/smarthome/schemas/binding-1.0.0.xsd) requires an `<author>` element.

wborn: This touches one of the subjects I'd like to see addressed in do's and don'ts on the compatibility page.

Just making a binding compatible/installable with OH2 doesn't make me the author/maintainer. Also I do not know if the original author (@mvolaart) is still around to maintain the binding. I think it is bad etiquette to just put the name of someone else here without consulting them first. Maybe @mvolaart can tell us if he wants to be named as author here?

Furthermore, there are also bindings that are no longer supported by the original authors. Though people may make them compatible with OH2 and submit PRs for this. Who should then be the author? I've also read that being an author/maintainer has responsibilities such as making a binding roadmap and reviewing changes etc. The task should not be taken lightly. In the future there could bindings without an author/maintainer.

As a sidenode. I work in a settings where *@*author tags are considered as a bad practice because everybody owns the code ;-)

wborn: Thanks for reviewing @watou and uploading the updated schema @kaikreuzer !

mvolaart:
@watou I'm still around. I agree with the way of working @kaikreuzer described. I think it is important original or actual maintainers can be contacted in case of questions / issues.

Most imported is that i'm glad to see that the binding is now OpenHAB2 compatible.

wborn:
Thanks for responding @mvolaart! What name can I use in the `<author>` tag? There is a convention to use "Firstname Lastname" for this tag.

watou:
Hi @mvolaart, would you like to have your real first name last name in the `<author>` element for display in the Paper UI, or should no author be listed for this binding under OH2?  With your answer, we can proceed and then merge this PR to add the DSMR binding to openHAB 2.  Thank you!

mvolaart:
Excuse me for the late answer (little bit busy at the moment).

My full name is Marcel Volaart and you can use this at the <auther> element

wborn:
Thanks a lot @mvolaart! @watou I've added it with a commit. So I think we are done with this now.

watou:
Thank you @wborn and @mvolaart!

PR #4724, opened 2016-10-23
Patch for compatibility with OH2:

lewie:
see: https://github.com/openhab/openhab2-addons/issues/503#issuecomment-254763216

The compatibility Layer of OH2 sends, one event for _Command_ AND one for _Update_.
OH1 sends, one event for _Command_ OR one for _Update_.

Therefore is a new configuration switch for openHAB 2 compatibility:

```
# Compatibility Mode for openHAB 2 (optional, defaults in openHAB 1 to false)
compatibilityOH2 = true
```

Needs to be updated too, but how to do it? They're not in the repository...

```
openhab/features/openhab-addons-external/src/main/resources/conf/knx.cfg
openhab-distro/distributions/openhab-offline/target/assembly/system/org/openhab/addons/openhab-addons-external/1.9.0-SNAPSHOT/openhab-addons-external-1.9.0-SNAPSHOT-knx.cfg
```

### DIFF
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
@@ -53,7 +53,7 @@ import tuwien.auto.calimero.process.ProcessListener;
  * @since 0.3.0
  *
  */
-public class KNXBinding extends AbstractBinding<KNXBindingProvider>implements ProcessListener, KNXConnectionListener {
+public class KNXBinding extends AbstractBinding<KNXBindingProvider> implements ProcessListener, KNXConnectionListener {
 
     private static final Logger logger = LoggerFactory.getLogger(KNXBinding.class);
 
@@ -108,7 +108,7 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider>implements Pr
     @Override
     protected void internalReceiveCommand(String itemName, Command command) {
         logger.trace("Received command (item='{}', command='{}')", itemName, command.toString());
-        if (!isEcho(itemName, command)) {
+        if (!KNXConnection.getCompatibilityOH2() && !isEcho(itemName, command)) {
             writeToKNX(itemName, command);
         }
     }
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/connection/KNXConnection.java
@@ -66,6 +66,9 @@ public class KNXConnection implements ManagedService {
     /** the KNX bus address to use as local sourceaddress in KNX bus */
     private static String sLocalSourceAddr = "0.0.0";
 
+    /** Compatibility Mode for openHAB 2 */
+    private static boolean sCompatibilityOH2 = false;
+
     /** the KNX bus address to use as local sourceaddress in KNX bus */
     private static boolean sIgnoreLocalSourceEvents = false;
 
@@ -131,7 +134,7 @@ public class KNXConnection implements ManagedService {
     /**
      * Returns the KNXNetworkLink for talking to the KNX bus.
      * The link can be null, if it has not (yet) been established successfully.
-     * 
+     *
      * @return the KNX network link
      */
     public static synchronized ProcessCommunicator getCommunicator() {
@@ -167,7 +170,7 @@ public class KNXConnection implements ManagedService {
 
     /**
      * Tries to connect either by IP or serial bus, depending on supplied config data.
-     * 
+     *
      * @return true if connection was established, false otherwise
      */
     public static synchronized boolean connect() {
@@ -320,7 +323,7 @@ public class KNXConnection implements ManagedService {
 
     /*
      * (non-Javadoc)
-     * 
+     *
      * @see org.osgi.service.cm.ManagedService#updated(java.util.Dictionary)
      */
     @Override
@@ -332,6 +335,11 @@ public class KNXConnection implements ManagedService {
             String readingBusAddrString = (String) config.get("busaddr");
             if (StringUtils.isNotBlank(readingBusAddrString)) {
                 sLocalSourceAddr = readingBusAddrString;
+            }
+
+            String compatOH2String = (String) config.get("compatibilityOH2");
+            if (StringUtils.isNotBlank(compatOH2String)) {
+                sCompatibilityOH2 = compatOH2String.equalsIgnoreCase("true");
             }
 
             String readingIgnLocEv = (String) config.get("ignorelocalevents");
@@ -465,6 +473,10 @@ public class KNXConnection implements ManagedService {
 
     public static String getLocalSourceAddr() {
         return sLocalSourceAddr;
+    }
+
+    public static boolean getCompatibilityOH2() {
+        return sCompatibilityOH2;
     }
 
     public static boolean getIgnoreLocalSourceEvents() {


bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
```diff
@@ -108,7 +108,7 @@ public void removeKNXTypeMapper(KNXTypeMapper typeMapper) {
     @Override
     protected void internalReceiveCommand(String itemName, Command command) {
         logger.trace("Received command (item='{}', command='{}')", itemName, command.toString());
-        if (!isEcho(itemName, command)) {
+        if (!KNXConnection.getCompatibilityOH2() && !isEcho(itemName, command)) {
```
watou: `isEcho` returns boolean but has the side effect of removing the ignoreEvent from the list.  With this change, there are cases where `isEcho` is no longer reached, likely leading to a memory leak and other problems when the manual switch is not set.

I would like to learn more about how there is a fundamental difference between the 1.x runtime, and the compat1x layer, in sending events.  How did you discover this difference, @lewie?  Also, is there a way to avoid requiring manual configuration of something that ought to be learnable at runtime without configuration?

kaikreuzer: This change is also not clear to me. 
What does "OH2 sends, one event for Command AND one for Update" mean? Where is the difference to OH1? Update events are sent by the auto-update binding and this behaves the very same way in OH1 and OH2.

lewie: The different behavior between and OH1 and OH2 compat1x is:

sending a _Command_ (sendCommand from Rule for example) ->
in OH1:
*internalReceiveCommand is called* (no matter if value is changed or not)

in OH2 compat1x:
*internalReceiveCommand is called AND directly after internalReceiveUpdate is called*
  (sent, no matter if value is changed or not)


sending an _Update_ (postUpdate) ->
in OH1:
*internalReceiveUpdate is called* (sent, no matter if value is changed or not)

in OH2 compat1x:
*internalReceiveUpdate is called* (sent, no matter if value is changed or not)

So running KNX1 bundle in OH2 causes two KNX-bus events. *KNXConnection.getCompatibilityOH2()* simply switches off the first of the two Events.


@watou, the Documentation of *List\<String\>* says:
> Removes the first occurrence of the specified element from this list, if it is present (optional operation). If this list does not contain the element, it is unchanged.

I believe there is not the risk of a memory leak at this point. Am I wrong?
It saves nothing if I check before removing whether the item is in the list, on the contrary.

`isEcho` has to delete a value from ignoreEventList first time it occurs, this is intended.

watou: > internalReceiveCommand is called AND directly after internalReceiveUpdate is called

If autoupdate is not suppressed, this is normal behavior for both OH1 and OH2.  I don't see how many other 1.x bindings would not be failing if this weren't true.  Does this binding attempt to suppress autoupdate logic but fails to do so on OH2?

> I believe there is not the risk of a memory leak at this point. Am I wrong?

Before this PR, the code was `if (B)`, so `B` is evaluated every time the `if` statement is reached.  In the code `if (A && B)`, if `A` is `false`, `B` is never evaluated, so no code removes the ignore event from the list, leading to a memory leak.

lewie: @watou, when many other bindings do it right, then its time to find a way for KNX1 too! ;-)

Lets have a closer look at autoupdate! Seams there could be the fault.
The different behavior between and OH1 and OH2 compat1x is observed with default configured items, autoupdate was not configured explicit.

Found autoupdate in [KNXGenericBindingProvider.autoUpdate](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProvider.java#L347)
Why shouldn't this work in OH2? Can you see any faults or better constructs to use here?

I do not understand the whole *autoupdate* mechanism comes together with internalReceiveCommand/internalReceiveUpdate, yet? Will have a closer look...


Yes, you are absolute right, *ignoreEventList* can grow unstopped in this way, it has to be flipped:
`if (!isEcho(itemName, command) && !KNXConnection.getCompatibilityOH2()) {
`

kaikreuzer: > Why shouldn't this work in OH2? 

The right question would be: Why doesn't it work on OH1, if autoupdate is NOT turned off? An update event should be there as well. The fix should be somewhere in the core framework, if there is a different behavior, it should NOT be necessary to do any change in the KNX binding.

lewie: @kaikreuzer, 

postUpdate behaves identically in OH 1 and 2.

Differences in log new installed OH 1.8.3 and OH2b4:
### OH1 sendCommand ON/OFF
```
2016-10-24 14:21:45.257 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='ON')
2016-10-24 14:21:45.261 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-24 14:21:45.263 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-24 14:21:45.263 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Wrote value 'ON' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-24 14:21:45.264 [WARN ] [.b.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.243 for Group address 1/1/10.

2016-10-24 14:21:50.017 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-24 14:21:50.022 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-24 14:21:50.024 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-24 14:21:50.024 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-24 14:21:50.026 [WARN ] [.b.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.243 for Group address 1/1/10.
```
### OH2 sendCommand ON/OFF
```
2016-10-24 15:16:20.005 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-24 15:16:20.005 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='OFF')
2016-10-24 15:16:20.005 [TRACE] [g.knx.internal.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-24 15:16:20.006 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-24 15:16:20.006 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-24 15:16:20.006 [WARN ] [.binding.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.244 for Group address 1/1/10.

2016-10-24 15:16:25.005 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='ON')
2016-10-24 15:16:25.005 [TRACE] [g.knx.internal.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-24 15:16:25.005 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='ON')
2016-10-24 15:16:25.006 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Wrote value 'ON' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-24 15:16:25.006 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-24 15:16:25.006 [WARN ] [.binding.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.244 for Group address 1/1/10.
```
My Item and Rules;
```
Switch TestSwitch 	 		"TestSwitch" 		 ()	 { knx="<1/1/10" }
```
```
import org.openhab.core.library.types.*
import org.openhab.core.persistence.*
import org.openhab.model.script.actions.*
/*
rule "knxTest_postUpdate"
	when
		Time cron "0/5 * * * * ?"
	then
	if(TestSwitch.state == OFF) {
		TestSwitch.postUpdate( ON )
	} else {
		TestSwitch.postUpdate( OFF )
	}	
end
*/
/*
rule "knxTest_sendCommand"
	when
		Time cron "0/5 * * * * ?"
	then
	if(TestSwitch.state == OFF) {
		TestSwitch.sendCommand( ON )
	} else {
		TestSwitch.sendCommand( OFF )
	}
end
*/
```

OH2 is much more faster! :-)
And for me OH2s behavior is correct.

The OH1 behavior is very old, because if not, we where stumbled over loosing/repeated telegrams as since compat1x layer in OH2.

@watou, 
I must not flip  `(!KNXConnection.getCompatibilityOH2() && !isEcho(itemName, command)) {`, it deletes the echo marker and Echo will be sent to KNX when called `internalReceiveUpdate` after `internalReceiveCommand`.

I stressed the KNX-bus and KNX1 binding with hundreds of fast changed calls. The ignoreEventList never got longer as 11 records and all these records where removed cleanly.

But to prevent ignoreEventList to grow unstopped and leak, I now use Set<String> instead of List<String>. now It cannot contain duplicate elements and is thus limited in size. 

When we find the way how OH1 and compat1x layer in OH2 works identically we do not need the compatibilityOH2 switch. I hope you two will help me, I miss your deep insights into the classes in background!

watou: > But to prevent ignoreEventList to grow unstopped and leak, I now use Set instead of List. 

There are still cases where the list could grow unbounded when using a Set, since item states could come in infinite variations.  Switching to a Set reverts [this commit](https://github.com/openhab/openhab/pull/2298/commits/04b092096b2ec9fc45ce8c249d939fcb1e83da46).  I hope to have more time to contribute more thoughts on this (but not at the moment unfortunately).

lewie: Hmm, not for switches but you are right for numerical values... I will think about how to restrict size.

lewie: @watou, 
I think about a simple cache, loosing its record data after a short period of time.... But I still have to ponder. But I believe it is superfluous, because in no test I've done, I could throw the ignoreEventList out of the rhythm.

One thing I can say, the [autoupdate](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/config/KNXGenericBindingProvider.java#L347) function in KNX 1.8 binding is never ever called, running in OH1 nor OH2.
Why? Is this behavior expected?
Is not called in OH1 and OH2, therefore I think this is not the cause for this difference behavior between OH1/OH2.

This [PR](https://github.com/openhab/openhab/pull/4179) cannot be the cause, it is only some months old. I know about this problem and the differences more about a year.
But I will test this!!

kaikreuzer: > One thing I can say, the autoupdate function in KNX 1.8 binding is never ever called, running in OH1 nor OH2.
> Why? Is this behavior expected?

I just checked this. The reason is that it isn't registered as an auto update provider. With [this change](https://github.com/openhab/openhab/pull/4728), it actually IS called. But does this mean that this code has never been active even in OH1? If so, I do not know, what effect https://github.com/openhab/openhab/pull/4728 might have on OH1 installations...

lewie: @kaikreuzer, 
AHHRGG.. how simple, I should have seen. Thank You!!

See my tests KNX1.8 with *AutoUpdateBindingProvider* registered:

### OH1 TestSwitch.sendCommand(ON/OFF): *Received command* and a second *Received command* but no expected *Received update* per pass
```
2016-10-26 06:44:35.289 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-26 06:44:35.297 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 06:44:35.300 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 06:44:35.300 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 06:44:35.301 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toType datapoint DPT = 1.001
2016-10-26 06:44:35.308 [TRACE] [.b.knx.internal.bus.KNXBinding] - Added event (item='TestSwitch', type='OFF') to the ignore event list
2016-10-26 06:44:35.310 [TRACE] [.b.knx.internal.bus.KNXBinding] - Processed event (item='TestSwitch', type='OFF', destination='1/1/10')
2016-10-26 06:44:35.316 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-26 06:44:35.319 [TRACE] [.b.knx.internal.bus.KNXBinding] - We received this event (item='TestSwitch', state='OFF') from KNX, so we don't send it back again -> ignore!

2016-10-26 06:44:40.009 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-26 06:44:40.013 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 06:44:40.015 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 06:44:40.015 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 06:44:40.017 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toType datapoint DPT = 1.001
2016-10-26 06:44:40.020 [TRACE] [.b.knx.internal.bus.KNXBinding] - Added event (item='TestSwitch', type='OFF') to the ignore event list
2016-10-26 06:44:40.021 [TRACE] [.b.knx.internal.bus.KNXBinding] - Processed event (item='TestSwitch', type='OFF', destination='1/1/10')
2016-10-26 06:44:40.023 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-26 06:44:40.027 [TRACE] [.b.knx.internal.bus.KNXBinding] - We received this event (item='TestSwitch', state='OFF') from KNX, so we don't send it back again -> ignore!
```

### OH1 TestSwitch.postUpdate(ON/OFF): *Received update* and long AFTER that a unwanted *Received command* per pass
```
2016-10-26 06:45:55.290 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='OFF')
2016-10-26 06:45:55.297 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 06:45:55.300 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 06:45:55.300 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 06:45:55.301 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toType datapoint DPT = 1.001
2016-10-26 06:45:55.305 [TRACE] [.b.knx.internal.bus.KNXBinding] - Added event (item='TestSwitch', type='OFF') to the ignore event list
2016-10-26 06:45:55.308 [TRACE] [.b.knx.internal.bus.KNXBinding] - Processed event (item='TestSwitch', type='OFF', destination='1/1/10')
2016-10-26 06:45:55.309 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-26 06:45:55.315 [TRACE] [.b.knx.internal.bus.KNXBinding] - We received this event (item='TestSwitch', state='OFF') from KNX, so we don't send it back again -> ignore!

2016-10-26 06:46:00.019 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='ON')
2016-10-26 06:46:00.022 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 06:46:00.024 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 06:46:00.024 [DEBUG] [.b.knx.internal.bus.KNXBinding] - Wrote value 'ON' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 06:46:00.027 [TRACE] [.o.b.k.i.dpt.KNXCoreTypeMapper] - toType datapoint DPT = 1.001
2016-10-26 06:46:00.029 [TRACE] [.b.knx.internal.bus.KNXBinding] - Added event (item='TestSwitch', type='ON') to the ignore event list
2016-10-26 06:46:00.031 [TRACE] [.b.knx.internal.bus.KNXBinding] - Processed event (item='TestSwitch', type='ON', destination='1/1/10')
2016-10-26 06:46:00.032 [TRACE] [.b.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='ON')
2016-10-26 06:46:00.037 [TRACE] [.b.knx.internal.bus.KNXBinding] - We received this event (item='TestSwitch', state='ON') from KNX, so we don't send it back again -> ignore!
```

### OH2 TestSwitch.sendCommand(ON/OFF): *Received command* and as expected a *Received update* per pass
```
2016-10-26 07:03:35.004 [INFO ] [.internal.AutoUpdateProviderDelegate] - AutoUpdateProviderDelegate: autoUpdate run for 'TestSwitch'
2016-10-26 07:03:35.004 [INFO ] [.internal.AutoUpdateProviderDelegate] - AutoUpdateProviderDelegate: autoUpdate null
2016-10-26 07:03:35.005 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='ON')
2016-10-26 07:03:35.005 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='ON')
2016-10-26 07:03:35.006 [TRACE] [g.knx.internal.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 07:03:35.007 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Wrote value 'ON' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 07:03:35.008 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 07:03:35.008 [WARN ] [.binding.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.244 for Group address 1/1/10.

2016-10-26 07:03:40.002 [INFO ] [.internal.AutoUpdateProviderDelegate] - AutoUpdateProviderDelegate: autoUpdate run for 'TestSwitch'
2016-10-26 07:03:40.002 [INFO ] [.internal.AutoUpdateProviderDelegate] - AutoUpdateProviderDelegate: autoUpdate null
2016-10-26 07:03:40.004 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received command (item='TestSwitch', command='OFF')
2016-10-26 07:03:40.004 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='OFF')
2016-10-26 07:03:40.004 [TRACE] [g.knx.internal.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 07:03:40.005 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 07:03:40.005 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 07:03:40.005 [WARN ] [.binding.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.244 for Group address 1/1/10.
```

### OH2 TestSwitch.postUpdate(ON/OFF): As expected a *Received update* per pass
```
2016-10-26 07:04:45.004 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='ON')
2016-10-26 07:04:45.004 [TRACE] [g.knx.internal.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 07:04:45.004 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Wrote value 'ON' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 07:04:45.004 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 07:04:45.004 [WARN ] [.binding.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.244 for Group address 1/1/10.

2016-10-26 07:04:50.003 [TRACE] [.binding.knx.internal.bus.KNXBinding] - Received update (item='TestSwitch', state='OFF')
2016-10-26 07:04:50.004 [TRACE] [g.knx.internal.dpt.KNXCoreTypeMapper] - toTypeClass looking for dptId = 1.001
2016-10-26 07:04:50.004 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Received groupWrite Event.
2016-10-26 07:04:50.004 [DEBUG] [.binding.knx.internal.bus.KNXBinding] - Wrote value 'OFF' to datapoint 'command DP 1/1/10 TestSwitch, DPT main 0 id 1.001, low priority'
2016-10-26 07:04:50.004 [WARN ] [.binding.knx.internal.bus.KNXBinding] - Ignoring local Event, received from my local Source address 15.15.244 for Group address 1/1/10.
```

## In OH1 KNX1.8 works really chaotic: 
To send a Command (rules sendCommand or Gui) first time is sent to KNX bus, then this Adress is blocked. No value update by sendCommand helps.
The only thing which can unblock is a postUpdate with a rule. postUpdate works as ecpected. After a postUpdate next sendCommand first time is sent to KNX bus, then this Adress is blocked again.

## In OH2 KNX1.8 works as expected.  Gui, sendCommand, postUpdate IN/OUT

## There is only an old known problem of the architecture of KNX1.8: 
The binding for writing KNX-Telegrams to calimero uses a function, which sets the KNX *READ-Flag*. 
The main side effect is, that values on the Bus are only sent after another or the same Groupadress was updated/sent on/from the KNX Bus.
Example: I want to send every second the value *OFF* to GA 1/1/10 to KNX Bus by a OH rule. First time when I send it to the KNX Bus, calimero will let pass it, second time it isn't, third time it isn't... . When any other GA is updated on KNX Bus (and comes through calimero) then ONLY NEXT TIME when I send my OFF to the KNX Bus, calimero will let pass it again, second time it isn't, third time it isn't... and so on.

But this is another issue, I just wanted to mention here again to avoid confusion for testers ;-)

What to do now?

lewie: Your second question:
https://github.com/openhab/openhab/blob/1.7/bundles/binding/org.openhab.binding.knx/OSGI-INF/knxgenericbindingprovider.xml
https://github.com/openhab/openhab/blob/1.8/bundles/binding/org.openhab.binding.knx/OSGI-INF/knxgenericbindingprovider.xml

No, it seams AutoUpdateBindingProvider has been never been active! 
I have not realized this over the years, what I would have spared. :-(

lewie: @kaikreuzer, @watou, Made further Tests. With AutoUpdateBindingProvider KNX1.8 seams to work well in OH1 too.

In OH2 the KNX1.8 bundle now works pretty good and as expected! (without "compatibilityOH2")

I am still doing detailed tests in a OH1 live system. Results will follow.

lewie:
@watou, @kaikreuzer,

register AutoUpdateBindingProvider (Thanks Kai) does the trick. KNX1.8 is running as expected in OH1 and OH2 now. So *compatibilityOH2* is obsolete, removed it.

Updated calimero (2.2 -> 2.3) to prevent "repeatFlag" sent to KNX-bus.

For Testers, please use: https://github.com/lewie/openhab/raw/knx1-patch-for-oh2-2-test/org.openhab.binding.knx-1.9.0-SNAPSHOT.jar

PR #4741, opened 2016-11-03
[GCal] updated authorization and Calendar API 

SirAd:
Hi,
google dropped support for old authorization mechanizm. This update implements OAuth2 and uses new Google Calendar API v3


bundles/io/org.openhab.io.gcal/.classpath
```diff
@@ -3,10 +3,14 @@
...
 	<classpathentry kind="lib" path="lib/jsr305.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.6"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/jackson-core-asl-1.9.11.jar"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/jackson-core-2.1.3.jar"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/google-oauth-client-1.21.0.jar"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/google-http-client-jackson2-1.21.0.jar"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/google-http-client-1.21.0.jar"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/google-api-services-calendar-v3-rev219-1.22.0.jar"/>
+	<classpathentry kind="lib" path="/home/adam/Development/openhab/bundles/io/org.openhab.io.gcal/lib/google-api-client-1.22.0.jar"/>
```
watou: Use only relative paths.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -135,69 +190,84 @@ protected void execute() {
...
+                        logger.debug("Got calendar {} CalendarID: {}", calendar_name, calendarID);
+                    }
+                }
+            }
 
-            CalendarService myService = new CalendarService("openHAB");
-            if (!StringUtils.isBlank(username) && !StringUtils.isBlank(password)) {
-                myService.setUserCredentials(username, password);
+            if (calendarID == null) {
+                logger.error("Calendar {} not found", calendar_name);
```
watou: reduce to warn or lower

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -135,69 +190,84 @@ protected void execute() {
...
+                return null;
             }
+        } else {
+            calendarID = calendar_name;
+        }
 
-            CalendarQuery myQuery = new CalendarQuery(feedUrl);
-            DateTime start = DateTime.now();
-            DateTime end = new DateTime(DateTime.now().getValue() + (2 * refreshInterval));
+        DateTime start = new DateTime(new Date());
```
watou: The default constructor uses the current time IIRC.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -135,69 +190,84 @@ protected void execute() {
...
-                checkIfFullCalendarFeed(feed.getEntries());
+                checkIfFullCalendarFeed(feed.getItems());
             }
 
             return feed;
-        } catch (AuthenticationException ae) {
-            logger.error("authentication failed: {}", ae.getMessage());
         } catch (Exception e) {
-            logger.error("downloading CalenerEventFeed throws exception: {}", e.getMessage());
+            logger.error("downloading CalenarEventFeed throws exception: {}", e.getMessage());
```
watou: *CalendarEventFeed

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -410,52 +479,52 @@ protected boolean createTriggerAndSchedule(JobDetail job, CalendarEventEntry eve
...
 
-                try {
-                    scheduler.scheduleJob(job, trigger);
-                    triggersCreated = true;
-                } catch (SchedulerException se) {
-                    logger.warn("scheduling Trigger '" + trigger + "' throws an exception.", se);
-                }
+        // List<When> times = event.getTimes();
+        // for (When time : times) {
+        EventDateTime date = isStartEvent ? event.getStart() : event.getEnd();
```
watou: remove commented out code

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -533,4 +607,134 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
...
+                    if (deviceToken.access_token != null) {
+                        logger.debug("Got access token");
+                        logger.debug("device access token: " + deviceToken.access_token);
+                        logger.debug("device token_type: " + deviceToken.token_type);
+                        logger.debug("device refresh_token: " + deviceToken.refresh_token);
+                        logger.debug("device expires_in: " + deviceToken.expires_in);
+                        break;
+                    }
+                    logger.debug("waiting for " + device.interval + " seconds");
+                    Thread.sleep(device.interval * 1000);
```
watou: Use parameterized logging instead of string concatenation in all logger calls.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -533,4 +607,134 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
...
+                dataCredential.setAccessToken(deviceToken.access_token);
+                dataCredential.setRefreshToken(deviceToken.refresh_token);
+                dataCredential.setExpirationTimeMilliseconds((long) deviceToken.expires_in * 1000);
+
+                datastore.set(TOKEN_STORE_USER_ID, dataCredential);
+
+                credential = loadCredential(TOKEN_STORE_USER_ID, datastore);
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
watou: Use only proper logging.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -533,4 +607,134 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
...
+                        device.expires_in);
+                logger.info(
+                        "################################################################################################");
+
+                logger.debug("Got access code");
+                logger.debug("user code :" + device.user_code);
+                logger.debug("device code :" + device.device_code);
+                logger.debug("expires in:" + device.expires_in);
+                logger.debug("interval :" + device.interval);
+                logger.debug("verification_url :" + device.verification_url);
```
watou: A small speed improve for the above block would be to enclose it in `if (logger.isDebugEnabled())`.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -59,12 +81,14 @@
...
+    private static final String TOKEN_STORE_USER_ID = "openhab";
 
     private static final Logger logger = LoggerFactory.getLogger(GCalEventDownloader.class);
 
-    private static String username = "";
-    private static String password = "";
-    private static String url = "";
+    private static String client_id = "923938667558-3agtjnqnlfq4ku2v6veoradnt88291du.apps.googleusercontent.com";
+    private static String client_secret = "ow_sZCtGTD316vnWm7zOGXfL";
+    private static String calendar_name = "primary";
```
SirAd: it's my client's credential - can be used as default

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
SirAd: Yes, we need current time to calculate startTime for fetching

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
SirAd: Typo fixing.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
SirAd: done

SirAd:
Suggestions implemented

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
watou: OK, but I worry that this creates a dependency on your credentials which might stop working someday.  I think it might make sense to either use credentials controlled by the openHAB Foundation as defaults, or simply not provide defaults.  (A similar situation recently cropped up with Twitter integration, and it's still unresolved due to unforeseen constraints.)  @kaikreuzer, what is your opinion?

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
watou: My (minor) point was that `new DateTime()` is identical in result to `new DateTime(new Date())`.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
watou: Is the "d" still missing?

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
watou: I don't see the change?

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
```diff
@@ -533,4 +605,136 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
...
+                return null;
+            }
+            credential.setAccessToken(stored.getAccessToken());
+            credential.setRefreshToken(stored.getRefreshToken());
+            credential.setExpirationTimeMilliseconds(stored.getExpirationTimeMilliseconds());
+
+            logger.debug("Loaded credential");
+            logger.debug("device access token: " + stored.getAccessToken());
+            logger.debug("device refresh_token: " + stored.getRefreshToken());
+            logger.debug("device expires_in: " + stored.getExpirationTimeMilliseconds());
```
watou: Use parameterized logging, like
```
logger.debug("device expires_in: {}", stored.getExpirationTimeMilliseconds())
```
for all logger calls.  This prevents unnecessary object creation and garbage collection.

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/internal/GCalEventDownloader.java
SirAd: There is no `new DateTime()` com.google.api.client.util.DateTime. I'm using `new DateTime(Date value)`.

watou: Sorry, my mistake regarding the class in question.  Please disregard.

watou:
Thanks for the updates.  IMHO, I would still like another maintainer to give an opinion on the source of default credentials, and if any are used regardless of controller, it should be prominently documented.

9037568:
Since we already have multiple problems with bindings failing due to
expired "default" credentials/accounts, I would vote against having such a
default.

Leave them empty and let the end user supply them.

My 2 drachmas only, and YMMV.

SirAd:
I can remove default credential if this is the decision. Just let me know.

watou:
I think it would be better to remove the default credentials, and clearly document how users can self-serve to supply their own.  That way there is no external dependency on a third party to keep the credentials valid.  I think that, if you've tested your changes, this PR can be merged.

SirAd:
I've removed default credentials. Will add instruction on wiki how to create account on google

watou:
@SirAd are you pretty confident this can be merged now, or would it make sense to enlist some testers from https://community.openhab.org to give it a good workout?

SirAd:
Hi,
I've posted new topic on community forum https://community.openhab.org/t/gcal-bindings-refreshed-version/16400 . Let's check if anybody wants to test it.

watou:
I tried again and it worked!  But I noticed an annoyance that perhaps you could fix in this PR: Too much logging at INFO level.

```
[INFO ] [g.internal.GCalEventDownloader] - created new startJob 'MyItem OFF' with details 'SchedulerJob [jobKey=kbr6mxxmmkg2xxxf5g4i3xxkk@google.com_start, jobGroup=gcal, 1 triggers=[Sat Nov 12 12:10:00 GMT 2016], content=send MyItem OFF'
[INFO ] [o.i.g.i.util.ExecuteCommandJob] - Command has been sent successfully.
```
I really think these should be DEBUG level log messages.  Others might also be better at DEBUG instead of INFO.

watou:
This PR lgtm and should be merged because it seems to work on the 1.8 runtime.  However, it does not work on the 2.0 runtime due to #4770.  Therefore, either the compat1x layer or this service will have to be further changed to be included in OH2.

Another maintainer: WDYT?

9037568:
@watou, I have a question about the method signature change here:

````
-public static CalendarEventFeed downloadEventFeed(String username, String password, String url, int refreshInterval)
+public static Events downloadEventFeed() {
````

In some projects it would not be permitted (or at least be considered poor form) to remove a public method without a deprecation period. Do we care about that here?

Aside from this, I think this PR is good to go.

**Comment by watou** [comment]
> downloadEventFeed()

@9037568: Good question.  I *think* it's OK to change the method signature in this case because I don't see that the class's package is exported in MANIFEST.MF and the the only OSGi service interface in OSGI-INF/eventdownloader.xml is ManagedService.  Also, with the OAuth2 implementation, the old arguments are now meaningless (afaik).  Unless changing the signature breaks a full build because some other project depends on it, I think it's OK.  Could the method be made private or package private?  @SirAd, WDYT?

I do wonder why the MANIFEST.MF is exporting com.google.gdata packages though (which predates this PR).  Could the entire `Export-Package` directive be removed, @SirAd?

**Comment by SirAd** [comment]
@watou: com.google.gdata packages packages are used by org.openhab.persistance.gcal. I'm working on  org.openhab.persistance.gcal to use oauth2 and Calendar API v3. Once it will be ready I remove it from MANIFEST.MF

**Comment by watou** [comment]
@SirAd: That's great news!  The idea that these services are all brought up to date, the org.openhab.io service provides function to the persistence service to eliminate any duplication of code or confguration, and finally working on both OH1 and OH2, will be a tremendous step forward.  Much appreciated.

**Comment by watou** [comment]
Thanks for the latest commit, @SirAd.  This is looking very good!

One small point: I believe it's considered "wrong" to export a `.internal` package in the `MANIFEST.MF` file.  So if the the persistence bundle needs services of the io bundle (which by itself makes perfect sense), it would have to be in a package not in or under `.internal`.

**Comment by SirAd** [comment]
Thanks for review. I will move code used in gcal persistence from `internal` to package. BTW, I'd like to update gcal wiki. Is there possibility to put screenshots in wiki? I did not found how to do it.

**Comment by watou** [comment]
Thanks @SirAd!

> Is there possibility to put screenshots in wiki?

You can put screenshots in the wiki like what is done [here](https://github.com/openhab/openhab/wiki/Actions#openwebif-action) (scroll down) using this syntax:

```
![](https://farm4.staticflickr.com/3882/15284270826_8cf0e637d8_z.jpg)
```

The question of where to permanently host the screenshot image files is a good one.  I've never heard the answer, but if you choose a place that's unlikely to change for years, that should do for now.

File: bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/auth/GCalGoogleOAuth.java [hunk 11]
```diff
@@ -0,0 +1,302 @@
...
+import com.google.api.services.calendar.CalendarScopes;
+import com.google.api.services.calendar.model.CalendarList;
+import com.google.api.services.calendar.model.CalendarListEntry;
+
+/**
+ * Service which downloads Calendar events, parses their content and creates
+ * Quartz-jobs and triggers out of them.
+ *
+ * @author Thomas.Eichstaedt-Engelen
+ * @since 0.7.0
```
watou: Please update this JavaDoc and take credit for your work! 😄

bundles/io/org.openhab.io.gcal/src/main/java/org/openhab/io/gcal/auth/GCalGoogleOAuth.java
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * Define a global instance of the JSON factory.
+     */
+    private static final JsonFactory JSON_FACTORY = new JacksonFactory();
+
+    /**
+     * <p>
+     * Setup OAuth2 client_id used for authentication
+     * </p>
```
watou: No need for `<p></p>`, and also the closing tag is discouraged (long story).

SirAd:
Hi,
I also updated Wiki page for gcal bindings, hope that's all for this PR.

watou:
Thank you for all of your nice work, @SirAd!

watou:
Would you like to replace the ConsoleInterpreter so all this works under openHAB 2, or should I?

SirAd:
Yes @watou , I will try to replace the ConsoleInterpreter. I will raise new PR when change will be implemented an tested

PR #4746, opened 2016-11-05
Fixed some typos in jdbc log messages.

9037568:
Also replaced a couple occurrences of string concatenation with parameterized log calls.


watou:
lgtm

9037568:
Thanks!

PR #1005, opened 2014-04-23
Binding for Philips TVs with jointspace API

Lenzebo:
I created a binding for my Philips TV into openhab. All their models (since 2011) support a json-based api called jointSPACE (see http://jointspace.sourceforge.net/projectdata/documentation/jasonApi/index.html for the reference what one can do with it) that allows controlling some functionality over ethernet.

Right now the binding can:
- Send Button commands
- Sets and Read Volume/Mute
- Set and Get Colors of Ambilight "Pixels"

PR #4757, opened 2016-11-09
[jointSPACE] Add the 1.9 binding to the OH2 distro

Lenzebo:
Hi 

After confirming that the current jointSPACE binding with the current snapshot of OH2 is still compatible (see #1005), I created the necessary entries and file to make it usefull as a Karaf feature for OH2
[information from here](http://docs.openhab.org/developers/development/compatibilitylayer.html)

As I do not have experience with Karaf, I hope I did not forget something.


kaikreuzer:
Great, thanks!

PR #4763, opened 2016-11-11
[fatekplc] Documentation moved to wiki page

slawekjaranowski:
Maintenance of documentation in one place is easier.


watou:
Thanks.  Indeed, multiple places for the same documentation means only one place is correct! :-)

PR #4765, opened 2016-11-11
In openHAB 2 all events are triggered twice to KNX bus

lewie:
KNX1.8 in OH2 does send all events twice to KNX bus. This behavior did not stop with [PR](https://github.com/openhab/openhab/pull/4736) including "AutoUpdateBindingProvider".

The reason is a different behavior of OH1 and OH2:
```
In OH1 
*postUpdate* causes once *internalReceiveUpdate* is triggered.
and 
*sendCommand* causes once *internalReceiveCommand* is triggered.

In OH2 
*postUpdate* causes once *internalReceiveUpdate* is triggered.
and 
*sendCommand* causes *internalReceiveCommand and after internalReceiveUpdate* is triggered. (->two KNX bus triggers)
```

Unfortunately, I could not determine the reason. It seams to be a difference between felix and equinox Event handling.
I could go back and compare callers 4 levels, but found no reason.
```
Callers before internalReceiveCommand OH2
level 1:  getClassName = 'org.openhab.binding.knx.internal.bus.KNXBinding'
level 1:  getFileName = 'KNXBinding.java'
level 1:  getLineNumber = '119'
level 1:  getMethodName = 'internalReceiveCommand'

level 2:  getClassName = 'org.openhab.core.binding.AbstractBinding'
level 2:  getFileName = 'AbstractBinding.java'
level 2:  getLineNumber = '97'
level 2:  getMethodName = 'receiveCommand'

level 3:  getClassName = 'org.openhab.core.events.AbstractEventSubscriber'
level 3:  getFileName = 'AbstractEventSubscriber.java'
level 3:  getLineNumber = '42'
level 3:  getMethodName = 'handleEvent'

level 4:  getClassName = 'org.apache.felix.eventadmin.impl.handler.EventHandlerProxy'
level 4:  getFileName = 'EventHandlerProxy.java'
level 4:  getLineNumber = '415'
level 4:  getMethodName = 'sendEvent'

Callers before internalReceiveCommand OH1
level 1: getClassName = 'org.openhab.binding.knx.internal.bus.KNXBinding'
level 1: getFileName = 'KNXBinding.java'
level 1: getLineNumber = '127'
level 1: getMethodName = 'internalReceiveCommand'

level 2: getClassName = 'org.openhab.core.binding.AbstractBinding'
level 2: getFileName = 'AbstractBinding.java'
level 2: getLineNumber = '97'
level 2: getMethodName = 'receiveCommand'

level 3: getClassName = 'org.openhab.core.events.AbstractEventSubscriber'
level 3: getFileName = 'AbstractEventSubscriber.java'
level 3: getLineNumber = '42'
level 3: getMethodName = 'handleEvent'

level 4: getClassName = 'org.eclipse.equinox.internal.event.EventHandlerWrapper'
level 4: getFileName = 'EventHandlerWrapper.java'
level 4: getLineNumber = '197'
level 4: getMethodName = 'handleEvent'

```

With this PR the cause of the different behavior of the two OH versions is not fixed.
But the KNX 1.8 behaves as expected in both openHAB versions now.

### DIFF
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
@@ -69,6 +69,9 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider>implements Pr
 
     private boolean mKNXConnectionEstablished;
 
+    /** if true knx is started in openHAB 2 */
+    private Boolean operatingEnvironmentIsFelix;
+
     private Map<String, DimmerThread> itemDimmerThreads = Collections
             .synchronizedMap(new HashMap<String, DimmerThread>());
 
@@ -107,9 +110,21 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider>implements Pr
      */
     @Override
     protected void internalReceiveCommand(String itemName, Command command) {
-        logger.trace("Received command (item='{}', command='{}')", itemName, command.toString());
-        if (!isEcho(itemName, command)) {
-            writeToKNX(itemName, command);
+        logger.trace(
+                "1. internalReceiveCommand from openHAB (item='{}', command='{}', operatingEnvironmentIsFelix = '{}'",
+                itemName, command.toString(), operatingEnvironmentIsFelix);
+
+        // openHAB1: After receiving a command from openHAB1, NO update event is triggered later.
+        // openHAB2: After receiving a command from openHAB2, ALWAYS update event is triggered later.
+        // In openHAB2 we ignore this command and wait for receiving the update event from openHAB,
+        // then only this is sent to the KNX bus by openHAB2
+
+        if (getEnvironmentIsFelix() == true) {
+            // ignore internalReceiveCommand in openHAB 2
+        } else {
+            if (!isEcho(itemName, command)) {
+                writeToKNX(itemName, command);
+            }
         }
     }
 
@@ -121,7 +136,7 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider>implements Pr
      */
     @Override
     protected void internalReceiveUpdate(String itemName, State newState) {
-        logger.trace("Received update (item='{}', state='{}')", itemName, newState.toString());
+        logger.trace("2. internalReceiveUpdate from openHAB (item='{}', state='{}')", itemName, newState.toString());
         if (!isEcho(itemName, newState)) {
             writeToKNX(itemName, newState);
         }
@@ -137,6 +152,15 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider>implements Pr
         } else {
             return false;
         }
+    }
+
+    private boolean getEnvironmentIsFelix() {
+        if (operatingEnvironmentIsFelix == null) {
+            StackTraceElement[] stacktrace = Thread.currentThread().getStackTrace();
+            logger.trace("4.  getEnvironmentIsFelix = '{}'", stacktrace[5].getClassName());
+            operatingEnvironmentIsFelix = stacktrace[5].getClassName().contains("felix");
+        }
+        return operatingEnvironmentIsFelix;
     }
 
     private void writeToKNX(String itemName, Type value) {


bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
```diff
@@ -107,9 +110,21 @@ public void removeKNXTypeMapper(KNXTypeMapper typeMapper) {
...
+                "1. internalReceiveCommand from openHAB (item='{}', command='{}', operatingEnvironmentIsFelix = '{}'",
+                itemName, command.toString(), operatingEnvironmentIsFelix);
+
+        // openHAB1: After receiving a command from openHAB1, NO update event is triggered later.
+        // openHAB2: After receiving a command from openHAB2, ALWAYS update event is triggered later.
+        // In openHAB2 we ignore this command and wait for receiving the update event from openHAB,
+        // then only this is sent to the KNX bus by openHAB2
+
+        if (getEnvironmentIsFelix() == true) {
+            // ignore internalReceiveCommand in openHAB 2
```
watou: I worry that relying on the auto-update logic, and relying on a stack trace to determine if it's running on Felix, are only working around a deeper root cause.  I think there may be a deeper issue in auto-update handling, or possibly in how echoes are detected, and this PR only masks it.

kaikreuzer: I share your worries - I am still of the opinion that this needs more thorough analysis; unfortunately, I didn't find time to look into it any further myself yet.

tkuehne71:
It would really be great if this could make it a bit higher on the prio list, since right now it kills the usability for OH2 in my KNX setup My motion Sensors get  constanly retriggered, meanigng that the light will never Go out. While comic it is still unaccecptable :)

kaikreuzer:
Just FTR: I run OH2 with KNX1 with a huge KNX setup without any problems. So I cannot confirm that any usability is killed...

tkuehne71:
hmm, as described in the issue referenced in this thread it seems to do so in my setup.

juelicher:
> So I cannot confirm that any usability is killed...

In this I have to disagree, at leas partially.

I have an OH2 test server and an OH1 productive system. Resending the packet immediately after receiving it triggered a consistency check in my OH1 rules which in turn stopped by gas meter evaluation. Admittedly this was not a very likely chain of events and I could solve it by changing the logic in the rules, but I assume that other logic engines which a running beside OH2 might have similar problems. 

In addition I believe that I observed at least once, that the packet resent by OH2 differed by one from the data received from the KNX sensor. I suspect, that this might have been caused b rounding differences (floating point value). This might also cause problems.

And last: One of my KNX sensors responds with a data packet when receiving on the GA it should send its values to (probably a bug in the sensor). If such devices are present resending the data by OH2 might not double the bus load but enhance it a lot more. In huge installations this might also cause problems.

So for me this is also a show stop for migrating my productive system to OH2.

martenss:
Same issue here.
Running Openhab 2.0.0 Release Build on openhabian Rpi3.

Openhab is consistently re-transmitting KNX telegrams for all incoming GAs linked to items - and not for the other GAs not configured in the item file.
Problem is reproduced both in TUNNEL and ROUTER mode, and while using different IP interfaces (MDT & Wiregate). ignorelocalevents=true doesn't seem to make a difference so I suspect this parameter is not recognized in the systems showing this behavior.

Quite annoying when using event-driven telegrams (quite common in home automation systems), as one need to check for changes instead of relying on simple event notifications.

My production Openhab 1.8 server doesn't show this behavior with same settings.

Note: the above-mentioned difference between incoming and outgoing GAs had been observed as well consistently for one partical item that had persistence issues (any value coming in resulted in OH2 writing 0 on the bus). This had been resolved by fixing the rrd4j persistence issue, and now also the same value is written back on the bus so this is just a configuration matter to be resolved by the end-user.

iOlli:
The same issue here.

I have created a simple setup with only one switch.
I tested TUNNEL and ROUTER.

In OH1 it work.

In OH2 it doesn't work. After switch OH2 sends permanently the last telegram. 

The option ingnorelocalevents=true doesn't fix the problem.

sucre87:
Same for me, sendCommand triggers everything twice.

Also with ingnorelocalevents=true

kaikreuzer:
I guess now that we have a KNX 2.x binding, we can close this as most likely nobody will follow up on it anymore.

PR #4772, opened 2016-11-13
xpl: add Contact item type

thefrip:


thefrip:
Made the change on version 1.8. Not sure exactly how to push it to master (not an expert Git user I must confess).

watou:
@thefrip You would have to checkout the master branch in your fork, create a branch for the change, change and test the code, and then submit a PR from your branch.  We cannot handle pull requests on the 1.8 branch because there is currently no plan for a 1.8.4 release (afaik).  Please close this PR and open a new one as described above.  The changed code would also have to conform to the formatting rules in the Eclipse IDE.

Aside from the above, looking at the code, it appears there is something very wrong previous to your changes where the binding is performing item.setState() calls, which bindings should never do.  I assume you are following the model of the existing code to add the Contact item, but do you know any history of why this binding would be setting item states directly?  I don't think such code should ever have been merged in the first place.

thefrip:
@watou Ok, will do it from master (I think I know how to do this). Not absolutely sure what you mean by conforming the formatting rules in the Eclipse IDE (changes were done through the IDE, so I would assume it should be there by default?).

Regarding the item.setState() calls, indeed I didn't touch that part, only the one related the new item type.

Thanks for your guidance.

PR #4773, opened 2016-11-13
add configurable refresh interval

J-N-K:
The current implementation of DMX transmitter is only transmitting if the universe has changed. This is checked every 35ms. The data is never re-transmitted if it does not change. If a receiver has been restarted after the last transmit it will not receive data until a channel changed. Some connections (as the upcoming E1.31) need to be refreshed by standrad definition.

This change introduces a "refresh" configuration option. The default is -1 (never refresh), which is compatible with the current implementation. A value of 0 would refresh every 35ms, another value would transmit data if it has been changed or the last transmission is more than (value) ms in the past.


J-N-K:
Mhm. This obviously is wrong.

PR #4775, opened 2016-11-13
[dmx] fixes #4774

### LINKED ISSUE #4774: [dmx] binding never refreshes if values do not change
## Expected Behavior
Binding should transmit from time to time even if no changed occured. Otherwise DMX receivers will not receive data if they were switched off or restarted.

## Current Behavior
DMX data is only transmitted if the universe changed.

## Possible Solution
PR #4773, although there seems to be a formatting issue.

1. switch off DMX receiver device
2. switch back on

J-N-K:
Adds configurable DMX transmit refresh. The "refresh" configuration option defaults to -1 (never refresh), which is compatible with the current implementation. A value of 0 would refresh every 35ms, another value would transmit data if it has been changed or the last transmission is more than (value) ms in the past.

### DIFF
--- a/bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
+++ b/bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
@@ -35,7 +35,8 @@ public class DmxController implements DmxService, ManagedService {
     private static Logger logger = LoggerFactory.getLogger(DmxController.class);
 
     private static int TRANSMIT_FREQUENCY_MS = 35;
-    private DmxRepeatMode repeatMode = DmxTransmitter.DmxRepeatMode.ALWAYS; // default is send every update
+    private DmxTransmitter.DmxRepeatMode repeatMode = DmxTransmitter.DmxRepeatMode.ALWAYS; // default is send every
+                                                                                           // update
 
     /** Thread in which the DMX transmitter is running **/
     private Timer transmitterTimer;
@@ -301,14 +302,14 @@ public class DmxController implements DmxService, ManagedService {
             // refresh timeout (i.e. interval between transmits if nothing changed)
             String configuredRepeatMode = ((String) config.get("repeatMode"));
             if (StringUtils.isNotBlank(configuredRepeatMode)) {
-                repeatMode = DmxTransmitter.DmxRepeatMode(configuredRepeatMode);
-                if (repeatMode==null) {
+                repeatMode = DmxTransmitter.DmxRepeatMode.fromString(configuredRepeatMode);
+                if (repeatMode == null) {
                     repeatMode = DmxTransmitter.DmxRepeatMode.ALWAYS;
                     logger.error("repeatMode {} not recognized, set to {}", configuredRepeatMode, repeatMode);
                 } else {
                     logger.debug("repeatMode set to {}", repeatMode.toString());
                 }
-                
+
                 if (transmitter != null) {
                     transmitter.setRepeatMode(repeatMode);
                 }
--- a/bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
+++ b/bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
  * Default DMX Service implementation
  *
  * @author Davy Vanherbergen
+ * @author Jan N. Klug
  * @since 1.2.0
  */
 public class DmxController implements DmxService, ManagedService {
@@ -34,6 +35,7 @@ public class DmxController implements DmxService, ManagedService {
     private static Logger logger = LoggerFactory.getLogger(DmxController.class);
 
     private static int TRANSMIT_FREQUENCY_MS = 35;
+    private int repeatMode = DmxTransmitter.REPEAT_MODE_ALWAYS; // default is send every update
 
     /** Thread in which the DMX transmitter is running **/
     private Timer transmitterTimer;
@@ -49,7 +51,6 @@ public class DmxController implements DmxService, ManagedService {
      */
     @Override
     public void start() throws Exception {
-
         logger.trace("Starting Dmx transmitter ...");
         transmitter = new DmxTransmitter(this);
         transmitterTimer = new Timer(true);
@@ -84,7 +85,6 @@ public class DmxController implements DmxService, ManagedService {
      */
     @Override
     public int getChannelValue(int channel) {
-
         int value = transmitter.getChannel(channel).getValue();
         logger.trace("Getting channel {} value: {}", channel, value);
         return value;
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
```diff
@@ -27,13 +27,15 @@
...
  * @author Davy Vanherbergen
+ * @author Jan N. Klug
  * @since 1.2.0
  */
 public class DmxController implements DmxService, ManagedService {
 
     private static Logger logger = LoggerFactory.getLogger(DmxController.class);
 
     private static int TRANSMIT_FREQUENCY_MS = 35;
+    private int transmitRefreshMs = 0; // default is send all updates
```
watou: This disagrees with the PR comment:

> The "refresh" configuration option defaults to -1 (never refresh)

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
```diff
@@ -49,9 +51,9 @@
      */
     @Override
     public void start() throws Exception {
-
         logger.trace("Starting Dmx transmitter ...");
         transmitter = new DmxTransmitter(this);
+        transmitter.setTransmitRefresh(transmitRefreshMs);
```
watou: Could the variables/members/methods all use the phrase `refreshInterval` so the changed code is easier to follow?

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxTransmitter.java
```diff
@@ -89,8 +96,18 @@ public void setSuspend(boolean suspend) {
     }
 
     /**
+     * change transmitter refresh cycle
+     *
+     * @param refreshInterval
+     *            interval in ms (if output did not change)
+     */
+    public void setTransmitRefresh(int refreshInterval) {
```
watou: `setRefreshInterval` as mentioned above?

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxTransmitter.java
```diff
@@ -37,11 +37,15 @@
...
 
+    private int refreshInterval = 0;
+    private long lastTransmit;
+
     /**
      * Default constructor.
      */
     public DmxTransmitter(DmxService service) {
         this.service = service;
+        lastTransmit = 0;
```
watou: Could `lastTransmit` be initialized to 0 above, instead of one private member being initialized that way, and the other initialized in the constructor?

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
J-N-K: fixed.

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxTransmitter.java
J-N-K: fixed

J-N-K:
I have incorporated all comments. However, after re-reading E1.11 (DMX 512) and E1.31 (sACN) I come to the conclusion that the "never-refresh-policy" is plain wrong when it comes to standards compliance. 

E1.11 states for DMX512 that there is a continous data stream (with a maximum MARK before BREAK of 1s), and defines that longer periods are considered as Data Lost (8.9 / . The standard does not define how a receiver should behave if that occurs, besides that it should remain operational for at least 60s. So after 60s without signal a standrad compliant receiver could even turn off.

E1.31 defines that if the output is not changing, the last packet should be re-transmitted at least three times. After that the sender is allowed to pause sending for 800-1000ms (6.6.2)

watou:
Hi @J-N-K, could you circulate your thoughts about proper refresh handling to other DMX users on the forum (unless you've already done that), and make sure your proposed changes don't meet thoughtful disagreement?  Your focus on this is probably a rare opportunity to get this right vis-a-vis standards and real-world operation.  So I will mark this PR as awaiting feedback until the related issues are all well considered.  Thanks for your contributions!

J-N-K:
Since nobody seems to object, how do we proceed on this one?

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
```diff
@@ -27,13 +27,15 @@
...
  * @author Davy Vanherbergen
+ * @author Jan N. Klug
  * @since 1.2.0
  */
 public class DmxController implements DmxService, ManagedService {
 
     private static Logger logger = LoggerFactory.getLogger(DmxController.class);
 
     private static int TRANSMIT_FREQUENCY_MS = 35;
+    private int repeatMode = DmxTransmitter.REPEAT_MODE_ALWAYS; // default is send every update
```
watou: Instead of `int`, it would be safer to use an `enum` type.  [Here](http://stackoverflow.com/a/2965252) is the construct that would seem to suit this optionally configured value.

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxTransmitter.java
```diff
@@ -27,16 +27,28 @@
  */
 public final class DmxTransmitter extends TimerTask {
 
+    /* REPEAT_INTERVAL is 750 ms (results in 800-1000ms) repetition time */
+    private static final int REPEAT_INTERVAL = 750;
+    private static final int REPEAT_COUNT = 3;
+
+    public static final int REPEAT_MODE_ALWAYS = 0;
+    public static final int REPEAT_MODE_NEVER = 1;
+    public static final int REPEAT_MODE_REDUCED = 2;
```
watou: These are more suitably an `enum` type.

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxTransmitter.java
```diff
@@ -89,8 +119,18 @@ public void setSuspend(boolean suspend) {
     }
 
     /**
+     * change transmitter refresh cycle
+     *
+     * @param refreshInterval
+     *            interval in ms (if output did not change)
+     */
+    public void setRepeatMode(int repeatMode) {
```
watou: `enum` type

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/core/DmxController.java
J-N-K: Good idea. A question regarding the coding style: should this enum be below DmxTransmitter (as are the constants now) or seperate?

watou:
The code looks good to me.  If you report back that this latest version of the PR is working reliably, and will continue to work the same (or better) for other existing users, I think it's ready to merge.  Also please update the wiki to reflect the `repeatMode` parameter and its detailed meaning.  Thanks!

J-N-K:
I'll test that for some days and report back. Do you think that positioning the enum in DmxTransmitter is a good place? If so, I'll squash all commits into one.

watou:
> Do you think that positioning the enum in DmxTransmitter is a good place?

There is no issue with nesting it where you did.  Java enums are implicitly static, so that's good too.

> If so, I'll squash all commits into one.

No need to unless you want to, since maintainers squash and merge in one step nowadays.

Thanks for your good work, @J-N-K!

J-N-K:
It's four weeks now since this is ready for merge. Please let me know if I should close this PR

PR #4786, opened 2016-11-16
Waterkotte Ecotouch heatpump binding for OH2

sibbi77:
This pull request adds some more recognized tags to the binding and provides the metadata for OH2.
I can squash the commits into one, if requested.

Open problems:
OH2 shows the binding under PaperUI/Configuration/Bindings, but changes to the configuration are ignored (AbstractActiveBinding<>.updated(Dictionary<String, ?> config) is called with config == null, if no ecotouch.cfg file is in place).
Only configuration via the provided ecotouch.cfg file works.

### DIFF
--- a/bundles/binding/org.openhab.binding.ecotouch/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.ecotouch/ESH-INF/binding/binding.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="ecotouch"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>Waterkotte EcoTouch Binding</name>
+    <description>This is the binding for the Waterkotte EcoTouch heatpump.</description>
+    <author>Sebastian Held</author>
+
+    <config-description>
+        <parameter name="refresh" type="integer">
+            <label>Refresh interval (ms)</label>
+            <description>Data refresh interval in ms.</description>
+            <default>60000</default>
+        </parameter>
+        <parameter name="ip" type="text">
+            <label>IP address</label>
+            <description>The IP address of the Waterkotte EcoTouch display unit.</description>
+            <default></default>
+        </parameter>
+        <parameter name="username" type="text">
+            <label>Username</label>
+            <description>Username of the display unit.</description>
+            <default>admin</default>
+        </parameter>
+        <parameter name="password" type="text">
+            <label>Password</label>
+            <description>Password of the display unit.</description>
+            <default>wtkadmin</default>
+        </parameter>
+    </config-description>
+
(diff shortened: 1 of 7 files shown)


watou:
It is common for `updated()` to be called with `config` being `null`.  But are you saying that attempts to use the Paper UI to configure the binding never take at all?

watou:
Thanks very much!  Aside from comments, lgtm.

sibbi77:
any attempt to configure the binding in the PaperUI only stores the configuration values at userdata/cache/org.eclipse.osgi/7/data/config/binding/ecotouch.config.
But this does not result in a valid config argument in the call to updated().
If I modify the conf/services/ecotouch.cfg and uncomment ip, username and password, the binding's refresh service starts.
Either ecotouch.cfg takes precedence over ecotouch.config, some ids inside binding.xml(?) not match or the framework does not propagate the UI config changes correctly (problem in compat layer?)

watou:
@kaikreuzer any thoughts on the above?

kaikreuzer:
I have a suspicion: I guess it simply ends up as a configuration for a different service pid.

I recently noticed that Paper UI sends binding configurations to a service pid `binding.<id>`, while openHAB 1 bindings expect their configuration as `org.openhab.<id>`. This would explain what you see.
I am nonetheless a bit puzzled, because I thought to remember seeing it once working. @watou When you did stuff like https://github.com/openhab/openhab/pull/4271/files, did you test it or did you blindly trust me?

watou:
@kaikreuzer: I thought I had tested it but at this point I can't vouch for my own testing.  

I would like to add a few more binding.xml files to bindings.  Can you see a way to support dynamic config keys like`home.port=443`, `office.ip=192.168.0.10`, etc. in the schema and UI?

kaikreuzer:
> I thought I had tested it but at this point I can't vouch for my own testing.

Ok, will have to find a solution for this then.

> Can you see a way to support dynamic config keys likehome.port=443, office.ip=192.168.0.10, etc. in the schema and UI?

No, we decided against this, because it is not really the way configurations are expected to be done in OSGi. To support such settings, we added the "expert mode" button in the Paper UI for configurations, so the user can still add it.

This mechanism was mainly used in OH1 for providing multiple instances of the same binding - in ESH, this is addressed by having Bridges/Things, so such configurations are not found not on the binding itself, but on the Bridges/Things that represent your instances ("home", "office", etc.).

watou:
Makes perfect sense.

watou:
> Ok, will have to find a solution for this then.

Could we add an optional service.pid setting to binding.xml, so that when it's present, the config is written to the given service PID instead of following the convention?  So all OH1 bindings would provide theirs.  I know it's duplication, but it seems like a fairly direct solution.

kaikreuzer:
@watou Good news, exactly such a mechanism is already in place! I successfully tested it with the Nest binding, see https://github.com/openhab/openhab/pull/4804/files. Could you update the OH1 bindings accordingly? I'll make sure to update the documentation to cover this.

watou:
> Could you update the OH1 bindings accordingly?

Great news.  I will update the other 1.x bindings.

watou:
@sibbi77: Could you add `<service-id>org.openhab.ecotouch</service-id>` to `ESH-INF/binding/binding.xml` and see if it resolves the issue you identified in https://github.com/openhab/openhab/pull/4786#issuecomment-261566477 ?

sibbi77:
Now it works as expected. Thanks!

PR #4789, opened 2016-11-17
[garadget] add config metadata

watou:
To address forum comment [here](https://community.openhab.org/t/garadget-binding/7650/23?u=watou).

### DIFF
--- a/bundles/binding/org.openhab.binding.garadget/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.garadget/ESH-INF/binding/binding.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="garadget"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>Garadget Binding</name>
+    <description>Monitor and control Garadget garage door futurizers using the Particle cloud-based API.</description>
+    <author>John Cocula</author>
+
+    <config-description>
+        <parameter name="username" type="text" required="true">
+            <label>Username</label>
+            <description>Username for accessing your account</description>
+        </parameter>
+        <parameter name="password" type="text" required="true">
+            <label>Password</label>
+            <description>Password for accessing your account</description>
+            <context>password</context>
+        </parameter>
+        <parameter name="granularity" type="integer">
+            <label>Poll Granularity</label>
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Thanks!

watou:
Thanks @kaikreuzer.  Are you aware of recent reports at the forum that configuring 1.x addons via Paper UI hasn't worked, and they have to resort to the .cfg file?  If both `ESH-INF/binding/binding.xml` and `<service>.cfg` are present, which source is expected to control?

kaikreuzer:
No, do you have a link?

watou:
Sorry the posts are bit hard to parse through:

* https://community.openhab.org/t/howto-manage-openhab-2-configurations/15940/3
* https://community.openhab.org/t/mqtt-and-oh2-not-working/1516/31
* https://community.openhab.org/t/mqtt-config-under-oh2/15497/9
* https://community.openhab.org/t/nest-binding-on-oh-2/13958/5

The 1.x addons that currently have ESH-INF metadata are:

```
$ find . -name ESH-INF -print
./bundles/binding/org.openhab.binding.dsmr/ESH-INF
./bundles/binding/org.openhab.binding.ecobee/ESH-INF
./bundles/binding/org.openhab.binding.garadget/ESH-INF
./bundles/binding/org.openhab.binding.gpio/ESH-INF
./bundles/binding/org.openhab.binding.nest/ESH-INF
./bundles/binding/org.openhab.binding.onewire/ESH-INF
./bundles/binding/org.openhab.binding.plugwise/ESH-INF
./bundles/binding/org.openhab.binding.powermax/ESH-INF
./bundles/binding/org.openhab.binding.wol/ESH-INF
./bundles/persistence/org.openhab.persistence.jdbc/ESH-INF
```

kaikreuzer:
Thanks for the links - as far as I can see, they are all about the fact that configuration that you once did in a cfg file is NOT removed once you remove it from the cfg file.
This is correct, the cfg files are now only used to set values, but configuration is not removed as it might have been edited through other channels (UI, console, etc.). Since very recent, users can manually remove them (without requiring an "osgi clean") by editing the files in userdata/etc.
I know that this is a bit unexpected and it definitely needs to be documented, but I don't really have a better idea right now. Maybe we could introduce some meta data in the cfg files that say whether existing config sets should be fully deleted when being processed?

But I cannot see any "reports that configuring 1.x addons via Paper UI hasn't worked" - do I overlook those?

watou:
I've either forgotten a link or was viewing the issue incorrectly.  A more surprise-free solution would be good, but it definitely requires some thought.  

I think that at the moment the user changes a configuration for service X via either UI or .cfg file, all config for service X should be replaced by it.  Maybe a hidden timestamp could be stored, and the .cfg file's last modified time would determine if it should replace what is in ConfigAdminManager?  
On startup, the hidden timestamp would be compared to the file last modified timestamp to determine if the config should be cleared and reloaded from file?

watou:
More articulate link describing issue here: https://github.com/openhab/openhab/pull/4786#issuecomment-261566477

PR #4797, opened 2016-11-19
[Plugwise] Reliability improvements

wborn:
Improves the binding reliability by:
* making sure CR is followed by LF when reading incoming packets, which
should result in less protocol message errors
* polling max 1 second for `AcknowledgeMessage`s so the `SendThread` does
not wait indefinitely when ACKs get lost

The most common scenario for the binding to stop working should now be solved. I.e. a received `AcknowledgeMessage` results in a protocol message error (e.g. because it is preceded by the LF of a previous message). Especially on Windows this seems to be an issue due to non-blocking reads.

### DIFF
--- a/bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/Stick.java
+++ b/bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/Stick.java
@@ -77,8 +77,12 @@ public class Stick extends PlugwiseDevice implements SerialPortEventListener {
     /** Plugwise protocol header code (hex) */
     private final static String PROTOCOL_HEADER = "\u0005\u0005\u0003\u0003";
 
+    /** Carriage return */
+    private final static char CR = '\r';
+    /** Line feed */
+    private final static char LF = '\n';
     /** Plugwise protocol trailer code (hex) */
-    private final static String PROTOCOL_TRAILER = "\r\n";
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/Stick.java
```diff
@@ -77,8 +77,12 @@
     /** Plugwise protocol header code (hex) */
     private final static String PROTOCOL_HEADER = "\u0005\u0005\u0003\u0003";
 
+    /** Carriage return */
+    private final static char CR = '\r';
+    /** Line feed */
+    private final static char LF = '\n';
     /** Plugwise protocol trailer code (hex) */
-    private final static String PROTOCOL_TRAILER = "\r\n";
+    private final static String PROTOCOL_TRAILER = new String(new char[] { CR, LF });
```
watou: I know you didn't introduce this, and it's very minor, but there is a recommended order of modifiers in Java which improves readability.  [Here](http://cr.openjdk.java.net/~alundblad/styleguide/index-v6.html#toc-modifiers) is one place that lists the common order:

1. Access modifier (public / private / protected)
1. abstract
1. static
1. final
1. transient
1. volatile
1. default
1. synchronized
1. native
1. strictfp

wborn:
Thanks for reviewing @watou! I think it should be fixed now.

watou:
Thanks @wborn!

PR #4806, opened 2016-11-21
Add service ids to ESH-INF/binding/binding.xml files

watou:
Following the example in #4804, add `<service-id>org.openhab.ID</service-id>` to all 1.x bindings that already provide an `ESH-INF/binding/binding.xml` file for configuration through the Paper UI.  This should result in the binding services finding the configuration at the service PIDs where they are expected to be.  

The GPIO binding was a special case because it did not explicitly set its `service.pid`, which I understand means that the component's name is used instead.  So I explicitly set the `service.pid` for the service component to match how other bindings name their service PIDs.


kaikreuzer:
Many thanks!

PR #4808, opened 2016-11-21
[fritzboxtr064] initial support Fritz!DECT 200

bbock:
This PR adds support for the Fritz!DECT 200 smart home power plugs to the fritzboxtr064 binding

Currenty, only switching mains on/off is supported. The additional features of the
Fritz!DECT 200 (power meter, thermometer) are currently unsupported.

Each Fritz!DECT has an unique identifier, which is called AIN. This identifier
can be found in the Fritz!Box GUI. This identifier is used to distinguish between
multiple Fritz!DECT 200 connected to the Fritz!Box.

An example item configuration line looks like this:

Switch mainswitch "Power" {fritzboxtr064="powerplug:08761-0123456"}

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
@@ -194,8 +194,7 @@ public class Tr064Comm {
                 // if input is mac address, replace "-" with ":" as fbox wants
                 if (itemMap.getItemCommand().equals("maconline")) {
                     dataInValue = dataInValue.replaceAll("-", ":");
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
```diff
@@ -195,6 +195,9 @@ public String getTr064Value(String request) {
                 if (itemMap.getItemCommand().equals("maconline")) {
                     dataInValue = dataInValue.replaceAll("-", ":");
                 }
+                if (itemMap.getItemCommand().equals("powerplug")) {
```
9037568: This should be an "else if"

watou:
Please be sure the wiki is up to date after this is merged, thanks!

PR #3589, opened 2015-12-23
SimpleBinary binding

docbender:
Hi, this binding enables connect devices through UART using simple binary protocol. Protocol is easy to implement into home made devices.

My binding repository can be found here: https://github.com/docbender/openHAB-SimpleBinary

PR #4809, opened 2016-11-21
Added SimpleBinary binding

docbender:
New merged binding pull request ready to review. Original pull request was #3589.


bundles/binding/org.openhab.binding.simplebinary/src/main/java/org/openhab/binding/simplebinary/internal/SimpleBinaryUART.java
```diff
@@ -0,0 +1,561 @@
...
+    /**
+     * Write data into device stream
+     *
+     * @param data
+     *            Item data with compiled packet
+     */
+    @Override
+    protected boolean sendDataOut(SimpleBinaryItemData data) {
+        // data line stabilization
+        if (LINE_STABILIZATION_TIME > 0) {
```
9037568: This block can never be executed. (The "variable" is always 0)

bundles/binding/org.openhab.binding.simplebinary/src/main/java/org/openhab/binding/simplebinary/internal/SimpleBinaryProtocol.java
```diff
@@ -0,0 +1,799 @@
...
+            Map<String, SimpleBinaryBindingConfig> itemsConfig, String deviceName, boolean letDataInBuffer)
+            throws NoValidCRCException, NoValidItemInConfig, UnknownMessageException, ModeChangeException {
+        byte devId = data.get();
+        byte msgId = data.get();
+        int address = 0;
+        byte[] rawPacket;
+        byte[] rawData = null;
+        byte crc = 0, calcCrc;
+
+        try {
```
9037568: This try block catches nothing. Why is it here?

bundles/binding/org.openhab.binding.simplebinary/src/main/java/org/openhab/binding/simplebinary/internal/SimpleBinaryUART.java
```diff
@@ -0,0 +1,561 @@
...
+                timeoutTask.cancel();
+            }
+
+            timeoutTask = new TimerTask() {
+                @Override
+                public void run() {
+                    timeoutTask = null;
+                    dataTimeouted();
+                    setWaitingForAnswer(false);
+                    // TODO: should i call it????
```
9037568: Remove or clarify

bundles/binding/org.openhab.binding.simplebinary/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,30 @@
...
+ org.openhab.core.types,
+ org.openhab.model.item.binding,
+ org.osgi.framework,
+ org.osgi.service.component,
+ org.osgi.service.event,
+ org.slf4j
+Export-Package: org.openhab.binding.simplebinary
+Bundle-DocURL: http://www.openhab.org
+Bundle-RequiredExecutionEnvironment: J2SE-1.5
+Service-Component: OSGI-INF/binding.xml, OSGI-INF/genericbindingprovider.xml
```
9037568: Bundle-ActivationPolicy should be set to lazy

9037568 (changes requested):
Thanks for this contribution.

In addition to my inline comments:

Please run all the source through the Eclipse formatter.

Also, in a number of places you are using this construct:
````
       if (logger.isDebugEnabled()) {
            logger.debug("execute() method is called!");
       }
````

This is really not worth doing. The wrapper that checks the logging level should only be used around operations that are expensive. You should consider eliminating these if conditionals anywhere that they're only wrapping a call to logger.x().

docbender:
Hello, 
thanks for your review.

About _logger_ section. I had a problem in production environment (logger level=WARN) with heavy traffic that sometimes totally random it just crash. I found that it was at _logger.debug(...)_ line.  With condition _logger.isDebugEnabled()_ this crashes disappeared.

Another one. Why should I change year in header? Some files was created in 2015. Same with openhab version. Pull request of this binding was pulled a year ago before 1.8.0 was released.

Finally complete all missing header doc would be long time run.

9037568:
> About logger section.

That's merely advisory. If you don't want to change those, it'll still be fine to merge.

> Another one. Why should I change year in header?

As far as I'm aware, the convention is to have the dates in the copyright headers be up to date based on the year the code is first merged. Certainly the openhab version in the "since" tags needs to be correct. It's also a quick and easy fix.

> Finally complete all missing header doc would be long time run.

If this is about the "@return" doc comments that are incomplete/empty, again I'd call those advisory. I think they'd still be ok to merge if you choose not to clean them up.

docbender:
Modified code pushed in. Thanks for review

9037568:
Thanks for these updates. There are still a few issues of commented code and copyright dates that need to be addressed, however.

9037568:
I just realized while perusing the wiki page for this binding that you need to add your configuration items to the [openhab_default.cfg](https://github.com/openhab/openhab/blob/master/distribution/openhabhome/configurations/openhab_default.cfg) file.

docbender:
I pushed openhab_default.cfg update. I also corrected logger strings and the wrong copyright date in OSGI files.

9037568:
Good progress. Still a few left.

PR #4815, opened 2016-11-24
Fix IllegalArgumentException with ColorItem MapDB persistence

wborn:
Because `ColorItem` extends `DimmerItem` the current `MapDBPersistenceService` implementation tries to store the item state as a percentage.

`item.getStateAs(PercentType.class)` then causes the following exception:

```
java.lang.IllegalArgumentException: Value must be between 0 and 100
```

Changing the if-then-else order results in the correct behavior.


watou:
Thanks @wborn!

PR #4822, opened 2016-11-25
Improved the handling of incorrect device id's

martinvw:
I accidently used 3:brightness and this does not lead to the logging of any error, even while running in debug mode, after I traced down the problem I now changed the behavior the message "Not a valid id on the bridge: {bridge}" is logged and no further attempts to execute commands or update the status from the bridge are made.

The reason the old code did not work was that a null object was always wrapped in some internal object (see the code in HueSetting), further in the execution a null pointer was triggered, because the internal state of bulb was inconsistent because nothing could be retrieved from the bridge. That NPE was (unfortunately) swallowed by openHAB's core classes.

The reason an exception is now thrown from the HueBuld is that otherwise a HueBulb object will be constructed and/or used while it has an invalid state (it refers to a non-existing light).


9037568:
lgtm. Do we need any community feedback on this?

martinvw:
I saw no other reports of this problem, but if someone wants to test this build I can append it here, do you want that?

9037568:
No need...

PR #4823, opened 2016-11-25
New "expire" binding. Just add to your item: expire="10s" or

micw:
expire="12h30m50s". This binding will set the state to Undefined if it's
not updated within the given time.

### DIFF
--- a/bundles/binding/org.openhab.binding.expire/.classpath
+++ b/bundles/binding/org.openhab.binding.expire/.classpath
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
+	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources"/>
+	<classpathentry kind="output" path="target/classes"/>
+</classpath>
--- a/bundles/binding/org.openhab.binding.expire/.classpath
+++ b/bundles/binding/org.openhab.binding.expire/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
 	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources"/>
--- a/bundles/binding/org.openhab.binding.expire/.classpath
+++ b/bundles/binding/org.openhab.binding.expire/.classpath
@@ -3,6 +3,5 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
-	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources"/>
 	<classpathentry kind="output" path="target/classes"/>
 </classpath>
--- a/bundles/binding/org.openhab.binding.expire/.gitignore
+++ b/bundles/binding/org.openhab.binding.expire/.gitignore
@@ -1,0 +1,3 @@
+/.classpath
+/.project
+/.settings/
--- a/bundles/binding/org.openhab.binding.expire/.gitignore
+++ b/bundles/binding/org.openhab.binding.expire/.gitignore
@@ -1,3 +1,0 @@
-/.classpath
-/.project
-/.settings/
--- a/bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
+++ b/bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
@@ -1,0 +1,7 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
+org.eclipse.jdt.core.compiler.compliance=1.7
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.source=1.7
--- a/bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
+++ b/bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
@@ -1,7 +1,7 @@
 eclipse.preferences.version=1
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
-org.eclipse.jdt.core.compiler.compliance=1.7
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
+org.eclipse.jdt.core.compiler.compliance=1.8
 org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
 org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
-org.eclipse.jdt.core.compiler.source=1.7
+org.eclipse.jdt.core.compiler.source=1.8
--- a/bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
@@ -1,0 +1,31 @@
+Manifest-Version: 1.0
+Private-Package: org.openhab.binding.expire.internal
+Ignore-Package: org.openhab.binding.expire.internal
+Bundle-License: http://www.eclipse.org/legal/epl-v10.html
+Bundle-Name: openHAB Expire Binding
+Bundle-SymbolicName: org.openhab.binding.expire
+Bundle-Vendor: openHAB.org
+Bundle-Version: 1.9.0.qualifier
+Bundle-ManifestVersion: 2
+Bundle-Description: This is the Expire binding of the open Home Aut
+ omation Bus (openHAB)
+Import-Package: org.apache.commons.lang,
+ org.openhab.core.binding,
+ org.openhab.core.events,
+ org.openhab.core.items,
+ org.openhab.core.library.items,
+ org.openhab.core.library.types,
+ org.openhab.core.transform,
+ org.openhab.core.types,
+ org.openhab.model.item.binding,
+ org.osgi.framework,
+ org.osgi.service.cm,
+ org.osgi.service.component,
+ org.osgi.service.event,
+ org.slf4j
+Export-Package: org.openhab.binding.expire,
+ org.openhab.binding.expire.internal;x-internal:=true
+Bundle-DocURL: http://www.openhab.org
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Service-Component: OSGI-INF/binding.xml, OSGI-INF/genericbindingprovider.xml
+Bundle-ClassPath: .
--- a/bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
@@ -23,8 +23,7 @@ Import-Package: org.apache.commons.lang,
  org.osgi.service.component,
  org.osgi.service.event,
  org.slf4j
-Export-Package: org.openhab.binding.expire,
- org.openhab.binding.expire.internal;x-internal:=true
+Export-Package: org.openhab.binding.expire
 Bundle-DocURL: http://www.openhab.org
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Service-Component: OSGI-INF/binding.xml, OSGI-INF/genericbindingprovider.xml
--- a/bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
@@ -25,6 +25,6 @@ Import-Package: org.apache.commons.lang,
  org.slf4j
 Export-Package: org.openhab.binding.expire
 Bundle-DocURL: http://www.openhab.org
-Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Service-Component: OSGI-INF/binding.xml, OSGI-INF/genericbindingprovider.xml
 Bundle-ClassPath: .
--- a/bundles/binding/org.openhab.binding.expire/OSGI-INF/binding.xml
+++ b/bundles/binding/org.openhab.binding.expire/OSGI-INF/binding.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2010-2015, openHAB.org and others.
+    Copyright (c) 2010-2016 by the respective copyright holders.
 
     All rights reserved. This program and the accompanying materials
     are made available under the terms of the Eclipse Public License v1.0
--- a/bundles/binding/org.openhab.binding.expire/OSGI-INF/binding.xml
+++ b/bundles/binding/org.openhab.binding.expire/OSGI-INF/binding.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2010-2015, openHAB.org and others.
+
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+-->
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.2.0" activate="activate" modified="modified" deactivate="deactivate" name="org.openhab.binding.expire.binding" immediate="true" configuration-pid="org.openhab.expire" configuration-policy="optional">
+	<implementation class="org.openhab.binding.expire.internal.ExpireBinding" />
+
+	<service>
+		<provide interface="org.osgi.service.event.EventHandler" />
+	</service>
+
+	<property name="event.topics" type="String" value="openhab/*" />
+
+	<reference bind="setEventPublisher" cardinality="1..1"
+		interface="org.openhab.core.events.EventPublisher" name="EventPublisher"
+		policy="dynamic" unbind="unsetEventPublisher" />
+	<reference bind="addBindingProvider" cardinality="1..n"
+		interface="org.openhab.binding.expire.ExpireBindingProvider" name="ExpireBindingProvider"
+		policy="dynamic" unbind="removeBindingProvider" />
+	
+</scr:component>
--- a/bundles/binding/org.openhab.binding.expire/pom.xml
+++ b/bundles/binding/org.openhab.binding.expire/pom.xml
@@ -11,7 +11,7 @@
 	<properties>
 		<bundle.symbolicName>org.openhab.binding.expire</bundle.symbolicName>
 		<bundle.namespace>org.openhab.binding.expire</bundle.namespace>
-		<deb.name>openhab-addon-binding-Expire</deb.name>
+		<deb.name>openhab-addon-binding-expire</deb.name>
 		<deb.description>${project.name}</deb.description>
 	</properties>
 
--- a/bundles/binding/org.openhab.binding.expire/pom.xml
+++ b/bundles/binding/org.openhab.binding.expire/pom.xml
@@ -13,6 +13,8 @@
 		<bundle.namespace>org.openhab.binding.expire</bundle.namespace>
 		<deb.name>openhab-addon-binding-expire</deb.name>
 		<deb.description>${project.name}</deb.description>
+		<maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
 	</properties>
 
 	<modelVersion>4.0.0</modelVersion>
--- a/bundles/binding/org.openhab.binding.expire/pom.xml
+++ b/bundles/binding/org.openhab.binding.expire/pom.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<parent>
+		<groupId>org.openhab.bundles</groupId>
+		<artifactId>binding</artifactId>
+		<version>1.9.0-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<bundle.symbolicName>org.openhab.binding.expire</bundle.symbolicName>
+		<bundle.namespace>org.openhab.binding.expire</bundle.namespace>
+		<deb.name>openhab-addon-binding-Expire</deb.name>
+		<deb.description>${project.name}</deb.description>
+	</properties>
+
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>org.openhab.binding</groupId>
+	<artifactId>org.openhab.binding.expire</artifactId>
+
+	<name>openHAB Expire Binding</name>
+
+	<packaging>eclipse-plugin</packaging>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.vafer</groupId>
+				<artifactId>jdeb</artifactId>
+			</plugin>
+		</plugins>
+	</build>
+
+</project>
--- a/bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
+++ b/bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010-2015, openHAB.org and others.
+ * Copyright (c) 2010-2016 by the respective copyright holders.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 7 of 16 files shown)


bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        for (ExpireBindingProvider provider : providers)
+        {
+            for (String itemName : provider.getItemNames())
+            {
+                Long nextExpireTs = nextExpireTsMap.get(itemName);
+                if (nextExpireTs == null || nextExpireTs <= System.currentTimeMillis())
+                {
+                    // value expired
+                    logger.info("Item {} was not updated for {} - setting state to undefined", itemName, "5s");
```
martinvw: Hello, nice idea, i see some neat use cases for this.

Could it be that hat this line you require a more dynamic thing instead of the 5s?

micw:
Thank you for the feedback. I 've fixed it.

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireGenericBindingProvider.java
```diff
@@ -54,14 +54,26 @@ public void processBindingConfiguration(String context, Item item, String bindin
         super.processBindingConfiguration(context, item, bindingConfig);
         ExpireBindingConfig config = new ExpireBindingConfig();
 
-        config.expiresAfterMs = parseDuration(bindingConfig);
+        Duration expiresAfter = parseDuration(bindingConfig);
```
martinvw: Isn't it way more easy to just store the string you have as an input here?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+     * method and must not be accessed anymore once the deactivate() method was called or before activate()
+     * was called.
+     */
+    private BundleContext bundleContext;
+
+    /**
+     * the refresh interval which is used to poll values from the Expire
+     * server (optional, defaults to 60000ms)
+     */
+    private long refreshInterval = 1000;
```
watou: comment disagrees with the default value.  Also, I don't think the current code has a way to change it via configuration (not that I can think of any reason to make it configurable), so I think the comment should be improved.

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        protected long expireAfterMs;
+
+        protected long nextExpireTs;
+    }
+
+    protected static Map<String, Long> nextExpireTsMap = new HashMap<>();
+
+    /**
+     * @{inheritDoc}
```
watou: {@inheritDoc}

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+                {
+                    // value expired
+                    logger.info("Item {} was not updated for {} - setting state to undefined", itemName, provider.getExpiresAfterAsText(itemName));
+                    eventPublisher.postUpdate(itemName, UnDefType.UNDEF); // set to undefined
+                    nextExpireTsMap.put(itemName, Long.MAX_VALUE); // disable expire trigger until next update
+                }
+
+            }
+        }
+    }
```
watou: Please put all the code through the Eclipse formatter as configured for openHAB.

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+        for (ExpireBindingProvider provider : providers)
+        {
+            for (String itemName : provider.getItemNames())
+            {
+                Long nextExpireTs = nextExpireTsMap.get(itemName);
+                if (nextExpireTs == null || nextExpireTs <= System.currentTimeMillis())
+                {
+                    // value expired
+                    logger.info("Item {} was not updated for {} - setting state to undefined", itemName, provider.getExpiresAfterAsText(itemName));
+                    eventPublisher.postUpdate(itemName, UnDefType.UNDEF); // set to undefined
```
watou: Suggestion: I can think of use cases for adding an optional state to the binding config string, so a state other than `UnDefType.UNDEF` could be used to indicate expiration.  So it could look like `expire="1d2h5m,No motion for 26 hours"` or `expire="1d,0"`.  You would use the `TypeParser` to post the correct State that corresponds to the item's `acceptedDataTypes`.  If the `,<state>` option wasn't present, it would set the item's state to `UnDefType.UNDEF` as you are doing here. WDYT?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+                    nextExpireTsMap.put(itemName, Long.MAX_VALUE); // disable expire trigger until next update
+                }
+
+            }
+        }
+    }
+
+    /**
+     * @{inheritDoc}
```
watou: {@inheritDoc}

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public void deactivate(final int reason)
+    {
+        this.bundleContext = null;
+        // deallocate resources here that are no longer needed and
+        // should be reset when activating this binding again
+    }
+
+    /**
+     * @{inheritDoc}
```
watou: {@inheritDoc}

bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,31 @@
...
+ org.openhab.core.transform,
+ org.openhab.core.types,
+ org.openhab.model.item.binding,
+ org.osgi.framework,
+ org.osgi.service.cm,
+ org.osgi.service.component,
+ org.osgi.service.event,
+ org.slf4j
+Export-Package: org.openhab.binding.expire,
+ org.openhab.binding.expire.internal;x-internal:=true
```
micw: I have no idea what to have in MANIFEST.MF. I just used that one that was created by the archetype. Feel free to change it to the project's requirements ;-)

bundles/binding/org.openhab.binding.expire/OSGI-INF/binding.xml
```diff
@@ -0,0 +1,28 @@
...
+<!--
+
+    Copyright (c) 2010-2015, openHAB.org and others.
+
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+-->
```
micw: Done

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: I missed this auto generated one. Fixed

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: What do you mean?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: I did. Got no changes...

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: Nice idea. Done.

bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
watou: Would you mind removing line 27 and the comma at the end of line 26?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
watou: The archetype produces incorrect "inheritDoc" tags when they should be `{@inheritDoc}`.

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
watou: Are you sure you're using the formatter as results from [these instructions](http://docs.openhab.org/developers/development/ide.html)?  Your code has opening curly braces on their own lines, but when I format code in my IDE, they move to the end of the preceding line.

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
watou: Thanks!

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /**
+     * @{inheritDoc}
+     */
+    @Override
+    protected void internalReceiveUpdate(String itemName, State newState)
+    {
+        if (newState != UnDefType.UNDEF)
```
watou: Should this check for `config.getExpiredState()` instead of UNDEF?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: All my eclipse settings are auto-generated by the installer. But for some reason the fresh eclipse contains our company's style as well - and uses it by default. Which style should I use?

micw: So please let me know which one it is. There is:

- Java conventions
- Eclipse
- Eclipse 2.1
- ESH

watou: It's set to ESH in my IDE, so I think that's the one.

watou:
Could you add this binding to the parent pom.xml?

micw:
sure.

watou:
This is great, I have motion switches that are only turned on by movement but not off, but I would like to have them switch off "on their own."  So all I will have to do for each of them is add `expire="2m,OFF"`!

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireGenericBindingProvider.java
```diff
@@ -0,0 +1,138 @@
...
+
+        String[] durationAndState = bindingConfig.split(",", 2);
+
+        Duration expiresAfter = parseDuration(durationAndState[0].trim());
+        config.expiresAfterMs = expiresAfter.toMillis();
+        config.expiresAfterAsText = formatDuration(expiresAfter);
+
+        if (durationAndState.length > 1) {
+            // use 2nd parameter as state. defaults to UNDEF
+            config.expiredState = TypeParser.parseState(item.getAcceptedDataTypes(), durationAndState[1].trim());
```
watou: The `TypeParser` can return `null`, so I think that warrants throwing a `BindingConfigParseException` that contains an explanation like "The given string '{}' could not be used for any accepted data type for item {}."

micw:
You're right, i have added this.

micw:
My personal use cases are ESP8266 based temperature sensors that somtimes loose contact (e.g. the one at my garage does not connect to wifi if there's lot of rain). I don't want to see outdated values.
After I have seen the rule based solutions I decided to create this binding ;-)

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void execute() {
+        for (ExpireBindingProvider provider : providers) {
+            for (String itemName : provider.getItemNames()) {
+                Long nextExpireTs = nextExpireTsMap.get(itemName);
+                if (nextExpireTs == null || nextExpireTs <= System.currentTimeMillis()) {
```
martinvw: I have one last small remark here, it seems that this line causes all 'expire-aware' items to be expired on start, which is especially for items which update less often (and get update ever few hours) potentially bad, because I recover that status on startup but after that it wel get erased here.

Or are the initial value updated also piped through this plug-in so that this problem will not occur.

watou: I think this binding should avoid posting updates in all cases except when it's been >= the expire time since a non-expiredState update has been received. @micw, WDYT?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: done

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: It depends on how initial state restores are handled. If it's sure that the restore is passed as state update to the binding, all is fine with this implementation. If not, @watou 's suggetion won't be sufficient because the restored value would never expire.

I changed it the following way: if an item has no entry in the expireTsMap, I create one with the initial expire time. So if ther's no update after initialisation, the item is expired after expire time instead and not immediatelly. It it would receive an update at start, this will work as well.

bundles/binding/org.openhab.binding.expire/.gitignore
```diff
@@ -0,0 +1,3 @@
+/.classpath
+/.project
+/.settings/
```
watou: You have .settings in your .gitignore, but this PR contains files from your .settings directory.  Most bindings don't have their own .gitignore.  Can you instead rely on the one higher in the tree?

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,156 @@
...
+                Long nextExpireTs = nextExpireTsMap.get(itemName);
+                if (nextExpireTs == null) {
+                    // set initial expireTs
+                    long expireAfterMs = provider.getExpiresAfterMs(itemName);
+                    nextExpireTsMap.put(itemName, System.currentTimeMillis() + expireAfterMs);
+                } else if (nextExpireTs <= System.currentTimeMillis()) {
+                    // value expired
+                    State newState = provider.getExpiredState(itemName);
+                    logger.info("Item {} was not updated for {} - setting state to {}", itemName,
+                            provider.getExpiresAfterAsText(itemName), newState);
```
watou: I think this should be at debug log level since it's normal operation for the binding.

micw: I'm not sure here. It triggers an ItemStateEvent that logs that the item state changed to UNDEF (or whatever configured). ItemStateEvent logs as Info. So IMO it would be good to see, why it was set to UNDEF.

Example log:
```

2016-11-26 20:41:17.098 [INFO ] [smarthome.event.ItemStateEvent:43   ] - Temperature_Test2 updated to 64
2016-11-26 20:41:17.098 [INFO ] [s.event.ItemStateChangedEvent :43   ] - Temperature_Test2 changed from UNDEF to 64
2016-11-26 20:41:25.696 [INFO ] [.expire.internal.ExpireBinding:137  ] - Item Temperature_Test1 was not updated for 10s  - setting state to undefined
2016-11-26 20:41:25.697 [INFO ] [smarthome.event.ItemStateEvent:43   ] - Temperature_Test1 updated to UNDEF
2016-11-26 20:41:25.698 [INFO ] [s.event.ItemStateChangedEvent :43   ] - Temperature_Test1 changed from 7 to UNDEF
```

bundles/binding/org.openhab.binding.expire/.gitignore
micw: I committed that file accidentally. I had a look to the other plugins - seems that people use to commit those eclipse files. If this is general project style, I just remove my gitignore. (personally I think, this files should not be committed for maven projects because it makes the project less portable between IDEs but I have impression that it might be almost impossible to develop without the auto-installed eclipse version).

micw: I agree

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: I agree. Changed.

watou:
Does anyone see a reason not to merge this PR yet?  @micw would you mind creating a wiki page for the new binding?

martinvw:
I'm testing the binding now, first it seemed to work but now it stopped functioning.

@micw did you test the most recent version on your openhab instance.

micw:
I run it on my eclipse. Works as expected for me. These are my test items:

String Temperature_Test1  "Temperature [%s]" { expire="10s,Undefined", http="<[https://www.random.org/integers/?num=1&min=1&max=100&col=1&base=10&format=plain&rnd=new:10000:default()]" }
String Temperature_Test2  "Temperature [%s]" { expire="10s,Undefined", http="<[https://www.random.org/integers/?num=1&min=1&max=100&col=1&base=10&format=plain&rnd=new:15000:default()]" }

martinvw:
I ran it on a switch which I manually switched now and then

Op zo 27 nov. 2016 om 21:04 schreef micw <notifications@github.com>

micw:
Tested with:
Switch Switch1 { expire="10s" }

The switch has 3 states with different visualisation in basicUI:
ON: Green symbol, Switch is on
OFF: Red symbol, Switch is off
UNDEF: Green symbol, Switch is off

I tested from both states, "ON" and "OFF" - both go to "UNDEF" after a while.

Please send me your Item definition to check.

watou:
Things to be aware of (might be good for the wiki page?):

* You can specify the expired state for UnDefType.UNDEF, but it's `Undefined` on openHAB 1.8 and `UNDEF` on openHAB 2 (from Eclipse SmartHome).  
* This will never work correctly for String items on openHAB 1.8, according to [this code](https://github.com/openhab/openhab/blob/1.8/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/StringItem.java#L36).  It will update to a StringType containg the string "Undefined".
* On openHAB 2 (and Eclipse SmartHome), you can correctly set a String item to UnDefType.UNDEF because the order of accepted data types was changed [here](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/items/StringItem.java#L37).

micw:
@watou Thank you. I was just starting to debug into this because I saw that expire="10s,UNDEF" was different than expire="10s" and saw the wrong order of the items.

I think about how to fix it because I'm more and more convinced to stay with OpenHAB 1.0 because I get the feeling that OpenHAB 2.0 is on a terrible wrong way doing easy things on very complex and hard ways (but that's another story that does not belong to this pull request ;-) )

micw:
@watou I added a workaround to get correct results on OpenHAB 1.x. ON will now produce a OnOffType, UNDEF an UnDefType, even for String items.

micw:
Maybe this workaround should be moved to typeParser ;-)

martinvw:
Most likely my bad then, my openHAB is running a little bit laggy and slow.

The only difference is that my switch was combined with a Zwave association, so your test should suffice.

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireGenericBindingProvider.java
```diff
@@ -73,6 +80,39 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
+        for (Class<? extends State> state : OPTIMAL_STATE_ORDER) {
+            if (states.remove(state)) {
+                orderedStates.add(state);
+            }
+        }
+        // Add remaining states to the list
+        orderedStates.addAll(states);
+
+        return orderedStates;
+    }
```
watou: It's a very elegant workaround, but IMHO it's better to work with the order of accepted data types as they are set in OH1 or ESH.  Please submit issues and pull requests to Eclipse SmartHome if there are problems or potential problems with the order (for example, the second accepted data type should be DateTimeType, but because it comes after StringType, it will never be reached).

watou: Sorry @micw, I closed #4838 as there will never be a release that would contain it.

micw:
I reverted the workaround. Is there anything left that prevents PR from being accepted? When will it be in debian repo? ;-)

bundles/binding/org.openhab.binding.expire/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+	<parent>
+		<groupId>org.openhab.bundles</groupId>
+		<artifactId>binding</artifactId>
+		<version>1.9.0-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<bundle.symbolicName>org.openhab.binding.expire</bundle.symbolicName>
+		<bundle.namespace>org.openhab.binding.expire</bundle.namespace>
+		<deb.name>openhab-addon-binding-Expire</deb.name>
```
watou: openhab-addon-binding-expire (lowercase "e")

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    protected String getName() {
+        return "Expire Refresh Service";
+    }
+
+    protected static class ExpireConfig {
+        protected long expireAfterMs;
+
+        protected long nextExpireTs;
+    }
```
watou: I think this static class isn't used?

watou:
> When will it be in debian repo? ;-)

No answer yet aside from Kai's comment: https://github.com/openhab/openhab/pull/4838#issuecomment-263547750

bundles/binding/org.openhab.binding.expire/pom.xml
micw: Done

bundles/binding/org.openhab.binding.expire/src/main/java/org/openhab/binding/expire/internal/ExpireBinding.java
micw: Right.

bundles/binding/org.openhab.binding.expire/.classpath
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
```
watou: Since the binding uses the `java.time` package, it needs `JavaSE-1.8`.

bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,7 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
```
watou: 1.8

bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,7 @@
...
+org.eclipse.jdt.core.compiler.compliance=1.7
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.source=1.7
```
watou: 1.8

bundles/binding/org.openhab.binding.expire/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,30 @@
...
+ org.openhab.core.types,
+ org.openhab.model.item.binding,
+ org.osgi.framework,
+ org.osgi.service.cm,
+ org.osgi.service.component,
+ org.osgi.service.event,
+ org.slf4j
+Export-Package: org.openhab.binding.expire
+Bundle-DocURL: http://www.openhab.org
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
```
watou: JavaSE-1.8

watou:
Sorry I missed the Java 1.8 dependency earlier.  Also, my Eclipse IDE complains about the missing `src/main/resources/readme.txt` like other bindings have.

micw:
I missed it as well. I added it in the files you mentioned plus the pom.xml

watou:
Last issue and then I think (reall, this time) that this PR can be merged:

res

micw:
;-)

micw:
How can I build+deploy my binding into my openhab2 instance?

watou:
> How can I build+deploy my binding into my openhab2 instance?

For testing, copy the jar file to your addons folder.  It should then appear as Active in bundle:list.  Then edit .items files to include the new binding.

To make the binding installable from a future OH2 build, make a new PR (after this one is merged) that is similar to [this one](https://github.com/openhab/openhab/pull/4384) and [this one](https://github.com/openhab/openhab/pull/4789).

watou:
FYI, I'm working on a new PR to remove the Java 1.8 dependency and a few other things...

micw:
I wonder why you want to do this. Java 1.7 is EOL since more than one year.
If you have to, you can use joda time that has a similar duration parser.

watou:
> I wonder why you want to do this. Java 1.7 is EOL since more than one year.

Because openHAB 1.8 must still support Java 1.7 environments, even though most users seem to be on Java 1.8.  So I broke the build when I merged your PR, since the build enforces this restriction, and have temporarily taken it out of the build.

I have a working PR that I will submit once I've tested it more thoroughly.

watou:
@micw: Please review #4845 to make sure I didn't break anything.

watou:
For convenience, [here](https://dl.dropboxusercontent.com/u/4286376/expire-java-7/org.openhab.binding.expire-1.9.0-SNAPSHOT.jar) is a JAR to test that matches #4845.

Issue #4559, opened 2016-07-27
[RRD4j] Persistence service returning wrong time.


CarstenMadsen:
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used: (e.g., openhab and addon versions)
* Environment name and version (e.g. Chrome 39, node.js 5.4, Java 8, ...):
* Operating System and version (desktop or mobile, Windows 10, Raspbian Jessie, ...):

CarstenMadsen:
When I use the rest api to get data from the RRD4j database, the second time value is wrong.

se [https://community.openhab.org/t/rrd4j-rest-persistence/12658]

This only happens when I set a start and endtime.

J-N-K:
https://github.com/openhab/openhab/blob/master/bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java#L238

Checking whether ts<start and not adding the value if this is the case should fix that. I can prepare a PR if needed.

J-N-K:
Is this still happening at your site @CarstenMadsen? I just tried and could not reproduce it. Perhaps you can send me a .rrd so I can check.

J-N-K:
Additional info: to reproduce on OH2: set &boundary=true. This is not happening with &boundary=false or without boundary at all

J-N-K:
Should be fixed now. Please check so this can be closed.

PR #4824, opened 2016-11-26
[RRD4J] Fix rrd interval #4559

J-N-K:
Again, there is a lot of reformatting. The only change besides that is in l. 238.

This fixes the data points with too early (#4559) or too late timestamps. However, it can still happen that the last datapoint/timestamp is duplicated. This outside the scope of the RRD4J binding but must be fixed in the REST implementation.

There are a lot of warning (e.g. l. 169) about wrong type casts in the logger. Can someone with deeper JAVA knowledge advice on that?

### DIFF
--- a/bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java
+++ b/bundles/persistence/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jService.java
@@ -235,7 +235,7 @@ public class RRD4jService implements QueryablePersistenceService {
 		long ts = result.getFirstTimestamp();
 		long step = result.getRowCount() > 1 ? result.getStep() : 0;
 		for (double value : result.getValues(DATASOURCE_STATE)) {
-		    if (!Double.isNaN(value)) {
+		    if (!Double.isNaN(value) && (ts>=start) && (ts<=end)) {
 			RRD4jItem rrd4jItem = new RRD4jItem(itemName, mapToState(value, itemName), new Date(ts * 1000));
 			items.add(rrd4jItem);
 		    }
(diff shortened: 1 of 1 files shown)


J-N-K:
I opened an [issue](https://github.com/eclipse/smarthome/issues/2538) for the double end-time.

J-N-K:
is there any reason that blocks merging?

9037568:
LGTM.  Has there been any verification from @CarstenMadsen or the community?

watou:
@J-N-K, I see the only non-formatting change in this branch currently is:

```diff
-		    if (!Double.isNaN(value)) {
+                    if (!Double.isNaN(value) && (((ts >= start) && (ts <= end)) || (start == end))) {
```

Please confirm that this change is still wanted and will not cause a regression.

> There are a lot of warning (e.g. l. 169) about wrong type casts in the logger. Can someone with deeper JAVA knowledge advice on that?

Line 169 is now:

```java
logger.warn("Could not persist '{}' to rrd4j database: {}", new String[] { name, e.getMessage() });
```

I recommend removing all the `new String[] { arg0, arg1 }` and replacing them with `arg0, arg1` since the logger can handle variable arguments, and we don't need to keep creating and garbage collecting the String or Object arrays.

J-N-K:
@watou, The change is what is intended. RRD4J returns one datapoint before and (id exists) after the requested time interval. This is obviously confusing and leads to wrong ordering of the datapoints in the REST API (when called with boundaries).

The loggers have been adjusted. I was just wondering if there is a reason the `new String[]` is used.

watou:
> I was just wondering if there is a reason the new String[] is used.

It might be that the logger API accepted an array after the format string, and then later added variable argument support.  The ideal usage now is to avoid temporary object construction for logging statements, so pass variable arguments instead of creating a new array.  Also, pass `someObject` instead of `someObject.toString()` as an argument, as `.toString()` would only be needed if the logger needed to actually format a log entry and `.toString()` might require work and/or temporary object creation.

J-N-K:
I changed that.

watou:
OK to be merged then, @J-N-K ?

J-N-K:
Yes. Works for me.

PR #4827, opened 2016-11-26
Attempt to fix #4820 recurring events not scheduled

### LINKED ISSUE #4820: caldav not scheduling some recurring events
in my quest for scheduling my heaters with caldav, I encountered a strange behaviour : some recurring (once in a week) events would work on first launch, and not work on the following schedule, while others (recurring more often, every day except week ends) would work "longer" ... until they dont trigger for unknown reason

## Expected Behavior
all recurring events should obviously happen within openhab2

## Current Behavior
some recurring events will simply be ignored

## Possible Solution
restarting openhab will make the events work automagically ... until their next recurring event

- set-up an event happening every monday at 8:00, ending at 9:00 triggering some visible action in your items
- set-up caldav io like this :
caldavio:chauffagecmd:reloadInterval=1
caldavio:chauffagecmd:preloadTime=1440
caldavio:chauffagecmd:disableCertificateVerification=False
caldavio:chauffagecmd:lastModifiedFileTimeStampValid=False
- set your system time to 7:55 (monday)
- start openhab : the event is found and scheduled (and works)
- when the event has triggered, set your system time 1 week in the future, at 7:55 ...

- wait ... and see nothing happen (event is not seen)

## Context
my recurring heating events will not fire, so I'm cold, this is NOT FUN :/

## Your Environment

* Version used:  openhab2 snapshot 
* Environment name and version Java 8
* Operating System and version Raspbian Jessie on RPI3

aschor:
Attempt to fix #4820 (from my tests, it's much better)

basically, the caldavReloader is not updating the scheduler and the eventlist when an event has not been modified on the caldav server (and contains the flag "last-modified") (in case the eventcontainer exists, and filetimestamp != valid, then a "addevents" is called but this did not actually UPDATE an already existing eventcontainer ... so not rescheduling new event recurrences)

Added a lot of trace and debug lines to be able to understand what was going on (might prove usefull for future use)

squashed commits : (not an easy task !) 
revert .classpath modifications introduced (javadoc was useful ! )
cosmetics + small perf improvement
remove lambda expression (jenkins was alergic to it)

sorry for the commit date, I can't figure out how to change that (committed while my system was 1 week in the future, to see if it was working ...)

### DIFF
--- a/bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/CalDavEvent.java
+++ b/bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/CalDavEvent.java
@@ -146,31 +146,63 @@ public class CalDavEvent {
     public int hashCode() {
         final int prime = 31;
         int result = 1;
-        result = prime * result
-                + ((calendarId == null) ? 0 : calendarId.hashCode());
+        result = prime * result + ((calendarId == null) ? 0 : calendarId.hashCode());
         result = prime * result + ((id == null) ? 0 : id.hashCode());
         return result;
     }
 
     @Override
     public boolean equals(Object obj) {
-        if (this == obj)
+        if (this == obj) {
             return true;
-        if (obj == null)
+        }
+        if (obj == null) {
             return false;
-        if (getClass() != obj.getClass())
+        }
+        if (getClass() != obj.getClass()) {
             return false;
+        }
         CalDavEvent other = (CalDavEvent) obj;
         if (calendarId == null) {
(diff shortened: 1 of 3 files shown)


aschor:
well, after all, it didn't fix the issue... this morning my heaters didn't switch on. Instead I had stack trace (at least I had this, with my modification) : 
org.quartz.JobExecutionException : cannot get recurence-event for config xxxx and eventid xxx and occurence; 1

=> I'll have to investigate further, it seems that I still don"t really get how the underlying mechanics work for the caldav.io, I might have not done something right in my attempt at fixing the issue :/

PR #4831, opened 2016-11-27
[KNX] Add log entry if DPT incompatible

### LINKED ISSUE #4229: KNX Events are not send to bus when DPT is set for item
When adding a DPT type to knx items like the one below causes the value not being send to the knx bus

Example item:
`Number iHeatingWestKidsRoomThermostatState "WestKidsRoom - Thermostat state [%s]" <heating> (gHeatingControlThermostatStates) { knx="5.001:1/7/27" }`

Example rule:
rule "iHeatingWestKidsRoom" when 
    Item iHeatingWestKidsRoomState received update
then 
    iHeatingWestKidsRoomThermostatState.sendCommand(new PercentType(10))
    logInfo("test", ""+iHeatingWestKidsRoomThermostatState)
end
##

J-N-K:
fixes #4229: DecimalType and DPT 5.001 are incompatible, expected is PercentType. Misuse was not logged but failed silently.


9037568:
Thanks, @J-N-K !

Issue #4837, opened 2016-11-29
Athmosphere transport notifications not working with OH 1.8.3


micw:
I run openhab 1.8.3 from debian repo. For some reason I had problems updating item states in the android app. I tracked it down to athmosphere transport notifications which seems not to work at all.

### Items
Switch test

### Sitemap
sitemap test label="Test" {
	Frame label="Test" {
		Switch item=test
	}
}

### Long-Polling:
curl -i http://127.0.0.1:8080/rest/sitemaps/default/test -H "X-Atmosphere-Framework: 1.0" -H "X-Atmosphere-Transport: long-polling" -H "X-Atmosphere-tracking-id: test1"

### Test:
Switch the switch on/off via UI

## Expected Behavior
Curl should receive a changed sitemap as soon as I change the switch status

## Current Behavior
Curl keeps waiting for response

## Steps to Reproduce (for bugs)
See above

## Context
The bug makes the android application unresponsive (shows item changes with large delays)

## Your Environment
Debian 8, OpenJDK 8 headless

9037568:
Can you provide a full debug log, @micw ?

PR #4838, opened 2016-11-29
Changed the order of accepted states in StringItem

micw:
When TypeParser is used with the list of accepted states of StringItem, it will process the items in accepted order. Since StringType accepts any String, the other types (DateTimeType, UnDefType) would never be processed. The String "Undefined" would be convertet in StringType("Undefined") rather than UnDefType.UNDEF.

### DIFF
--- a/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/StringItem.java
+++ b/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/items/StringItem.java
@@ -33,9 +33,9 @@ public class StringItem extends GenericItem {
 	private static List<Class<? extends Command>> acceptedCommandTypes = new ArrayList<Class<? extends Command>>();
 
 	static {
-		acceptedDataTypes.add(StringType.class);
-		acceptedDataTypes.add((DateTimeType.class));
 		acceptedDataTypes.add(UnDefType.class);
+		acceptedDataTypes.add(DateTimeType.class);
+		acceptedDataTypes.add(StringType.class);
 
 		acceptedCommandTypes.add(StringType.class);
 	}


watou:
I don't think this pull request can be merged, unfortunately, because 1) it's against the 1.8 branch, and there currently is no plan to release a 1.8.4 distro, and 2) the core components have been removed from the master branch since it's discontinued.  Essentially, the openHAB 1.x core is frozen as development, testing and documentation resources are all focused on OH2/ESH (and the 1.x addons).

Correcting the accepted data types order for `StringItem` in *Eclipse SmartHome*, so `DateTimeType` appears before `StringType`, might still be a worthwhile PR!

I think that for the expire binding, leaving `UnDefType.UNDEF` as the default that isn't specified, will still achieve expiring a String item with Undefined/UNDEF, and as the default expire state, it doesn't have to be specified and thus doesn't risk producing a `StringType` with the string value `Undefined`.

micw:
Will we have a 1.9.x release?

watou:
> Will we have a 1.9.x release?

I don't think so!  I would kind of like to see a 1.9.0 full distro, where the core is identical to 1.8.3 but all the addons are updated.  However, I acknowledge that this is a lot of work to get right and I don't think almost anyone would have the time, experience and login credentials to get it done.  A step back from that would be to have Debian packages for all the updated 1.x addons once they reach release state, so existing Debian installs could be easily upgraded, but again, huge work, minimal value, and no one to do it.  So....

kaikreuzer:
> Correcting the accepted data types order for StringItem in Eclipse SmartHome, so DateTimeType appears before StringType, might still be a worthwhile PR!

👍 

> Will we have a 1.9.x release?

Yes, but only of what is within this repo now: openHAB 1 addons.

watou:
> Yes, but only of what is within this repo now: openHAB 1 addons.

^^^clearer answer than mine! :slightly_smiling_face:

kaikreuzer:
Not necessarily. It does not answer whether there will be debian packages for those add-ons :-)

So far, I have published 1.9.0 pre-builds only as Maven artifacts: https://bintray.com/openhab/mvn/openhab/1.9.0.b4#files/org/openhab

Not sure if the build produced debian packages as well that could be easily published to the apt-repo? Probably you or @theoweiss are the better ones to answer that!

micw:
So there's no chance to make a progress here until 2.0 is ready? Because I'd like to get other things fixed as well, e.g. #4837 
It does not seem to be an option to switch to 2.0 at the moment (e.g. there are complex rule-based _workarounds_ for switching on/of a light by astro events - one of the most basic things a home automation thing should support).

micw:
BTW, where should developers discuss such things? Doesn't seem to be the right place here but in irc channel I did not meet a single developer...

watou:
> It does not seem to be an option to switch to 2.0 at the moment (e.g. there are complex rule-based workarounds for switching on/of a light by astro events - one of the most basic things a home automation thing should support).

The Astro 2.0 binding still lacks capabilities of the 1.x binding IIRC, but you can install the astro1 binding under OH2 in the meantime.  It's true that OH2 is still in beta, but it's come a long way and the next release should be very capable.

Regarding #4837, I think the best thing to do regarding Atmosphere is to run the other direction, which means another reason to try to migrate your config to OH2.

watou:
> BTW, where should developers discuss such things? Doesn't seem to be the right place here but in irc channel I did not meet a single developer...

I've never logged into the irc channel myself, but I think https://community.openhab.org has proven to be a great way to communicate, and I'm all for having more developer conversations there.

PR #4845, opened 2016-11-30
[expire] Remove Java 1.8 dependency; add command support, minor refactoring

watou:
small bug fixes, re-added to parent pom.xml

### DIFF
--- a/bundles/binding/org.openhab.binding.expire/.classpath
+++ b/bundles/binding/org.openhab.binding.expire/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
 	<classpathentry kind="output" path="target/classes"/>
--- a/bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
+++ b/bundles/binding/org.openhab.binding.expire/.settings/org.eclipse.jdt.core.prefs
@@ -1,7 +1,7 @@
 eclipse.preferences.version=1
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
-org.eclipse.jdt.core.compiler.compliance=1.8
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
+org.eclipse.jdt.core.compiler.compliance=1.7
 org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
 org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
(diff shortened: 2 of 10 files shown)


watou:
For convenience, [here](https://dl.dropboxusercontent.com/u/4286376/expire-java-7/org.openhab.binding.expire-1.9.0-SNAPSHOT.jar) is a JAR to test that matches this PR.

micw:
@watou, unfortunately I won't find the time to review before mid of next week. At 1st glance:
Watching commands as well as state changes for resettings the expire trigger make sense to me.
Using a command to expire the item looks also like a good idea but here things might become complicated:
If you trigger a command, the item will probably change it's state. But you don't know which state the command will cause. So the state change will reset the expire timer. This might lead to the follwing situation:
1. expires
2. the command is triggered
3. the state is updated, it re-sets the expire time
4. (=1.) the item expires again

I'm not sure if this happens (I don't know enough about openhab's internal handling). It depends mainly on the behaviour of 3. - will internalReceiveUpdate be triggered if the command did not change the state? If yes, this can build a loop with lots of unnecessary status updates. If no, the loop would stop after 2nd iteration.

A better solution would it be if we'd know the initiator of a command and the initiator of a state change.

watou:
Thank you, @micw for looking into my changes!

> will internalReceiveUpdate be triggered if the command did not change the state?

In my observation an `internalReceiveCommand` is not immediately followed by an `internalReceiveUpdate` which I think is odd because if the item is updated, the binding ought to be informed.  So posting either the command or state `OFF` to an item bound to either `expire="1s,OFF"` or `expire="1s,command=OFF"` ought to (with the current state of this PR) cancel any expiration trigger.  Or posting any other command or state will set up the trigger.

The current state of this PR is to take all incoming `State` or `Command` (via `internalReceiveUpdate or internalReceiveCommand`) and compare it to whichever `expireCommand` or `expireState` is set.  If they are equal, future expirations are cancelled.  If they aren't, the expiration is either set up or moved further into the future.  So one way or another, the item will in the future either receive an update or command that cancels the trigger, or continue to receive non-matching states or commands that move the expiration further into the future.

If you get another chance to read through this PR to critique the logic, it would be much appreciated.  I would like to merge this soon if possible.

micw:
@watou, is there an easy way to create some kind of unit/integration test that starts a minimal openhab core to test that behaviour? I'd prefer this a lot over manual reviewing/testing.

Can we simply turn the project into a maven project to enable test scope? I have seen that if there are tests, there are expra projects for these - this might be the reason that we have so few tests here ;)

watou:
> @watou, is there an easy way to create some kind of unit/integration test that starts a minimal openhab core to test that behaviour? I'd prefer this a lot over manual reviewing/testing.

That would be heaven!  I'm not aware of such a thing.  This binding would be ideal for such a test stand because it doesn't directly rely on external devices or bridges to operate.  It's impractical for many bindings, but not this one.

> Can we simply turn the project into a maven project to enable test scope? I have seen that if there are tests, there are expra projects for these - this might be the reason that we have so few tests here ;)

Any ideas you have to improve automated testing should probably go to https://community.openhab.org/c/organisation/code or https://www.eclipse.org/forums/index.php/f/271/ .  To have the build more thoroughly verify proper function would have definitely stopped many regressions.

kaikreuzer:
> That would be heaven! I'm not aware of such a thing. 

https://www.eclipse.org/smarthome/documentation/development/testing.html

See examples for how it is used:
https://github.com/eclipse/smarthome/pull/2436/files
https://github.com/eclipse/smarthome/pull/2247/files
https://github.com/openhab/openhab2-addons/pull/1434/files

Note: All only available on ESH/OH2 build.

watou:
> Note: All only available on ESH/OH2 build.

OK, thanks @kaikreuzer for the links!  Can tests make use of items, rules, custom configuration, things like that?

kaikreuzer:
I general yes, but you will need to bootstrap all the right bundles and feed them with the configuration you need. The framework is meant for unit test running on an OSGi framework, not really for integration tests on a "final distro".

PR #4766, opened 2016-11-11
[caldav] add filter-category-any

aschor:
rebased on 2016/12/03 and squashed my commits into 1.

added a "filter-category-any" configuration that will allow a item
configuration to have filters that match any of an event filter. Allows
an event to command a GROUP, and have an "awareness" on all individual
items of the group that an event concerning them will happen

example : 
String cuisine_NextEventName1 "cuisine proch. evt. 1 [%s]" <calendar> {
caldavPersonal="calendar:chauffagecmd type:UPCOMING eventNr:1 value:NAME
filter-category-any:'cuisine,salon,rezdechaussee'"}
String cuisine_NextEventName2 "cuisine proch. evt. 2 [%s]" <calendar> {
caldavPersonal="calendar:chauffagecmd type:UPCOMING eventNr:2 value:NAME
filter-category-any:'cuisine,salon,rezdechaussee'"}
String cuisine_NextEventName3 "cuisine proch. evt. 3 [%s]" <calendar> {
caldavPersonal="calendar:chauffagecmd type:UPCOMING eventNr:3 value:NAME
filter-category-any:'cuisine,salon,rezdechaussee'"}

will match all 3 events below :
event1 with category "cuisine"
event2 with category "salon"
event3 with categories "salon,cuisine,rezdechaussee"

PR #4855, opened 2016-12-03
[caldav] fix #4820 recurring events not scheduled

### LINKED ISSUE #4820: caldav not scheduling some recurring events

aschor:
Fix #4820, tested live on my home heaters

rebased on upstream/master on 2016/12/03

The issue was that some known recuring events, under specific confitions would not have their schedule updated ; the reload code is calling the function "addEventToMap" for every new or modified event, but the event was already in the map, thus the call was ignored. Fixed that. At least this case does not occur at my home, all my heaters (zwave controlled) have worked perfectly this week (never before).

Added a lot of trace and debug lines to be able to understand what was
going on

revert .classpath modifications
cosmetics + small perf improvement
remove lambda expression
(jenkins alergic)

### DIFF
--- a/bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/job/EventReloaderJob.java
+++ b/bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/job/EventReloaderJob.java
@@ -16,6 +16,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
@@ -38,7 +39,10 @@ import org.quartz.Job;
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
 import org.quartz.JobKey;
+import org.quartz.Scheduler;
 import org.quartz.SchedulerException;
+import org.quartz.Trigger;
+import org.quartz.impl.matchers.GroupMatcher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


watou:
Hi @anschor, both this PR and #4766 report that they fix #4820.  Are both required or is one obsolete?

aschor:
hi, sorry, there was a typo on my copy paste ! this one is the only one fixing the bug, the other is a feature

bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/job/EventReloaderJob.java
```diff
@@ -209,7 +256,14 @@ public synchronized void loadEvents(final CalendarRuntime calendarRuntime, final
                     continue;
                 }
 
-                oldEventIds.remove(filename);
+                // oldEventIds.remove(filename); //BUGGY lazy "remove" ??(first occurence) :
```
9037568: Should remove commented code in this line.

9037568:
Aside from my one comment, this LGTM.

aschor:
hi @9037568  , comment removed. Thanks for the review

aschor:
looking at the doc again : it could be that I fixed this : 

- Known Problems (or limitations) :  
You have to set the preloadInterval to a higher or equal value as the recurring events in the calendar exists.

aschor:
one week later, I have not restarted openhab once. no missed event anymore. I'm happy with this

aschor:
thanks for merging !

PR #4862, opened 2016-12-04
[eBus] Convert linefeed from CRLF to LF only

csowada:
Adjust this binding to the openhab 2 dev guidelines


watou:
I checked out the pr/4862 and performed `git diff -b master` and it produced no output.  Thanks!

PR #4868, opened 2016-12-05
made all openHAB 1 features follow the id-ends-with-1 naming conventi…

kaikreuzer:
…on (with the exception of actions)

### DIFF
--- a/features/openhab-addons-legacy/src/main/feature/feature.xml
+++ b/features/openhab-addons-legacy/src/main/feature/feature.xml
@@ -43,7 +43,7 @@
     <configfile finalname="${openhab.conf}/services/netatmo.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/netatmo</configfile>
   </feature>
 
-  <feature name="openhab-binding-networkhealth" description="NetworkHealth Binding (1.x)" version="${project.version}">
+  <feature name="openhab-binding-networkhealth1" description="NetworkHealth Binding (1.x)" version="${project.version}">
     <feature>openhab-runtime-base</feature>
     <feature>openhab-runtime-compat1x</feature>
     <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.networkhealth/${project.version}</bundle>
--- a/features/openhab-addons/src/main/feature/feature.xml
+++ b/features/openhab-addons/src/main/feature/feature.xml
@@ -111,7 +111,7 @@
 
   <!-- binding -->
 
-  <feature name="openhab-binding-alarmdecoder" description="Alarm Decoder Binding" version="${project.version}">
+  <feature name="openhab-binding-alarmdecoder1" description="Alarm Decoder Binding" version="${project.version}">
     <feature>openhab-runtime-base</feature>
     <feature>openhab-runtime-compat1x</feature>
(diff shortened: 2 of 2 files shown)


kaikreuzer:
This change prevents naming clashes when OH2 versions of add-ons are introduced. So far this resulted in the id of the OH1 add-on being changed (e.g. from "exec" to "exec1" as done [here](https://github.com/openhab/openhab/pull/4816/files)), which means that it breaks the setup of users that expect the OH1 version to be installed.

This breaks it once now, but keeps it stable in future, so imho this step makes sense (fyi @ThomDietrich and @rkoshak).

ThomDietrich:
Interesting. That's definitely a simple but clear change.
Just to mention it once, `openhab1x-binding-networkhealth` (or similar) is not an option? I have no clue about namespace related details this might be in conflict with...

kaikreuzer:
> openhab1x-binding-networkhealth (or similar) is not an option?

This would incur many code changes as "openhab-" is used as a prefix in many places.
An option might be "openhab1x-binding1-networkhealth", so that it actually becomes a new add-on category - this might fit well to my suggestion in the last bullet point of https://github.com/openhab/openhab-docs/pull/148#issuecomment-264850049.

watou:
Would it make sense to use a format that continues the hierarchical nature of the feature names and hyphen separation, like `openhab-binding-1-enocean`?  Or is it time to invent new names that better describe the basic difference between the two types of bindings, like `openhab-binding-item-nest` and `openhab-binding-thing-astro`?  The latter idea respects the notion that 1.x bindings are likely to linger on for awhile, and the `1` in the name could lead to confusion about what versions of the runtime are compatible.  Just food for thought.

ThomDietrich:
I guess the whole discussion depends on what's about to come next and how the information will be used.
I remember that the old exec binding could still be used in addons.cfg by adding "exec1". How would the other schema work in that situation? Will we run into similar cases in the future, with addons using specific features of OH 2.5 for examples? Depending on these questions here are some variations, I prefer the idea behind 2 and 4. 

1. `openhab-binding-enocean1`
2. `openhab-binding-enocean_v1`
3. `openhab-binding-1-enocean`
4. `openhab-binding-v1.x-enocean`

rkoshak:
The Astro binding has set a precedent (1.9 is astro1 where the 2.0 is just astro). And I will add that I figured that out without looking it up (in fact at the time I wasn't clear where I would look anyway). I just guessed and was right.

I don't really have a strong opinion on this matter except that having some consistent pattern will help make choosing and discovering bindings clearer and easier to intuit.

I think when users can correctly guess what the right thing to use is so much the better. Though my anecdote is hardly evidence that appending a 1 is a better approach.

watou:
The least disruptive approach is this PR, LGTM.   Just an opportunity to think about bigger naming issues, and possibly avoid confusion with present or future bindings whose names already end in digits.

kaikreuzer:
> The least disruptive approach is this PR, LGTM.

I agree, that's how we had it so far (for everything that was already declared "legacy" and it didn't really cause any issues. So I would also keep it simple - the current PR does not require any changes in the feature installation infrastructure either, so I clearly prefer it :-)

ThomDietrich:
I'm also okay with the current state.

PR #4884, opened 2016-12-10
initial contribution of openhabcloud bundle (based on my.openHAB bundle)

kaikreuzer:
See https://github.com/openhab/openhab/wiki/openHAB-Cloud-Connector

Note that this bundle is not included in the build (like the my.openHAB bundle) as it does not compile against the target platform, since it requires OH 1.8 bundles. I will provide a binary jar on the wiki.


kaikreuzer:
Well, if nobody else wants to merge it, I do it myself ;-)

marziman:
I just saw it and you already merged it :-)

marziman:
But looks good. We should communicate that the docu on my.oh is deprecated

kaikreuzer:
Well, the docs on my.oh are still valid for the (old) my.oh bundle, which still exists and will be used until end of January.

PR #4887, opened 2016-12-11
[Z-Wave] Smart power plug (NAS-WR01Z) added to the database

druciak:
Power plug added: http://products.z-wavealliance.org/products/1783

### DIFF
--- a/bundles/binding/org.openhab.binding.zwave/database/products.xml
+++ b/bundles/binding/org.openhab.binding.zwave/database/products.xml
@@ -3769,4 +3769,17 @@
             <ConfigFile>logic/zhc5010.xml</ConfigFile>
         </Product>
     </Manufacturer> 
+    <Manufacturer>
+        <Id>0258</Id>
+        <Name>Shenzhen NEO Electronics</Name>
+        <Product>
+            <Reference>
+                <Type>0003</Type>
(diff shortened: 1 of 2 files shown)


cdjackson:
It doesn't look like this is in the online database - it would be good to add it if you have the XML produced by OH?

My only comment on the file is the labels of the associations look very long and probably should be split to a short label, and a longer ```<Help> ``` element.

druciak:
What database do you mean? I put a link in my comment to a db.
Link to product page: http://szneo.com/en/products/show.php?id=189
Attached manual: [Power-plug-Z-wave1.pdf](https://github.com/openhab/openhab/files/644799/Power-plug-Z-wave1.pdf)
Regarding associations labels you're right, will try to make them shorter.

cdjackson:
I meant this database -:
http://www.cd-jackson.com/index.php/zwave/zwave-device-database/ <http://www.cd-jackson.com/index.php/zwave/zwave-device-database/>

It’s used for OH1 and OH2 definitions.

druciak:
Never heard about this database, but looks nice. 
Anyway the device is already there: http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/397

I also made assoc. labels shorter, I hope now there're better.

PR #4888, opened 2016-12-11
[influxdb] Quote retention policy and measurement

gedejong:
Quotes are added around the retention policy and measurement
identifiers. The lexer of influxdb treats 'default' (which
is also the default retention policy) as a reserved word.

This will also allow querying items which have reserved
characters in their names.

Tested on openhab2. Fixes at least:
https://github.com/openhab/openhab/issues/4840


theoweiss:
Thanks @gedejong ! LGTM

gedejong:
@9037568 : did some Eclipse formatting. Would prefer the appends to be vertically aligned, but alas!

PR #4889, opened 2016-12-12
Replace string casting of config values

watou:
For every current binding that has an `ESH-INF/binding/binding.xml` file with configuration parameters, replace all constructs like
```java
String configValue = (String) config.get(“configKey”);
```
with
```java
String configValue = Objects.toString(config.get(“configKey"), null);
```
Because OH2 runtime, unlike OH1 runtime, the ConfigAdmin dictionary can have non-String config values.

### DIFF
--- a/bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseBinding.java
+++ b/bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseBinding.java
@@ -25,6 +25,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang.IllegalClassException;
+import org.apache.commons.lang.ObjectUtils;
 import org.joda.time.DateTime;
 import org.openhab.binding.plugwise.PlugwiseBindingProvider;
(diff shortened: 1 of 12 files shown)


bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseBinding.java
```diff
@@ -106,7 +107,7 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
 
     private Stick setupStick(Dictionary<String, ?> config) {
 
-        String port = (String) config.get("stick.port");
+        String port = ObjectUtils.toString(config.get("stick.port"), null);
```
kaikreuzer: Why ObjectUtils and not Objects anymore here in this class?

kaikreuzer: Ah, thanks for the explanation! While I also do not see a reason to keep it on Java 1.6, I agree that your change is the least invasive - so let's go for it!

kaikreuzer:
Oops, sorry, just saw now that you asked for a review by @9037568, while I have it already merged - sorry ;-)

PR #4892, opened 2016-12-12
Add panasonictv binding to compatible 1.x addons

dominicdesu:
Signed-off-by: Dominic Lerbs <dominicdesu@hallojapan.de>

### DIFF
--- a/features/openhab-addons-external/src/main/resources/conf/panasonictv.cfg
+++ b/features/openhab-addons-external/src/main/resources/conf/panasonictv.cfg
@@ -1,0 +1,4 @@
+################################ Panasonic TV Binding #######################################
+#
+# IP address of a Panasonic TV instance
+# <id>=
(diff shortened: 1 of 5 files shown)


features/openhab-addons-external/src/main/resources/conf/panasonictv.cfg
```diff
@@ -0,0 +1,4 @@
+################################ Panasonic TV Binding #######################################
+#
+# IP address of a Panasonic TV instance
+# <id>=
```
watou: A commented out example would probably be very helpful here.  Also, the `##### Panasonic TV Binding #####` heading isn't needed.

dominicdesu:
@watou valid point, I have updated the config. Is it fine like this or should I rebase into a single commit?

watou:
All good.  Thanks @dominicdesu !

PR #4896, opened 2016-12-14
Weather: Fixed exception during scheduler restart

gerrieg:
Signed-off-by: Gerhard Riegler <gerhard.riegler@gmail.com>


9037568:
Thanks, @gerrieg !

PR #4898, opened 2016-12-15
Add support for IT message to support the Flamingo FA500 and others

martinvw:
Add support for IT message to support the Flamingo FA500 and others

Running it in my production env since the 24th of november, this month it will be switching the christmas lights ;-)

Fixes https://github.com/openhab/openhab2-addons/issues/1019

### DIFF
--- a/bundles/binding/org.openhab.binding.rfxcom.test/src/test/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5MessageTest.java
+++ b/bundles/binding/org.openhab.binding.rfxcom.test/src/test/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5MessageTest.java
@@ -9,8 +9,16 @@
 package org.openhab.binding.rfxcom.internal.messages;
 
 import org.junit.Test;
+import org.openhab.binding.rfxcom.RFXComValueSelector;
 import org.openhab.binding.rfxcom.internal.RFXComException;
-import org.openhab.binding.rfxcom.internal.messages.RFXComBaseMessage.PacketType;
+import org.openhab.core.library.types.OnOffType;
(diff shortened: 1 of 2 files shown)


martinvw:
@Jamstah would you like to review this small change :-)

Jamstah:
Of course:
You still have a todo at the bottom of the tests to add real tests, and it looks like you added one. I'd say remove the todo, or at least change it to say "add _more_ real tests" because these messages look complicated enough to deserve a few more.

The commands having a supported subtype addition is a new concept, could be worth explaining in a comment, ie, we do this because multiple subtypes use the same command ids.

Did you run it through the source formatter? I'm surprised to see * imports, I thought they were discouraged (and generally, I don't like them anyway).

Otherwise, LGTM. Changes are substantial enough that you should probably take an author credit too.

Jamstah:
Is there (or will there be) a openhab2 equivalent?

martinvw:
Yes there will be an OH2 equivalent but I prefer to introduce after my other pull-request is merged otherwise adding the test is not possible / trivial.

I might have auto-formatted with Intellij and Eclipse does not have an opinion about it, now all should be fine again.

PS: i believe squashing of commits is not needed any more, they can do it via GitHub now.

martinvw:
Note to self: I should add some example to the wiki at https://github.com/openhab/openhab/wiki/RFXCOM-Binding, which could also make it searchable by google.

martinvw:
Done: wiki is also updated!

If https://github.com/openhab/openhab2-addons/pull/1247 is merged I will add this code also to OH2

Jamstah:
Great, LGTM.

9037568:
Thanks, @martinvw !

PR #4900, opened 2016-12-15
Update for Plex binding

idserda:
This fixes two issues in the current Plex binding:

- Latest Plex Media Server (>= 1.3.2.3112) sends a different message format over the websocket connection. The binding now supports both this and the earlier format.
- No websocket connections were possible when authentication was enabled in PMS. The token used for other http calls is now also sent as a query parameter when connecting to the websocket.


watou:
Thanks @idserda, please report here when the PR passes the testing cycle you started [here](https://community.openhab.org/t/openhab2-plex/14036/42?u=watou).  Code change looks good (but I haven't tested it myself).

watou:
Give the go-ahead when you think this change has gotten sufficient field testing, and thanks for your PR!

idserda:
All testing was successful, this PR can be merged.

watou:
Thank you very much, @idserda!

Issue #4038, opened 2016-02-13
Netatmo binding in OH2: first values only set after 5 minutes


lolodomo:
When running the netatmo binding in OH2, you have to wait for 5 minutes to get the first values.
In OH1, you get values directly at startup.
##

watou:
I believe I've seen something similar with the Ecobee binding under OH2.

lolodomo:
5 minutes is the default delay between each refresh. That means that there is something wrong at the first call and then that's OK.

lolodomo:
I enabled the DEBUG logs and looked at the code.
The problem is that the binding is trying to get an access token while the items are not yet loaded. That should not be a problem except that the code is doing nothing if there is not at least an item linked to a weather station or a camera (NetatmoBinding - execute).
Here is the code from method refreshAccessToken:
```
        if (this.getWeather() == false && this.getCamera() == false) {
            logger.debug("There are not any items configured for this binding!");
            return;
        }
```
And this is the log entry I can find in the logs (There are not any items configured for this binding!).
I have no idea how it could be fixed. The problem is bery noticeable because the refresh cycle is 5 minutes by default for this binding. But it is a general problem for all bindings when the bindings are started before the items are loaded.

teichsta:
@lolodomo the binding startup order "problem" is known for a long time and won't be solved in the 1.x core anymore. I would propose the decrease the refresh cycle in order to mitigate the problem in 1.x. Anyhow thanks for sharing your thoughts and letting us know about your problems.

lolodomo:
Just to be clear, the problem is when using OH2

kaikreuzer:
If I understand it right, what @lolodomo is asking for is to modify the binding in a way that it correctly reacts on the OH2 startup order as well, so that it does not miss the first update cycle.

teichsta:
aaaah, didn't get that, sorry! So the question would be about removing the ManagedService implementation and use the latest activate() implementation through DS?

robnielsen:
According to the second comment, above,  this isn't specific to the Netatmo binding.

kaikreuzer:
It only means that the problem is not only in the Netatmo binding. Nonetheless, the problem is most likely within the code of the Netatmo binding and needs to be fixed there.
I assume that the problem is related to things like https://community.openhab.org/t/solved-oh2-bindingchanged-allbindingschanged-not-called-after-items-loading/7126. 
[...]

PR #4906, opened 2016-12-16
Netatmo: fix for issue #4038

robnielsen:
This fixes an issue #4038 where it takes 5 minutes for Netatmo items to get the initial values. Implemented solution similar to what @watou implemented for Ecobee

### DIFF
--- a/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
+++ b/bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
@@ -15,6 +15,7 @@ import java.util.Dictionary;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.openhab.binding.netatmo.NetatmoBindingProvider;
 import org.openhab.binding.netatmo.internal.authentication.OAuthCredentials;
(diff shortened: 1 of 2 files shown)


9037568:
LGTM.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
```diff
@@ -85,41 +98,52 @@ protected long getRefreshInterval() {
...
+                // Start the execution of the weather station and/or camera binding
+                if (bWeather) {
+                    weatherBinding.execute(oauthCredentials, this.providers, this.eventPublisher);
+                }
+                if (bCamera) {
+                    cameraBinding.execute(oauthCredentials, this.providers, this.eventPublisher);
+                }
+
+                // schedule the next poll at the standard refresh interval
+                schedulePoll(this.refreshInterval);
```
watou: Just to confirm that no matter what, line 146 is executed (as opposed to being skipped by an exception being thrown earlier, etc.

Another possibly nice thing to add would be to do something like

```
   /**
     * {@inheritDoc}
     */
    @Override
    public void allBindingsChanged(BindingProvider provider) {
        if (provider instanceof NetatmoBindingProvider) {
            schedulePoll(0);
        }
        super.allBindingsChanged(provider);
    }
```
so just editing your .items file will cause a the poll to happen sooner than the normal refresh cycle.

watou:
Aside from a suggestion to @robnielsen, also LGTM.

bundles/binding/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/NetatmoBinding.java
robnielsen: @watou, yes that is correct, line 146 is always executed, unless an exception is thrown. It needed to be somewhere after line 125, and seemed clearer to put at the end of the loop.

I've also added allBindingsChanged() per your suggestion, and squashed the commits.

watou: So just to be clear, it's OK if an exception is thrown that skips `schedulePoll(this.refreshInterval);`, so it will try again every five seconds (by default)?  As long as this works the way you think it should, LGTM.

robnielsen: @watou, Yes it's OK the way it currently is.

watou:
Thanks Rob!

PR #4908, opened 2016-12-17
Feature for Plex binding

idserda:
This PR adds a feature for the Plex binding and makes it configurable in the Paper UI.

### DIFF
--- a/bundles/binding/org.openhab.binding.plex/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.plex/ESH-INF/binding/binding.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="plex"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>Plex Binding</name>
+    <description>The Plex Binding connects to a Plex Media Server to track the state of multiple clients.</description>
+    <author>Jeroen Idserda</author>
+
+    <service-id>org.openhab.plex</service-id>
+
+    <config-description>
+        <parameter name="host" type="text" required="true">
+            <context>network-address</context>
+            <label>Host</label>
+            <description>IP address of the Plex server</description>
+        </parameter>
+        <parameter name="port" type="text">
+            <label>Port</label>
+            <description>Port that the Plex server is running on</description>
+            <default>32400</default>
+        </parameter>
+        <parameter name="refresh" type="integer">
+            <label>Refresh</label>
+            <description>Refresh interval in ms, only used for polling online/offline status of clients.</description>
+            <default>5000</default>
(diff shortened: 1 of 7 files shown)


bundles/binding/org.openhab.binding.plex/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,44 @@
...
+    <description>The Plex Binding connects to a Plex Media Server to track the state of multiple clients.</description>
+    <author>Jeroen Idserda</author>
+
+    <service-id>org.openhab.plex</service-id>
+
+    <config-description>
+        <parameter name="host" type="text" required="true">
+            <context>network-address</context>
+            <label>Host</label>
+            <description>IP address of the Plex server</description>
```
watou: Could this be changed to support hostnames like your recent change?  (And that would mean the `<context>` element would no longer apply.)

bundles/binding/org.openhab.binding.plex/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,44 @@
...
+        </parameter>
+        <parameter name="port" type="text">
```
watou: `type="integer"`

bundles/binding/org.openhab.binding.plex/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,44 @@
...
+        </parameter>
+        <parameter name="refresh" type="integer">
+            <label>Refresh</label>
+            <description>Refresh interval in ms, only used for polling online/offline status of clients.</description>
+            <default>5000</default>
+            <unitLabel>ms</unitLabel>
+        </parameter>
+        <parameter name="username" type="text">
+            <label>Username</label>
+            <description>If you're using Plex Home you need to supply the username and password of your Plex account here. If you don't want to enter your credentials you can also directly set your account token below instead.</description>
```
watou: Possible to shorten the description a bit?  I think it won't format well in the UI.

features/openhab-addons-external/src/main/resources/conf/plex.cfg
```diff
@@ -0,0 +1,19 @@
+# IP address of the Plex server
```
watou: or hostname?

watou:
With the support to configure the binding from the Paper UI, we enter a mode where some configuration elements could be other basic types like Integer and Boolean.  Therefore, all the constructs [here](https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.plex/src/main/java/org/openhab/binding/plex/internal/PlexBinding.java#L167) like 

```
(String) configuration.get("X");
```
should be changed to
```
Objects.toString(configuration.get("X"), null);
```
In this way, the configuration dictionary can contain values of any type and the binding remains unconcerned.

idserda:
Thanks for you feedback! Made the requested changes.

bundles/binding/org.openhab.binding.plex/ESH-INF/binding/binding.xml
watou: Better to have more description than not, so that's good!

watou:
Thank you, @idserda!  If [my comment](https://community.openhab.org/t/openhab2-plex/14036/62?u=watou) equates to nothing of concern, then this PR LGTM.

PR #4909, opened 2016-12-17
[zwave] Decrease logging for "does not support SECURITY_REPORT"

watou:
I have many lines in my log like

```
2016-12-17 14:20:52.586 [INFO ] [z.i.p.i.ZWaveNodeStageAdvancer] - NODE 2: does not support SECURITY_REPORT, proceeding to next stage.
```
It seems like this could be at DEBUG log level.


cdjackson:
Agreed.

watou:
Thanks!

Issue #3922, opened 2016-01-30
insteonplm binding does not work with openhab2


jtmoderate876:
I've got OH2b1 build #106 (2016.01.27) working with ecobee and a few things but...

I get the following error for the insteonplm binding after a couple minutes which worries me that the new, more rigorous requirements of eclipse or whatever means it needs to be patched/fixed.

```
Exception in thread "Thread-36" java.lang.NoSuchMethodError: org.apache.http.impl.client.DefaultHttpClient.execute(Lorg/apache/http/client/methods/HttpUriRequest;)Lorg/apache/http/client/methods/CloseableHttpResponse;
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.getURL(HubIOStream.java:213)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.bufferStatus(HubIOStream.java:108)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.poll(HubIOStream.java:156)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.run(HubIOStream.java:226)
    at java.lang.Thread.run(Thread.java:744)
Exception in thread "/hub2/Mirabell:BAxfhx6l@192.168.1.120:25105,poll_time=1000 Writer" java.lang.NoSuchMethodError: org.apache.http.impl.client.DefaultHttpClient.execute(Lorg/apache/http/client/methods/HttpUriRequest;)Lorg/apache/http/client/methods/CloseableHttpResponse;
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.getURL(HubIOStream.java:213)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.bufferStatus(HubIOStream.java:108)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.poll(HubIOStream.java:156)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream.write(HubIOStream.java:139)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream$HubOutputStream.flushBuffer(HubIOStream.java:308)
    at org.openhab.binding.insteonplm.internal.driver.hub.HubIOStream$HubOutputStream.write(HubIOStream.java:302)
    at java.io.OutputStream.write(OutputStream.java:75)
    at org.openhab.binding.insteonplm.internal.driver.IOStream.write(IOStream.java:68)
[...]

Issue #4630, opened 2016-09-08
[insteonplm] Does not connect with Insteon Hub 2245-222


larizzo:
Using binding org.openhab.binding.insteonplm-1.9.0-SNAPSHOT.jar the binding will not connect to my Insteon Hub 2245-222. It is not able to download the modem database or communicate with the hub.

 It does seem to work for the USB PLM as mentioned in this thread [4573](https://github.com/openhab/openhab/pull/4573#issuecomment-238106983) but there is at least one other [person](https://github.com/openhab/openhab/pull/4573#issuecomment-239878895) reporting the same issue as [myself](https://github.com/openhab/openhab/pull/4573#issuecomment-238573507).

This issue is happening in openhab 1.8.3 and OH 2. I've only uploaded the logs for openhab 1.8.3 but I could upload for 2 if needed.

When I drop back to the 1.8.3 binding I have no issue, I'm not exactly sure which PR caused the issue but it wasn't the latest one as I tried the 1.9.0 snapshot from before the PR and have the same issue.

Related converstation: [https://github.com/openhab/openhab/pull/4573](https://github.com/openhab/openhab/pull/4573)

Logs:
[insteonplm-2016-09-08.txt](https://github.com/openhab/openhab/files/462536/insteonplm-2016-09-08.txt)
[openhab-2016-09-08.txt](https://github.com/openhab/openhab/files/462537/openhab-2016-09-08.txt)
##

tpett:
I did a little bit of digging in git to try and help out. Just looking at the diff between v1.8.3 and master is way too much because someone touched pretty much every line in every file when they did some code format cleanup (b85c6575ad8565807a2ee9b6a82e4e4669188c4d). I filtered that commit out though and there aren't that many interesting changes to filter through.

Here are the commands I ran. I don't know if it is possible to do this using Github UI, but it is trivial in git command line:

```
git diff v1.8.3..5a845a94e02e7cf9daf7c0a40ecbdaf898190e43 bundles/binding/org.openhab.binding.insteonplm
git diff 03b62452009363556bcecb7da8e6a97cbf92ef01..master bundles/binding/org.openhab.binding.insteonplm
```

To make things easier I exported patches for each of these diffs and put them in a gist here: https://gist.github.com/tpett/f55d42b7d7f48afeaeb68a0a1416c6c7

I don't know the codebase well enough to spot the problem, but a developer experienced with this binding might know where to dig right away. I hope this helps!

larizzo:
[...]

PR #4925, opened 2016-12-22
insteonplm: fix for insteon hub with openHAB 2

robnielsen:
The binding now includes httpclient-4.1.3.jar which fixes a NoSuchMethodError exception that occurs with newer versions of httpclient.

This should resolve https://github.com/openhab/openhab/issues/3922 and https://github.com/openhab/openhab/issues/4630.

The fix has been tested by myself and wdesponts.


larizzo:
I was also able to test also and it does indeed resolve #3922 and #4630.

PR #4930, opened 2016-12-24
[Squeezebox] Add capability to retry failed connections at startup

### LINKED ISSUE #4611: openhab doesn't recognize squeezebox server at server boot
After booting my debian server all processes starts up. Especially squeezebox server (LMS) needs some minutes to come up. In general, openhab starts faster than LMS. I assume that this is the reason, why openhab does not recognize squeezebox server. Later, when squeezbox server has started up, openhab obviously doesn't care about. Just a reboot of openhab solves this and LMS is connected to OH.
I would expect, that the squieezebox binding look on regular basis after the squeezebox server.

I run OH 1.8.3 on a debain 8.5 server. Logitech Media Server Version: 7.9.0.

Is this a problem, which is known or do just I have it?
##

9037568:
Add capability to retry failed connections at startup.

Fixes #4611.


watou:
LGTM, but small suggestion: Would it be more concise to combine the `connect` and `retryConnect` methods into the `connect` method?  Not that there is anything wrong with your code!

9037568:
Thanks, John.  The reason I like them separate is to keep connect as a small bit of code that just does one connection attempt, while retryConnect keeps the full retry logic.

PR #4936, opened 2016-12-27
[satel] Added ability to revert user code

druciak:
This PR adds new action method `satelResetUserCode()` which reverts user code set by `satelSetUserCode(String)` to the value configured in *openhab.cfg*.


watou:
I added a line to the wiki but feel free to improve it further, and thanks @druciak!

druciak:
Thanks a lot! :)

PR #4944, opened 2016-12-28
Fix: Changed the Base64 Encoder/Decoder to be java 7 compatible

Markinus:
This should fix now the compiling on java 7.


watou:
The binding code was deleted from the master branch, so it will have to be restored, or better you could submit that same branch from your fork as a new PR, and add this Java 7 compatibility change as a new commit.

PR #4947, opened 2016-12-28
* NEW: Bindung for the communication with the KM200 

Markinus:
Buderus Gateway Logamatic web KM200 (/ KM100 / KM50 should work too, tested on KM200)

  It is possible to get and set string and float values. There are string, float, date and switch items supported.

* All log outputs are now parameterized
  Fixed some function headers and outputs
  Changed the binding configuration style

* Added some new predefined services

* Some more fixes in the logging strings

* Fixed wrong type casting

* Fix: Changed the Base64 Encoder/Decoder to be java 7 compatible

* Enabled direct access to the switching programs with json over strings
  It is usable in scripts and rules.


9037568:
Thanks for updating this, @Markinus !

Issue #4777, opened 2016-11-13
[denon] Marantz M-CR611 - not updating states of items


rapiertg:
I have configured Marantz in this way both on OH 1.8 and current snapshot of 2.0:
```
avr2000.host=192.168.1.26
avr2000.update=http
refresh=2000
```
And everything works in one direction. Eg. Commands are executed on the device - but states of items are not send to openhab. I mean that if I turn it off with my remote, change Volume, Input or mute then the Input Web Interface are not reflected. It is also not showing current TRACK, ALBUM etc. This is weird as I can get these info by sending POST message to _http://192.168.1.26/goform/AppCommand.xml_:
```
<?xml version="1.0" encoding="utf-8"?>
<tx>
 <cmd id="1">GetVolumeLevel</cmd>
 <cmd id="1">GetMuteStatus</cmd>
 <cmd id="1">GetSourceStatus</cmd>
</tx>
```
and get:
```
<?xml version="1.0" encoding="utf-8" ?>
<rx>
<cmd>
<volume>-73.0</volume>
<disptype>ABSOLUTE</disptype>
<dispvalue>7</dispvalue>
</cmd>
<cmd>
<mute>off</mute>
</cmd>
<cmd>
<source>Internet Radio</source>
</cmd>
</rx>
```
Current playback info can be seen xml file in http://192.168.1.26/goform/formNetAudio_StatusXml.xml?ZoneName=MAIN+ZONE

Would it be possible to make this little device work both sides with denon binding?

9037568:
Can you provide a full debug log, @rapiertg ?

sjcliffe:
I've recently purchased this model so would like to see this functionality too.

sjcliffe:
Startup log:

2016-12-21 13:07:05.232 [DEBUG] [.denon.internal.DenonActivator] - Denon binding has been started.
2016-12-21 13:07:05.393 [DEBUG] [.i.DenonGenericBindingProvider] - denon:hifi item Hifi_Power bound to property PW
2016-12-21 13:07:05.395 [DEBUG] [.i.DenonGenericBindingProvider] - denon:hifi item Hifi_Mute bound to property MU
2016-12-21 13:07:05.396 [DEBUG] [.i.DenonGenericBindingProvider] - denon:hifi item Hifi_Volume bound to property MV
[...]

Issue #3243, opened 2015-10-04
Denon Binding: Encoding error in get javax.xml.bind.UnmarshalException: null


nesys:
Hi Team,

just updated from 1.6.2 to 1.7.1 (Binding 1.7.1, but also tried with 1.8.0 snapshot) and during startup the log is saying:

2015-10-03 22:41:34.863 [WARN ] [.denon.internal.DenonConnector] - Encoding error in get
javax.xml.bind.UnmarshalException: null
    at com.sun.xml.internal.bind.v2.runtime.unmarshaller.UnmarshallerImpl.handleStreamException(UnmarshallerImpl.java:468) ~[na:1.8.0_31]
    at com.sun.xml.internal.bind.v2.runtime.unmarshaller.UnmarshallerImpl.unmarshal0(UnmarshallerImpl.java:402) ~[na:1.8.0_31]
    at com.sun.xml.internal.bind.v2.runtime.unmarshaller.UnmarshallerImpl.unmarshal(UnmarshallerImpl.java:371) ~[na:1.8.0_31]
    at org.openhab.binding.denon.internal.DenonConnector.getDocument(DenonConnector.java:549) [bundlefile:na]
    at org.openhab.binding.denon.internal.DenonConnector.setConfigProperties(DenonConnector.java:474) [bundlefile:na]
    at org.openhab.binding.denon.internal.DenonConnector.getInitialState(DenonConnector.java:189) [bundlefile:na]
    at org.openhab.binding.denon.internal.DenonConnector.connect(DenonConnector.java:147) [bundlefile:na]
    at org.openhab.binding.denon.internal.DenonBinding.updated(DenonBinding.java:189) [bundlefile:na]
    at org.eclipse.equinox.internal.cm.ManagedServiceTracker$1.run(ManagedServiceTracker.java:183) [org.eclipse.equinox.cm_1.0.400.v20120522-1841.jar:na]
    at org.eclipse.equinox.internal.cm.SerializedTaskQueue$1.run(SerializedTaskQueue.java:36) [org.eclipse.equinox.cm_1.0.400.v20120522-1841.jar:na]
Caused by: com.sun.istack.internal.SAXParseException2: unexpected element (uri:"", local:"html"). Expected elements are <{}device_Info>
    at com.sun.xml.internal.bind.v2.runtime.unmarshaller.UnmarshallingContext.handleEvent(UnmarshallingContext.java:726) ~[na:1.8.0_31]
[...]

PR #4952, opened 2016-12-30
Denon: Enhancements/fixes for #3243 and #4777.

9037568:
Avoid a null pointer exception in 4777.
Enhance logging a bit and eliminate noisy failure in 3243.

### DIFF
--- a/bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
+++ b/bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
@@ -137,73 +137,75 @@ public class DenonBinding extends AbstractActiveBinding<DenonBindingProvider>imp
      */
     @Override
     public void updated(Dictionary<String, ?> config) throws ConfigurationException {
-        if (config != null) {
-            logger.debug("Denon binding updated");
+        logger.debug("Denon binding updated");
 
-            Enumeration<String> keys = config.keys();
-
-            while (keys.hasMoreElements()) {
-
-                String key = keys.nextElement();
-                if (CONFIG_SERVICE_PID.equals(key)) {
-                    continue;
-                }
-
-                String[] parts = key.split("\\.");
-                String value = ((String) config.get(key)).trim();
-
-                if (parts.length == 1) {
-                    String option = parts[0];
-                    if (CONFIG_REFRESH.equals(option)) {
-                        refreshInterval = Integer.valueOf(value);
-                    }
-                } else {
-                    String instance = parts[0];
-
-                    DenonConnectionProperties connection = connections.get(instance);
-                    if (connection == null) {
-                        connection = new DenonConnectionProperties();
-                        connection.setInstance(instance);
-                        connections.put(instance, connection);
-                    }
-
-                    String option = parts[1].trim();
-
-                    if (CONFIG_HOST.equals(option)) {
-                        connection.setHost(value);
-                    } else if (CONFIG_UPDATE_TYPE.equals(option)) {
-                        connection.setTelnet(value.equals(CONFIG_UPDATE_TYPE_TELNET));
-                        connection.setHttp(value.equals(CONFIG_UPDATE_TYPE_HTTP));
-
-                        if (!value.equals(CONFIG_UPDATE_TYPE_TELNET) && !value.equals(CONFIG_UPDATE_TYPE_HTTP)) {
-                            logger.warn("Invalid connection type {} for instance {}, using default", value, instance);
-                        }
-                    }
-                }
-            }
-
-            boolean isActiveBinding = false;
-
-            for (Entry<String, DenonConnectionProperties> entry : connections.entrySet()) {
-                DenonConnectionProperties connection = entry.getValue();
-
-                logger.debug("Denon receiver configured at {}", connection.getHost());
-                DenonConnector connector = new DenonConnector(connection, new DenonPropertyUpdatedCallback() {
-                    @Override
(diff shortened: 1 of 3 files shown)


bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
```diff
@@ -137,73 +137,75 @@ protected void removeBindingProvider(DenonBindingProvider bindingProvider) {
      */
     @Override
     public void updated(Dictionary<String, ?> config) throws ConfigurationException {
-        if (config != null) {
-            logger.debug("Denon binding updated");
+        logger.debug("Denon binding updated");
 
-            Enumeration<String> keys = config.keys();
+        if (config == null) {
+            logger.warn("Denon binding has no configuration defined. Check the configuration file.");
```
watou: Is this warning necessary?

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
```diff
@@ -137,73 +137,75 @@ protected void removeBindingProvider(DenonBindingProvider bindingProvider) {
...
+        while (keys.hasMoreElements()) {
+            String key = keys.nextElement();
+            if (CONFIG_SERVICE_PID.equals(key)) {
+                continue;
+            }
 
-                String[] parts = key.split("\\.");
-                String value = ((String) config.get(key)).trim();
+            String[] parts = key.split("\\.");
+            String value = ((String) config.get(key)).trim();
```
watou: It's possible that the configuration can have non-String values, so instead of `(String) config.get(key)`, you could safely use `Objects.toString(config.get(key), null)`.

Also, is `trim` really needed?  If the code reading a .cfg file uses [this](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.Reader)), trimming the string shouldn't be needed.

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonConnector.java
```diff
@@ -449,7 +450,12 @@ private void updateMainZone() {
             stateCache.put(DenonProperty.POWER_MAINZONE.getCode(),
                     mainZone.getPower().getValue() ? OnOffType.ON : OnOffType.OFF);
             stateCache.put(DenonProperty.MUTE.getCode(), mainZone.getMute().getValue() ? OnOffType.ON : OnOffType.OFF);
-            stateCache.put(DenonProperty.SURROUND_MODE.getCode(), new StringType(mainZone.getSurrMode().getValue()));
+            if (mainZone.getSurrMode() == null ) {
```
watou: Run the Eclipse formatter.

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
9037568: Certainly it seems like it's needed. If the configuration doesn't define at least one receiver, nothing's going to work, unless I'm misreading something.

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
9037568: Agreed. Will update this.

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonBinding.java
watou: `updated` can be called with a `null` `Dictionary`, but this doesn't mean that there is no configuration file.  [Here](https://nakvic-dev.blogspot.ie/2010/11/osgi-declarative-managed-services-and.html) is an example outside the world of openHAB that shows a case where `updated` silently does nothing if passed a `null` `Dictionary`.  There are also many bindings that handle this situation the same silent way.  So I worry that this log message will heighten user anxiety without pointing to an actual problem.

9037568: I removed the warning because it didn't exist in the original code (before my changes), so it's effectively no change to not have it there.

However I don't understand why we're claiming it's not needed. If udpated() is passed a null config, the binding will not work. Which means it will fail silently.

Interestingly, I notice that in the current code, if updated is passed a correct configuration, then later receives a null, the binding remains "properly configured", because that flag never gets cleared...

watou: > However I don't understand why we're claiming it's not needed. If udpated() is passed a null config, the binding will not work. Which means it will fail silently.

Yes, but it will probably also be called with a non-`null` configuration around the time that the service is registered.

A ManagedService can have its `updated` method called by the ConfigAdminManager with a `null` dictionary, but its doing so doesn't mean there is really no configuration.  So it should be treated as a "nothing" and ignored.  I don't know if this is an implementation quirk or if there is some intended meaning behind it, but in practice, the only calls to `updated` that should do anything are ones where the `Dictionary` is not `null`.

watou:
LGTM. Merge when ready!

9037568:
Thanks, John.

PR #4960, opened 2017-01-04
[HttpBinding]: Implement post with body content

### LINKED ISSUE #2044: [HttpBinding] does not send post request body
Hi,
I am using the http binding to change the status of my relais. The item configuration is;

> Switch gWZ_Light  "Light Test"    <switch>    (gWZ_Mm) {
>   http=">[ON:POST:http://192.168.178.20/?6=1] >[OFF:POST:http://192.168.178.20/?6=0]"
> }

The configuration seems to be correct, as I can see in the logs and in the code.  
But the data which is send to the server does not contain any request body. So the server receives:

> POST /?6=1 HTTP/1.1
> User-Agent: Jakarta Commons-HttpClient/3.1
> Host: 192.168.178.20
> Content-Length: 0
> <line break (correct due to specification)>
> <line break (not correct, here should be the post data)>

This leads to the problem, that the server rejects the POST request, because POST data must be send in the request body and not as GET-URL-Sring.  
So I am not sure if my configuration is incorrect or the POST method implementation is incorrect.

Could anyone take a look please?

Cheers,
Iwan
##

9037568:
Implement a means of sending an HTTP POST with body content.

Fixes #2044.


bundles/binding/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/HttpBinding.java
```diff
@@ -303,8 +308,31 @@ private void formatAndExecute(String itemName, Command command, Type value) {
...
+        if (StringUtils.isNotBlank(httpMethod) && StringUtils.isNotBlank(url)) {
+            if (httpMethod.equals("POST") && StringUtils.isNotBlank(body)) {
+                try {
+                    stream = IOUtils.toInputStream(body, "UTF-8");
+                } catch (IOException ioe) {
+                    logger.error("Failed to convert the specified body into an acceptable input stream.", ioe);
+                    logger.error("Body contents: {}", body);
+                }
+                logger.debug("Executing url '{}' via method {}, with body content '{}'", url, httpMethod, body);
+                HttpUtil.executeUrl(httpMethod, url, provider.getHttpHeaders(itemName, command), stream, "text/plain",
```
watou: Will `text/plain` be suitable in all likely use cases for posting body content?

9037568: I would think so, but there are probably some deficient apps out there that insist on having a different content-type in the header.

There doesn't seem to be a very good way of allowing the end user to specify this in a binding, though. I considered adding new POST commands that would carry different content-types (eg. POSTJ for json, POSTX for xml).  But that seems like an ugly bit of hackery.

watou: I agree it's very limiting to fit everything sensibly inside a binding config string.  In the case where someone needs a different content type, the answer should be to use the [HTTP action](https://github.com/openhab/openhab/wiki/Actions#http-actions) triggered from a rule.  In fact, the workaround for the absence of this PR is:

.items
```
String PostBody
```

.rules
```
rule PostBodyToSomewhere
when
  Item PostBody received command
then
  val url = "http://server/path"
  sendHttpPostRequest(url, "text/plain", receivedCommand.toString)
end

watou:
@9037568:
Weird how HttpGenericBindingProvider.java has conflicts with master now.  Hmmm.  Aside from that, is this ready to merge Chris?

9037568:
Yes, I'd say this is ready to be merged.

watou:
I resolved the one odd conflict before merge.

watou:
Would you mind updating the wiki? ;-) Thanks.

9037568:
Thanks, John. Wiki is on my todo list here shortly.

watou:
@9037568 : I updated the wiki further [here](https://github.com/openhab/openhab/wiki/Http-Binding/_compare/9a9cfe101fa7e3b42460f57aedc2fac2f3055fa5...3c1550bf65aa60388f5f96a76023515ffcba12e8).  Please feel free to make further improvements.  Thanks Chris.

PR #4970, opened 2017-01-07
[diyonxbee] Fix parsing messages when the key part is sent in another message as the value

juri8:


bundles/binding/org.openhab.binding.diyonxbee/src/main/java/org/openhab/binding/diyonxbee/internal/DiyOnXBeeBinding.java
```diff
@@ -369,47 +369,74 @@ protected void internalReceiveUpdate(String itemName, State newState) {
...
+
+	void processResponse(final String message, final String remoteAddress) {
+		logger.debug("received message: '{}' from '{}'", message, remoteAddress);
+		
+		int startIdx = 0;
+		do {
+			final int idxEquals = message.indexOf('=', startIdx);
+			final int idxEnd = message.indexOf(ITEM_SEPARATOR, startIdx);
+			
+			if(idxEquals > 0 && idxEnd > 0) {
```
watou: Please apply the formatter in the Eclipse IDE.

watou:
cherry picked onto 1.9.x so it will appear in OH 2.0.0 release.  Thanks @juri8!

PR #4971, opened 2017-01-08
Added the capability to add and remove switch cycles

Markinus:
+ Add a new cycle with the max. cycle number +1 (only to the end)
+ Delete a cycle if both cwitching points are set to max time (1430Min., only the last one)


watou:
Hi @Markinus, please report when this PR is ready for review/merge, and thanks!

Markinus:
Hi, it is. You can check and merge it. Wiki is updated, too.

PR #4975, opened 2017-01-08
KNX-Binding: Changed warn to debug in log message 'Ignoring local events...'

sebastiankutschbach:
Ignoring local events is the correct behaviour of the binding. This message only blows up my and probably logfiles of other users.

Signed-off-by: Sebastian Kutschbach <sebastian.kutschbach@freenet.de>


watou:
Thanks @sebastiankutschbach.

PR #4981, opened 2017-01-10
TinkerForge: Fix AmbientLightV2 Lux values

theoweiss:
AmbientLightV2 measures 1/100 Lux values not 1/10 Lux.
The auto formatter wants even more changes ;-( .

### DIFF
--- a/bundles/binding/org.openhab.binding.tinkerforge/src/main/java/org/openhab/binding/tinkerforge/internal/model/BrickStepperSubIds.java
+++ b/bundles/binding/org.openhab.binding.tinkerforge/src/main/java/org/openhab/binding/tinkerforge/internal/model/BrickStepperSubIds.java
@@ -19,7 +19,7 @@ import org.eclipse.emf.common.util.Enumerator;
  * A representation of the literals of the enumeration '<em><b>Brick Stepper Sub Ids</b></em>',
  * and utility methods for working with them.
  * <!-- end-user-doc -->
- * 
+ *
  * @see org.openhab.binding.tinkerforge.internal.model.ModelPackage#getBrickStepperSubIds()
  * @model
  * @generated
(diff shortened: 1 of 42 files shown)


9037568:
There are a number of instances of code changes to blocks like this one where the curly braces have been removed. That seems to be contrary to the code convention. Did your formatter do this?

I don't see any of these changes when I run the Eclipse formatter...


````
if (eNotificationRequired()) {
             eNotify(new ENotificationImpl(this, Notification.SET, ModelPackage.BRICKLET_OLED_CONFIGURATION__INVERT, oldInvert, invert));
}
````

**Comment by theoweiss** [comment]
I've rerun the formatter and as well see no changes. Do I have to apply any updates to the ide? I use an oomph installed eclipse. The initial installation is quite old and if I execute "Check for Updates" lots of updates are proposed.

Formatter settings:
eclipseformatter

Another option is to disable the formatter for the generated code:
https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.tinkerforge/model/tinkerforge.xcore#L11 setting:

```
codeFormatting="false"
```

???

9037568:
I don't know about the IDE setup. I dislike that everything depends on installing the IDE from scratch in just the right way... but that's a digression for another time.

Kai had suggested [over here in this thread](https://github.com/openhab/openhab-distro/pull/344#issuecomment-265725006) to use [the official oomph profile](https://github.com/openhab/openhab-distro/blob/master/launch/openHAB2.setup).

theoweiss:
> Kai had suggested over here in this thread to use the official oomph profile.

That's what I did a year ago. What's really not obvious to me is how to upgrade the ide.

I've reverted the format changing commit.

theoweiss:
@9037568 is there any chance to get this into 2.0.0?

watou:
cherry picked onto 1.9.x branch so it should appear in OH 2.0.0

theoweiss:
Thanks @watou that's great.

PR #4950, opened 2016-12-30
[http] Configure binding from Paper UI

watou:
Signed-off-by: John Cocula <john@cocula.com>

PR #5000, opened 2017-01-16
[km200] Bug fixes; install/config in OH2

Markinus:
And preparations for OH2

### DIFF
--- a/bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
@@ -4,38 +4,36 @@
         xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
         xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
 
-    <name>km200 Binding</name>
-    <description>The KM200 Binding is communicating with a Buderus Logamatic web KM200 / KM100 / KM50. It is possible to recive and send parameters like string or float values.</description>
+    <name>KM200 Binding</name>
+    <description>The KM200 binding is communicating with a Buderus Logamatic web KM200 / KM100 / KM50. It is possible to receive and send parameters like string or float values.</description>
     <author>Markus Eckhardt</author>
     
-    <service-id>org.openhab.binding.km200</service-id>
+    <service-id>org.openhab.km200</service-id>
 
     <config-description>
-        <parameter name="ip4_address" type="text">
+        <parameter name="ip4_address" type="text" required="true">
             <label>IP4 Address</label>
+            <context>network-address</context>
             <description>IP4 Address of the KMXXX device</description>
-            <default>"192.168.XXX.XXX"</default>
         </parameter>
-        <parameter name="PrivKey" type="text">
+        <parameter name="PrivKey" type="text" required="false">
            <label>Private Key</label>
            <description>Option 1:Private en-/decryption key built from MD5Salt, GatewayPassword and PrivatePassword</description>
-           <default></default>
-        </parameter>
-        <parameter name="MD5Salt" type="text">
+         </parameter>
+        <parameter name="MD5Salt" type="text" required="false">
             <label>MD5Salt</label>
             <description>Option 2: MD5Salt for the building of the private key</description>
-            <default></default>
         </parameter>
-        <parameter name="GatewayPassword" type="text">
+        <parameter name="GatewayPassword" type="text" required="false">
            <label>Gateway Password</label>
+           <context>password</context>
            <description>Option 2: GatewayPassword for the building of the private key (Sticker on the device without '-')</description>
-           <default></default>
         </parameter>
-        <parameter name="PrivatePassword" type="text">
+        <parameter name="PrivatePassword" type="text" required="false">
            <label>Private Password</label>
+           <context>password</context>
            <description>Option 2: Private Password for the building of the private key (Set in the App)</description>
-           <default></default>
-        </parameter>
+         </parameter>
     </config-description>
 
 </binding:binding>
--- a/bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
+++ b/bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="km200"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>km200 Binding</name>
+    <description>The KM200 Binding is communicating with a Buderus Logamatic web KM200 / KM100 / KM50. It is possible to recive and send parameters like string or float values.</description>
+    <author>Markus Eckhardt</author>
+    
+    <service-id>org.openhab.binding.km200</service-id>
+
+    <config-description>
+        <parameter name="ip4_address" type="text">
+            <label>IP4 Address</label>
+            <description>IP4 Address of the KMXXX device</description>
+            <default>"192.168.XXX.XXX"</default>
+        </parameter>
+        <parameter name="PrivKey" type="text">
+           <label>Private Key</label>
+           <description>Option 1:Private en-/decryption key built from MD5Salt, GatewayPassword and PrivatePassword</description>
+           <default></default>
+        </parameter>
+        <parameter name="MD5Salt" type="text">
+            <label>MD5Salt</label>
+            <description>Option 2: MD5Salt for the building of the private key</description>
+            <default></default>
+        </parameter>
+        <parameter name="GatewayPassword" type="text">
+           <label>Gateway Password</label>
+           <description>Option 2: GatewayPassword for the building of the private key (Sticker on the device without '-')</description>
+           <default></default>
+        </parameter>
+        <parameter name="PrivatePassword" type="text">
+           <label>Private Password</label>
+           <description>Option 2: Private Password for the building of the private key (Set in the App)</description>
+           <default></default>
+        </parameter>
+    </config-description>
+
+</binding:binding>
--- a/bundles/binding/org.openhab.binding.km200/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.km200/META-INF/MANIFEST.MF
@@ -6,11 +6,13 @@ Bundle-Version: 1.9.0.qualifier
 Bundle-Activator: org.openhab.binding.km200.internal.KM200Activator
 Bundle-Vendor: openHAB.org
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
-Import-Package: com.google.common.net;version="10.0.1",
+Import-Package: com.google.common.io,
+ com.google.common.net,
+ org.apache.commons.codec.binary,
  org.apache.commons.collections,
- org.apache.commons.httpclient;version="3.1.0",
- org.apache.commons.httpclient.methods;version="3.1.0",
- org.apache.commons.httpclient.params;version="3.1.0",
+ org.apache.commons.httpclient,
+ org.apache.commons.httpclient.methods,
+ org.apache.commons.httpclient.params,
  org.apache.commons.io,
  org.apache.commons.lang,
  org.apache.commons.net,
@@ -31,8 +33,6 @@ Import-Package: com.google.common.net;version="10.0.1",
  org.slf4j
 Export-Package: org.openhab.binding.km200
 Service-Component: OSGI-INF/genericbindingprovider.xml, OSGI-INF/km200binding.xml
-Require-Bundle: org.apache.commons.codec,
- com.google.guava
 Bundle-ClassPath: .,
  lib/json-20160810.jar
 Bundle-ActivationPolicy: lazy
--- a/bundles/binding/org.openhab.binding.km200/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.km200/META-INF/MANIFEST.MF
@@ -27,12 +27,12 @@ Import-Package: com.google.common.net;version="10.0.1",
  org.osgi.framework,
  org.osgi.service.cm,
  org.osgi.service.component,
- org.osgi.service.event
+ org.osgi.service.event,
+ org.slf4j
 Export-Package: org.openhab.binding.km200
 Service-Component: OSGI-INF/genericbindingprovider.xml, OSGI-INF/km200binding.xml
 Require-Bundle: org.apache.commons.codec,
- com.google.guava,
- org.slf4j.api
+ com.google.guava
 Bundle-ClassPath: .,
  lib/json-20160810.jar
 Bundle-ActivationPolicy: lazy
--- a/bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
+++ b/bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
@@ -11,6 +11,7 @@ package org.openhab.binding.km200.internal;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.math.BigDecimal;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -160,20 +161,22 @@ class KM200Comm {
                 for (int i = 0; i < maxNbrGets && statusCode != HttpStatus.SC_OK; i++) {
                     // Execute the method.
                     statusCode = client.executeMethod(method);
-                    // Check the status and the forbidden 403 Error.
-                    if (statusCode != HttpStatus.SC_OK) {
-                        String statusLine = method.getStatusLine().toString();
-                        if (statusLine.contains(" 500 ")) { /* Unknown problem with the device, wait and try again */
-                            logger.warn("HTTP GET failed mit error 500, try Again");
+                    // Check the status
+                    switch (statusCode) {
+                        case HttpStatus.SC_OK:
+                            break;
+                        case HttpStatus.SC_INTERNAL_SERVER_ERROR:
+                            /* Unknown problem with the device, wait and try again */
+                            logger.warn("HTTP GET failed: 500, internal server error, repeating.. ");
                             Thread.sleep(2000L);
                             continue;
-                        } else if (statusLine.contains(" 403 ")) {
+                        case HttpStatus.SC_FORBIDDEN:
+                            /* Service is available but not readable */
                             byte[] test = new byte[1];
                             return test;
-                        } else {
+                        default:
                             logger.error("HTTP GET failed: {}", method.getStatusLine());
                             return null;
-                        }
                     }
                 }
                 device.setCharSet(method.getResponseCharSet());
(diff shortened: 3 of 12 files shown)


bundles/binding/org.openhab.binding.km200/META-INF/MANIFEST.MF
```diff
@@ -27,12 +27,12 @@ Import-Package: com.google.common.net;version="10.0.1",
  org.osgi.framework,
  org.osgi.service.cm,
  org.osgi.service.component,
- org.osgi.service.event
+ org.osgi.service.event,
+ org.slf4j
 Export-Package: org.openhab.binding.km200
 Service-Component: OSGI-INF/genericbindingprovider.xml, OSGI-INF/km200binding.xml
 Require-Bundle: org.apache.commons.codec,
```
watou: You must not use `Require-Bundle` but instead only use `Import-Package`.

Markinus:
Ok, changed.

Markinus:
Binding is working now with OH2 in lagacy mode.

watou:
Excellent!

Markinus:
I reverted my last commit with the project renaming. 
But, if I don't do it then I cannot import the new skaleton for OH2 binding (Eclipse is not doing it because a project with this name is already existing...) What is the right way?

Markinus:
The homatic binding is done in exact this way. That's the reason for my decision.

watou:
Oh yes, the project namespace clash in the IDE.  I thought you had changed the name of the bundle, not the project.  

Are you proceeding directly to authoring an OH2 binding?  If so, that's great!  But it might make sense to abandon the 1.x version now before people start using it; otherwise you will have two codebases to maintain but with little value in having both.

Markinus:
Ok, so I will push renaming commit in again.

Yes, but this will take time. I have to lern how this whole discovery stuff works and think about structures. 

Btw, do I have to add the bindings to the feature.xml file?

watou:
> Btw, do I have to add the bindings to the feature.xml file?

To make the binding installable in openHAB 2, it must have a feature defined.  If it has a .cfg file, that also needs to be added.  [This PR](https://github.com/openhab/openhab1-addons/pull/4913/files) is an example of doing that.

Further, if your binding has a configuration and you would like it to be configurable via the PaperUI, the .JAR needs to include an `ESH-INF/binding/binding.xml` file.  [This PR](https://github.com/openhab/openhab1-addons/pull/4950/files) is an example of doing that.  If you are the author of the binding, consider adding your real name (not email or alias) as an `<author>Markus Eckhardt</author>` to binding.xml.

Markinus:
I'm done with this pull request. Check and merge it, please.

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="km200"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>km200 Binding</name>
```
watou: KM200 Binding

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+    <description>The KM200 Binding is communicating with a Buderus Logamatic web KM200 / KM100 / KM50. It is possible to recive and send parameters like string or float values.</description>
```
watou: *binding *receive

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+    <author>Markus Eckhardt</author>
+    
+    <service-id>org.openhab.binding.km200</service-id>
+
+    <config-description>
+        <parameter name="ip4_address" type="text">
+            <label>IP4 Address</label>
+            <description>IP4 Address of the KMXXX device</description>
+            <default>"192.168.XXX.XXX"</default>
```
watou: Are you sure a quoted string with X's in it is a valid default here?  Best to not have a default unless it can be concretely correct.  Please see [this link](https://www.eclipse.org/smarthome/documentation/development/bindings/xml-reference.html#xml-structure-for-configuration-descriptions) for the full XML definition for `<config-description>` and browse other 1.x add-on source for other examples.

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+        </parameter>
+        <parameter name="PrivKey" type="text">
+           <label>Private Key</label>
+           <description>Option 1:Private en-/decryption key built from MD5Salt, GatewayPassword and PrivatePassword</description>
+           <default></default>
```
watou: Don't use empty defaults.  See doc link mentioned above.

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+        </parameter>
+        <parameter name="MD5Salt" type="text">
+            <label>MD5Salt</label>
+            <description>Option 2: MD5Salt for the building of the private key</description>
+            <default></default>
```
watou: Don't use empty defaults.  See doc link mentioned above.

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+        </parameter>
+        <parameter name="GatewayPassword" type="text">
+           <label>Gateway Password</label>
+           <description>Option 2: GatewayPassword for the building of the private key (Sticker on the device without '-')</description>
+           <default></default>
```
watou: Don't use empty defaults.  See doc link mentioned above.

watou: Use `<context>password</context>` to obscure typing of the password.

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+        </parameter>
+        <parameter name="PrivatePassword" type="text">
+           <label>Private Password</label>
+           <description>Option 2: Private Password for the building of the private key (Set in the App)</description>
+           <default></default>
```
watou: Don't use empty defaults.  See doc link mentioned above.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -155,17 +157,23 @@ public static String bytesToHex(byte[] data) {
             method.addRequestHeader("User-Agent", "TeleHeater/2.2.3");
 
             try {
-                // Execute the method.
-                int statusCode = client.executeMethod(method);
-                // Check the status and the forbidden 403 Error.
-                if (statusCode != HttpStatus.SC_OK) {
-                    String statusLine = method.getStatusLine().toString();
-                    if (statusLine.contains(" 403 ")) {
```
watou: Please use the correct method for getting the HTTP response status code.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -155,17 +157,23 @@ public static String bytesToHex(byte[] data) {
...
-                    } else {
-                        logger.error("HTTP GET failed: {}", method.getStatusLine());
-                        return null;
+                for (int i = 0; i < maxNbrGets && statusCode != HttpStatus.SC_OK; i++) {
+                    // Execute the method.
+                    statusCode = client.executeMethod(method);
+                    // Check the status and the forbidden 403 Error.
+                    if (statusCode != HttpStatus.SC_OK) {
+                        String statusLine = method.getStatusLine().toString();
+                        if (statusLine.contains(" 500 ")) { /* Unknown problem with the device, wait and try again */
```
watou: Please use the correct method for getting the HTTP response status code.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -155,17 +157,23 @@ public static String bytesToHex(byte[] data) {
...
+                            logger.warn("HTTP GET failed mit error 500, try Again");
```
watou: Please use English and proper English upper- and lowercase. ;-)

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -155,17 +157,23 @@ public static String bytesToHex(byte[] data) {
...
+                            Thread.sleep(2000L);
+                            continue;
+                        } else if (statusLine.contains(" 403 ")) {
```
watou: Please use the correct method for getting the HTTP response status code.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -175,6 +183,9 @@ public static String bytesToHex(byte[] data) {
             } catch (HttpException e) {
                 logger.error("Fatal protocol violation: {}", e);
                 e.printStackTrace();
```
watou: Please never use `printStackTrace` but only use log4j methods where needed.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -175,6 +183,9 @@ public static String bytesToHex(byte[] data) {
             } catch (HttpException e) {
                 logger.error("Fatal protocol violation: {}", e);
```
watou: If the last argument is a `Throwable` as in this case, the stacktrace is logged but the `{}` will be printed.  If you only want the exception method in one line, use `e.getMessage()` instead of `e` and the message will substitute the `{}` in the message.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -175,6 +183,9 @@ public static String bytesToHex(byte[] data) {
...
                 e.printStackTrace();
+            } catch (InterruptedException e) {
+                logger.error("Sleep was interrupted: {}", e);
+                e.printStackTrace();
```
watou: See above two comments.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -175,6 +183,9 @@ public static String bytesToHex(byte[] data) {
...
             } catch (IOException e) {
                 logger.error("Fatal transport error: {}", e);
                 e.printStackTrace();
```
watou: See above two comments.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -368,12 +379,12 @@ public void initObjects(String service) {
...
                     logger.debug("initDevice: type float value: {}", decodedData);
-                    valObject = new Float(nodeRoot.getDouble("value"));
+                    valObject = Float.parseFloat(nodeRoot.get("value").toString());
                     newObject.setValue(valObject);
                     if (nodeRoot.has("minValue") && nodeRoot.has("maxValue")) {
                         List<Float> valParas = new ArrayList<Float>();
-                        valParas.add(new Float(nodeRoot.getDouble("minValue")));
-                        valParas.add(new Float(nodeRoot.getDouble("maxValue")));
+                        valParas.add(Float.parseFloat(nodeRoot.get("minValue").toString()));
+                        valParas.add(Float.parseFloat(nodeRoot.get("maxValue").toString()));
```
watou: `BigDecimal` is usually a much better solution than `Float` or `Double`, as the exact value can be represented.  Please use `BigDecimal` exclusively going between the KMXXX and item states.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -666,7 +677,7 @@ public State parseJSONData(String decodedData, String type, String item, KM200Bi
 
                 case "floatValue": /* Check whether the type is a single value containing a float value */
                     logger.debug("state of type float value: {}", decodedData);
-                    Float fVal = new Float(nodeRoot.getDouble("value"));
+                    Float fVal = Float.parseFloat(nodeRoot.get("value").toString());
```
watou: Please see above comment regarding Floats and Doubles.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -775,91 +786,83 @@ public State getVirtualState(KM200CommObject object, Class<? extends Item> itemT
...
-                switch (virtService) {
-                    case "weekday":
-                        if (itemType.isAssignableFrom(StringItem.class)) {
-                            String val = sPService.getActiveDay();
-                            if (val == null) {
-                                return null;
-                            }
-                            state = new StringType(val);
-                        } else {
-                            logger.error("Bindingtype not supported for day service: {}", itemType.getClass());
```
watou: Please use `logger.warn` for problems that don't lead to system stability issues.  If the binding recovers from a situation, don't use a more severe level than `WARN`.  Other similar lines below.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200SwitchProgramService.java
```diff
@@ -345,11 +321,24 @@ void updateSwitches(JSONObject nodeRoot) {
             /* Update the list of switching points */
             removeAllSwitches();
             JSONArray sPoints = nodeRoot.getJSONArray("switchPoints");
+            logger.debug("sPoints: {}", nodeRoot.toString());
```
watou: You never need to call `.toString()` on logger arguments and should not, as you can leave it up to the logger to call `.toString()` on any `java.lang.Object` only when needed.  Some `.toString()` methods are expensive, or at least create short-lived objects that add burden to the Java garbage collection.

features/openhab-addons-external/pom.xml
```diff
@@ -69,6 +69,7 @@
                                 <artifact><file>src/main/resources/conf/jdbc.cfg</file><type>cfg</type><classifier>jdbc</classifier></artifact>
                                 <artifact><file>src/main/resources/conf/jpa.cfg</file><type>cfg</type><classifier>jpa</classifier></artifact>
                                 <artifact><file>src/main/resources/conf/jointspace.cfg</file><type>cfg</type><classifier>jointspace</classifier></artifact>
+				<artifact><file>src/main/resources/conf/km200.cfg</file><type>cfg</type><classifier>km200</classifier></artifact>
```
watou: Please correct the formatting of this file using the Eclipse IDE formatter.

features/openhab-addons/src/main/feature/feature.xml
```diff
@@ -365,6 +365,13 @@
     <configfile finalname="${openhab.conf}/services/jointspace.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/jointspace</configfile>
   </feature>
 
+  <feature name="openhab-binding-km100_1" description="KM200 Binding" version="${project.version}">
```
watou: The feature name should be `openhab-binding-km1001`.

bundles/binding/org.openhab.binding.km200/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,41 @@
...
+<binding:binding id="km200"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+    <name>km200 Binding</name>
+    <description>The KM200 Binding is communicating with a Buderus Logamatic web KM200 / KM100 / KM50. It is possible to recive and send parameters like string or float values.</description>
+    <author>Markus Eckhardt</author>
+    
+    <service-id>org.openhab.binding.km200</service-id>
```
watou: The `<service-id>` needs to be identical to the service PID defined [here](https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.km200/OSGI-INF/km200binding.xml#L21), so it needs to be `org.openhab.km200`.

bundles/binding/org.openhab.binding.km200/src/main/java/org/openhab/binding/km200/internal/KM200Comm.java
```diff
@@ -379,12 +377,12 @@ public void initObjects(String service) {
 
                 case "floatValue": /* Check whether the type is a single value containing a float value */
                     logger.debug("initDevice: type float value: {}", decodedData);
-                    valObject = Float.parseFloat(nodeRoot.get("value").toString());
+                    valObject = BigDecimal.valueOf(nodeRoot.getDouble("value"));
```
watou: Not a reason to require a change to this PR, but if you could get the value from JSON without passing it through `double`, as in getting it as the string as it appears in the JSON payload, then the `BigDecimal` would be able to represent it exactly.  By becoming a `double`, it could lose the accuracy as it appears in JSON.  However, this JSON library may not let you do this.

Markinus:
Yes, you're right. I thoght that the BigDecimal is able to parse  a value from a string (like Double or Float) but it isn''t possible.

watou:
You can `new BigDecimal(String value)` no problem: 

http://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#BigDecimal(java.lang.String)

Markinus:
I overseen something. The jsonLib has a function getBigDecimal(). I will change this lines and push it.

watou:
> The jsonLib has a function getBigDecimal()

Perfect!

watou:
@Markinus LGTM, but please confirm this latest version works as intended on openHAB 2, and then it can be merged.  Thanks for everything!

Markinus:
The binding is working with OH2 in lagacy mode, some users tested it too. But I don't know what I have to replace in the OH2 installation for testing of the PaperUI configuration.

watou:
@Markinus  Thank you!  Small comment about terminology: the binding runs on OH2 via the compat1x layer.  The term "legacy" is only used to refer to 1.x bindings for which there is a 2.0 replacement and thus aren't listed unless you enable "Include legacy 1.x bindings" in PaperUI under Configuration->System->Add-on Management.  Since there is not yet a 2.0 replacement, this binding ought to be installable and configurable in the next OH2 build that includes it.

@kaikreuzer please consider cherry picking https://github.com/openhab/openhab1-addons/commit/5e58cd1d8392bec120d31e328dbf46044c960146 onto the 1.9.x branch.

watou:
I've cherry picked the commit onto 1.9.x now.

PR #5006, opened 2017-01-17
[mail] Set sslSmtpPort if ssl=true; added optional timeout parameters

### LINKED ISSUE #5004: [mail] If ssl=true, no way to override default SSL port 465
An openHAB 2 user was setting `port=` in mail.cfg, but no matter what, the mail action was attempting to use port 465.  Apache Commons Email 1.4 needs to have that port explicitly overridden ([code](https://commons.apache.org/proper/commons-email/apidocs/src-html/org/apache/commons/mail/Email.html#line.1704)), so a new configuration option is needed for users who use `ssl=true` in their configurations.

See forum discussion [here](https://community.openhab.org/t/sending-mail-does-not-work-on-snapshot-release/20225).

watou:
Fixes #5004.

Also replaced String casting in case configuration comes from Paper UI, console or a source other than .cfg file.


9037568:
LGTM.

PR #5007, opened 2017-01-18
[HTTP binding] Fix logic for handling headers in binding config.

### LINKED ISSUE #4992: [HTTP Binding] JSON Body impossible with parseOutBindingConfig
## Expected Behavior
```http='>[*:POST:http://192.168.178.2/animation:{"name": "rainbow"}]'```
shall send a POST request with appended json data.

## Current Behavior
It tries to parse HTTP Headers from some "endsWith trimmed } and contains {"
https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/HttpGenericBindingProvider.java#L267

## Possible Solution

Refactor Regexes (modular) according to documentation, e.g. match(3) as url+headers and match(4) as body.
Parse URL headers from match(3) string.

compile current git master from https://github.com/openhab/openhab/blob/master/bundles/binding/org.openhab.binding.http

run with any Outgoing (e.g. Switch) Item 
```Switch Rainbow (Animations)  { http='>[*:POST:http://192.168.178.2/animation:{"name": "rainbow"}]', autoupdate="false"}```

## Context
OpenHAB --> REST --> nodeMCU --> Lamp

## Your Environment
Home Environment. Running openHAB on Raspberry pi 3 and arch Linux with jre 1.8_u112.
Compiled **1.9.0-SNAPSHOT** Version of HTTP Bindings and added to stable 1.8.3 Server instance.

9037568:
Add new test for same.
Fixes #4992.

### DIFF
--- a/bundles/binding/org.openhab.binding.http.test/src/test/java/org/openhab/binding/http/internal/HttpGenericBindingProviderTest.java
+++ b/bundles/binding/org.openhab.binding.http.test/src/test/java/org/openhab/binding/http/internal/HttpGenericBindingProviderTest.java
@@ -194,7 +194,6 @@ public class HttpGenericBindingProviderTest {
         Assert.assertTrue(config.get(HttpGenericBindingProvider.IN_BINDING_KEY).headers.containsKey("header2"));
         Assert.assertTrue(config.get(HttpGenericBindingProvider.IN_BINDING_KEY).headers.contains("value1"));
         Assert.assertTrue(config.get(HttpGenericBindingProvider.IN_BINDING_KEY).headers.contains("value2"));
-
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


watou:
Hi Chris,  I'm not able to understand how this change discriminates between the syntax for HTTP headers and being able to post arbitrary data in the POST body.  I see you added a test that looks like a spec for HTTP headers, but #4992 is asking for the ability to post JSON data in the body of the request.  I might just be too dense at the moment to understand it.  Would you mind verbalizing how this change is supposed to work, all the while being fully compatible with binding config strings that predate the POST body enhancement?  Thanks very much.

9037568:
Absolutely no problem. But I won't be able to fill that in until tomorrow at least.

watou:
Thanks Chris.  No worries.

But note the schedule [here](https://community.openhab.org/t/upcoming-releases/18503/39?u=watou) for making the cut on inclusion in the 1.9.0 release from this repo, and thus the openHAB 2.0.0 release.

watou:
Chris, would you mind sending me an email to first@last.com, when you get a chance?  I want to keep you up to date (or let you know how to keep yourself up to date in case you don't already 😄 ) regarding repo and time topics.  Many thanks.

9037568:
Email sent if I interpreted your email address instructions correctly.

watou:
Thanks; I've replied!

9037568:
Here's an explication of the core logic and how this fix applies. 

First, the essential problem found by this ticket is that a JSON POST body causes an out binding config string to be treated as if it contained a set of header definitions and no POST body.

The out binding config strings are parsed by separating them into 3 parts:
1) Command
2) Http Method
3) URL

This was true before the change to allow POST body content and hasn't changed.

The URL in part 3 has for some time been allowed to include a set of headers at its end formatted like this:
````
http://a.b.c.d/something.htm{header1=value1}
````

These header values were removed from the URL by checking if the URL ended with a right curly brace.  My mistake was not examining that logic when I implemented the POST body logic.  So what happens without the fix for this ticket is that if the URL contains a POST body that is JSON, and ends with a right curly brace, it triggers the header parsing logic and skips over the POST body logic.  So you get an incorrect URL and no POST body.

The fix is to take the entirety of the URL part 3 and parse it into sections. The code then checks the full URL for header definitions and for POST body content.  It therefore eliminates the problem where a JSON POST body gets treated as if it were header definitions.

watou:
Thank you for the clear explanation.  So this means that the previous ability to provide headers remains in place, but with this PR the POST body content can contain JSON (or any textual data?) as the POST body content?  If so, should this PR be merged and cherry picked onto the 1.9.x branch so it can be included in the OH 2.0.0 release?

9037568:
That's correct, all previous functionality should remain unaffected, but a wide variety of POST body content should now be possible.

My testing has not been exhaustive as the matrix is too large, but I did verify that with the new fix both headers and JSON POST body content work as expected.

I think this is solid enough to be merged and included in the OH 2.0.0 release.

watou:

PR #5009, opened 2017-01-18
[Plugwise] Power measurement improvements

wborn:
This PR improves the Plugwise Circle/Stealth power measurements as follows:
* Increase precision by using `PULSES_PER_KW_SECOND = 468.9385193` which is also used by all other Plugwise implementations on the internet, instead of `PULSE_FACTOR = 2.1324759f` which is: `1/(468.9385193/1000)` but with loss of precision
* Increase precision by using doubles instead of floats in calculations
* Use seconds correction that is part of the `PowerInformationResponseMessage`
* Fix wrong `Energy` date/time values being used in `PowerBufferResponseMessage`
* Corrected log address calculations in `InformationResponseMessage`, `PowerBufferRequestMessage`, `PowerBufferResponseMessage` because the original address calculation would cause the same logs returned multiple times (inspired by [Plugwise-2-pi](https://github.com/SevenW/Plugwise-2-py))
* Make sure Circle/Stealth last hour power consumption is always the last hour (also after startup) and is updated only after calibration took place

Power measurements now exactly match the decimals found in the Plugwise Source database (though with more precision).

### DIFF
--- a/bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/Circle.java
+++ b/bundles/binding/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/Circle.java
@@ -35,8 +35,7 @@ import org.slf4j.LoggerFactory;
  * Circles maintain current energy usage by counting 'pulses' in a one or eight-second interval. Furthermore, they
  * store hourly energy usage as well in a buffer. Each entry in the buffer contains usage for the last 4 full hours
  * of consumption. In order to convert pulses to power (Watt) or KWh you need to apply a formula that uses some
- * calibration
- * information
+ * calibration information.
  *
  * @author Karel Goderis
  * @since 1.1.0
@@ -45,7 +44,9 @@ public class Circle extends PlugwiseDevice {
 
     private static Logger logger = LoggerFactory.getLogger(Circle.class);
 
-    private static final float PULSE_FACTOR = 2.1324759f;
+    private static final double PULSES_PER_KW_SECOND = 468.9385193;
+
+    private static final double PULSES_PER_W_SECOND = (PULSES_PER_KW_SECOND / 1000);
 
     private static final int POWER_STATE_RETRIES = 3;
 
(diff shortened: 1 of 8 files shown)


watou:
Hi @wborn, did you consider and reject using `BigDecimal` since accuracy and precision are factors here?

```java
    private static final BigDecimal PULSES_PER_KW_SECOND = new BigDecimal("468.9385193");

    private static final BigDecimal PULSES_PER_W_SECOND = PULSES_PER_KW_SECOND.movePointLeft(3);
```
etc.

wborn:
I just rewrote the calculations to `BigDecimal`s and found that the percentage error of using doubles in the calculations is: 0.07984247352300347106 %

So I would prefer using `double`s over `BigDecimal`s because:
* the error is small
* the code is easier to understand with doubles
* double calculations use less CPU/memory
* the calculations now have exactly the same results (but with more decimals) as stored in the Plugwise Source database. This also makes it easier to compare and validate the correctness of the Plugwise Binding calculations.

watou:
That sounds like a valuable exercise you went through.  Do you think this PR is ready to merge?

wborn:
Yes I've been using this code without any issues for a week myself now.

Now that the measurements are the same as Plugwise Source, I store the "last hour" kWh values in influxdb and use them in Grafana to draw nice charts. :-)

watou:
Should https://github.com/openhab/openhab1-addons/commit/6ee86c1d3d577cd06fbbe7bf70a411da4e10c375 be cherry picked to the 1.9.x branch so it appears in the 2.0.0 release (I'm guessing yes, but WDYT)?

wborn:
Thanks for merging @watou . Yes it would be nice when it is cherry picked so it also to appears in the 2.0.0 release. I think it will be my last Plugwise 1.9.x contribution since it is more or less feature complete for me now. My next contribution might be a real openHAB 2.0 Plugwise Binding. ;-)

watou:
> My next contribution might be a real openHAB 2.0 Plugwise Binding. ;-)

That will be very well received!  Thanks so much for your contributions.  I will make sure this PR gets into the 2.0.0 path.

watou:
Cherry picked onto 1.9.x branch.

PR #5011, opened 2017-01-19
removed myopenhab and openhabcloud bundles as both are not buildable in 1.9 anymore

kaikreuzer:


kaikreuzer:
@watou this should also be picked for the 1.9.x branch, although it will have no impact on the build results.

watou:
Thanks, will do.

kaikreuzer:
So don't forget to merge, I'll come up with another important PR afterwards ;-)

watou:
picked onto 1.9.x

PR #5026, opened 2017-01-23
insteonplmbinding: added support for venstar and smoke bridge

berndpfrommer:
insteonplm binding: added support for venstar thermostat and smoke bridge devices


watou:
Thanks @berndpfrommer!  I merged this and then realized that [this file](https://github.com/openhab/openhab1-addons/blob/master/features/openhab-addons-external/src/main/resources/conf/insteonplm.cfg#L12) might also need the same change you made to openhab_default.cfg.

berndpfrommer:
Yes, please commit the change as you suggested, thanks!

PR #5031, opened 2017-01-23
Fix some errors in SNMP binding

J-N-K:
- NPE when address invalid (thows BindingConfigException instead)
- string concatenation in logger
- updated snmp4j from 2.2.2 to 2.5.4 (many fixes in library)


9037568:
Thanks, @J-N-K !

watou:
Since this is a bugfix, should it appear in the next 1.9 maintenance release so users have access it to it sooner?

PR #5036, opened 2017-01-26
bugfix: padding of temperature control hex string

bbesser:
The FHT binding would not send correct control commands for low temperature values.

(
Presumably, this bug was hidden for quite a long time, since virtually nobody sets low temperatures to heat their homes :-)
)

**Detailed bug description**
Temperature control command are hex strings following the "FHT80b syntax: THHHHCCAA" given in http://culfw.de/commandref.html#cmd_T. (Note: Temp commands can only be sent to FHT80b devices, i.e. the buggy method setDesiredTemperature(...) will not be called for commands requiring other syntax.)

Low temp values were converted to one-digit hex strings, whereas two-digit hex-strings are required, cf. the trailing AA part of the command string.

**Question**
This bug also exists in the 1.9 branch and in master. How to proceed? (It seems that the FHT binding does not yet exist in openhab2-addons, hence openhab2-addons can be ignored for now.)


watou:
Please re-submit this PR against the master branch, as there are no 1.8 maintenance releases planned.  Since it's a bug on master and 1.9.x, I will cherry pick your fix from master onto 1.9.x so it will appear in all future releases.  Thanks!

PR #5037, opened 2017-01-26
bugfix: padding of FHT temperature control hex string

bbesser:
The FHT binding would not send correct control commands for low temperature values.

(
Presumably, this bug was hidden for quite a long time, since virtually nobody sets low temperatures to heat their homes :-)
)

Detailed bug description
Temperature control command are hex strings following the "FHT80b syntax: THHHHCCAA" given in http://culfw.de/commandref.html#cmd_T. (Note: Temp commands can only be sent to FHT80b devices, i.e. the buggy method setDesiredTemperature(...) will not be called for commands requiring other syntax.)

Low temp values were converted to one-digit hex strings, whereas two-digit hex-strings are required, cf. the trailing AA part of the command string.


watou:

PR #5039, opened 2017-01-26
[gcal] Ensure number updates and commands survive round trip

watou:
The Google Calendar Scheduler is instructed to send a whole number for a PercentType like 50, but was sending 50.0 because the StreamTokenizer was identifying a number (TT_NUMBER) which means it gives a `double`, whose string representation of whole numbers ends in ".0".  This change causes TT_NUMBER tokens to never happen, so numbers are always their string representation, which can be dealt with

[Forum discussion](https://community.openhab.org/t/presence-simulation-with-google-calendar-scheduler/19281/19?u=watou)


watou:

PR #5049, opened 2017-01-29
[ComfoAir] Optimize serial connection

csowada:
- Add RXTX CPU workaround
- Change Thread to SchedulerExecutorService
- Remove flush() on send, can cause blocking forever

### DIFF
--- a/bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirConnector.java
+++ b/bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirConnector.java
@@ -53,7 +53,7 @@ public class ComfoAirConnector {
 
     /**
      * Open and initialize a serial port.
-     * 
+     *
      * @param portName
      *            e.g. /dev/ttyS0
      * @param listener
@@ -132,7 +132,7 @@ public class ComfoAirConnector {
 
     /**
      * Prepare a command for sending using the serial port.
-     * 
+     *
      * @param command
      * @return reply byte values
      */
@@ -278,7 +278,7 @@ public class ComfoAirConnector {
     /**
      * Generate the byte sequence for sending to ComfoAir (incl. START & END
      * sequence and checksum).
-     * 
+     *
      * @param command
      * @param data
      * @return response byte value block with cmd, data and checksum
@@ -319,7 +319,7 @@ public class ComfoAirConnector {
 
     /**
      * Calculates a checksum for a command block (cmd, data and checksum).
-     * 
+     *
      * @param block
      * @return checksum byte value
      */
@@ -341,7 +341,7 @@ public class ComfoAirConnector {
 
     /**
      * Cleanup a commandblock from quoted 0x07 characters.
-     * 
+     *
      * @param processBuffer
      * @return the 0x07 cleaned byte values
      */
@@ -367,7 +367,7 @@ public class ComfoAirConnector {
 
     /**
      * Escape special 0x07 character.
-     * 
+     *
      * @param cleanedBuffer
      * @return escaped byte value array
      */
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirConnector.java
```diff
@@ -390,7 +404,7 @@ private boolean send(byte[] request) {
 
         try {
             outputStream.write(request);
-            outputStream.flush();
+            // remove outputStream.flush(), can hang forever
```
watou: please remove commented out code

bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirConnector.java
```diff
@@ -423,4 +437,15 @@ private String dumpData(byte[] data) {
...
+    private class CPUWorkaroundThread implements SerialPortEventListener {
+        @Override
+        public void serialEvent(SerialPortEvent event) {
+            try {
+                logger.trace("RXTX library CPU load workaround, sleep forever");
+                Thread.sleep(Long.MAX_VALUE);
+            } catch (InterruptedException e) {
+            }
+        }
+    }
```
watou: I'm concerned about this workaround, only because I'm unaware of other bindings that use the serial port having to implement a workaround of this sort.  Can you offer how this binding's use of the serial port is necessarily different from, say, how the Z-Wave binding uses it?

csowada: This is common workaround for the buggy rxtx library. Zwave and eBUS binding uses the same workaround. I've attached the source in zwave binding.

https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java#L348

https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.zwave/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java#L1468

bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirConnector.java
csowada: No problem, I will remove the commented out code.

bundles/binding/org.openhab.binding.comfoair/src/main/java/org/openhab/binding/comfoair/handling/ComfoAirConnector.java
big-col: @csowada do you think my issue [https://github.com/openhab/openhab2-addons/issues/1777](url)  could be related to the problem you are fixing here?  my issue is around Powermax binding, but it sounds similar to me?

csowada: No, I don't think so.

paulianttila: This is common workaround. See more details from #1868.

watou: Thanks very much for educating me about the workaround. 😄

watou:

PR #5053, opened 2017-01-29
zwave updates

dmize:
Added Zooz ZEN06 Smart Plug
Updated paramter 43 of aeon/zw056.xml to include unknown value


dmize:
Thanks @cdjackson

PR #5058, opened 2017-02-02
Fixed incorrect gcal test

watou:
Test was expecting the incorrect ".0" in the parse result.  See #5039


watou:
cherry picked onto 1.9.x

PR #5060, opened 2017-02-03
[Cups] Upgrade cups binding to the newest cups4j library

9037568:
This library update has added support for slf4j logging within the library (though it doesn't appear to be used yet).

There's also one small multibyte character bug fix.


watou:
There is still an 0.6.3 in the `.classpath` file. 😄

9037568:
Oops :)

watou:
😄  Does this fix a bug and should be in 1.9.1, or is this just good maintenance and can wait for 1.10.0?

9037568:
No, no bug fixing involved. It can wait for 1.10.0.

PR #5062, opened 2017-02-03
Intro paragraph cleanup; add Sapp binding README.md contents.

watou:


ThomDietrich:
Since you made me aware of this PR: While you are at it, do you think it's worth rewriting all intro sentences to not start with "This binding is for" or "The Asterisk binding is used..."? Seems utterly redundant, wdyt?

Thinking of the docs overview table, I think the intro paragraph should shortly describe what the binding is aimed at, ideally with one or more links to the related product or project... 
The CoolMasterNet binding is **imho** a good example:

> The CoolMasterNet binding is used to control [CoolMasterNet HVAC bridge devices](https://coolautomation.com/products/coolmasternet/), using the ‘ASCII I/F’ plaintext TCP control protocol.

watou:
I've reduced quite a lot of "fluff" from the wiki versions of these docs, and further reduction (wihout losing important information) is absolutely a good idea.  I'm in trouble with the partner for all the hours I've spent so far on the 1.x documentation, so I would like to merge this as is, and invite you to polish these README.md files further (obviously respecting the intent of the original authors, etc.).

ThomDietrich:
Sure thing! As always, we can only do as much for an open source project.
I'm in the same corner as you right now. Dunno when I'll find the time to do the said above for openhab2-addons.

LGTM

PR #5065, opened 2017-02-04
[MQTT] Allow longer Client ID values

### LINKED ISSUE #5055: mqtt clientID is not taken from the opehab.cfg but automatically generated
I defined clientID for mqtt as this is required by external mqtt broker:

mqtt:cayenne.url=tcp://mqtt.mydevices.com:1883
mqtt:cayenne.clientid=01155d90-e731-11e6-a971-4db18559717d

tcp connection works fine. it is connected but not authorized. Cayenne requires valid clientID to be sent. 

In parallel I configured Chromium addon MQTT lens to test the same connection. I used the same  and authentication works without any problems. 

I took network trace to see what is the difference in sent requests.
It seems to MQTT field is set as: 
Client ID: owca.1485813787709
instead of the one configured in openhab.cfg

in Chrome browser addon MQTT Lens in network trace correct one is set and field takes values as below: 
Client ID: 01155d90-e731-11e6-a971-4db18559717d

## Expected Behavior
openhab is taking mqtt clientID that is set in openhab.cfg 

## Current Behavior
openhab ignores mqtt.broker.clientID setting and generates random one (<username>.<numbers>)

1. set mqtt clientID in configuration
2. sniff for traffic when connecting to the broker
3. check what is sent
4.

## Context
I tried to connect openhab to the mydevices.com service over mqtt 

## Your Environment

* Version used: openhab 1.8.3
* Environment name and version 
* Operating System and version: Sabayon Linux amd64 16.08

9037568:
Update the MQTT binding to allow Client ID values longer than 23 characters.

Also minor enhancements to configuration files (README.md, etc.)

Fixes #5055.


bundles/binding/org.openhab.binding.mqtt/README.md
```diff
@@ -21,18 +25,19 @@ This binding can be configured in the file `services/mqtt.cfg`.
...
-| `<broker>`.pwd |      | Password to authenticate with the broker. |
-| `<broker>`.qos | 0    | Set the quality of service level for sending messages to this broker. Possible values are 0 (Deliver at most once),1 (Deliver at least once) or 2 (Deliver exactly once). |
-| `<broker>`.retain | false | true or false. Defines if the broker should retain the messages sent to it. |
-| `<broker>`.async | true | true or false. Defines if messages are published asynchronously or synchronously. |
-| `<broker>`.keepAlive | 60 | Integer. Defines the keep alive interval in seconds. |
+| Property             | Default | Required | Description                                                 |
+|----------------------|---------|:--------:|-------------------------------------------------------------|
+| `<broker>`.url       |         | Yes      | URL to the MQTT broker, e.g. tcp://localhost:1883 or ssl://localhost:8883 |
+| `<broker>`.clientId  |         | Yes/No   | Client ID to use when connecting to the broker. If not provided a random default is generated. Usually restricted to 23 characters in length. (see the `allowLongerClientIds` setting) |
```
watou: I think Required should be "No" or "depends on broker" or something clearer.

bundles/io/org.openhab.io.transport.mqtt/src/main/java/org/openhab/io/transport/mqtt/MqttService.java
```diff
@@ -91,7 +91,13 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
             } else if (property.equals("async")) {
                 conn.setAsync(Boolean.parseBoolean(value));
             } else if (property.equals("clientId")) {
+                if (value.length() > 65535) {
+                    logger.warn("clientId must be less than 65536 characters long");
+                    throw new ConfigurationException("clientId", "clientId must be less than 65536 characters long");
+                }
                 conn.setClientId(value);
+            } else if (property.equals("allowLongerClientIds")) {
+                conn.setAllowLongerClientIds(Boolean.parseBoolean(value));
```
watou: Doesn't this depend on the config property `allowLongerClientIds` having been read before `clientId` for the logic in `MqttBrokerConnection` to correctly enforce the length of `clientId`?

bundles/binding/org.openhab.binding.mqtt/README.md
9037568: I think Yes* is a better description, since it's required but has a random default if not specified by the user (see latest push). WDYT?

bundles/io/org.openhab.io.transport.mqtt/src/main/java/org/openhab/io/transport/mqtt/MqttService.java
9037568: I'm not clear on this one. Are you saying the broker connections can start up before the config has been fully read in?

watou: I'm saying that I read the code that 

```
allowLongerClientIds=true
clientId=0123456789012345678901234
```

will behave differently from

```
clientId=0123456789012345678901234
allowLongerClientIds=true
```

which should not be -- both should act in the same way, doing what the user intends.

bundles/binding/org.openhab.binding.mqtt/README.md
```diff
@@ -28,7 +28,7 @@ In order to consume or publish messages to an MQTT broker, you need to define al
 | Property             | Default | Required | Description                                                 |
 |----------------------|---------|:--------:|-------------------------------------------------------------|
 | `<broker>`.url       |         | Yes      | URL to the MQTT broker, e.g. tcp://localhost:1883 or ssl://localhost:8883 |
-| `<broker>`.clientId  |         | Yes/No   | Client ID to use when connecting to the broker. If not provided a random default is generated. Usually restricted to 23 characters in length. (see the `allowLongerClientIds` setting) |
+| `<broker>`.clientId  | random  | Yes*     | Client ID to use when connecting to the broker. If not provided a random default is generated. Usually restricted to 23 characters in length. (see the `allowLongerClientIds` setting) |
```
watou: What does Yes* mean?  With a "random default is generated," I would think No* would be more appropriate, but with "*" given below the table to explain.  Marking something with an asterisk but not explaining below is one of my pet peeves! 😄

bundles/io/org.openhab.io.transport.mqtt/src/main/java/org/openhab/io/transport/mqtt/MqttService.java
9037568: I've been attempting to test this out, but I'm having trouble getting the updated MQTT IO Transport to be used by the platform.  I tried dropping the jar into the addons directory of the 1.8.3 install, but that didn't work.

Is it that the 1.9.0 version can't be used with the 1.8.x platform?

watou: I think I know why -- the IO transport bundle is part of the core set of bundles, so you would have to find and remove the built-in one, and drop its replacement in the addons folder like you do with binding JAR files.

9037568: That was the problem. I had tried removing the built-in jar, but I put the new jar back in the same directory instead of in the addons folder. Thanks!

9037568: I have now tried various orderings of the keys in openhab.cfg, and I have yet to find any inconsistent results. No matter whether the `allowLongerClientIds` key is specified before or after the `clientId` key, we still get the proper behaviour when the broker starts up:

21:59:28.001 [INFO ] [o.i.t.m.i.MqttBrokerConnection:112  ] - Starting MQTT broker connection 'ruby'
21:59:28.003 [DEBUG] [o.i.t.m.i.MqttBrokerConnection:322  ] - Creating new client for 'tcp://io.adafruit.com:1883' using id 'ThisIsAReallyLongClientId' and file store 'C:\Users\me\AppData\Local\Temp\/ruby'

Does this fully address your concern?

PR #5070, opened 2017-02-06
 bugfix: FHT time update message

bbesser:
Bug: FHT80b would not receive time update, although the openhab log tells that messages have been sent.

Explanation: culfw (i.e. the firmware running on the required rf-device) silently dropped those messages, since they were malformed.

Bugfix: FHT control messages for FHT80b start with the letter T, not with F, see http://culfw.de/commandref.html#cmd_T.

Note: Please cherry pick to 1.9


watou:
cherry picked onto 1.9.x

PR #5077, opened 2017-02-09
add tested IEC 62056-21 meter 1.x version to OH2

kreutpet:
added 1.x binding according instruction 
http://docs.openhab.org/developers/development/compatibilitylayer.html#how-to-add-a-successfully-tested-1x-add-on-to-the-distribution

it has been discussed here https://community.openhab.org/t/iec-62056-21-meter-binding/18382/11

PR #5079, opened 2017-02-11
[IEC] Code cleanup

9037568:
Update logging statements to use parameterized logging.
Convert config gets to use Objects.toString().
Eclipse formatting.

### DIFF
--- a/bundles/binding/org.openhab.binding.iec6205621meter/src/main/java/org/openhab/binding/iec6205621meter/internal/Iec6205621MeterBinding.java
+++ b/bundles/binding/org.openhab.binding.iec6205621meter/src/main/java/org/openhab/binding/iec6205621meter/internal/Iec6205621MeterBinding.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -93,7 +94,6 @@ public class Iec6205621MeterBinding extends AbstractActiveBinding<Iec6205621Mete
     }
 
     private final Meter createIec6205621MeterConfig(String name, MeterConfig config) {
-
         Meter reader = null;
         reader = new Meter(name, config);
         return reader;
@@ -116,7 +116,7 @@ public class Iec6205621MeterBinding extends AbstractActiveBinding<Iec6205621Mete
                         Map<String, DataSet> dataSets;
                         if ((dataSets = cache.get(meterName)) == null) {
                             if (logger.isDebugEnabled()) {
-                                logger.debug("Read meter: " + meterName + "; " + reader.getConfig().getSerialPort());
+                                logger.debug("Read meter: {}; ", meterName, reader.getConfig().getSerialPort());
                             }
                             dataSets = reader.read();
                             cache.put(meterName, dataSets);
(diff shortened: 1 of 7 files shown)


watou:
I know you didn't touch [line 180](https://github.com/9037568/openhab/blob/ab3bd35839dae96dc28f962d45c6cbc3d3659b0c/bundles/binding/org.openhab.binding.iec6205621meter/src/main/java/org/openhab/binding/iec6205621meter/internal/Iec6205621MeterBinding.java#L180), but I don't think the `logger.warn` is needed, since it's nothing the user should do anything about.  The file `OSGI-INF/binding.xml` should have something like

```
 configuration-pid="org.openhab.iec6205621meter" configuration-policy="require"
```

so the binding isn't started without a configuration, and config being `null` can be quietly ignored without the logger.warn message.  [This](http://blog.vogella.com/2016/09/26/configuring-osgi-declarative-services/) is interesting reading.

9037568:
I can certainly change that. I built a new version with the change, but I'm not sure it's doing what you expect.

With no configuration specified, we don't hit the line 180 clause you referenced.  But we also don't get any error logged to indicate to the user that they've misconfigured the binding.  Is that acceptable?

We still get [this](https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.iec6205621meter/src/main/java/org/openhab/binding/iec6205621meter/internal/Iec6205621MeterActivator.java#L34), though:
>    07:46:46.233 [DEBUG] [b.i.i.Iec6205621MeterActivator:34   ] - iec6205621meter  binding has been started.

With a configuration specified, we get that line plus these 2:
>   07:46:46.249 [INFO ] [o.b.i.i.Iec6205621MeterBinding:207  ] - Created reader <meter-id1> with  IEC 62056-21Meter DeviceConfig [serialPort=/dev/ttyS0, initMessage=null, baudRateChangeDelay=0, echoHandling=true]!
   07:46:46.249 [INFO ] [.service.AbstractActiveService:169  ] - iec6205621meter Refresh Service has been started

If this is all ok, I'll push the changes up.

watou:
> we also don't get any error logged to indicate to the user that they've misconfigured the binding. Is that acceptable?

I think so, since there's no point starting the service if there is no configuration specified.  When the user saves a `services/iec6205621meter.cfg` with at least one config property, the binding should start and try to use the provided config.

> We still get this, though:
> > 07:46:46.233 [DEBUG] [b.i.i.Iec6205621MeterActivator:34 ] - iec6205621meter binding has been started.

That's expected, because the bundle is activated as it should be.  The binding service, however, is not started until it has a configuration.

So I think your change is behaving correctly and should be added to this PR.

watou:
And with the above mentioned commit, this PR should be good to merge.  Should this be on 1.10.0 or 1.9.1 milestone?

9037568:
Your call on the milestone. It's a minor change.

bundles/binding/org.openhab.binding.iec6205621meter/src/main/java/org/openhab/binding/iec6205621meter/internal/Iec6205621MeterBinding.java
```diff
@@ -177,11 +177,6 @@ protected void removeBindingProvider(Iec6205621MeterBindingProvider bindingProvi
     @Override
     public void updated(Dictionary<String, ?> config) throws ConfigurationException {
 
-        if (config == null || config.isEmpty()) {
```
watou: I'm not positive that `updated` won't be called with a `null` for the `config` parameter, for whatever reason.  It's safest to simply return in that case.

9037568: Rather pointless to claim the config is required if it can be passed in as null isn't it? I hate OSGI.

watou:
> Your call on the milestone. It's a minor change.

Since it's not really a fix of a bug but rather adding a Karaf feature and binding configuration, I think it makes sense to target 1.10.0.  That being said, it would be sad if it took a long time for users to benefit from this PR and #5077.  @kaikreuzer, WDYT?

bundles/binding/org.openhab.binding.iec6205621meter/src/main/java/org/openhab/binding/iec6205621meter/internal/Iec6205621MeterBinding.java
watou: > Rather pointless to claim the config is required if it can be passed in as null isn't it?

Yes indeed.  My advice to leave the null check might be more superstition based on fuzzy past observations than how any OSGi SCR mechanisms actually work when configured properly.  Unless it's completely without merit, this kind of "defensive programming" can just avoid stupid frustration and time wasting later.

PR #5097, opened 2017-02-16
[modbus] test for int32_swap valuetype

ssalonen:
Introducing test case for int32_swap. After this there is at least one test for the "swapped" data types of modbus binding.


watou:
Thanks Sami!

PR #5105, opened 2017-02-20
[Dropbox]: Restore original keys to config files

9037568:
Update the Dropbox config files to restore the original appkey keys.

@watou, could you please cherry pick to 1.9x? Thanks!


watou:
Cherry picked to 1.9.x.

Please add a line to features/openhab-addons-external/pom.xml so dropbox.cfg makes it into the distro and installing the feature won't fail.  If you merge that, I will cherry it pick to 1.9.x too.

9037568:
Updated in #5106.

PR #5114, opened 2017-02-24
[Squeeze] Fix playing time error

### LINKED ISSUE #3542: Squeeze Server uses PercentType for playing time
Here's the error:

```
Exception in thread "Squeeze Server Listener" java.lang.IllegalArgumentException: Value must be between 0 and 100
    at org.openhab.core.library.types.PercentType.validateValue(PercentType.java:48)
    at org.openhab.core.library.types.PercentType.<init>(PercentType.java:33)
    at org.openhab.binding.squeezebox.internal.SqueezeboxBinding.numberChangeEvent(SqueezeboxBinding.java:245)
    at org.openhab.binding.squeezebox.internal.SqueezeboxBinding.currentPlayingTimeEvent(SqueezeboxBinding.java:173)
    at org.openhab.io.squeezeserver.SqueezePlayer.fireCurrentPlayingTimeEvent(SqueezePlayer.java:400)
    at org.openhab.io.squeezeserver.SqueezePlayer.setCurrentPlayingTime(SqueezePlayer.java:225)
    at org.openhab.io.squeezeserver.SqueezeServer$SqueezeServerListener.handleStatusMessage(SqueezeServer.java:703)
    at org.openhab.io.squeezeserver.SqueezeServer$SqueezeServerListener.handlePlayerUpdate(SqueezeServer.java:654)
    at org.openhab.io.squeezeserver.SqueezeServer$SqueezeServerListener.run(SqueezeServer.java:559)
```

This is an issue because of (in Binding)

```
public void currentPlayingTimeEvent(PlayerEvent event) {
    numberChangeEvent(event.getPlayerId(), CommandType.PLAYTIME, event.getPlayer().getCurrentPlayingTime());
}
```

...

```
private void numberChangeEvent(String playerId, CommandType commandType, int newState) {
    logger.debug("SqueezePlayer " + playerId + " -> " + commandType.getCommand() + ": " + Integer.toString(newState));
    for (String itemName : getItemNames(playerId, commandType)) {
        eventPublisher.postUpdate(itemName, new PercentType(newState));
    }
}
```

Which is no bueno when the "newState" is the number of seconds in a song (often over 100). My suggestion would be to change it to a string type.
##

9037568:
Fix the playing time to use a number and not a percent.

Rename the numberChangeEvent() method to percentChangeEvent().
Add a new numberChangeEvent() method that uses DecimalType instead of PercentType.

Fixes #3542.

### DIFF
--- a/bundles/binding/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/SqueezeboxBinding.java
+++ b/bundles/binding/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/SqueezeboxBinding.java
@@ -154,7 +154,7 @@ public class SqueezeboxBinding extends AbstractBinding<SqueezeboxBindingProvider
                         logger.warn("Unsupported command type '{}'", bindingConfig.getCommandType());
                 }
             } catch (Exception e) {
-                logger.warn("Error executing command type '" + bindingConfig.getCommandType() + "'", e);
+                logger.warn("Error executing command type '{}'", bindingConfig.getCommandType(), e);
             }
         }
     }
@@ -171,12 +171,12 @@ public class SqueezeboxBinding extends AbstractBinding<SqueezeboxBindingProvider
 
     @Override
     public void volumeChangeEvent(PlayerEvent event) {
-        numberChangeEvent(event.getPlayerId(), CommandType.VOLUME, event.getPlayer().getVolume());
+        percentChangeEvent(event.getPlayerId(), CommandType.VOLUME, event.getPlayer().getVolume());
     }
 
     @Override
     public void currentPlaylistIndexEvent(PlayerEvent event) {
-        numberChangeEvent(event.getPlayerId(), CommandType.CURRTRACK, event.getPlayer().getCurrentPlaylistIndex());
+        percentChangeEvent(event.getPlayerId(), CommandType.CURRTRACK, event.getPlayer().getCurrentPlaylistIndex());
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/SqueezeboxBinding.java
```diff
@@ -171,12 +171,12 @@ public void muteChangeEvent(PlayerEvent event) {
...
     @Override
     public void volumeChangeEvent(PlayerEvent event) {
-        numberChangeEvent(event.getPlayerId(), CommandType.VOLUME, event.getPlayer().getVolume());
+        percentChangeEvent(event.getPlayerId(), CommandType.VOLUME, event.getPlayer().getVolume());
     }
 
     @Override
     public void currentPlaylistIndexEvent(PlayerEvent event) {
-        numberChangeEvent(event.getPlayerId(), CommandType.CURRTRACK, event.getPlayer().getCurrentPlaylistIndex());
+        percentChangeEvent(event.getPlayerId(), CommandType.CURRTRACK, event.getPlayer().getCurrentPlaylistIndex());
```
watou: I assume you are using PercentType here because it was before and you just renamed the method to be clear, but would this (and others here) be a case where DecimalType ([parent class of PercentType](https://github.com/openhab/openhab1-addons/blob/1.8/bundles/core/org.openhab.core.library/src/main/java/org/openhab/core/library/types/PercentType.java#L20)) is really the more appropriate State?

9037568: You're probably right. I was trying to keep the changes to a minimum. Do you think it's worth changing?

watou: I don't know this binding but it looks like PercentType wouldn't apply here, but instead an unrestricted DecimalType makes more sense.

9037568:
Agreed. I'll change these.

9037568:
Cherry-picked to 1.9.1

PR #5117, opened 2017-02-26
[CalDAV] Various configuration fixes

### LINKED ISSUE #5116: [Caldav] No warnings in log for missing required configuration settings
## Expected Behavior

If the binding is not configured with the required settings, warnings should be generated in openhab.log.

## Current Behavior

If the binding is not configured with the required settings, no warnings are generated in openhab.log.

## Possible Solution

Add warnings.

1. Add the caldav jars to the addons directory
2. Do not configure caldav settings
3. Start OH

## Context

Attempting to set up caldav binding for the first time.

## Your Environment

Version: OH 1.8.3 with 1.8.3 bindings
(problem exists also in 1.9.* and 1.10.* bindings)

9037568:
If config is null, log a warning and return.
Switch to using Objects.toString() for reading config values.
Drop the thrown exceptions for missing username/password, since those values are not required.
Add a warning to the log output if the url is not specified in the configuration.
Use the default TimeZone if the one specified in configuration is invalid.
Misc. fixes in openhab_default.cfg and caldavio.cfg.

Fixes #5116

### DIFF
--- a/bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/CalDavLoaderImpl.java
+++ b/bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/CalDavLoaderImpl.java
@@ -143,12 +143,12 @@ public class CalDavLoaderImpl extends AbstractActiveService implements ManagedSe
     @Override
     public void updated(Dictionary<String, ?> config) throws ConfigurationException {
         if (config == null) {
-            log.warn("No configuration exists for CalDAV IO.");
+            log.debug("Update was called with a null configuration for CalDAV IO.");
             return;
         }
 
+        log.debug("Update was called for CalDAV IO.");
         CompatibilityHints.setHintEnabled(CompatibilityHints.KEY_RELAXED_PARSING, true);
-
         Map<String, CalDavConfig> configMap = new HashMap<String, CalDavConfig>();
 
         Enumeration<String> iter = config.keys();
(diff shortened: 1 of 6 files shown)


bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/CalDavLoaderImpl.java
```diff
@@ -140,96 +142,100 @@ public void shutdown() {
...
-                if (configMap.get(id).getUsername() == null) {
-                    throw new ConfigurationException(PROP_USERNAME, PROP_USERNAME + " must be set");
+            log.trace("processing configuration parameter: {}", key);
+            if (key.equals(PROP_TIMEZONE)) {
+                String newTimeZone = Objects.toString(config.get(key), null);
+                log.debug("overriding default timezone {} with {}", defaultTimeZone, newTimeZone);
+                defaultTimeZone = DateTimeZone.forID(newTimeZone);
+                if (defaultTimeZone == null) {
+                    log.warn("Invalid timezone value: {}", newTimeZone);
+                    throw new ConfigurationException("CalDAV IO", "Invalid timezone value: " + newTimeZone);
```
9037568: Since we have/use a default timezone already, should we really throw an exception if the timezone in the configuration file isn't valid? Wouldn't it be better to log a warning and proceed to use the default?

bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/CalDavLoaderImpl.java
```diff
@@ -140,96 +142,100 @@ public void shutdown() {
...
                 }
-                if (configMap.get(id).getPassword() == null) {
-                    throw new ConfigurationException(PROP_PASSWORD, PROP_PASSWORD + " must be set");
+                log.debug("found timeZone: {}", defaultTimeZone);
+                continue;
+            }
+            String[] keys = key.split(":");
+            if (keys.length != 2) {
+                log.warn("Unable to parse configuration parameter: {}", key);
+                throw new ConfigurationException("CalDAV IO", "Unable to parse configuration parameter: "+key);
```
9037568: Here again it seems like we should not throw an exception.

bundles/io/org.openhab.io.caldav/src/main/java/org/openhab/io/caldav/internal/CalDavLoaderImpl.java
9037568: And now I see that I've misconfigured this throw statement. So if we keep the throw, it'll have to be fixed up...

9037568: Updated to use the default TimeZone if the one specified in configuration is invalid.

PR #5126, opened 2017-02-28
[Z-Wave] Database update: Fibaro Wall Plug (Z-Wave+)

bodiroga:
* Added Fibaro Wall Plug (Z-Wave+) (FGWPE102)


cdjackson:
If I remember correctly, there are two versions of this device - it might be worth adding them both as I think they are quite different (or maybe I'm remembering incorrectly and I can't check the database right now).

bodiroga:
Hi Chris!

You are totally right, I have checked the database and it seems that the "0602:1001" device has 2 variants depending of its firmware version. The 1st version, the one with firmware version from 2.1 to 2.5, has the same configuration parameters as the previous (non Z-Wave+) Wall Plug. The 2nd one, the version that I own and the one to which the PR is related to, is the new model and the one with different and reordered configuration parameters.

I will update the PR tomorrow and include the relevant information about the other version.

Many thanks for your support and sorry for the confusion.

Aitor

cdjackson:
No problem  - thanks Aitor.

bodiroga:
Done!

cdjackson:
Thanks

PR #5141, opened 2017-03-13
Telegram.sendTelegram() now returns "false" if an exception is thrown.

### LINKED ISSUE #5140: Add boolean return value to Telegram.sendTelegram() after an exception
The action Telegram.sendTelegram() does not return `false` if an exception occurred (HttpException, IOException). But it does return `false` if i.e. it receives a response from the server which is not equal to "200 OK".

The action should be extended such that it also returns `false` after an exception.

planetrenner-martin:
Closes #5140

Signed-off-by: Martin Renner <martin+github@planet-renner.net>


9037568:
Thanks, @planetrenner-martin.

PR #5150, opened 2017-03-24
[cardio2e] Initial contribution

magd1978:
First complete stable version.
To do: README.md, JavaDoc.

PR #5151, opened 2017-03-24
[cardio2e] Added config file and registered in POM.xml

magd1978:


hakan42:
Depends on #5150

9037568:
These changes need to be merged with the PR in #5150.

PR #5152, opened 2017-03-24
[caldav] Fix race condition causing NPE

gedejong:
This fixes a problem described in https://community.openhab.org/t/caldav-binding-setup-issue/21651 caused by a race-condition of the binding not having been initialized at the right time.

Should find some time to do a new openhab 2 port, but for now we're fine.


9037568:
Thanks, @gedejong !

PR #5054, opened 2017-01-29
insteonplm: add support for smoke bridge (this time for real)

berndpfrommer:
Apologies, I messed up on PR #5026: it did not contain the smoke bridge code.
This PR should fix it.

PR #5169, opened 2017-04-22
insteonplm: fixed bug introduced with PR #5054

berndpfrommer:
Tested: the thermostat works with this fix. Whether it still works for the smoke bridge I don't know since I don't own the device.


9037568:
Thanks, @berndpfrommer !

PR #5176, opened 2017-04-27
Update GPIOGenericBindingProvider.java

jokercho:
I'm proposing additional property in the GPIO binding item configuration which defines the initial value of the GPIO pin. With the old implementation it was always defined initially to LOW (0). There are cases when it is necessary the initial value to be HIGH (1). There is support for this in org.openhab.io.gpio but wasn't implemented in the configuration part of the GPIO binding.

### DIFF
--- a/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOGenericBindingProvider.java
+++ b/bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOGenericBindingProvider.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
  * <p>
  * Allowed item types in the configuration are "Switch" and "Contact".
  * "Switch" is used for output pins, "Contact" - input pins.
+ * 
  * </p>
  *
  * <p>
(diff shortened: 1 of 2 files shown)


bundles/binding/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/GPIOGenericBindingProvider.java
```diff
@@ -176,6 +178,17 @@ public void processBindingConfiguration(String context, Item item, String bindin
                     throw new BindingConfigParseException("Unsupported value for activelow (" + value
                             + ") in configuration string '" + bindingConfig + "'");
                 }
+            } else if (key.compareToIgnoreCase("initialValue") == 0) {
+                if (value.compareToIgnoreCase("high") == 0) {
+                    config.direction = GPIOPin.DIRECTION_OUT_HIGH;
+                } else if (value.compareToIgnoreCase("low") == 0) {
+                    config.direction = GPIOPin.DIRECTION_OUT_LOW;
+                } else {
+                    logger.error("Unsupported value for initialValue (" + value + ") in configuration string '"
```
9037568: Need to use parameterized logging instead of string concatenation.

jokercho: I tried to make changes with minimal modifications. I just copied logging from other parameters and didn't change the way it is used. If it is necessary I'll do this but then only this modification will be with parameterized logging and will differ from the other places in this class.

9037568:
Thanks, @jokercho !

PR #5184, opened 2017-05-03
[Serial binding] OH2 Compatibility issue with RegexTransformation service

marekhalmo:
When Serial binding is used in OH2 there is no compatible interface to fill the regex transformation to and therefore the Regex functionality is absent if serial binding is used in OH2

**Explanation**
This is because OH2 [RegexTransformationService](https://github.com/eclipse/smarthome/tree/master/extensions/transform/org.eclipse.smarthome.transform.regex) implements interface org.eclipse.smarthome.core.transform.TransformationService but Serial binding [requires implementation of interface org.openhab.core.transform.TransformationService](https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.serial/OSGI-INF/serialbinding.xml) available only in OH1

Because there is no way of including org.eclipse.smarthome.core.transform.TransformationService in the OH1 addons and OH2 will never have org.openhab.core.transform.TransformationService regex implementation the use of transformation service has to be removed and [substituted with local code](https://github.com/marekhalmo/openhab/blob/30d13f1c4913b08e0a90a1149051d357c69f7850/bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java) which is a slightly modified [RegexTransformationService](https://github.com/eclipse/smarthome/tree/master/extensions/transform/org.eclipse.smarthome.transform.regex) 
This has also the benefit of REGEX functionality working even without the Regex transformation installed

**Note**
The OH2 RegexTransformationService can't fully cover the required cases for serial devices as serial device with fast communication can send multiple commands to buffer.

_Example_
Device sends  temperature in format "temp:%C" (e.g. "temp:15C\n")
The corresponding regex would be: REGEX(temp:([0-9]{1,3}C)
If device sends temperature too fast, the Serial binding would receive "temp:15C\ntemp:16C\n"
in this case the original regex won't match the regex and therefore won't update the item at all. 
This way both values are ignored!

Original issue and discussion can be found [here](https://github.com/openhab/openhab-core/issues/128)

This patch fixes **both** issues

A jar for everybody to test can be found [here](http://www.maniacs.sk/oh/org.openhab.binding.serial-1.10.0-SNAPSHOT.jar)


bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,130 @@
+/**
+ * Copyright (c) 2010-2016 by the respective copyright holders.
```
9037568: 2017

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,130 @@
...
+
+import org.openhab.core.transform.TransformationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Pattern matcher to match and cache patterns
+ *
+ * @author Marek Halmo
+ * @see RegexPatternMatcher#getMatches getMatches()
```
9037568: Add an "@since 1.10.0" below "@author"

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return
+     */
+    private static Pattern getPattern(String regExpression) {
+        cache(regExpression);
+        return patternCache.get(regExpression);
+    }
+
+    /**
+     * Returns true if given regExpression is a substitution pattern
+     * This fact is indicated by giving
```
9037568: "indicated by giving" ... what? Please complete this sentence.

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param regExpression
+     * @return
+     */
+    private static boolean isSubtitution(String regExpression) {
+        cache(regExpression);
+        return patternCache.get(regExpression) == null;
+    }
+
+    /**
+     * Caches given regexpression, if the expression is substitution pattern it marks is as substitution
```
9037568: "Caches given regExpression. If the expression is a substitution pattern, it marks it as a substitution."

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param regExpression regular expression to match or a substitution in form of "s/<regex>/result/g" (replace all)
+     *            or "s/<regex>/result/" (replace first)
+     * @param source text to search in
+     * @return
+     * @throws TransformationException if regExpression or source is null
+     * @see org.eclipse.smarthome.transform.regex.internal.RegExTranformationService RegExTranformationService
+     */
+    public static String[] getMatches(String regExpression, String source) throws TransformationException {
+        if (regExpression == null || source == null) {
+            throw new TransformationException("the given parameters 'regex' and 'source' must not be null");
```
9037568: "The given" ...

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialDevice.java
```diff
@@ -280,39 +280,35 @@ public void serialEvent(SerialPortEvent event) {
...
                                                 eventPublisher.postUpdate(entry.getKey(), state);
                                             } catch (NumberFormatException e) {
                                                 logger.warn("Unable to convert regex result '{}' for item {} to number",
                                                         new String[] { result, entry.getKey() });
                                             }
-                                        } catch (TransformationException e) {
-                                            logger.error("Unable to transform!", e);
                                         }
+                                    } catch (TransformationException e) {
+                                        logger.error("Unable to transform!", e);
```
9037568: This should be a warn instead of an error.

marekhalmo:
Thanx Chris.

I will check the forums when i have a bit of time.. i posted the jar here and on an issue report.. so a good feedback from users using regex would be great. I don't want to break compatibility..

marekhalmo:
Hello,
the community test request [here](https://community.openhab.org/t/serial-binding-testers-needed-for-regex-compatibility-test-for-new-version-of-regex-transformation/28268)

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialDevice.java
marekhalmo: Not shure, TransformationException is thrown only if you provide null pattern or result

9037568: See the logging [guidelines](http://docs.openhab.org/developers/development/guidelines.html#e-logging).

marekhalmo: OK, based on guidelines you are right.. changing to warn

marekhalmo:
Fixed, could you review again?

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialBinding.java
```diff
@@ -135,7 +140,7 @@ public String getBindingType() {
     @Override
     public void validateItemType(Item item, String bindingConfig) throws BindingConfigParseException {
         if (!(item instanceof SwitchItem || item instanceof StringItem || item instanceof NumberItem
-                || item instanceof RollershutterItem)) {
+                || item instanceof RollershutterItem || item instanceof ContactItem || item instanceof DimmerItem)) {
             throw new BindingConfigParseException("item '" + item.getName() + "' is of type '"
```
9037568: Because of the changes to supported item types, this throw message should probably be updated to something like "Items of type {} are not supported."

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+ * Pattern matcher to match and cache patterns
+ *
+ * @author Marek Halmo
+ * @SInCE 1.10.0
+ * @see RegexPatternMatcher#getMatches getMatches()
+ *
+ */
+public class RegexPatternMatcher {
+    private static final Logger logger = LoggerFactory.getLogger(RegexPatternMatcher.class);
+    private static HashMap<String, Pattern> patternCache = new HashMap<String, Pattern>();
```
9037568: new HashMap<>();

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final Pattern SUBST_PATTERN = Pattern.compile("^s/(.*?[^\\\\])/(.*?[^\\\\])/(.*)$");
+
+    /**
+     * Returns compiled pattern for specific regeExpression that is stored in patternCache
```
9037568: "regeExpression" should say "regExpression", or possibly "regular expression".

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+                } else {
+                    result = source.trim().replaceFirst(regex, substitution);
+                }
+
+                return new String[] { result };
+            }
+        }
+
+        Matcher matcher = getPattern(regExpression).matcher(source.trim());
+        List<String> results = new ArrayList<String>();
```
9037568: = new ArrayList<>();

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialBinding.java
```diff
@@ -173,32 +161,43 @@ public void processBindingConfiguration(String context, Item item, String bindin
...
-                    downCommand = substring.substring(5, substring.length() - 1);
-                } else if (substring.startsWith("STOP(")) {
-                    stopCommand = substring.substring(5, substring.length() - 1);
+        String format = null;
+
+        int parameterSplitterAt = bindingConfig.indexOf(",");
+
+        if (parameterSplitterAt > 0) {
+            String[] split = bindingConfig.substring(parameterSplitterAt + 1, bindingConfig.length() - 1).split("\\),");
+            if (split.length > 0) {
```
9037568: Redundant. Should remove this if block.

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialBinding.java
marekhalmo: I will modify the message to include proper types, but i thing the user needs to be informed of the wrong item name so i will keep the old format.

bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/RegexPatternMatcher.java
marekhalmo: Good point

marekhalmo:
Ok.. fixes made. I added even more comments so the functionality is clear + found&fixed a small bug for the replacements.

Feel free to review again

marekhalmo:
Hello, the reporter of the issue tested the jar, and says it's working. There is not much attention to the thread.. What is next?

Will this get merged?

9037568:
Yes, looks good to me. Thanks, @marekhalmo !

marekhalmo:
Thank you!

marekhalmo:
Hello @kaikreuzer , can you point me to the issue with this? Why did you label this with a bug?

PR #5187, opened 2017-05-06
[weather] Add feels like temperature for ForecastIO

kabili207:
Signed-off-by: Andrew Nagle <kabili@zyrenth.com>


9037568:
Thanks, @kabili207 !

PR #5207, opened 2017-06-04
[Satel] Binding deactivation fix

druciak:
This PR fixes a deadlock when stopping the binding.


9037568:
I don't have enough knowledge of multithreading/concurrency.

@druciak, can you provide a summary of what this does and why it's needed?

druciak:
I think you have enough knowledge in this case :) 

There was a mistake in this commit, for some reason "synchronized" keyword on *close* method - which I was going to remove - was still there. Now I fixed that, thanks for spotting this mistake.

Basically "close" call did a "close" on watchdog object, which called "disconnect" method in the same SatelModule class, also synchronized. Since this was in a separate thread, we had a deadlock.
The fix removes "synchronized" from method and makes watchdog "close" outside "synchronized" block, which resolves the issue.

9037568:
Thanks!

PR #5235, opened 2017-07-02
Remove dependencies to openhab-esh-addons

pfink:
@kaikreuzer: This is the PR regarding openhab-esh-addons that we discussed this today. openhab-esh-addons does no longer exist, currently openhab1-addons just works because of a second ugly glitch (openhab-esh-addons is currently taken from 2.0.0-SNAPSHOT - even for release versions).

@mrguessed: The mios binding was the only one that still had dependencies to openhab-esh-addons which I removed from the featues.xml. As you're the main contributor of this binding, could you please review if the transformation features I removed are still in use? Build is fine, but maybe I broke something on runtime. If yes, I hope it's easy to fix...

We need this cleanup for an automated openHAB release process.

Cheers,
Patrick

### DIFF
--- a/features/openhab-addons-verify/pom.xml
+++ b/features/openhab-addons-verify/pom.xml
@@ -86,7 +86,7 @@
                                 <descriptor>mvn:org.eclipse.smarthome/esh-core/${esh.version}/xml/features</descriptor>
                                 <descriptor>mvn:org.eclipse.smarthome/esh-tp/${esh.version}/xml/features</descriptor>
                                 <descriptor>mvn:org.eclipse.smarthome/esh-ext/${esh.version}/xml/features</descriptor>
-                                <descriptor>mvn:org.openhab.core/openhab-core/${ohc.version}/xml/features</descriptor>
+                                <descriptor>mvn:org.openhab.core/openhab-core/${oh2.version}/xml/features</descriptor>
                                 <descriptor>mvn:org.openhab.addons/openhab-addons/${project.version}/xml/features</descriptor>
                             </descriptors>
                             <distribution>org.apache.karaf.features:framework</distribution>
--- a/features/openhab-addons-verify/pom.xml
+++ b/features/openhab-addons-verify/pom.xml
@@ -52,13 +52,6 @@
             <type>xml</type>
         </dependency>
         <dependency>
-            <groupId>org.openhab.core</groupId>
-            <artifactId>openhab-esh-addons</artifactId>
-            <version>${ohc.version}</version>
-            <classifier>features</classifier>
-            <type>xml</type>
-        </dependency>
-        <dependency>
             <groupId>org.openhab.addons</groupId>
             <artifactId>openhab-addons</artifactId>
             <version>${project.version}</version>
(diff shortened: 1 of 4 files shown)


kaikreuzer:
> Build is fine, but maybe I broke something on runtime.

Yes, unfortunately this change will break the functionality during runtime. 
This is indeed a very nasty exceptional case here, where a 1.x binding has a dependency on ESH features - it actually isn't even a compile time dependency, only one at runtime.

Only two solutions that I see here:
1. Don't define this dependency, but mention in the documentation that users need to install those transformations manually.
2. Remove the feature verification from the openhab1-addons build and only verify them during the distro build - which would mean that problems are not noticed in openhab1 builds anymore, but only when we try to assemble the distro.

@mrguessed How bad would the impact of option 1 be from your point of view?

pfink:
Option 3. would be to publish a release for openhab-esh-addons (or another artifact that contains the necessary dependencies). The build process is just not working because snapshot dependencies are invalid for creating a release.

kaikreuzer:
>  The build process is just not working because snapshot dependencies are invalid for creating a release.

No, I think the build process is not working because we refer to an artifact that isn't built at all (as it is only built by and for the distro).

The issue is really that all those addons should be independent of each other, so having this dependency here is not really allowed. If (1) is an acceptable option, I would go for it. Otherwise, (2) is probably the best workaround (which will also tremendously speed up the build time).

pfink:
> No, I think the build process is not working because we refer to an artifact that isn't built at all (as it is only built by and for the distro).
Yes, I know that it isn't built within openhab1-addons. The snapshot build does currently only work because 2.0.0-SNAPSHOT of openhab-esh-addons is available within the snapshot repo. The unleash release build does not work because it does not allow snapshot dependencies, so if a release of openhab-esh-addons would be available within the repository, it would work (of course as you said this would be still a bad smell, but not worse than status quo).

Regarding Option 2.: You told me, openhab-esh-addons does not exist anymore as of version 2.1. So how would it help to remove the feature verification within openhab1-addons? Wouldn't openhab-disto run into the same issue? Anyway, a cleaner option would may be to keep the feature verification for snapshot builds and only skip it for the Jenkins release build so that developers are informed about verification failures when building only openhab1-addons. Of course, the Jenkins release build doesn't need to verify the artifacts twice and therefore can skip it within openhab1-addons.

kaikreuzer:
> Wouldn't openhab-disto run into the same issue?

No, because the features that were once defined in openhab-esh-addons are available in openhab-distro, so verification will succeed.

> the Jenkins release build doesn't need to verify the artifacts 3 times

Well, the Jenkins snapshot build shouldn't do it either. Is there a mechanism to inform that some subsequent steps have failed? We should imho always see which commit have cause a pipeline build to be executed, so having the validation at one place should be enough.

pfink:
> Is there a mechanism to inform that some subsequent steps have failed?

What exactly do you mean?

kaikreuzer:
I mean that if you have a pipeline a->b which triggers because of a code change in a, but fails on b, that it can be easily seen that the code change in a was the reason for the build failure.

kaikreuzer:
@pfink I have created #5267 as a successor - this is for now definitely the simplest solution and it will speed up the build deliberately at the same time. If we feel later that it is too risky to not do a feature verification here as it causes issues at other places, we can still look at options to address this.

I hope this will make your work on the build pipeline easier :-)

PR #5239, opened 2017-07-06
[influxdb] fix state type calculation of group items

### LINKED ISSUE #5238: [influxdb] invalid states on startup
When using a `restoreOnStartup` strategy with influxdb, the log is full of errors like these:

```
2017-07-06 10:04:16.867 [ERROR] [rthome.core.library.items.NumberItem] - Tried to set invalid state 6799.360000000001 on item weather_brightness_max of type NumberItem, ignoring it
```

This happens for all of my group items. As it turns out, it tries to set StringItems because GroupItem does not match any of [these](https://github.com/openhab/openhab1-addons/blob/master/bundles/persistence/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/InfluxDBPersistenceService.java#L504). 

This happens on OH 2.1 and OH 2.2-SNAPSHOT build #980.

sjka:
...as it did not look at the base item before, so it was falling back to
StringType for all group items.

fixes #5238
Signed-off-by: Simon Kaufmann <simon.kfm@googlemail.com>


9037568:
Thanks, @sjka !

PR #5243, opened 2017-07-10
[influxdb] Added possibility to store all measurements in a single measurement

joek:
In influxdb it is sometimes good to have all data in a single table to do more complicated queries. At least if some external tools are used, e.g. grafana. If the `measurement` property is set, it will switch to the single measurement schema. If not, all will stay like it is.


dominikkv:
Nice! This would simplify data handling.
Maybe the `measurement` property should be renamed to a more intuitive name, like  `combine_measurements`?
And, do you think it is possible to group the measurements, for example, on bindings or item tags? So all item values from one binding or tag would end up in one measurement. I don't know if these information are avaiblable on this level...

joek:
Hi @dominikkv,
sorry for the late answer. I am on parental leave and not checking anything on github frequently.

I will rename the ```measurement``` name. 

In regards to your second request. I would prefer to have multiple influxdb instances (like in mqtt) configured. With this the user will be able to define everything by himself in the persistence file.

Nevertheless, I am not sure, if I am able to do it. Java is not my prefered programming language so it is a lot work for me to do it.

Cheers

PR #5270, opened 2017-07-28
[HTTP Binding] Don't log an error when the "format" configuration setting is found

lucasnetau:
Http Binding logs the following error if the config key "format" is included in http.cfg

> [ERROR] [ab.binding.http.internal.HttpBinding] - given config key 'format' does not follow the expected pattern '<id>.<url|updateInterval>'

This PR adds the format config key to the list of excluded keys from additional processing


9037568:
Thansk, @lucasnetau !

PR #5271, opened 2017-08-01
[eBUS] Fix escaping issue if slave CRC is 0xAA

csowada:
This PR fixes a CRC calculation issue if the CRC byte 0xAA is expanded to 0xA9 0x01

Started from community thread, already got feedback from user.
https://community.openhab.org/t/ebus-binding/3399/163


9037568:
Thanks, @csowada !

PR #5278, opened 2017-08-19
[KNX1] post command only

lewie:
In openHAB 1.x calling postUpdate AND sendCommand was sent to KNX bus.
While in openHAB2 in some scenarios this generates a lot of Echos and unwanted bus traffic.
In openHAB2 default behavior postUpdate is ignored, calling postUpdate() in rules are no longer sent to the KNX bus.


lewie:
Found bug...

PR #5279, opened 2017-08-20
[KNX1] postCommand only

lewie:
In openHAB 1.x calling postUpdate AND sendCommand was sent to KNX bus.
While in openHAB2 in some scenarios this generates a lot of Echos and
unwanted bus traffic.
In openHAB2 default behavior postUpdate is ignored, calling postUpdate()
in rules are no longer sent to the KNX bus.

### DIFF
--- a/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
+++ b/bundles/binding/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/bus/KNXBinding.java
@@ -122,7 +122,7 @@ public class KNXBinding extends AbstractBinding<KNXBindingProvider> implements P
     @Override
     protected void internalReceiveUpdate(String itemName, State newState) {
         logger.trace("Received update (item='{}', state='{}')", itemName, newState.toString());
-        if (!isEcho(itemName, newState)) {
+        if (KNXConnection.getPostUpdateToKNX() && !isEcho(itemName, newState)) {
             writeToKNX(itemName, newState);
         }
     }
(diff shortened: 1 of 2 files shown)


kaikreuzer:
> In openHAB 1.x calling postUpdate AND sendCommand was sent to KNX bus.

Which is fully correct and if this is changed it potentially breaks many installations out there. Or am I missing something?

udo1toni:
I may be wrong, but as far as I understood, postUpdate() should never send to the bound bus. 
knx does not know to differentiate commands and states, in fact, every sent state from openHAB is a command in knx.
knx separates both commands and states only through group addresses.

lewie:
@kaikreuzer, you are not fully correct. Sending both postUpdate AND sendCommand to the bus every command in KNX will be received twice. Once by internalReceiveCommand ond second by internalReceiveUpdate.

In openHAB 1, basically internalReceiveCommand and after, internalReceiveUpdate are called, which is logical. 

In openHAB 2 this behavior has changed, the order is arbitrary. So there is no simple way to ignore the second value sent by internalReceiveUpdate.

But you know, we talked about that many times.

@udo1toni this is not fully correct too, because the state GAs in a Dimmer  Item for example needs the internalReceiveUpdate.

I now learned by more tests, my really brutal wise to switch off internalReceiveUpdate does not work.

Therefore my next suggestion to solve this vexing problem: https://github.com/openhab/openhab1-addons/pull/5288

udo1toni:
State GAs shall only be set from real knx actuators. 
Given a dimmer, there is  at least one GA for switching ON/OFF, one GA which represents the state (read only), one GA from INCREASE/DECREASE (which will typically not work as expected when sending this GA as openHAB will never send a STOP command, this is needed by most of knx dimmers, called START/STOP dimming) and mostly common one GA for absolute dimming and one GA (read only) for absolute dimm level state. openHAB shall not send on any read only GA but only send read requests on these GAs.

PR #5294, opened 2017-09-02
[weather] Replaced org.apache.commons.httpclient dependency with ESH HttpUtil

cweitkamp:


kaikreuzer:
Just to note: This won't have any real effect, since `org.openhab.io.net.http.HttpUtil` is still using the Apache HTTP client. This is done to keep full backward compatibility for all 1.x add-ons.

cweitkamp:
I see. Wasn't aware of that. Thanks for mentioning it. So this is only a minor code clean up.

9037568:
Thanks, @cweitkamp !

PR #5301, opened 2017-09-11
[JDBC persistence] Decreased logging level

paulianttila:
Changed logging level from warn to debug for UnDefType items as they
pollute logs.


9037568:
Thanks, @paulianttila !

PR #5313, opened 2017-09-20
rework manifest

maggu2810:
Please have a look at the manifest yourself, too -- it has been done by a tool.

Related to: https://github.com/eclipse/smarthome/pull/4308

### DIFF
--- a/bundles/action/org.openhab.action.astro/META-INF/MANIFEST.MF
+++ b/bundles/action/org.openhab.action.astro/META-INF/MANIFEST.MF
@@ -1,16 +1,21 @@
 Manifest-Version: 1.0
-Private-Package: org.openhab.action.astro.internal
-Ignore-Package: org.openhab.action.astro.internal
+Bundle-ActivationPolicy: lazy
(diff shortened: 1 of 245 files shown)


kaikreuzer:
Hi @maggu2810, sorry, I didn't see this PR here as I am hardly tracking the openhab1 repo...
By now, a few files are in conflict - could you resolve them? Then we should merge right away.

maggu2810:
done

kaikreuzer:
Thanks - unfortunately, we are now in a deadlock with https://github.com/openhab/openhab1-addons/pull/5372 now as we do not have the JDT annotations in the TP, which makes Jenkins fail.

Do we actually need those imports on openHAB1? None of the code extends anything from ESH, so it should not involve any null annotations.

maggu2810:
If we don't need the nullness annotations at all, I don't see any reason for this PR at all.
No need to change the manifest as nothing is added (just reordered).

kaikreuzer:
Ok, then sorry for the hassle - it's probably easiest to close it for now (I am not too keen on introducing the use of null annotations on openhab1 as well right now...)

PR #5323, opened 2017-10-12
[KNX] Fixed README markdown syntax for docs.openhab.org

mojoaxel:
There was a new line in the markdown link syntax that broke the GIRA link.
minor fix.


9037568:
It appears this "broken" link only occurs when viewing the README via docs.openhab.org. Viewing it here, it's fine.

Issue #5328, opened 2017-10-14
[intertechno] Hang of CUL after a series of commands


Mik-:
I'm using this rule:
```
rule "Good evening"
when
    Item Command_Good_Evening received update ON
then
    logInfo(filename, "Good evening command received update ON")
    
    // First set state to OFF to force sending the command to the switch
    Livingroom_Lights_A.postUpdate(OFF)
    Livingroom_Lights_A.sendCommand(ON)

    Livingroom_Lights_B.postUpdate(OFF)
    Livingroom_Lights_B.sendCommand(ON)

    Livingroom_Lights_C.postUpdate(OFF)
    Livingroom_Lights_C.sendCommand(ON)

    Kitchen_Lights_A.postUpdate(OFF)
    Kitchen_Lights_A.sendCommand(ON)

    Kitchen_Lights_B.postUpdate(OFF)
    Kitchen_Lights_B.sendCommand(ON)

    Hall_Lights_A.postUpdate(OFF)
    Hall_Lights_A.sendCommand(ON)

    Command_Good_Evening.postUpdate(OFF)
end
```

All switches are bound to Intertechno with a serial CUL attached.

After almost every execution of this rule, the CUL didn't respond anymore, but sends permanently some commands over RF. The action LED (not the serial communication LEDs) is continuously blinking.
I've to stop the Intertecho and CUL bundles, plug off and on the CUL and start the bundles again.
I've used the current culfw and a-culfw on the CUL.

After adding some trace information in the CUL transport, I figured out, that there are overlapping commands send to the CUL. After every normal command a credit report is triggered. Sometimes three normal commands are sent to the CUL and after receiving the respones of that commands, three credit report requests are sent. I assumed, that this is the culprit.

I build a throttle in that way, that the next command waits on the response of the former command. (See pull-request #5329). This works perfectly fine, in my case.

I don't know, if the transport is still working with other bindings than intertechno and maybe I'm completely wrong with the remedy of  my problem.

I would be pleased, if someone can check the changes with other bindings or have any suggestions to this topic.

Cheers
Mik

9037568:
Have you raised this question in the [community forum](http://community.openhab.org/)? If not, you should.

Mik-:
I did, but unfortunately didn't received any response...

9037568:
I think I found them:
[...]

PR #5329, opened 2017-10-14
[CUL transport] Throttle commands, which are sent to the CUL

Mik-:
I've build a throttle, which waits with sending a new command for the response of the former command to prevent race conditions.

See issue #5328

I've tested this with the Intertechno binding, but no other, which use the CUL transport


Mik-:
I've got some feedback on this topic from the community and it's working there, at least with one other binding than Intertechno, which I've tested myself.
See https://community.openhab.org/t/cul-transport-fix-for-issue-with-cul-transport-race-condition/35385

9037568:
Thanks, @Mik- !

PR #5335, opened 2017-10-20
Deactivated the static-code-analysis tool.

VelinYordanov:
I believe currently the static-code-analysis tool can be executed but doesn't log the errors in this repo.
With this change it doesn't execute at all.

Signed-off-by: VelinYordanov <velin.iordanov@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,9 +51,6 @@
 
     <properties>
 
-		<!-- we deactivate the static analysis tool as the openHAB 1 code does not comply to these rules -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -130,6 +127,17 @@
                         </supportedProjectTypes>
                     </configuration>
                 </plugin>
+				
+				<!-- we deactivate the static analysis tool as the openHAB 1 code does not comply to these rules -->
```
kaikreuzer: could you please fix the indentation? Thanks!

Issue #5309, opened 2017-09-17
[weather1] Request: add support for the free version of forecasts for OpenWeatherMap


dominikkv:
Wrong url for weather forecast request for OpenWeatherMap. There is no "daily" in the url, see documentation https://openweathermap.org/forecast5

## Expected Behavior
Working forecast for OpenWeatherMap (http://docs.openhab.org/addons/bindings/weather1/readme.html#forecast)

## Current Behavior
Forecast is broken, see karaf log:
```
13:16:22.182 [TRACE] [nal.provider.AbstractWeatherProvider] - OPENWEATHERMAP[openweather]: request : http://api.openweathermap.org/data/2.5/forecast/daily?lat=48.xxx&lon=7.yyy&lang=de&cnt=5&mode=json&units=metric&APPID=123notthateasy
13:16:22.239 [TRACE] [nal.provider.AbstractWeatherProvider] - OPENWEATHERMAP[openweather]: response: {"cod":401, "message": "Invalid API key. Please see http://openweathermap.org/faq#error401 for more info."}
13:16:22.239 [ERROR] [nal.provider.AbstractWeatherProvider] - OPENWEATHERMAP[openweather]: Can't retreive weather data: HTTP/1.1 401 Unauthorized
```

## Possible Solution
See merge request https://github.com/openhab/openhab1-addons/pull/5308

## Steps to Reproduce (for bugs)
1. Install binding weather1
2. Create /conf/services/weather.cfg with appropriate content like:
```
apikey.OpenWeatherMap=123456789

location.openweather.name=Openweather
location.openweather.latitude=48.xxxx
location.openweather.longitude=7.yyyy
location.openweather.provider=OpenWeatherMap
location.openweather.language=de
location.openweather.updateInterval=10
location.openweather.units=si
```
3. Add any weather item 
4. Open karaf console, set trace level: `log:set TRACE org.openhab.binding.weather`
5. After 10 minutes, see log entries above

## Your Environment
* Version used: openhab2, openhab-binding-weather1 1.10.0
* Operating System and version: Docker image openhab/openhab:2.1.0-amd64 hosted on an Ubuntu 17.04

9037568:
The error you're getting is because you're using an invalid API key. There's nothing wrong with the binding.

And the forecast URL certainly does have 'daily' in the path. [See here](http://openweathermap.org/forecast16#16days).

dominikkv:
OK, there are indeed 2 API endpoints
[...]

PR #5308, opened 2017-09-17
Weather binding: OpenWeatherMap: remove "daily" from forecast url

dominikkv:
There is no "daily" in the forecast url as the documentation (https://openweathermap.org/forecast5) shows.

The binding tries to get wheather forecast information from OpenWeatherMap, but the response is `401 Unauthorized`. Turning debuggin on in karaf shows:
```
13:16:22.182 [TRACE] [nal.provider.AbstractWeatherProvider] - OPENWEATHERMAP[openweather]: request : http://api.openweathermap.org/data/2.5/forecast/daily?lat=48.xxx&lon=7.yyy&lang=de&cnt=5&mode=json&units=metric&APPID=123notthateasy
13:16:22.239 [TRACE] [nal.provider.AbstractWeatherProvider] - OPENWEATHERMAP[openweather]: response: {"cod":401, "message": "Invalid API key. Please see http://openweathermap.org/faq#error401 for more info."}
13:16:22.239 [ERROR] [nal.provider.AbstractWeatherProvider] - OPENWEATHERMAP[openweather]: Can't retreive weather data: HTTP/1.1 401 Unauthorized
```
A quick lookup in the OpenWeatherMap documentation reveals that the url this addon uses is wrong.

Please merge this pull request to fix the url.

Thanks
Dominik

PR #5344, opened 2017-10-28
[OpenWeatherMap] Change FORECAST path to get free 5 day forecast

kstruessmann:
I just changed the URL for the forecast information. The original URL is calling up the 16 day forecast information which is not available for Free API users.

### DIFF
--- a/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/provider/OpenWeatherMapProvider.java
+++ b/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/provider/OpenWeatherMapProvider.java
@@ -19,7 +19,7 @@ import org.openhab.binding.weather.internal.parser.JsonWeatherParser;
  */
 public class OpenWeatherMapProvider extends AbstractWeatherProvider {
     private static final String URL = "http://api.openweathermap.org/data/2.5/weather?lat=[LATITUDE]&lon=[LONGITUDE]&lang=[LANGUAGE]&mode=json&units=metric&APPID=[API_KEY]";
-    private static final String FORECAST = "http://api.openweathermap.org/data/2.5/forecast/daily?lat=[LATITUDE]&lon=[LONGITUDE]&lang=[LANGUAGE]&cnt=5&mode=json&units=metric&APPID=[API_KEY]";
+    private static final String FORECAST = "http://api.openweathermap.org/data/2.5/forecast?lat=[LATITUDE]&lon=[LONGITUDE]&lang=[LANGUAGE]&cnt=5&mode=json&units=metric&APPID=[API_KEY]";
 
     public OpenWeatherMapProvider() {
         super(new JsonWeatherParser());


cweitkamp:
See #5308

kstruessmann:
Hi, 
I've seen the other pull request but I don't understand why it is closed and not merged. If you see #5309 there ist an open issue. If the binding needs to support either the free 5 day forecast or the 16 day payed forecast, maybe two providers for Openwethermap are a possible solution? So you have one free provider and one payed provider?

9037568:
I would say the best solution is to have a configuration option to choose between the free or commercial APIs.

cweitkamp:
I would love to see a working forecast for OWM provider to be part of this binding. Even better without anyone has to pay for it.

I agree, that it should be configurable for the user which version he wants to use. But that is not simply possible by switching the URLs.

If you have a closer look at the [API documentation](http://openweathermap.org/api) of OWM you will see that the 5 day API provides weather data for every 3 hours. We have to adopt that in the binding too. Currently the binding parses every result set as a separate day. Meaning that you will currently see the weather data for the next 3 hours as forecast for tomorrow, the weather data for the next 6 hours as forecast for the day after tomorrow, and so on.

How do you think can we archive that? Should we skip the result sets for the next 3, 6, ..., 21 hours and use the result set for the next 24 hours as forecast for tomorrow, skip 27, 30, ...,  45 hours again and use the result set of 48 hours as forecast for the day after tomorrow, and so on? Wdyt? This could end in a forecast which could change every few hours.

Another approach could be to use a specific time of the day (e.g. 12 o'clock) as forecast.

A third approach could be to break the general logic of the binding to return 3 hours forecast instead of daily forecast (especially if you're using OWM, but maybe other provider support other forecast options as well).

kstruessmann:
Okay that means to me that I have to do two different things. The first one is the extension of the config file to the possibility to choose between the free version and the commercial version. 
How would you implement that? I'm thinking about something like the APIKEY2 for Hamweather but called APIKEYTYPE where you can choose between the two options (free and commercial).

The second things is the forecast. I would like to change the integer to float so you are able to do something like forecast=1,5 meaning tomorrow at 12 a.m. or forecast=1,375 meaning tomorrow at 9 a.m.
But if you use the commercial API version you would have to request data 3 times (todays data, five day forecast, 16 day forecast).

9037568:
Another thing to think about is that as the amount of work here gets larger, at some point it might no longer be worth doing for OH1, but might be a better idea to implement an OH2/ESH binding.

This is not a comment favoring one over the other. Just an item for consideration.

9037568:
>  I'm thinking about something like the APIKEY2 for Hamweather but called APIKEYTYPE where you can choose between the two options (free and commercial).

I would favor `apitype` or `licencetype`.

kstruessmann:
I was thinking about it once again and am now convinced that we don’t need an apitype. If your adapt the forest in a way you get every three hours the forecast information for 5 days, you have to use the free API starting point anyways. That means if your key is only for the five day forecast you are not getting any data for the sixth day.

9037568:
@kstruessmann does this mean you're not making any changes? If so, I'll close this ticket.

PR #5349, opened 2017-10-31
[weather] Fixed parsing of rain value in forecast data

cweitkamp:
- The modifier `DOUBLE_3H` is not needed to parse the forecast data.
- Added mapping for parsing snow from forecast data

### DIFF
--- a/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/model/Precipitation.java
+++ b/bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/model/Precipitation.java
@@ -27,8 +27,7 @@ public class Precipitation {
             @Provider(name = ProviderName.WUNDERGROUND, property = "current_observation.precip_1hr_metric"),
             @Provider(name = ProviderName.WUNDERGROUND, property = "qpf_allday.mm"),
             @Provider(name = ProviderName.OPENWEATHERMAP, property = "rain.3h", converter = ConverterType.DOUBLE_3H),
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.weather/src/main/java/org/openhab/binding/weather/internal/model/Precipitation.java
```diff
@@ -27,8 +27,7 @@
             @Provider(name = ProviderName.WUNDERGROUND, property = "current_observation.precip_1hr_metric"),
             @Provider(name = ProviderName.WUNDERGROUND, property = "qpf_allday.mm"),
             @Provider(name = ProviderName.OPENWEATHERMAP, property = "rain.3h", converter = ConverterType.DOUBLE_3H),
-            @Provider(name = ProviderName.OPENWEATHERMAP, property = "rain", converter = ConverterType.DOUBLE_3H),
-            @Provider(name = ProviderName.OPENWEATHERMAP, property = "rain.1h"),
```
9037568: Why is this rain.1h value removed? Is it unused?

cweitkamp: In the API docs I can't find anything about a "rain.1h" property. My latest API calls didn't show it either in the responses. I don't know if it produces overhead if we leave it in there.

9037568: Good enough.

9037568:
Thanks, @cweitkamp !

PR #5350, opened 2017-11-02
[tacmi] Add binding to openhab2 distribution

Wolfgang1966:
As I now have the TA CMI binding running with openhab 2 for several months, I think it is safe to add it to the distribution.

Included is a small change of one scale factor which turned out to be necessary after a firmware change of the C.M.I, and a slightly better error handling when the listening socket can't be created.


kaikreuzer:
Perfect, thanks!

Wolfgang1966:
Whow, that was fast :-)

PR #5372, opened 2017-12-01
compile against 2.1.0 target platform

kaikreuzer:
This should solve the current build problems.
As openHAB 1 add-ons do not depend on any changes in ESH/openHAB2 target platform, it should be safe to always compile against the older 2.1.0 version of it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
     <parent>
         <groupId>org.openhab</groupId>
         <artifactId>pom-tycho</artifactId>
-        <version>2.2.0-SNAPSHOT</version>
+        <version>2.1.0</version>
         <relativePath></relativePath>
     </parent>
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
@pfink It seems that the Jenkins build pipeline does not like this change - while the "old" build now works as expected, the pipeline still fails as I assume it still sets the pom versions in a way that openHAB1 is compiled against the latest target platform.

Do you have any advice on how to make it work?

pfink:
@kaikreuzer: Should be no problem, there is a flag if the parent version should be updated. I started a test run to check whether it works to just set it to `false`: https://openhab.ci.cloudbees.com/view/Sandbox%20Release%20Pipeline/job/sandbox-openhab2-release/294/console

kaikreuzer:
It failed again, but this time with another message:
```
[ERROR] No plugin found for prefix 'unleash' in the current project and in the plugin groups
```
Shall we restart it once more?

pfink:
Oh yes, this cannot work because we introduced the maven unleash plugin in the 2.2.0-SNAPSHOT parent pom, it was not present in the 2.1.0 release. Workaround would be to add the unleash plugin to the pom here as well: #5375

pfink:
Started another releasebuild now: https://openhab.ci.cloudbees.com/view/Sandbox%20Release%20Pipeline/job/sandbox-openhab2-release/295/

PR #5390, opened 2017-12-20
[Telegram] Correct chat id in README.md for Telegram action

mwl:
With chat.id it's also possible to add the bot to a group


9037568:
Thanks, @mwl !

PR #5421, opened 2018-01-03
Fatek PLC Switch item with two registers

### LINKED ISSUE #5420: [Fatek PLC] Switch item does not work with 2 registers
## Expected Behavior

According to docs, when i want to display value of REG1, but control REG2, i should use this syntax:

plcName:[!]REG1[[!]REG2]
    REG1 is used for read state of switch
    REG2 is used for change state of switch, if is not specified - REG1 is used

However this is incorrect, it cannot ever work without REG1/REG2 separator, which should be probably ":"

## Current Behavior

When i use space, ; or : as separator, only REG1 is both displayed and controlled.

## Possible Solution

Fix the binding, so it use correct REG separator and fix the documentation.

Not possible to reproduce without Fatek PLC present, but this code in Items should work, but it does not:
Switch Test                 	    	    "Test"                                            { fatekplc="plc1:M150:M151" }

## Context

Showing status of a pump, but not controlling it directly.

## Your Environment

openHAB 2.1.0-1 (Release Build) on RPI3/OpenHABian
binding-fatekplc1 - 1.10.0

Thanks.

slawekjaranowski:
Fixes #5420

Signed-off-by: Slawomir Jaranowski <s.jaranowski@gmail.com> (github: slawekjaranowski)

### DIFF
--- a/bundles/binding/org.openhab.binding.fatekplc/README.md
+++ b/bundles/binding/org.openhab.binding.fatekplc/README.md
@@ -59,7 +59,7 @@ We specified only one register.  If register name starts with "!" (exclamation m
 ### Switch Items
 
 ```
-plcName:[!]REG1[[!]REG2]
+plcName:[!]REG1[:[!]REG2]
 ```
 
 * REG1 is used for read state of switch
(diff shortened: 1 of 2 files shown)


9037568:
Thanks, @slawekjaranowski !

PR #5427, opened 2018-01-07
Correct feature names

wborn:
Some recently merged bindings don't have a "1" suffix in their feature name. 

This will also resolve the "404 Not Found" errors when you click on the respective bindings in Paper UI to show their documentation.


9037568:
Thanks, @wborn !

Issue #5433, opened 2018-01-11
[CUL] CUL transport never unloads after disconnection of consuming bindings


Mik-:
## Expected Behavior
If all bindings has stopped, wich are consuming the CUL transport, the CUL transport should also unload.

## Current Behavior
The CUL transport  stays active.

## Possible Solution
This [pull request](https://github.com/openhab/openhab1-addons/pull/5329) to prevent a race condition, adds a separate send thread, which binds to the transport itself. This self-listening must be considered in then `hasListeners()` method of the `AbstractCULHandler`.

## Your Environment
* Version used: milestone 1.12.0

9037568:
Closed via merged PR.

PR #5434, opened 2018-01-11
[CUL] Bugfix issue 5433

Mik-:
This is a bugfix for #5433. The CUL transport now unloads after stopping last consuming binding.


9037568:
Thanks, @Mik- !

PR #5435, opened 2018-01-12
Add 7590 as tested

cribskip:
As I'm using the 7590 successfully, I've added it to the list in order for others to know.


9037568:
Thanks, @cribskip !

Issue #5441, opened 2018-01-18
[FritzboxTr064]: Still problems with "Could not read response from FritzBox. Unauthorized! ..."


oprobst:
I think issue [#4548](https://github.com/openhab/openhab1-addons/issues/4548) is not finally solved.
Maybe the implemented HTTP authorization reset reduces failure effects in some cases. But on my system I still got 401 responses from the Fritzbox. 

And finally after the reset the further progress results in an `Auth Scheme is null` failure as listed in [#5331](https://github.com/openhab/openhab1-addons/issues/5331)

## Expected Behavior
No communication failure at all... :-)

## Current Behavior
The current implementation does an error prone call to the fritz box. It appears only under load. The request to the fritzbox is responded by 401 (Unauthorized). 
Log output is `Could not read response from FritzBox. Unauthorized! Check User/PW in config. Create user for tr064 requests`.

The error handling of the 401 response causes a complete addon outage until openhab restarts. So we've got two issues. The unstable failure handling will be solved by [#5331](https://github.com/openhab/openhab1-addons/issues/5331). 

This ticket is about preventing the failure cause. 

## Possible Solution
I think the implementation of calling the Web Service is not completely thread safe. The attribute `Tr064Comm _fboxComm` is shared between the inherited methods `protected void execute()` and `protected void internalReceiveCommand(String itemName, Command command)`.
Even without deeper investigation I assume a concurrency problem in sharing this variable. Both methods are called by the framework. And it is very likely that it is not the same thread calling or any kind of synchronization active... 

Two options: 
1. Synchronize the http request in between the calling methods. This will create a blocking dependency between the framework threads. (Maybe not the best idea at all...)
2. Open a separate connection for each method (or even better establish a connection pool, but maybe we should not crack a walnut with a sledgehammer).

**I'm already implemented the fix and currently verifying this locally. If it solves the problem, I will prepare a pull request for option 2.**
 
## Steps to Reproduce (for bugs)
1. Configure the addon to do a lot of calls to the fritzbox by configuring few items.
2. Write a rule also triggering a lot of calls in parallel.
3. After some time, the request will be responded by 401. 

## Your Environment

* Version used: 1.11.0 on openhab 2.2 
* Environment name and version (e.g. Chrome 39, node.js 5.4, Java 8, ...):  
* Operating System and version (desktop or mobile, Windows 10, Raspbian Jessie, ...): Debian Jessie

oprobst:
[...]

PR #5442, opened 2018-01-18
[fritzboxtr064] Separate connection for execute and receiveCommand #5441

oprobst:
fix for #5441 
Separated Tr064Comm attribute in two instances for receiveCommand and Execution to prevent concurrent thread access.


oprobst:
Nope, this workaround didn't work very well. The error appears less often, but still happens from time to time. 
I'll try to use the the PooledConnectionManager instead. Therefore I'm closing this pull request until I found a better solution.

Issue #5331, opened 2017-10-16
[FritzboxTr064]: Error while executing background thread FritzboxTr064 Refresh Service


staehler:
After some time running openhab this error appears:
```
2017-10-16 11:21:12.694 [ERROR] [b.core.service.AbstractActiveService] - Error while executing background thread FritzboxTr064 Refresh Service
java.lang.IllegalStateException: Auth scheme is null
        at org.apache.http.util.Asserts.notNull(Asserts.java:52) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.auth.HttpAuthenticator.ensureAuthScheme(HttpAuthenticator.java:229) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.auth.HttpAuthenticator.generateAuthResponse(HttpAuthenticator.java:184) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:262) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:184) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:88) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:184) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:82) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.openhab.binding.fritzboxtr064.internal.Tr064Comm.readSoapResponse(Tr064Comm.java:456) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
        at org.openhab.binding.fritzboxtr064.internal.Tr064Comm.getTr064Values(Tr064Comm.java:247) [209:org.openhab.binding.fritzboxtr064:1.11.0.201710050110]
[...]

PR #5443, opened 2018-01-18
[Fritzboxtr064] Close connection and re-init in case of failure (#5331)

oprobst:
fix for #5331 

Improved error handling: Closing and resetting connection in case of error and initialize again.

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
@@ -461,8 +461,8 @@ public class Tr064Comm {
 
             // Check for (auth-) error
             if (slResponse.getStatusCode() == 401) {
-                logger.error(
-                        "Could not read response from FritzBox. Unauthorized! Check User/PW in config. "
+                logger.warn(
+                        "Could not read response from FritzBox. Unauthorized! Check user and password in config. "
                         + "Verify configured user for tr064 requests. Reason from Fritzbox was: {}", slResponse.getReasonPhrase());
                       
                 postSoap.releaseConnection();
@@ -496,7 +496,7 @@ public class Tr064Comm {
         } finally {
             // Make sure connection is released. If error occurred make sure to print in log
             if (exceptionOccurred) {
-                logger.error("Releasing connection to FritzBox because of error!");
+                logger.warn("Releasing connection to FritzBox because of error.");
                 resetHttpClient();
             } else {
                 logger.debug("Releasing connection");
@@ -519,7 +519,7 @@ public class Tr064Comm {
     	try {
 			_httpClient.close();
 		} catch (IOException e) {
-			logger.error("Failed to close connection to fritzbox at {}. This might result in still open, but dead connections waiting for a timeout.", _url, e);
+			logger.debug("Failed to close connection to fritzbox at {}. This might result in still open, but dead connections waiting for a timeout.", _url, e);
 		}
 		_httpClient = createTr064HttpClient(_url);		
 	}
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
@@ -462,8 +462,11 @@ public class Tr064Comm {
             // Check for (auth-) error
             if (slResponse.getStatusCode() == 401) {
                 logger.error(
-                        "Could not read response from FritzBox. Unauthorized! Check User/PW in config. Create user for tr064 requests");
-                _httpClientContext.getTargetAuthState().reset();
+                        "Could not read response from FritzBox. Unauthorized! Check User/PW in config. "
+                        + "Verify configured user for tr064 requests. Reason from Fritzbox was: {}", slResponse.getReasonPhrase());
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
```diff
@@ -462,8 +462,11 @@ private SOAPMessage readSoapResponse(String soapActionHeader, SOAPMessage reques
             // Check for (auth-) error
             if (slResponse.getStatusCode() == 401) {
                 logger.error(
```
9037568: Please convert this to a WARN instead of ERROR.

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
```diff
@@ -494,7 +497,7 @@ private SOAPMessage readSoapResponse(String soapActionHeader, SOAPMessage reques
             // Make sure connection is released. If error occurred make sure to print in log
             if (exceptionOccurred) {
                 logger.error("Releasing connection to FritzBox because of error!");
```
9037568: Please convert this to a WARN instead of ERROR.

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
```diff
@@ -505,7 +508,23 @@ private SOAPMessage readSoapResponse(String soapActionHeader, SOAPMessage reques
...
+     */
+    private void resetHttpClient() {
+    	logger.trace("Drop client for fritzbox and setup connection again.");
+		if (_httpClientContext.getTargetAuthState() != null) {
+			_httpClientContext.getTargetAuthState().reset();
+		}
+    	try {
+			_httpClient.close();
+		} catch (IOException e) {
+			logger.error("Failed to close connection to fritzbox at {}. This might result in still open, but dead connections waiting for a timeout.", _url, e);
```
9037568: This probably should not be higher than INFO. I'd argue for DEBUG.

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
oprobst: :+1: This wasn't mine, but I changed it.

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
oprobst: :+1:

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
oprobst: :+1: We're not in any kind of high load situation where this could be critical, so I'm fine with DEBUG.

oprobst:
> please run the code through the Eclipse formatter.

The eclipse oxygen builtin once? This one is using tabs instead of spaces as the rest of the code. Gives quite a lot of changes. I executed the Eclipse Formatter but configured it to use spaces instead (even if this is not my personal preference... 😃) .

9037568:
>The eclipse oxygen builtin once?

Depends how you set up your environment; usually it's installed with Eclipse. Looks good now, though. Thanks!

9037568:
Thanks, @oprobst !

staehler:
Could it be that this patch influences the call recognition? I updated from 2.2.0-20171109111930-1 to 2.3.0-20180129232532-1 and since than my item
`Switch Active_Call "Telefon klingelt [%s]"	 <phone> (gPhone,gMapDB) {fritzboxtr064="callmonitor_ringing" }`
only recognises some of the events, but not all. So the item Active_Call remains ON after a phone call or it remains OFF during a phone call :-(

BTW: my openhab log showes sometimes the following error
`2018-02-04 10:44:35.519 [ERROR] [ntime.internal.engine.RuleEngineImpl] - Rule 'Fritzbox Active Call ON': Index is out of range`
where the rule lookes like
```
rule "Fritzbox Active Call ON"
when
    Item Active_Call changed from OFF to ON or
then
    if (Active_Call.state==ON) {

		val incCall = fboxIncomingCallResolved.state as StringListType
		val caller = incCall.getValue(1)

		var String text= "Eingehender Anruf: " + caller
	    logInfo("RSS", text)		
end
```
I think that the rule error has nothing to do with the item error, but maybe it gives a hint or you see an obvious mistake in it? This rule error happend also before the upgrade sometimes.
Do you have any idea, what goes wrong?

Before the update the Active_Call switched as expected.

Kind regards,
Michael

9037568:
>Could it be that this patch influences the call recognition?

@staehler would you raise this as a new defect, and attach a debug log?

Thanks!

oprobst:
Very unlikely that your index out of range issue was introduced by this fix. I did just a very small change to re-init the socket connection after a failure. You might check your log if there was a connection problem somewhere before, which might give a hint.
Nevertheless +1 for opening a new defect.

PR #5484, opened 2018-02-02
[Fritzboxtr064] Synchronized access to fritzbox.

oprobst:
Prevent mixing up Digest nonce during concurrent remote calls. Actually this change prevents concurrent calls at all. 
This could happend if the function was called by more than one threat the same time.


9037568:
I'll start a review. Have you gotten any community members to try out your enhanced version? If not, please drop a thread in the community forum to request it.

PR #5490, opened 2018-02-09
[Telegram Action] Add option for Markdown or HTML formatting

memtew:
The Telegram Bot API supports basic formatting for messages. My commit add Markdown formatting mode by default to Telegram action.

### DIFF
--- a/bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
+++ b/bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
@@ -66,8 +66,8 @@ public class Telegram {
 
     private static Map<String, TelegramBot> groupTokens = new HashMap<String, TelegramBot>();
 
-    public static void addToken(String group, String chatId, String token) {
-        groupTokens.put(group, new TelegramBot(chatId, token));
+    public static void addToken(String group, String chatId, String token, String parse) {
+        groupTokens.put(group, new TelegramBot(chatId, token, parse));
     }
 
     @ActionDoc(text = "Sends a Telegram via Telegram REST API - direct message")
@@ -89,7 +89,7 @@ public class Telegram {
         postMethod.getParams().setParameter(HttpMethodParams.RETRY_HANDLER,
                 new DefaultHttpMethodRetryHandler(HTTP_RETRIES, false));
         NameValuePair[] data = { new NameValuePair("chat_id", groupTokens.get(group).getChatId()),
-                new NameValuePair("text", message), new NameValuePair("parse_mode", "Markdown") };
+                new NameValuePair("text", message), new NameValuePair("parse_mode", groupTokens.get(group).getParse()) };
         postMethod.setRequestBody(data);
 
         try {
--- a/bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
+++ b/bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
@@ -89,7 +89,7 @@ public class Telegram {
         postMethod.getParams().setParameter(HttpMethodParams.RETRY_HANDLER,
                 new DefaultHttpMethodRetryHandler(HTTP_RETRIES, false));
         NameValuePair[] data = { new NameValuePair("chat_id", groupTokens.get(group).getChatId()),
-                new NameValuePair("text", message) };
+                new NameValuePair("text", message), new NameValuePair("parse_mode", "Markdown") };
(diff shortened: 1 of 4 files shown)


memtew:
But I don't know how to ad an option without breaking current behavior. Maybe through telegram.cfg? My current version of commit doesn't breaks anything.

memtew:
After day of testing I'm found breaking behavior - Markdown parse mode throw error if your message have an alone _ or * symbol.

9037568:
>But I don't know how to add an option without breaking current behavior. Maybe through telegram.cfg?

Yes, the telegram.cfg would need to be updated with a new key such as `messageFormat` or something similar. Then you'd need to update the action code to fetch the value of the new key and use it appropriately.

9037568:
>After day of testing I'm found breaking behavior - Markdown parse mode throw error if your message have an alone _ or * symbol.

Post the error and let's see if we can work out how to fix it.

memtew:
> Post the error and let's see if we can work out how to fix it.

Error is simple: if message includes (it doesn't matter where - in variable name like _door_Hall_, in text) unclosed symbols from Markdown (_ or * for example) Telegram api throws error:
`21:50:11.888 [WARN ] [hab.action.telegram.internal.Telegram] - Method failed: HTTP/1.1 400 Bad Request`

memtew:
Added new option to telegram.cfg:
```
#Markdown or HTML
bot.parseMode=Markdown
```
If option not present or blank - Telegram Action just don't use it.

9037568:
>Added new option to telegram.cfg

When you do this, please make sure to also modify the openhab_default.cfg to reflect the new setting.

memtew:
> In addition to addressing the inline comments, please also run the Eclipse formatter on all source.
Sorry, but I don't have Eclipse installed.

9037568:
>Sorry, but I don't have Eclipse installed.
Then you'll need to install it or find another way to comply with the formatting requirements.

memtew:
> Then you'll need to install it or find another way to comply with the formatting requirements.

Ok, it was hard, but Eclipse successfully installed on VM. But after Source->Format nothing was changed.

bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/Telegram.java
```diff
@@ -66,10 +66,14 @@
 
     private static Map<String, TelegramBot> groupTokens = new HashMap<String, TelegramBot>();
 
-    public static void addToken(String group, String chatId, String token, String parse) {
-        groupTokens.put(group, new TelegramBot(chatId, token, parse));
+    public static void addToken(String group, String chatId, String token) {
+        groupTokens.put(group, new TelegramBot(chatId, token));
     }
 
+    public static void addToken(String group, String parse) {
```
9037568: I don't believe this method does anything. The bot created by this method will never function, will it?

9037568:
>Ok, it was hard, but Eclipse successfully installed on VM. But after Source->Format nothing was changed.

Probably indicating your Eclipse was incorrectly installed. But I can fix the formatting before merging if you can't get it to work.

memtew:
> @9037568 @memtew
> Hi Aydar, hi Chris,
> 
> I've seen this PR and just want to ask, if there is any update on it?
> If not, would it be possible to update the necessary things here to bring up the needed support? (I see only a few lines, which has to be changed, but I'm not familiar with java)
> 
> Best regards,
> Patrick

Hello, Patrick,

My knowledge was not enough to finish the job correctly, so I quit the job. Maybe someone else can do it right. At the moment I use the package that I compiled from the source myself. But in this form, the package does not pass the test of Chris.

pschaumburg:
Hi Aydar,

Thank you for your answer. Maybe @9037568 can give us a short answer :)

Best regards,
Patrick

pschaumburg:
@memtew 
I've done some small thoughts in my own fork with your whole idea and nearly all code in it, but I really don't know, if all of it works. Could you check, if this would work for you?
https://github.com/pschaumburg/openhab1-addons/commit/800c2c5d4d1f94ab59314e27b48ce31ccfbef092
Of if possible, to create a jar file out of it (I will figure out, if all works fine for me).

9037568:
I believe [this](https://github.com/openhab/openhab1-addons/pull/5490#pullrequestreview-96033478) is the last remaining item, but it's been so long I can't tell for sure.

thorstenbe:
I created a new fork since this is labeled as stale and all issues should be fixed. Formatting works while sending photos as well and I could create a new pull request. Would this be ok?

9037568:
That'd be fine.

thorstenbe:
Please have a look at #5765

9037568:
Replaced by #5765

PR #5498, opened 2018-02-18
[FS20] Add remark about adding openhab user to dialout group

dwaldhans:
User `openhab` needs to be in the `dialout` group so it can access a CUL device unter `/dev/ttyACM0`. Otherwise an exception occurs.

Signed-off-by: David Waldhans <public@waldhans.at> (github: dwaldhans)


9037568:
Thanks, @dwaldhans !

PR #5501, opened 2018-02-21
Send base64 image with Telegram Action

BasvanH:
You can send an Image Item with the Telegram Action


BasvanH:
It's my first commit to Openhab, I do not know why the check has failed. Any help?

9037568:
This change appears to be the cause of the build failure. Reverting this change should fix the problem.

````
<version>1.13.0-SNAPSHOT</version>
````

PR #5510, opened 2018-03-05
Update products.xml

donald1985:


cdjackson:
This will not work without the device file. Please also add that to this pr.

donald1985:
Hi Chris, I tried now for 30min to merge the two pull requests, but no success, now I have separated them again.

cdjackson:
Ok, I’ll take a look. When I initially commented I didn’t see there were two PRs which is what prompted the comment...

PR #5515, opened 2018-03-12
Rename Eclipse projects for bindings that also exist in OH2

wborn:
This should fix the issue that only the OH1 or the OH2 version of these projects can be automatically imported in Eclipse.

Maybe we should add the 1 suffix to all binding projects to prevent future issues? The same has been done with the feature names.

### DIFF
--- a/bundles/binding/org.openhab.binding.rfxcom.test/.project
+++ b/bundles/binding/org.openhab.binding.rfxcom.test/.project
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>org.openhab.binding.rfxcom.test</name>
+	<name>org.openhab.binding.rfxcom.test1</name>
 	<comment></comment>
 	<projects>
 	</projects>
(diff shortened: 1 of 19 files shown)


bundles/binding/org.openhab.binding.rfxcom.test/.project
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>org.openhab.binding.rfxcom.test</name>
+	<name>org.openhab.binding.rfxcom.test1</name>
```
9037568: Shouldn't this be `org.openhab.binding.rfxcom1.test` ?

9037568:
Do we need feedback from @kaikreuzer on this?

bundles/binding/org.openhab.binding.rfxcom.test/.project
wborn: That seems more logical indeed so I'll rename it.

wborn:
> Do we need feedback from @kaikreuzer on this?

Yes it would be nice to know how other people think about:

> Maybe we should add the 1 suffix to all binding projects to prevent future issues? The same has been done with the feature names.

kaikreuzer:
> Maybe we should add the 1 suffix to all binding projects to prevent future issues? The same has been done with the feature names.

I agree that would make sense as we do not always have to track the required adaptions once something gets merged in openhab2-addons.

PR #5516, opened 2018-03-14
[Pushbullet] Added ability to push notes to a channel

jsetton:
Added the ability to push notes to all subscribers of a channel based on the recipient format. Also, added the ability to push notes without specifying a recipient as the Pushbullet API will default to broadcast them to all the user account's devices.

### DIFF
--- a/bundles/action/org.openhab.action.pushbullet/README.md
+++ b/bundles/action/org.openhab.action.pushbullet/README.md
@@ -6,7 +6,10 @@ The Pushbullet action allows you to notify iOS, Android & Windows 10 Phone & Des
 
 The following is a valid action call that can be made when the plugin is loaded.
 For specific information on each item, see the [Pushbullet API](https://docs.pushbullet.com/).
+The recipient can either be an email address or a channel tag.
+If it is not specified or invalid, the note will be broadcast to all of the user account's devices.
 
+- `sendPushbulletNote(String title, String message)`
 - `sendPushbulletNote(String receiver, String title, String message)`
 - `sendPushbulletNote(String botname, String receiver, String title, String message)`
 
--- a/bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
+++ b/bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
@@ -18,6 +18,9 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 
+import javax.mail.internet.AddressException;
+import javax.mail.internet.InternetAddress;
+
 import org.openhab.action.pushbullet.internal.model.Push;
 import org.openhab.action.pushbullet.internal.model.PushResponse;
 import org.openhab.core.scriptengine.action.ActionDoc;
@@ -95,6 +98,14 @@ public class PushbulletAPIConnector {
         return sendPush(DEFAULT_BOTNAME, recipient, title, message, "note");
     }
 
+    @ActionDoc(text = "Sends a message to PushbulletAPIConnector to all user's devices from default bot")
+    public static boolean sendPushbulletNote(
+            @ParamDoc(name = "title", text = "title of the message") String title,
+            @ParamDoc(name = "message", text = "the message to be sent") String message) {
+        logger.debug("Trying to send a note to all user's devices from the default bot");
+        return sendPush(DEFAULT_BOTNAME, null, title, message, "note");
+    }
+
     /**
      * Inner method handling all the pushes.
      *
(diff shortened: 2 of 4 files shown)


bundles/action/org.openhab.action.pushbullet/README.md
```diff
@@ -6,7 +6,9 @@ The Pushbullet action allows you to notify iOS, Android & Windows 10 Phone & Des
 
 The following is a valid action call that can be made when the plugin is loaded.
 For specific information on each item, see the [Pushbullet API](https://docs.pushbullet.com/).
+The recipient can either be an email address or a channel tag. If it is not specified or invalid, the note will be broadcast to all of the user account's devices.
```
hakan42: Tiny detail, could you please make sure that every sentence is on its own line?

hakan42:
Looks fine to me. Could you perhaps open and clonse this PR so Jenkins rebuilds it?

The javax.measure error could be caused by the recent Units of Measure feature that came in with the ESH core. Failing a successful build, could you rebase this branch against master and see if that fixes it?

bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
```diff
@@ -126,7 +133,16 @@ private static boolean sendPush(String botName, String recipient, String title,
...
-        push.setEmail(recipient);
+
+        if (recipient != null) {
+            if (isValidEmail(recipient)) {
+                logger.trace("    Recipient is an email address");
+                push.setEmail(recipient);
+            } else if (isValidChannel(recipient)) {
+                logger.trace("    Recipient is a channel tag");
+                push.setChannel(recipient);
+            }
```
9037568: else  ?

What happens if neither of these is satisfied? Shouldn't this then throw?

bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
```diff
@@ -126,7 +133,16 @@ private static boolean sendPush(String botName, String recipient, String title,
         push.setTitle(title);
         push.setBody(body);
         push.setType(type);
-        push.setEmail(recipient);
+
+        if (recipient != null) {
```
9037568: else ?

What if recipient is null? Shouldn't there be a throw in that case?

bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
jsetton: If neither of these is satisfied, no specific target will be included in the Pushbullet API query.

Per the [API documentation](https://docs.pushbullet.com/#target-parameters), if no target is specified, the notification will be broadcast to all of the user's devices, which is considered as the default behavior.

bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
jsetton: Same reason as above. If the recipient is null, we would fallback to the default behavior.

jsetton:
I have updated the README file per @hakan42 recommendation and also rebased the branch against master. All checks are now passing.

bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
jsetton: I added an log warn to let the user know that the provided recipient was invalid.

bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java
```diff
@@ -126,7 +136,19 @@ private static boolean sendPush(String botName, String recipient, String title,
...
+            if (isValidEmail(recipient)) {
+                logger.trace("    Recipient is an email address");
+                push.setEmail(recipient);
+            } else if (isValidChannel(recipient)) {
+                logger.trace("    Recipient is a channel tag");
+                push.setChannel(recipient);
+            } else {
+                logger.warn("Invalid recipient: {}", recipient);
```
9037568: These two should be DEBUG, not WARN.

jsetton: Shouldn't we let the user know that he specified an invalid recipient? Sending these two in DEBUG will hide this information.

9037568: I overlooked the distinction between 'not specified' and 'specified, but invalid'.
With the view that this warning is for the latter case, I agree with you.

9037568:
Thanks, @jsetton !

PR #5532, opened 2018-03-28
Minor update to readme with Image Item example

BasvanH:
Added example for sending Image Item with telegram action. I had some diffeculty finding the way to do it, so adding it to the readme would possibly help others.

### DIFF
--- a/bundles/action/org.openhab.action.telegram/README.md
+++ b/bundles/action/org.openhab.action.telegram/README.md
@@ -126,3 +126,16 @@ then
     sendTelegramPhoto("bot1", base64Image, "sent from Openhab")
 end
 ```
+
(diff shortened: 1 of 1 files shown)


BasvanH:
Please hold, found out it's not working as expected.

bundles/action/org.openhab.action.telegram/README.md
```diff
@@ -126,3 +126,16 @@ then
     sendTelegramPhoto("bot1", base64Image, "sent from Openhab")
 end
 ```
+
+If you would like to send a Image Item.
```
9037568: "an ImageItem"

9037568:
Thanks, @BasvanH !

PR #5547, opened 2018-04-08
[FritzboxTr064] Fixed some examples in FritzBox TR064-Readme

Hunv:
Adjusted presence.map content to real values
Changed <present> icon to <presence>
Added remark to install Map Transformation Addon for using mapping of presence.map
Fixed variable in item configuration

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/README.md
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/README.md
@@ -118,8 +118,6 @@ Use a map for presence detection item:
 Create file `transform/presence.map` and add:
 
 ```
-1=present
-0=not present
 OPEN=present
 CLOSED=not present
 NULL=unknown
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/README.md
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/README.md
@@ -128,7 +128,7 @@ NULL=unknown
 Now, as item configuration use (the addon "Map Transformation" must be installed):
 
 ```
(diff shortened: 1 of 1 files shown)


9037568:
Please explain these changes.

Hunv:
I just installed this binding and recognized that the documentation is not correct.
1. The real values of the call state are "OPEN", "CLOSED" and "NULL" (before the first status change)
2. I didn't know about the Map Transformation plugin, so I struggled with the "MAP"-stuff a while. The hint will save some time for other newbees
3. I am not really aware why and what the difference is, but the %d is not working while %s is working in the item definition
4. the Iconname is "presence" not "present"

9037568:
Do @tensberg, @gitbock, or @staehler have any comments regarding this proposed change?

tensberg:
I never tried this out myself until today. The config works nicely with the corrections Hunv has made. Thanks @Hunv!

Before this is merged, the 1=present / 0=not present can be removed from the README, it's obsolete.

FYI Hunv regarding 3: %d is for decimal numbers, while %s is for strings, this is why %s is the right one for the real values.

bundles/binding/org.openhab.binding.fritzboxtr064/README.md
```diff
@@ -120,12 +120,15 @@ Create file `transform/presence.map` and add:
 ```
 1=present
```
tensberg: I thisnk you should remove the 1 and 0 lines while you're at it, these values are not used.

Hunv:
Hi, I updated the file and removed the old 0 and 1 for the presence.map-file.

9037568:
Thanks @Hunv !

PR #5551, opened 2018-04-11
change IndustrialDigitalOutSubIDs

Rosi2143:
add check for
* OUT1_VALUE
* OUT2_VALUE
* OUT3_VALUE
to IndustrialDigitalOutSubIDs

### DIFF
--- a/bundles/binding/org.openhab.binding.tinkerforge/src/main/java/org/openhab/binding/tinkerforge/internal/model/IndustrialDigitalOutSubIDs.java
+++ b/bundles/binding/org.openhab.binding.tinkerforge/src/main/java/org/openhab/binding/tinkerforge/internal/model/IndustrialDigitalOutSubIDs.java
@@ -204,6 +204,12 @@ public enum IndustrialDigitalOutSubIDs implements Enumerator {
         switch (value) {
             case OUT0_VALUE:
                 return OUT0;
+            case OUT1_VALUE:
+                return OUT1;
+            case OUT2_VALUE:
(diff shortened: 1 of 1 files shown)


9037568:
This change fails to compile. Also, from what I can see, the file being modified here is auto-generated and not meant to be edited.

@theoweiss, any input?

theoweiss:
Chris you are right this is generated code which is not meant to be edited. 
It's defined here:
https://github.com/openhab/openhab1-addons/blob/18ba124225ff4292d42be76b79a08f60aa07d501/bundles/binding/org.openhab.binding.tinkerforge/model/tinkerforge.xcore#L1327

It's obvious that the code generation did not the right thing here. But I'm not sure if the method changed by this PR is really used by the binding. @Rosi2143 It would be good to know which bug this PR should fix. Any description and/or log entries would be helpful. Is there an existing issue? If not please open an issue.

BTW: I worked on the binding some weeks ago to add new devices but was not able to reliable generate valid code. I've tried diverse solutions and workarounds and asked the XCore (this is the framework I used) developers for help but with no success. Therefore I decided to start working on the reimplementation of the binding  as openHAB 2 add-on without using XCore. Therefore only bugfixes to the openhab1 tinkerforge binding would be possible which do not require a regeneration of the model code.

TL;DR
Close this PR without merging and open an issue with a bug description.

9037568:
@Rosi2143 please open a new issue if you have a problem that this PR was intended to solve.

PR #5560, opened 2018-04-24
 [openenergymonitor] Added value cache to minimize item updates

paulianttila:


9037568:
Thanks, @paulianttila !

PR #5562, opened 2018-04-25
[TCP Binding][Documentation] Fix wording of UDP binding

jpmens:
There was a three-year old note in the documentation which nobody had clarified. I am not a developer, but just stumbled over the documentation simply not matching what 2.3 does so my wording fixes at least that.

However: a developer urgently needs to address whether all of this is still true ...

closes https://github.com/openhab/openhab-docs/pull/670


9037568:
Probably the same as #5593, but I have not verified with UDP.

PR #5564, opened 2018-04-28
Add the Samsung TV Binding to the legacy bindings

martinvw:
Because the old binding uses a different protocol the TV's supported by both bindings might vary

Signed-off-by: Martin van Wingerden <martinvw@mtin.nl> (github: martinvw)

### DIFF
--- a/features/openhab-addons-legacy/src/main/feature/feature.xml
+++ b/features/openhab-addons-legacy/src/main/feature/feature.xml
@@ -122,6 +122,12 @@
     <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.rwesmarthome/${project.version}</bundle>
   </feature>
 
+  <feature name="openhab-binding-samungtv1" description="Samsung TV Binding (1.x)" version="${project.version}">
+    <feature>openhab-runtime-base</feature>
+    <feature>openhab-runtime-compat1x</feature>
+    <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.samungtv/${project.version}</bundle>
+  </feature>
+
   <feature name="openhab-binding-satel1" description="Satel Binding (1.x)" version="${project.version}">
     <feature>openhab-runtime-base</feature>
     <feature>openhab-runtime-compat1x</feature>


kaikreuzer:
Ah, sorry, I just read:

> Looks like there was never a feature for the openHAB 1 Samsung TV Binding

So all is fine then.

martinvw:
If was my first time on this file should I also have added:

```
<configfile finalname="${openhab.conf}/services/samsungtv.cfg" override="false">mvn:${project.groupId}/openhab-addons-external/${project.version}/cfg/samsungtv</configfile>
```

kaikreuzer:
Yes, I missed that (it seem to be missing for rwesmarthome as well...).

Please check the docs on the exact steps that need to be done: https://docs.openhab.org/developers/development/compatibilitylayer.html#how-to-add-a-successfully-tested-1x-add-on-to-the-distribution

martinvw:
> RFXCOM binding seems to be missing as well

Indeed, but that one ports all functionalities to OH2 so I would not propose to add it? Wdyt?

martinvw:
> Please check the docs on the exact steps that need to be done: 

Now everyone is awake and I have a dentist appointment.

Tonight will be about SamsungTV bindings and GDPR :-)

paulianttila:
>Indeed, but that one ports all functionalities to OH2 so I would not propose to add it? Wdyt?

I have understand that there are still many users which use openHAB 1 as transition is always painful. If openHAB 2 support all legacy bindings, it makes transition much more easier to users as they don't need convert configurations immediately, right?

kaikreuzer:
> I have understand that there are still many users which use openHAB 1 as transition is always painful. 

If they are using openHAB 1, there is no need to add it to the legacy binding feature - because this is the metadata that openHAB2 needs in order to include it in the openHAB2 distro.

And as nobody so far requested/tested the RFXCOM 1.x binding with openHAB 2 (otherwise it would already be included), there does not seem to be any real need for it. Also note that including it means that someone has successfully tested it and confirms that it works nicely (see https://docs.openhab.org/developers/development/compatibilitylayer.html#how-to-use-openhab-1x-add-ons-that-are-not-part-of-the-distribution)

features/openhab-addons-legacy/src/main/feature/feature.xml
```diff
@@ -122,6 +122,12 @@
     <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.rwesmarthome/${project.version}</bundle>
   </feature>
 
+  <feature name="openhab-binding-samungtv1" description="Samsung TV Binding (1.x)" version="${project.version}">
```
kaikreuzer: When you are dealing with it tonight, you might also want to replace "samungtv1" with "samsungtv1", so that [the build won't break](https://openhab.ci.cloudbees.com/job/PR-openHAB-Distribution/522/console) any longer ;-)

features/openhab-addons-legacy/src/main/feature/feature.xml
```diff
@@ -122,6 +122,12 @@
...
+    <feature>openhab-runtime-base</feature>
+    <feature>openhab-runtime-compat1x</feature>
+    <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.samungtv/${project.version}</bundle>
```
kaikreuzer: Especially here :-)

paulianttila:
If I remember correctly, there was a compatibility list in very beginning, where developers/users was reporting if OH1 bindings was successfully run in OH2. And I reported both samsungtv and rfxcom be compatible.

I think the reason why e.g. rfxcom binding and also this samsung tv binding was removed was that they caused some problems as bundle names was same in both OH1 and OH2 (or something like that). Both bindings were introduced in 2015, so  OH2 was under heavy development.

kaikreuzer:
> And I reported both samsungtv and rfxcom be compatible.

Ok, if they work, they should/can be added for sure.

I cannot remember if they were added and removed again - the bundle names should be no issue, we have the same situation for all legacy bindings...

paulianttila:
> I cannot remember if they were added and removed again - the bundle names should be no issue, we have the same situation for all legacy bindings...

Not nowadays, but in that time there was problems. Was e.g. Karaf even introduced in that time?

features/openhab-addons-legacy/src/main/feature/feature.xml
martinvw: I hope this was not the reason my 1.x tests did fail, maybe I will look into it later but for now it becomes to me unlikely that the binding was part of the distribution recently and I get back to thinking I broke something :-)

PR #5565, opened 2018-04-28
Revert "Add the Samsung TV Binding to the legacy bindings"

martinvw:
Reverts openhab/openhab1-addons#5564

Sorry. A short test suggests that its not working so I rather not spent time on this but on making the OH2 binding behave properly :-) 

Signed-off-by: Martin van Wingerden <martinvw@mtin.nl>

### DIFF
--- a/features/openhab-addons-legacy/src/main/feature/feature.xml
+++ b/features/openhab-addons-legacy/src/main/feature/feature.xml
@@ -122,12 +122,6 @@
     <bundle start-level="80">mvn:org.openhab.binding/org.openhab.binding.rwesmarthome/${project.version}</bundle>
   </feature>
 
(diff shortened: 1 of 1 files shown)


martinvw:
Note that I also have a PR to properly include it but I could net the configuration to function properly.

It did not pick up my `services/samsungtv.cfg`

kaikreuzer:
> It did not pick up my services/samsungtv.cfg

Yes, because you missed adding the cfg file here: https://github.com/martinvw/openhab/blob/21dac9e402deda4f6382d78cfd8d8076e62a3ddf/features/openhab-addons-external/pom.xml#L32-L170

martinvw:
I mean at runtime not a build time, I assume this was/is only to package an example file, is that correct?

kaikreuzer:
Yes, to package the file that contains the parameters with documentation.
Ok, if it isn't working at all at runtime, then we should not add the feature.

PR #5577, opened 2018-05-21
[ComfoAir][Documentation] - Add undocumented features

### LINKED ISSUE #521: Flexible HTML5-based UI for tablets
_From [kai.openhab](https://code.google.com/u/kai.openhab/) on March 29, 2011 09:55:19_

The current WebApp.Net UI of openHAB has a row-based approach which works fine for smartphones, but does not use the screen of tablets efficiently.

The bigger tablet screens require a more flexible UI framework and configuration possibilities. SenchaTouch might be a good candidate to use here.

It should be tried to make use of the openHAB sitemap metamodel here as well, which possibly will have to be extended by some new properties and widget types. Ideally, currently existing sitemaps should render in a nice way on the new UI as well, so they can be used for both UIs at the same time.

_Original issue: http://code.google.com/p/openhab/issues/detail?id=24_

Maurits28:
I added the device commands from the [issue](https://github.com/openhab/openhab-docs/issues/521) opened by binderth. 
Also rearranged the sequence to make it more logical.

Closes openhab docs issue #521.

Signed-off-by: Maurits Jaspers <mauritsjas@gmail.com> (github: Maurits28)

### DIFF
--- a/bundles/binding/org.openhab.binding.comfoair/README.md
+++ b/bundles/binding/org.openhab.binding.comfoair/README.md
@@ -40,9 +40,18 @@ where `<device-command>` is replaced with the ComfoAir command from the list bel
 | incomming_fan                   | Read         |  |
 | outgoing_fan                    | Read         |  |
 | filter_running                  | Read         |  |
+| filter_error                    | Read         |  |
 | filter_error_intern             | Read         |  |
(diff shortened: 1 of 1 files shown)


9037568:
Thanks, @Maurits28 !

PR #5582, opened 2018-05-27
[Weather][ForecastIo] Get UV-Index and Wind Gust from forecast.io

azielke:
This patch enables uvindex and wind gust values from forecast.io


9037568:
Thanks, @azielke !

PR #5591, opened 2018-06-03
[MapDB] Fix new items not immediately persisted, upgrade to mapdb-1.0.9

wborn:
* Fix new items not immediately persisted. Without this fix the database transaction would only be committed when the state of an existing item changes and is stored again.
* Upgrade to mapdb-1.0.9 (also used in ESH). For the changelog see also: http://www.mapdb.org/changelog-archive

---

See also these community threads where people most likely run into this issue:
* [mapDB: problem with restoreOnStartup](https://community.openhab.org/t/mapdb-problem-with-restoreonstartup/11995?u=wborn)
* [Getting mapdb persistence to work with openhab2](https://community.openhab.org/t/getting-mapdb-persistence-to-work-with-openhab2/17998?u=wborn)


9037568:
Thanks, @wborn !

PR #5594, opened 2018-06-08
[TCP Binding][Documentation] Clarify requirement of services/files

jpmens:
UDP/TCP bindings don't do anything without services/ files. Clarify that. (I am not a developer; I determined this empirically...)

(and BTW: didn't I patch this already in some other respository recently ....... ?)

### DIFF
--- a/bundles/binding/org.openhab.binding.tcp/README.md
+++ b/bundles/binding/org.openhab.binding.tcp/README.md
@@ -10,7 +10,7 @@ The TCP & UDP Bindings act as a network client or as a network server.
 
 The TCP and UDP bindings can be configured in the files `services/tcp.cfg` and `services/udp.cfg`, respectively.
 
-> Note: This is optional for the configuration and not necessary for receiving data. Item-defintions are enough for receiving data. (Developer confirm? 20150128). There's a bug in the binding that requires at least one udp configuration to be defined or the binding will not send UDP messages.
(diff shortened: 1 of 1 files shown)


jpmens:
Yes, I did already do this https://github.com/openhab/openhab1-addons/pull/5562

9037568:
>UDP/TCP bindings don't do anything without services/ files.

This statement does not appear to be accurate. I was able to get a TCP socket listening with no tcp.cfg file.

jpmens:
> This statement does not appear to be accurate 

"does not appear" doesn't appear to be very accurate. Either it works or it doesn't and my experience is it doesn't.

jpmens:
So the text in the current binding documentation, including the note to ask developers whether that’s actually true, remains as it is? ~~I wish I felt like actually tracking down how long that paragraph’s been in there, but I don’t.~~ That note’s been in there since January 2017 if I’m not mistaken. Sad.

9037568:
This makes it sound like there's a problem with the update logic.

My guess is that you'd also see the port opened if this:

````
$ echo port=9002 | sudo tee services/tcp.cfg
port=9002
````

was changed to this:

````
$ echo port=9002 | sudo tee services/tcp.cfg
yada=yada
````

9037568:
Closing for lack of activity.

jpmens:
Whose lack of activity is responsible for this being closed?

jpmens:
Never mind; I am no longer interested in this; half a year for a patch to two sentences is too much for me.

PR #5596, opened 2018-06-09
[MQTT] Fix UTF-8 encoding issues.

### LINKED ISSUE #5569: [MQTT] special characters not supported
## Expected Behavior
Sending a message via MQTT which contains special characters such as German umlaut or accented characters should be converted to the correct enconding for openHAB.

## Current Behavior
Currently those messages are not received correctly. As an example when I send any text to my Echo Dot via MQTT and openHAB by sending it as topic `openHAB/in/RenesEchoDot_Erinnere/command` the message is not received correctly. Symptom: the Dot does not read the word which contains the umlaut but rather spells it and omits the umlaut. The Alexa App on my phone shows the umlaut as a question mark in a black square.

This does not happen when I trigger a reminder in a rule using `RenesEchoDot_Erinnere.sendCommand` using the exact same text. In this case the app shows everything correctly and the Dot reads everything including umlauts correctly.

The same behavior has been described in the openHAB forum at https://community.openhab.org/t/incorrect-character-encoding-in-mqtt-binding/39889

## Possible Solution
I guess this has something to do with what is described here https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_9.0.0/com.ibm.mq.pro.doc/q003040_.htm

> The MQTT Java™ client has a convenient MqttMessage.toString method. The method treats the message payload as being encoded in the local platform default character set, which is generally UTF-8. It converts the payload to a Java String. Java has a String method, getBytes that converts a string into a byte array encoded using the local platform default character set. Two MQTT Java programs exchanging text in the message payload, between platforms with the same default character set do so easily and efficiently in UTF-8.

## My Environment

* Version used: openHAB 2.2, MQTT binding 1.11.0
* Environment name and version: Java 8
* Operating System and version: Debian 8.10

9037568:
Also clean up logging, eliminate a warning, and fix a few common programming errors.

Fixes #5569.

### DIFF
--- a/bundles/binding/org.openhab.binding.mqtt/META-INF/MANIFEST.MF
+++ b/bundles/binding/org.openhab.binding.mqtt/META-INF/MANIFEST.MF
@@ -23,6 +23,7 @@ Bundle-SymbolicName: org.openhab.binding.mqtt
 Bundle-DocURL: http://www.openhab.org
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Service-Component: OSGI-INF/mqtt-*.xml
+Bundle-ActivationPolicy: lazy
 Bundle-ClassPath: .
 Require-Bundle: org.openhab.io.transport.mqtt;bundle-version="1.3.0"
 Export-Package: org.openhab.binding.mqtt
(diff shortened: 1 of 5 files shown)


wawa79:
I installed latest build and still the charset issue. See what is logged in the console when I send a command to change value of  Echo_Living_Room_TTS to 'Ceci est un test accentué':
`22:12:11.881 [INFO ] [smarthome.event.ItemStateChangedEvent] - Echo_Living_Room_TTS changed from  to Ceci est un test accentuÃ©`

9037568:
>I installed latest build and still the charset issue.

First off, the latest build won't have this change yet (assuming you mean the builds from cloudbees).
Second, this won't change the Karaf console output. For that you'll have to log a [defect against ESH](http://github.com/eclipse/smarthome/issues/).

wawa79:
The JAR I installed came from https://openhab.jfrog.io/openhab/libs-pullrequest-local/
You are right: the console might not change its display. The important thing is that MQTT binding get fixed. (as it currently crashes my Evenghost MQTT client when it receive the inconsistent 'state' change message from OpenHAB)

PR #5607, opened 2018-06-21
[vdr1] change logLevel for could not connect to DEBUG

udo1toni:
VDR is not a typical always on service, so it's very common that vdr is not available.

Signed-off-by: Udo Hartmann udo1toni@gmail.com


9037568:
Thanks, @udo1toni !

PR #5615, opened 2018-07-02
[EpsonProjector] Adjustments for direct projector TCP/IP connection implemented

tbraune42:
Until now it was only possible to operate an Epson projector with a SERIAL2TCP adapter. Because of my changes this now also goes without adapter


MaroXausa:
how to install you're binding in openhab?

svedue:
Any news on this? Will this be part of 2.4?
I'd love to use it. 
Is there any easy way to install as local addon?

9037568:
Has there been any community testing of this change?

9037568:
@svedue were you using this binding before this change?

tbraune42:
The addon has already been tested and works so far. I wouldn't know why you shouldn't take this into the repro

Issue #5623, opened 2018-07-19
[Zibase] garbled messages


bchabrier:
Debug traces show garbled messages from the Zibase, as in:

> 2018-07-17 13:09:17.503 [DEBUG] [inding.zibase.internal.ZibaseBinding] - ZIBASE MESSAGE: Received radio ID (<rf>ZWAVE ZB10</rf> <dev>Info</dev> Class=0x<cla>60</cla> Op=0x<op>0D</op> data (I3...Ix)=0x03 03 31 05 01 44 00 00 0A 53  Batt=<bat>Ok</bat>): <id>I600D_ZB10</id>
> 2018-07-17 13:09:17.510 [DEBUG] [inding.zibase.internal.ZibaseBinding] - Found event from ID null
> 2018-07-17 13:09:35.904 [DEBUG] [inding.zibase.internal.ZibaseBinding] - ZIBASE MESSAGE: Received radio ID (<rf>ZWAVE ZC7</rf> <dev>Low-Power Measure</dev> Total Energy=<kwh>8.9</kwh>kWh Power=<w>00</w>W Batt=<bat>Ok</bat>): <id>PZC7</id>t>Ok</bat>): <id>I600D_ZB10</id>

It seems that the previous message appears as if the buffer was not cleared.

## Expected Behavior

The following output would be expected:
> 2018-07-17 13:09:17.503 [DEBUG] [inding.zibase.internal.ZibaseBinding] - ZIBASE MESSAGE: Received radio ID (<rf>ZWAVE ZB10</rf> <dev>Info</dev> Class=0x<cla>60</cla> Op=0x<op>0D</op> data (I3...Ix)=0x03 03 31 05 01 44 00 00 0A 53  Batt=<bat>Ok</bat>): <id>I600D_ZB10</id>
> 2018-07-17 13:09:17.510 [DEBUG] [inding.zibase.internal.ZibaseBinding] - Found event from ID null
> 2018-07-17 13:09:35.904 [DEBUG] [inding.zibase.internal.ZibaseBinding] - ZIBASE MESSAGE: Received radio ID (<rf>ZWAVE ZC7</rf> <dev>Low-Power Measure</dev> Total Energy=<kwh>8.9</kwh>kWh Power=<w>00</w>W Batt=<bat>Ok</bat>): <id>PZC7</id>

## Possible Solution

Fix proposed in #5624, cleaning the buffer.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

[...]

PR #5624, opened 2018-07-19
[Zibase] Cleaning the zibase message buffer: fix of #5623

bchabrier:
This is a fix of issue #5623: garbled zibase messages

The buffer was indeed not cleared, causing the previous messages to remain visible if longer.

### DIFF
--- a/bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java
+++ b/bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java
@@ -153,6 +153,12 @@ public class ZibaseListener extends Thread {
                     ZbResponse zbResponse = new ZbResponse(receivePacket.getData());
                     logger.debug("ZIBASE MESSAGE: " + zbResponse.getMessage());
                     publishEvents(zbResponse);
+
+                    // reset buffer
+                    for (int i = 0; i < receiveData.length; i++) {
+                        receiveData[i] = 0;
+                    }
+
                 }
                 zibase.hostUnregistering(listenerHost, listenerPort);
 
--- a/bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java
+++ b/bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java
@@ -13,6 +13,7 @@ import java.net.DatagramPacket;
 import java.net.DatagramSocket;
 import java.net.SocketException;
 import java.net.UnknownHostException;
+import java.util.Arrays;
 import java.util.Vector;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java
```diff
@@ -153,6 +153,12 @@ public void run() {
                     ZbResponse zbResponse = new ZbResponse(receivePacket.getData());
                     logger.debug("ZIBASE MESSAGE: " + zbResponse.getMessage());
                     publishEvents(zbResponse);
+
+                    // reset buffer
```
9037568: This is really the wrong way to do this.

What should be done is to use the length of the received data to create a zbResponse from the correct set of bytes.

bchabrier: Thanks a lot for your prompt review!

Unfortunately the length of the received data is not known. `receivePacket` is filled in by the socket receive, and as it is not clean, there is no way to know the number of bytes received.

bchabrier: After investigation, zbReponse expects a byte array of 470 bytes. So I could create one each time:
>                   byte[] readData = new byte[470];
>                   System.arraycopy(receivePacket.getData(), 0, readData, 0, receivePacket.getLength());
>                   ZbResponse zbResponse = new ZbResponse(readData);

However, I don't think it is that more efficient than the reset buffer approach as it allocates a new byte array for each new message...

What would you advise?

9037568: I'm ok with your Arrays.fill optimization. Not sure if it'd be more efficient to just write all 470 bytes instead of checking the length; probably not a lot of difference either way.

9037568:
Need to rerun Jenkins...

9037568:
Apparently there's a missing import:

````
01:47:21.926  |  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.openhab.binding.zibase: Compilation failure: Compilation failure:
01:47:21.927  |  [ERROR] /scratch/jenkins/workspace/PR-openHAB1-Addons/bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java:[158,1] 1. ERROR in /scratch/jenkins/workspace/PR-openHAB1-Addons/bundles/binding/org.openhab.binding.zibase/src/main/java/org/openhab/binding/zibase/internal/ZibaseListener.java (at line 158)
01:47:21.927  |  [ERROR] Arrays.fill(receivePacket.getData(), 0, receivePacket.getLength(), (byte) 0);
01:47:21.927  |  [ERROR] ^^^^^^
01:47:21.927  |  [ERROR] Arrays cannot be resolved
````

9037568:
Thanks, @bchabrier !

Issue #5334, opened 2017-10-19
[dynamodb] Instructions for completing the validation for inclusion in the openHAB2 distribution


ssalonen:
The [Persistence docs](http://docs.openhab.org/addons/persistence.html) say that:
> Some openHAB 1 persistence services have not yet completed validation for inclusion in the distribution; however, they may indeed work properly under openHAB 2.

How to complete this validation in practice?

I have personally tested the version in PR #5333 quickly using openHAB2, and it seems to work fine... Is there more guidance on the testing? I wonder why so many persistence services have been left out from the distribution?

Should I as the dynamodb addon developer move this addon to use smarthome packages? I notice that all (?) persistence addons still rely on compat layer?

## Context

Community discussion re missing persistence services 
https://community.openhab.org/t/amazon-dynamodb-persistence-binding-not-in-the-addons-download/17616/2?u=ssalonen

Also mentioned here: https://community.openhab.org/t/new-amazon-dynamodb-persistence-addon-alpha-experimental/9856

## Your Environment

* Version used: (e.g., openhab and addon versions)
* Environment name and version (e.g. Chrome 39, node.js 5.4, Java 8, ...):
* Operating System and version (desktop or mobile, Windows 10, Raspbian Jessie, ...):

9037568:
I'm not clear on where these would migrate _to_. OH2 addons or ESH?

ssalonen:
Thanks for the reply! I think I used a bit wrong wording... There are two topics as addon developer which are unclear to me:
- what should I do to "complete the validation" for getting this persistence addon as part of openHAB2 distribution?
- this persistence bundle (as all other persistence bundles in openHAB1-addons) do not use smarthome packages, and thus need the openHAB1 compatibility layer to be installed? If it would use smarthome packages, it indeed would make sense to move this somewhere else.

From my perspective it would be OK to move this to even ESH, I think the licenses should be clear for the actual addon code. Naturally there are the dependencies in lib with different license.

9037568:
I don't know the answers to these questions. I think we need input from someone better connected, like @kaikreuzer or @watou.

kaikreuzer:
Hi @ssalonen, I guess this is the info you are looking for: http://docs.openhab.org/developers/development/compatibilitylayer.html#how-to-add-a-successfully-tested-1x-add-on-to-the-distribution

ssalonen:
@kaikreuzer thanks! So in order to "complete the validation" I should follow the steps to create
[...]

PR #5635, opened 2018-08-02
[dynamodb] Introduce Karaf support

ssalonen:
This PR introduces the changes discussed in #5334. I highly recommend reading that thread for background.

Currently DynamoDB is not part of the official addon bundle since it not supporting up-to-par with new openHAB2 packaging requirements. It is working though without any issues using textual configuration.

To meet openHAB2 requirements, this PR introduces:
- Karaf feature `openhab-persistence-dynamodb`
- PaperUI configuration support (actually I think this might the first persistence to have this)

Furthermore, `service.pid` is now declared, which seems to help with the textual configuration (in case someone prefers it over UI). I also had to adjust some OSGI parameters to make configuration reloading work as it should with the UI (once again, details in #5334).

I hope that with these changes, the DynamoDB persistence would be once again as part of the official bundle.


9037568:
I don't see any problems, but I'd like to see if @maggu2810 or @kaikreuzer have any input...

PR #5636, opened 2018-08-03
moved modbus1 binding to legacy now that we have a 2.x version

kaikreuzer:


9037568:
Thanks, @kaikreuzer !

PR #5647, opened 2018-08-22
[velux] Support for Velux Solar Blinds added

gs4711:
New product supported by this binding: solar blinds.


9037568:
Thanks, @gs4711 !

PR #5649, opened 2018-08-29
moved OneWire binding to legacy add-ons

kaikreuzer:


martinvw:
Hello @kaikreuzer,

I see that you moved the onewire binding to the legacy bindings, what is the openHAB 2 alternative?

We got this issue:

https://github.com/openhab/openhab2-addons/issues/3943#issuecomment-421717835

> ... - based on the documentation - OneWire and OneWireGPIO seems to be different bindings. First on is using One Wire Server, second is using only local IO ports.

martinvw:
Okay solved, it was moved to ESH instead :-)

PR #5655, opened 2018-09-06
[fritzboxtr064] fixed name resolution of address book entries

andre77:
most important thing here is the trimming of the address book entries. sometimes there are some special characters like linebreaks, which did not match correctly.
additionally there is a little refactoring to avoid boilerplate code


andre77:
@kaikreuzer thanks for merging
currently i'm using the OH version 2.3.0-SNAPSHOT Build #1240
the changes of the PR  are on the master branch, so what is actually necessary now to get a version over the repository to use it with my OH installation?

PR #5656, opened 2018-09-07
[fritzboxtr064] replaced the trim method, to get rid of white spaces in between the tel numbers

andre77:
fixed name resolution of address book entries.
the name resolution did not work 100% correct, if there are some white spaces in between the number, i.e. like this '+49 40 123 456 789'

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/PhonebookManager.java
+++ b/bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/PhonebookManager.java
@@ -55,7 +55,7 @@ public class PhonebookManager {
                 if (phoneBookTel == null) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


9037568:
Why? I don't see any defects stating that name resolution doesn't work.
Have you had any community testing for this change?

andre77:
just put some white spaces into a tel number inside your address book on the fritz box ( like this '+49 40 123 456 789') and make a call from this number. the current version will not find a match because of the white spaces

9037568:
Then don't put the whitespaces in.
I don't use or even have a fritzbox.

andre77:
@9037568 
its not forbidden to put white spaces into the tel numbers. it makes them even more readable. so why not?
most of the phones like for example iphone even add a white space automatically to separate the prefix of the tel number.
but why are you worried about the changes for a fritz box binding, if you even don't have one? :-)
or are u using the binding with another device, which supports the TR064 protocol?

9037568:
Thanks, @kaikreuzer, for the confirmation.
Thanks for the patch, @andre77 !

Issue #5658, opened 2018-09-09
[influxdb] InfluxDB persistence does not work with #1356 and #1357


sihui62:
With the snapshot versions mentioned above persistence for InfluxDB does not work anymore.

Related Community Thread:

https://community.openhab.org/t/snapshot-build-1357-items-persisted-by-influxdb-are-not-restored/51122

9037568:
This will probably have to be fixed in ESH. The persistence bundle here hasn't changed.

sihui62:
Thx, I think you are right. After writing this issue I realized that mqtt and caldav are not working properly, too. 
Reverting to snapshot #1338 solved the problem for me.

kaikreuzer:
The log says:
```
Tried to set invalid state xxx (StringType)
```
So although nothing has changed about InfluxDB persistence service, it seems as if it has always already posted wrong values and now it is only checked and noticed by the framework.
I'd claim it needs to be fixed in the InfluxDB service.

sihui62:
> Maybe there's a startup order issue that the itemRegistry is still missing?

That was one of my other problems with # 1356 regarding the caldav binding: log message "item registry not set" on default logging level.
Also my mqtt action switch was sending "1" regardless it I set the switch to ON or OFF where it normally sends ON/OFF.
No idea if it has anything to do with it, so I wanted just to mention it. Reverting back to an older snapshot #1338 solved all problems.

kaikreuzer:
@sihui62 I have no good idea what might have changed the startup behavior, but in any case the code seems to be not correct and thus the whole construct is fragile. https://github.com/openhab/openhab1-addons/pull/5659 should hopefully address this for the InfluxDB service. Other add-ons will have to be treated independently.

kaikreuzer:
@sihui62 I have meanwhile found the real cause for this problem - and have created a fix for it with https://github.com/openhab/openhab-core/pull/397. All should be fine with tonight's snapshot again, I would hope.

PR #5659, opened 2018-09-10
Make the ItemRegistry a static reference

kaikreuzer:
Fixes https://github.com/openhab/openhab1-addons/issues/5658


9037568:
I don't have a clue what this does... but ok, if it fixes the problem.

ssalonen:
So the static reference is the important thing? Or the cardinality change from 0..1 to 1..1?

PR #5665, opened 2018-09-19
[mysql-persistence] Do not log plain passwords

martinvw:
Fixes: https://github.com/openhab/openhab-distro/issues/762

Signed-off-by: Martin van Wingerden <martin@martinvw.nl>

### DIFF
--- a/bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java
+++ b/bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java
@@ -553,7 +553,7 @@ public class MysqlPersistenceService implements QueryablePersistenceService {
         } catch (Exception e) {
             logger.error(
                     "mySQL: Failed connecting to the SQL database using: driverClass={}, url={}, user={}, password={}",
-                    driverClass, url, user, password, e);
+                    driverClass, url, user, password.replaceAll(".", "*"), e);
         }
     }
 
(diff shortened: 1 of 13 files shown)


bundles/persistence/org.openhab.persistence.mysql/java/org/openhab/persistence/mysql/internal/MysqlPersistenceService.java
```diff
@@ -553,7 +553,7 @@ private void connectToDatabase() {
         } catch (Exception e) {
             logger.error(
                     "mySQL: Failed connecting to the SQL database using: driverClass={}, url={}, user={}, password={}",
-                    driverClass, url, user, password, e);
+                    driverClass, url, user, password.replaceAll(".", "*"), e);
```
kaikreuzer: You are still exposing some information about the password here (its length), which should also be avoided, see the comment at https://github.com/eclipse/smarthome/issues/4583#issuecomment-349715045.

So I'd suggest to either put 8 asterisks or maybe simply remove the password part completely here as it anyhow is not helpful.

martinvw: @kaikreuzer, I like that, shall I also remove it then in all other places in this repo.

Should I also apply it for trace logging? or is that configurable and thus okay.

kaikreuzer: Yes, can be removed from other places as well from the logging.
I'd also remove it from trace statements.

martinvw:
Hi @kaikreuzer, its updated!

CrispinP:
Thank you

martinvw:
@CrispinP thank you for reporting it. 

Note that I also created follow up issue for openHAB2 addons

https://github.com/openhab/openhab2-addons/issues/3990

PR #5671, opened 2018-09-29
[Persistence][Documentation] Small change to highlight the mandatory use of the everyMinute strategy

vzorglub:
I have hignlighted the fact that the `everyMinute` strategy must be used,
There have been a number of recent queries regarding non persisted data with the rrd4j database where the user had not set the everyMinute strategy.


9037568:
Thanks, @vzorglub !

PR #5678, opened 2018-10-09
added "1.x" to mqtt binding label

kaikreuzer:
Related to https://github.com/openhab/openhab-distro/pull/781
Not yet moving it to legacy addons - will do this once we can confirm that the new binding is fully integrated and working.


9037568:
Thanks, @kaikreuzer !

PR #5680, opened 2018-10-20
[Fritzbox][Documentation] update readme: provide config example, remove syntax error

stefan-muc:
When using FritzBox binding, I stumbled across those issues in documentation - would be good if those could be resolved.

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzbox/README.md
+++ b/bundles/binding/org.openhab.binding.fritzbox/README.md
@@ -59,6 +59,6 @@ The following items switch DECT, WIFI, GUEST_WIFI and the answering machine 0 (d
 ```
 Switch DECT {fritzbox="dect"}
 Switch WIFI {fritzbox="wlan"}
-Switch GWIFI {fritzbox="guestwlan"} (Version 1.7.0)
+Switch GWIFI {fritzbox="guestwlan"} // (Version 1.7.0)
 Switch TAM0 {fritzbox="tam0"}
 ```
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.fritzbox/README.md
```diff
@@ -10,6 +10,15 @@ This binding can be configured in the `services/fritzbox.cfg` file.
 | password |         |   Yes    | Password to your Fritz!Box |
 | user     |         | when user management is enabled | User of your Fritz!Box. It is a good practice to create an additional user for openHAB. |
 
+
+As a result, your config might look like the following:
```
9037568: Probably better to leave these examples out, since they're incorrect for the current version of OH.

bundles/binding/org.openhab.binding.fritzbox/README.md
```diff
@@ -50,6 +59,6 @@ The following items switch DECT, WIFI, GUEST_WIFI and the answering machine 0 (d
 ```
 Switch DECT {fritzbox="dect"}
 Switch WIFI {fritzbox="wlan"}
-Switch GWIFI {fritzbox="guestwlan"} (Version 1.7.0)
+Switch GWIFI {fritzbox="guestwlan"} // (Version 1.7.0)
```
9037568: Let's just drop the version comment entirely, since it's no longer relevant.

9037568 (changes requested):
Thanks for improving the docs. Please see my inline comments.

bundles/binding/org.openhab.binding.fritzbox/README.md
stefan-muc: If the example is wrong, then please modify it and explain it how it is done better?

Because the example was missing I had to dig Google to find https://github.com/openhab/openhab1-addons/wiki/fritz-box-binding to get which syntax the binding is using.

9037568: The correct way in OH2 is

````
ip=<ipaddress>
password=<password>
user=<username>
````

PR #5686, opened 2018-10-30
[CUL] Request credit info when opening device

hadogenes:
After #5329 the _credit10ms_ is 0 and it cannot send anything, because the credit status will be never updated

### DIFF
--- a/bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/AbstractCULHandler.java
+++ b/bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/AbstractCULHandler.java
@@ -164,6 +164,7 @@ public abstract class AbstractCULHandler<T extends CULConfig> implements CULHand
         
         registerListener(sendThread);
         sendThread.start();
+        requestCreditReport();
     }
 
     @Override
--- a/bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/AbstractCULHandler.java
+++ b/bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/AbstractCULHandler.java
@@ -253,8 +253,10 @@ public abstract class AbstractCULHandler<T extends CULConfig> implements CULHand
             return;
         } else if (data.matches("^\\d+\\s+\\d+")) {
             processCreditReport(data);
+            return;
         }
         notifyDataReceived(data);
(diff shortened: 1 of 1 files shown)


johgoe:
looks like a revert of #5329 but because the change in #5329 in method requestCreditReport() which queues the command instead of synchronuous writing, makes the differnce. 

@9037568 Can you care about this issue as you was the merger of #5329?

9037568:
I don't have a deep enough knowledge of CUL and this binding to be able to resolve these changes against the earlier ones in #5329.

I suggest hashing it out in the forum in either a new thread or [the original](https://community.openhab.org/t/cul-transport-fix-for-issue-with-cul-transport-race-condition/35385/8).

MooLMihaly:
Are there any new on the pull request? All CUL users are affected.

MooLMihaly:
Are there any news on the pull request? All CUL users are affected.

MichaelKlemm:
Looks fine to me. Why the PRs taking that long?

MichaelKlemm (approved):

bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/AbstractCULHandler.java
```diff
@@ -252,8 +253,10 @@ protected void processNextLine(String data) {
             return;
         } else if (data.matches("^\\d+\\s+\\d+")) {
             processCreditReport(data);
+            return;
```
MichaelKlemm: This return prevents "notifyDataReceived" is called. Doing so, the SendingThread will wait until Timeout (2s).

MichaelKlemm (changes requested):

MooLMihaly:
A new month without progress again. Can we somehow support the request to be accepted finally? All CUL users are affected.

MichaelKlemm:
> A new month without progress again. Can we somehow support the request to be accepted finally? All CUL users are affected.

Have you seen my comment? I think that PR could be improved. 
Nevertheless you‘re right, the support could be better. Maybe it’s because of legacy stuff.

bundles/io/org.openhab.io.transport.cul/src/main/java/org/openhab/io/transport/cul/internal/AbstractCULHandler.java
hadogenes: Before #5329 it works that way, I think that credit data is special that don't need to notify

PR #5689, opened 2018-10-31
[velux] Support for new firmware v2 added

gs4711:
The Velux 2.x firmware provide full access to any IO-Home compatible devices not limited to Velux and includes many different features: It is now possible to manipulate the settings of any IO-Home controled device explicitely without the previous scene definition.


gs4711:
replaced by #5690.

PR #5707, opened 2018-11-13
[PLCBus] Fix addon crash when two commands are sent in quick succession

pit3k:
This caused two different openhab threads to concurrently receive
bytes from the serial port. In turn, each thread received every other byte,
and so both threads saw corrupted frames. Even worse - threads would get
stuck waiting for response, and entire openhab was unresponsive until
restarted.

Adding synchronized helped. I've also improved logging of error (which is
now not being thrown in this scenario).


9037568:
Thanks, @pit3k !

PR #5710, opened 2018-11-17
[TCP Binding] Enhancements for README.md

arberg:
Signed-off-by: Alex R. Berg, developlantern@gmail.com, github: arberg

Better example, info that its broken with links, added unclear stuff - strange to have in wiki, but better than leaving user in wilderness I believe. Lets just be honest, if somebody knows what it means and feels like specified that would be great.

### DIFF
--- a/bundles/binding/org.openhab.binding.tcp/README.md
+++ b/bundles/binding/org.openhab.binding.tcp/README.md
@@ -2,9 +2,11 @@
 
 The TCP and UDP bindings provide basic support for simple ASCII-based protocols. They send and receive data as ASCII strings. Data sent out is by default padded with a CR/LF. This should be sufficient for many home automation devices that take simple ASCII-based control commands, or that send back text-based status messages.
 
-The TCP part of the binding has a built-in mechanism to keep connections to remote hosts alive, and will reset connections at regular intervals to overcome the limitation of "stalled" connections or remote hosts.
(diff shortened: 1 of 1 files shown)


9037568:
>But its gotta be quick for me, otherwise I won't do it.

I get the picture. I'll leave this open for a while just in case you decide to spend a few seconds on it.

arberg:
Hi Chris, I figured out GitHub kept the changes in a branch, so I could pull that and edit locally. That helped a bit with diffing and applying changes. I have added/updated with most of your comments. I may have missed some given the nature of the task. I chose not to follow your guideline regarding breaking the syntax (line 47) into two examples for IN and OUT. I kept it this way to avoid duplicating the explanations. I now see you probably meant replace '<direction>' with '<'. I still don't think that help much. Instead I have improved much upon the examples, splitting those into IN and OUT. I elaborate the difference between direction IN/OUT and command/update (data in/out). Its very easy to get confused, I certainly was and was mislead by other confused people on the forum.

If you want find minor edits and its easy to do, please feel free to just make the edits, and send me the whole changed file, and just leave inline comments for bigger things, if any.

I can of cause delete the section on known isuses and alternative solutions, but I'm quite sure that would not help the adopters of OpenHAB.

Best Alex

9037568:
Thanks, @arberg !

PR #5699, opened 2018-11-10
[persistence/mysql] Update mysql connector

martinvw:
Is backwards compatible, quoting: "MySQL Connector/J 8.0 is highly recommended for use with MySQL Server 8.0, 5.7, 5.6, and 5.5."

Closes #5698

Signed-off-by: Martin van Wingerden <martin@martinvw.nl>

PR #5714, opened 2018-11-21
[Persistence/Mysql] Corrected project config for mysql related to #5699

martinvw:
Replaced library in classpath because of errors in Eclipse, related to #5699


9037568:
Thanks @martinvw !

PR #5717, opened 2018-11-27
[Prowl] Correct configuration info in README

bgilmer77:
-  Corrects prowl configuration file location
-  Corrects prowl configuration file name

Signed-off-by: Brad Gilmer <brad@gilmer.tv> (github:bgilmer77)

### DIFF
--- a/bundles/action/org.openhab.action.prowl/README.md
+++ b/bundles/action/org.openhab.action.prowl/README.md
@@ -1,6 +1,6 @@
 # Prowl Actions
 
-Prowl lets you use push notifications on iOS devices (please check openhab.cfg for required configuration settings):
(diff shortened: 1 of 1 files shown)


bgilmer77:
@9037568, I am open to anything that improves the situation.  If I am not mistaken, Prowl configuration was documented previously, but it looks like that got dropped.  I have it up and running, so I must have found the information somewhere.  Given that I was a pretty active contributor at the time I started using Prowl, I suspect that if I had found configuration information somewhere else, I would have created a PR to add it.

I agree that configuration information is not parenthetical.  It is core information that should be available on this page.  I am completely under water at the moment, but I had to reinstall OH on a system, and when I went looking for the configuration information here, I could not find it.  But at least I put in a PR.  If you would like to modify the PR, please feel free to do so.  It may be several weeks before I can get to it.

Thank you for the feedback.
Brad

9037568:
Thanks, @bgilmer77 !

PR #5740, opened 2018-12-21
[ekey] Correct README links to ekey.net

jd1:
I was browsing through the addons and noticed that the links are broken.


9037568:
Nice! Thanks for the updates, @jd1 !

PR #5747, opened 2018-12-29
correct the format of item type table

wilfcoles:
just adding the missing header separation line on the item type table


wilfcoles:
Yvw. Thanks for putting the effort into the binding :)

Wilf

PR #5754, opened 2019-01-06
[ucprelayboard] Correct feature name

wborn:
Adds the "1" suffix to the ucprelayboard binding feature name which solves issues like "404 Not Found" when clicking on the binding name in Paper UI to show the documentation.


9037568:
Thanks, @wborn !

PR #5755, opened 2019-01-06
[Fritzbox] Update README.md with notes about removed Telnet support

cubistico:
Add note that telnet is not supported any more and that the TR064 binding should be used

Signed-off-by: Metin Savignano ms@savignano.net (github: cubistico)

### DIFF
--- a/bundles/binding/org.openhab.binding.fritzbox/README.md
+++ b/bundles/binding/org.openhab.binding.fritzbox/README.md
@@ -1,5 +1,11 @@
 # Fritz!Box Binding
 
+## Prerequisites
+
+Note that this binding requires you to enable telnet on your FRITZ!Box, for which, due to security reasons, support has been dropped in firmware version 06.25 and later.
+
+Instead, you may use the [Fritzbox binding (using TR06 protocol)](./fritzboxtr0641).
+
 ## Binding Configuration
 
 This binding can be configured in the `services/fritzbox.cfg` file.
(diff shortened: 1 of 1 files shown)


cubistico:
Please check if the relative link is correct as I did it.

bundles/binding/org.openhab.binding.fritzbox/README.md
```diff
@@ -1,5 +1,11 @@
 # Fritz!Box Binding
 
+## Prerequisites
```
9037568: I suggest merging this section with the `Prepare your Fritzbox` section below, and put the note about the firmware support below the `Enable Telnet` bullet.

bundles/binding/org.openhab.binding.fritzbox/README.md
```diff
@@ -1,5 +1,11 @@
...
+
+Note that this binding requires you to enable telnet on your FRITZ!Box, for which, due to security reasons, support has been dropped in firmware version 06.25 and later.
+
+Instead, you may use the [Fritzbox binding (using TR06 protocol)](./fritzboxtr0641).
```
9037568: This link doesn't work.

9037568: The only kind of link that works is an absolute link. Not sure what you mean by "with the versioning".
I'd argue for linking to the docs page at https://www.openhab.org/addons/bindings/fritzboxtr0641/

bundles/binding/org.openhab.binding.fritzbox/README.md
9037568: It's fine for it to be the first section. But the two sections should be combined, since they both describe prerequisites.

cubistico: Understood now. Please review my changes and resolve if okay.

9037568:
Thanks, @cubistico !

PR #5765, opened 2019-01-11
[Telegram] Support for formatted messages

thorstenbe:
This pull requests adds the ability to write messages that contain markdown or html tags, see https://core.telegram.org/bots/api#markdown-style.
It is configurable if a bot should use formatted messages in the `telegram.cfg`. This pull request is inspired by #5490 but since the other pull request is stale I started a complete rewrite.

### DIFF
--- a/bundles/action/org.openhab.action.telegram/README.md
+++ b/bundles/action/org.openhab.action.telegram/README.md
@@ -42,11 +42,14 @@ The URL can be specified using the http, https, and file protocols.
 
 The action can be configured in `services/telegram.cfg`.
 
-| Property            | Default | Required | Description                           |
-|---------------------|---------|:--------:|---------------------------------------|
-| bots                |         | Yes      | Comma-separated list of `<bot-name>`s |
-| `<bot name>.chatId` |         | Yes      | chat id                               |
-| `<bot name>.token`  |         | Yes      | authentication token                  |
+| Property                | Default | Required | Description                                                  |
+|-------------------------|---------|:--------:|--------------------------------------------------------------|
+| bots                    |         | Yes      | Comma-separated list of `<bot-name>`s                        |
+| `<bot name>.chatId`     |         | Yes      | chat id                                                      |
+| `<bot name>.token`      |         | Yes      | authentication token                                         |
+| `<bot name>.parseMode`  |         | No       | Support for formatted messages, values: `Markdown` or `HTML` |
+
+See https://core.telegram.org/bots/api#markdown-style for formatting options
 
 ### Configuration example
 
@@ -55,6 +58,7 @@ bots=bot1,bot2
 
 bot1.chatId=22334455
 bot1.token=xxxxxxxxxxx
+bot1.parseMode=Markdown
 
 bot2.chatId=654321
 bot2.token=yyyyyyyyyyy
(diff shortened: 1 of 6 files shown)


bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/TelegramActionService.java
```diff
@@ -62,8 +62,17 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
             for (String bot : bots) {
                 String chatIdKey = String.format("%s.chatId", bot);
                 String tokenKey = String.format("%s.token", bot);
+                String parseKey = String.format("%s.parseMode", bot);
+
                 if (config.get(chatIdKey) != null && config.get(tokenKey) != null) {
-                    Telegram.addToken(bot, (String) config.get(chatIdKey), (String) config.get(tokenKey));
+
+                    String chatId = (String) config.get(chatIdKey);
```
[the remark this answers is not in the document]
thorstenbe: Thank you for your review and your improvements. I fixed these lines of code.

bundles/action/org.openhab.action.telegram/README.md
```diff
@@ -42,11 +42,14 @@ The URL can be specified using the http, https, and file protocols.
...
-|---------------------|---------|:--------:|---------------------------------------|
-| bots                |         | Yes      | Comma-separated list of `<bot-name>`s |
-| `<bot name>.chatId` |         | Yes      | chat id                               |
-| `<bot name>.token`  |         | Yes      | authentication token                  |
+| Property                | Default | Required | Description                                                  |
+|-------------------------|---------|:--------:|--------------------------------------------------------------|
+| bots                    |         | Yes      | Comma-separated list of `<bot-name>`s                        |
+| `<bot name>.chatId`     |         | Yes      | chat id                                                      |
+| `<bot name>.token`      |         | Yes      | authentication token                                         |
+| `<bot name>.parseMode`  |         | No       | Support for formatted messages, values: `Markdown` or `HTML` |
```
9037568: This needs to specify what the default is when the parameter is not configured.

Suggest something like:

````
Message formatting to be used. Supported values: `Markdown` or `HTML`. Default is Markdown.
````

File: bundles/action/org.openhab.action.telegram/src/main/java/org/openhab/action/telegram/internal/TelegramActionService.java [hunk 3]
```diff
@@ -72,4 +83,8 @@ public void updated(Dictionary<String, ?> config) throws ConfigurationException
             isProperlyConfigured = true;
         }
     }
+
+    private String getConfigValue(Dictionary<String, ?> config, String key) {
```
9037568: This method provides no value and should be removed.

You're also using the Objects.toString method incorrectly, as you've omitted the second `null` parameter.

thorstenbe: This one is fixed. I thought that the default value (if none is given) is `null` but it is `"null"` as String.

bundles/action/org.openhab.action.telegram/README.md
thorstenbe: I added a bit more text that hopefully it is better understandable that if no value is set then no formatting is used at all.

thorstenbe:
Do I have to do anything that this pull request gets merged? (Sorry for the dumb question but it is my first one)

9037568:
Thanks, @thorstenbe !

PR #5766, opened 2019-01-12
[Anel] Update README.md

yveshanoulle:
I added the name of the binding that needs to be used in addons.cfg.
It helps people if they see the name of to be used in the documentation.

### DIFF
--- a/bundles/binding/org.openhab.binding.anel/README.md
+++ b/bundles/binding/org.openhab.binding.anel/README.md
@@ -32,6 +32,19 @@ The binding can be configured in the file `services/anel.cfg`.  In the table bel
 | `cachePeriod` | 0 | No | Cache the state for `cachePeriod` minutes so only changes are posted (optional, defaults to 0 = disabled).  Example: if period is 60, once per hour all states are posted to the event bus; changes are always and immediately posted to the event bus. |
 
 
+## Addons.cfg Configuration
+
+If you don't want to use PaperUI, you can use addons.cfg to tell openhab to load this binding.
+(Be carefull to never combine PaperUI and this configuration file. 
+([![Read more on addons.cfg here](https://www.openhab.org/docs/configuration/addons.html#through-configuration-files))
+
+The name for this binding to use in addons.cfg is anel1.
+so you would have something like: 
+
+```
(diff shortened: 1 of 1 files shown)


Confectrician:
Tough it looks ok so far it is also a bit duplicated content.
But the relevant docs article is linked aside, so why not.

LGTM

9037568:
Duplicate content indeed. I say not here.

yveshanoulle:
I don't understand.
Nowhere is the information about what to fill into  addons.cfg for a certain binding.
so please explain what info is duplicated and show me where it is duplicated.
I did find it in any of the documentation.
Maybe I added too much context and that is duplicated, right now I don't understand what I have duplicated.

Confectrician:
https://www.openhab.org/docs/configuration/addons.html#through-configuration-files

Explains how you can find out the id through the url of the addons documentation page.

grafik

Which would lead to the docs page for anel:
grafik

What we could do, is change the order of the two identification methods in the installation docs.
It is possibly easier to quickly open the addons page to get the id, than doing it via console.

Confectrician:
Can you link the community thread?
Maybe we should clarify this too over there.

yveshanoulle:
Thanks.
I did not see that (or this page has evolved since I used it.)
I still worry because this explains how to find info, yet if I'm working with a certain binding that is not mentioned here, I doubt I will come to this page.
Adding a binding to addons.cfg is not something I do very often. I remember the last time I needed that, I tried out a few names because the binding did not mention it, and I did not know about this link between url and name of the binding. 

I think there are more places where paperui as a solution  was mentioned. This was one on a question of me
https://community.openhab.org/t/solved-initial-setup-package-selection-using-addons-cfg/43288/2?u=yves

Confectrician:
>I did not see that (or this page has evolved since I used it.)

It was already part of the initial commit done by me in June 2018, but that doesn't matter.

>I still worry because this explains how to find info, yet if I'm working with a certain binding that is not mentioned here, I doubt I will come to this page.

You are not the first one to notice this gap.
Please have a look at https://github.com/openhab/website/issues/152 where we are already discussing ways to improve this.

yveshanoulle:
> June 2018
I started using addons.cfg before june.

PR #5775, opened 2019-01-18
[InsteonPLM] Add OnLevel feature to Insteon PLM 2477D dimmer

crxporter:
This will add the ability to add on level items into OpenHAB for the Insteon PLM binding. Items should be number type with a range from 0 to 255. The feature works the same as other features in the binding (LEDBrightness and RampRate). 

RampRate is Insteon message 0x05. LEDBrightness is Insteon message 0x07. OnLevel (new) is Insteon message 0x06.


9037568:
LGTM, but let's see if @berndpfrommer has any comments.

9037568:
Thanks, @crxporter !

PR #5787, opened 2019-02-11
[Anel] Update README with better sendCommand alternative

semperor:
Please always use the object method instead of the static method. This is safer and less confusing for newcomers. See https://www.openhab.org/docs/configuration/rules-dsl.html#manipulating-item-states


JueBag:
A little side-note: Naming the binding in the title of the PR would help identifying your work upfront.

JueBag:
Thanks

9037568:
Thanks for all the README enhancements, @semperor !

PR #5792, opened 2019-02-11
[Cups] Update README with better sendCommand alternative

semperor:
Please always use the object method instead of the static method. This is safer and less confusing for newcomers. See https://www.openhab.org/docs/configuration/rules-dsl.html#manipulating-item-states
Additionally changed pre to ``` for better readability on openhab.org

### DIFF
--- a/bundles/binding/org.openhab.binding.cups/README.md
+++ b/bundles/binding/org.openhab.binding.cups/README.md
@@ -42,19 +42,21 @@ Each item binding should have this format:
 
 ## Examples
 
-- Number  Print_Jobs_Queued   "Unfinished print jobs"   (FF_Office)   { cups="MX-870#NOT_COMPLETED" }
-
+```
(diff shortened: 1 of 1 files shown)


bundles/binding/org.openhab.binding.cups/README.md
```diff
@@ -51,32 +51,34 @@ Each item binding should have this format:
 The CUPS Binding can be used to switch on a printer if there are print jobs in the queue and switch it off if the queue is empty.
 
 #### Items
-
-    Number Print_Jobs_Queued  "Unfinished print jobs"  (FF_Office)   { cups="MX-870#NOT_COMPLETED" }
-    Switch Printer            "Printer"
-
+```
+Number Print_Jobs_Queued  "Unfinished print jobs"  (FF_Office)   { cups="MX-870#NOT_COMPLETED" }
```
9037568: This is a duplicate of the item shown on line 45, so please delete line 45.
Also, please move the example on line 47 down here.

PR #5798, opened 2019-02-11
[fritzboxtr064] Update README.md

semperor:
Added my current fritz box model


9037568:
I don't think we're going to be adding to this list.

semperor:
Ok. Maybe the list should be removed?

PR #5799, opened 2019-02-12
[Horizon] Update README with better sendCommand alternative

semperor:
Please use the object method instead of the static method. :) see https://www.openhab.org/docs/configuration/rules-dsl.html#manipulating-item-states

Furthermore, it should be considered whether the sendCommand method should be used.


9037568:
For future reference, if you're planning to make the same relatively minor types of changes to a bunch of files (e.g. converting one format to another), please submit one PR with all of the changes.

semperor:
> For future reference, if you're planning to make the same relatively minor types of changes to a bunch of files (e.g. converting one format to another), please submit one PR with all of the changes.

I also thought about it. But it seemed to me better this way, as the bindings are maintained by different people.

PR #5814, opened 2019-02-12
[fritzbox] Add a hint about the state of the Fritzbox bindings

mrumpf:
As a new user it is quite hard to understand what the relationship
between the different FRITZ!Box bindings is. The note tries to
explain the current situation to get new users started faster.

Signed-off-by: Michael Rumpf <michael@rumpfonline.de> (github: mrumpf)


9037568 (changes requested):
A few changes, please.

PR #5818, opened 2019-02-20
[Misc.] Fix some alerts raised by LGTM

1m2c3t4:
LGTM raised some alerts during analysis of openhab1-addons: (https://lgtm.com/projects/g/openhab/openhab1-addons/alerts/?mode=list).

Fixed some of the alerts:
- ArrayIndexOutOfBounds in HarmonyHubDiscovery
- ContradictoryTypeChecks in various classes
- Equals on incomparable types in OneWireDecimalTypeMultiplyModifier and UCPRelayBoardBinding
- Hashed value without hashCode definition in XMLMessageReader
- Inconsistent hashCode and equals in Field
- Missing format argument in various classes
- Reference equality of boxed types in ZWaveSceneConverter
- Type mismatch on container access in Tr064Comm


bundles/binding/org.openhab.binding.autelis/src/main/java/org/openhab/binding/autelis/internal/AutelisGenericBindingProvider.java
```diff
@@ -57,7 +57,7 @@ public String getBindingType() {
      */
     @Override
     public void validateItemType(Item item, String bindingConfig) throws BindingConfigParseException {
-        if (!(item instanceof SwitchItem || item instanceof DimmerItem || item instanceof NumberItem
+        if (!(item instanceof SwitchItem || item instanceof NumberItem
```
9037568: Please explain the removal of the `DimmerItem` checks throughout...

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/DmxGenericBindingProvider.java
```diff
@@ -92,7 +92,7 @@ public String getBindingType() {
      */
     @Override
     public void validateItemType(Item item, String bindingConfig) throws BindingConfigParseException {
-        if (!(item instanceof SwitchItem || item instanceof DimmerItem || item instanceof ColorItem)) {
+        if (!(item instanceof SwitchItem || item instanceof DimmerItem)) {
```
9037568: Why was the `ColorItem` check removed here?

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
```diff
@@ -179,7 +179,7 @@ public String getTr064Value(ItemConfiguration request) {
 
             String itemCommand = itemConfiguration.getItemCommand();
 
-            if (values.containsKey(itemCommand)) {
+            if (values.containsKey(itemConfiguration)) {
```
9037568: Explain this change, please.

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/message/Field.java
```diff
@@ -8,6 +8,8 @@
  */
 package org.openhab.binding.insteonplm.internal.message;
 
+import java.util.Objects;
```
9037568: Please explain the changes made to this file.

bundles/binding/org.openhab.binding.lgtv/src/main/java/org/openhab/binding/lgtv/internal/LgtvBinding.java
```diff
@@ -204,8 +204,6 @@ private String convertOpenHabCommandToDeviceCommand(Command command, String cmdT
         } else if (command instanceof DecimalType) {
             deviceCmd = String.format(cmdTemplate, ((DecimalType) command).intValue());
 
-        } else if (command instanceof PercentType) {
```
9037568: Explain or revert this change.

bundles/binding/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoBinding.java
```diff
@@ -181,8 +181,6 @@ private String convertOpenHabCommandToDeviceCommand(Command command, String cmdT
         } else if (command instanceof DecimalType) {
             deviceCmd = String.format(cmdTemplate, ((DecimalType) command).intValue());
 
-        } else if (command instanceof PercentType) {
```
9037568: Explain or revert this change.

bundles/persistence/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/InfluxDBPersistenceService.java
```diff
@@ -478,8 +478,6 @@ private String stateToString(State state) {
             value = ((OnOffType) state) == OnOffType.ON ? DIGITAL_VALUE_ON : DIGITAL_VALUE_OFF;
         } else if (state instanceof OpenClosedType) {
             value = ((OpenClosedType) state) == OpenClosedType.OPEN ? DIGITAL_VALUE_ON : DIGITAL_VALUE_OFF;
-        } else if (state instanceof HSBType) {
```
9037568: Explain or revert this change.

bundles/persistence/org.openhab.persistence.influxdb08/java/org/openhab/persistence/influxdb08/internal/InfluxDBPersistenceService.java
```diff
@@ -221,8 +221,6 @@ public void store(Item item, String alias) {
     State state = null;
     if (item instanceof DimmerItem || item instanceof RollershutterItem) {
       state = item.getStateAs(PercentType.class);
-    } else if (item instanceof ColorItem) {
```
9037568: Explain or revert all the changes to this file.

bundles/persistence/org.openhab.persistence.jpa/java/org/openhab/persistence/jpa/internal/JpaHistoricItem.java
```diff
@@ -111,8 +111,6 @@ public static HistoricItem fromPersistedItem(JpaPersistentItem pItem, Item item)
             state = OpenClosedType.valueOf(pItem.getValue());
         } else if (item instanceof RollershutterItem) {
             state = PercentType.valueOf(pItem.getValue());
-        } else if (item instanceof ColorItem) {
```
9037568: Explain or revert this change.

bundles/persistence/org.openhab.persistence.mongodb/java/org/openhab/persistence/mongodb/internal/MongoDBPersistenceService.java
```diff
@@ -285,8 +285,6 @@ private void disconnectFromDatabase() {
                 state = OpenClosedType.valueOf(obj.getString(FIELD_VALUE));
             } else if (item instanceof RollershutterItem) {
                 state = new PercentType(obj.getInt(FIELD_VALUE));
-            } else if (item instanceof ColorItem) {
```
9037568: Explain or revert this change.

bundles/binding/org.openhab.binding.autelis/src/main/java/org/openhab/binding/autelis/internal/AutelisGenericBindingProvider.java
1m2c3t4: `DimmerItem` is a subclass of `SwitchItem`. In the above code (and in other places) the test `item instanceof DimmerItem` is being made after it has been established that `item` is not an instance of `SwitchItem` (and therefore not an instance of any subclass of `SwitchItem`, including `DimmerItem`). So in this case it's a redundant check. In some other cases, there is also dead code following the check.

bundles/binding/org.openhab.binding.dmx/src/main/java/org/openhab/binding/dmx/internal/DmxGenericBindingProvider.java
1m2c3t4: `ColorItem` is a subclass of `DimmerItem`. So this is another redundant check, because the test for `ColorItem` will only be evaluated if the test for `DimmerItem` fails. And if it isn't a `DimmerItem`, it certainly can't be a `ColorItem`. In fact, the test for `DimmerItem` is itself redundant, since `DimmerItem` is a subclass of `SwitchItem`. So this code should really only be testing for `SwitchItem`, which would make the code clearer and simpler.

bundles/binding/org.openhab.binding.fritzboxtr064/src/main/java/org/openhab/binding/fritzboxtr064/internal/Tr064Comm.java
1m2c3t4: The `values` map is declared as `Map<ItemConfiguration, String>`, so it is clear that any key will need to be of type `ItemConfiguration`. `itemCommand` is declared as a String, so I assume the original code had a typo here. I assumed that the intention was to check whether `itemConfiguration` was present as a key.

bundles/binding/org.openhab.binding.insteonplm/src/main/java/org/openhab/binding/insteonplm/internal/message/Field.java
1m2c3t4: The class overrides `equals` but does not override `hashCode`. This breaks the contract whereby objects that are considered equal should have the same hashCode. There are other cases of this in the code base, but this one is the most problematic because `Field` is used as a key in a hashing structure in `XMLMessageReader`. LGTM provides more details on this error at https://lgtm.com/rules/2980070/

bundles/binding/org.openhab.binding.lgtv/src/main/java/org/openhab/binding/lgtv/internal/LgtvBinding.java
1m2c3t4: `PercentType` is a subclass of `DecimalType`, and the test for `command instanceof PercentType` is only evaluated after it has been found to be not an instance of `DecimalType` (and consequently not an instance of any subclass of `DecimalType`). The test will always evaluate to false, so there is dead code following. The change I made removes the redundant check and the dead code.

bundles/binding/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoBinding.java
1m2c3t4: See above under LgtvBinding

bundles/persistence/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/InfluxDBPersistenceService.java
1m2c3t4: `HSBType` is a subclass of `PercentType`, which is in turn a subclass of `DecimalType`. So this check is being made after it has been established that `state` is not an instance of `HSBType`. The change removes the redundant check and the dead code, preserving the semantics of the code. On the other hand, perhaps there is a bug and the code is not doing what was intended. If that is the case, the test for `HSBType` should be done before the test for `DecimalType`. Which solution do you prefer ?

bundles/persistence/org.openhab.persistence.influxdb08/java/org/openhab/persistence/influxdb08/internal/InfluxDBPersistenceService.java
1m2c3t4: Here `ColorItem` is a subclass of `DimmerItem` so again the check is redundant and there is dead code. I have removed the dead code and the redundant check, but again an alternative solution would be to check for `ColorItem` before checking for `DimmerItem`. Which is better here ? Similarly, there are checks in this class for `HSBType`, which are done after checks for its superclass. Again, I've removed the redundant code, but could instead move the check for `HSBType` earlier if you prefer that .

bundles/persistence/org.openhab.persistence.jpa/java/org/openhab/persistence/jpa/internal/JpaHistoricItem.java
1m2c3t4: Again this is a redundant check and dead code, because the check for `ColorItem` comes after the check for its superclass. Here again, we could instead move the check for `ColorItem` before the check for `DimmerItem` if you prefer that.

bundles/persistence/org.openhab.persistence.mongodb/java/org/openhab/persistence/mongodb/internal/MongoDBPersistenceService.java
1m2c3t4: The check for `ColorItem` is redundant because it comes after the check for `DimmerItem`. Again, I can move the check for `ColorItem` earlier if that is a solution you prefer.

1m2c3t4:
@9037568 thanks for reviewing. I have addressed your comments, but there a few places where different solutions are possible and I'm not sure which you'd prefer. Please let me know your preference.

bundles/persistence/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/InfluxDBPersistenceService.java
9037568: I guess we'll leave it as you have it, since that's the way it's been operating in practice.

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonGenericBindingProvider.java
```diff
@@ -53,10 +53,10 @@ public String getBindingType() {
      */
     @Override
     public void validateItemType(Item item, String bindingConfig) throws BindingConfigParseException {
-        if (!(item instanceof SwitchItem || item instanceof DimmerItem || item instanceof StringItem)) {
+        if (!(item instanceof SwitchItem || item instanceof StringItem)) {
             throw new BindingConfigParseException("item '" + item.getName() + "' is of type '"
                     + item.getClass().getSimpleName()
-                    + "', only Switch-, Dimmer- and StringItems are allowed - please check your *.items configuration");
```
9037568: Restore to original. Also apply to the other similar throws.

1m2c3t4:
@9037568 thanks, I have restored the exception messages to their previous form, and also removed some unused imports.

bundles/binding/org.openhab.binding.denon/src/main/java/org/openhab/binding/denon/internal/DenonGenericBindingProvider.java
1m2c3t4: This and other cases have been restored to original

9037568:
Thanks, @1m2c3t4 !

PR #5822, opened 2019-02-27
[Twitter] Remove reference to Dropbox from README

ItIsFelix:
Since there is no official binding for "Dropbox", as far as I know, this information is not helpful because you can not make any reference.


9037568:
Thanks, @ItIsFelix !

PR #5828, opened 2019-03-11
[anel] Added Masking of passwords in log entries

paphko:


yveshanoulle:
@paphko does this have the side effect that the password is replaced with stars, so when you do that before the try, will the tray fail because of a wrong password? 

When the password is replaced with stars, does it even need to be in the log?

paphko:
@yveshanoulle the password is only masked for logging, so it doesn't cause any side effects.

yveshanoulle:
what needs to happen to unblock this?

PR #5836, opened 2019-03-19
Fix OpenWeather API key failure

ptrooms:
Why:
* OpenWeather discontinued the the Free Plan and by this the old code generates an unconditonal error using API as FORECAST is always used / opened.

This change addresses the need by:
* returning null in the FORECAST call


ptrooms:
Appologies for doing this on the primary branch...... I had to pull on my own fork.
Still learning, in git'ting.....
Kind regards, Peter

PR #5846, opened 2019-04-13
[pushover] Add HTML and Monospace formatting option

bruestel:
* Added message formatting features (https://pushover.net/api#html)

Signed-off-by: Jonas Brüstel <jonas@bruestel.net> (github: bruestel)


bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java
```diff
@@ -468,6 +486,13 @@ private static String pushover0(String apiKey, String user, String message, Stri
                             attachment, e.getMessage());
                 }
             }
+
+            if (htmlFormatting) {
+                parts.add(new StringPart(MESSAGE_KEY_HTML, MESSAGE_VALUE_HTML, UTF_8_ENCODING));
```
cweitkamp: Please use `StandardCharsets.UTF_8.name()`. Would you be so kind to replace it everywhere in this file?

cweitkamp:
I maybe see an issue raising. Currently we check for the length of the message plus the title and suppress to send something when the length exceeds 512 characters.

https://github.com/openhab/openhab1-addons/blob/9f1ee25844da5bb2fa3bd0d088b719c63d1a5f19/bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java#L362

This will happen very quickly when one will start using HTML in his messages. IIRC the used thresholds are not correct anymore. According to https://pushover.net/api#limits we can increase them to 1024 chars for the message and 250 chars for the title.

bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java
bruestel: changed it

bruestel:
Changed title and message length constraints.

bruestel:
CI crashed. How can I trigger a new build?

cweitkamp:
Thanks for the changes.

> How can I trigger a new build?

You can close the PR and reopen it :wink:.

bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java
```diff
@@ -45,7 +46,7 @@
 
     private static final String JSON_API_URL = "https://api.pushover.net/1/messages.json";
     private static final String JSON_CANCEL_API_URL = "https://api.pushover.net/1/receipts/{receipt}/cancel.json";
-    private static final String UTF_8_ENCODING = "UTF-8";
+    private static final String UTF_8_ENCODING = StandardCharsets.UTF_8.name();
```
cweitkamp: That is okay but not the idea behind my comment. I maybe could have explained it a little bit more in detail.

I think we do not need this constant here. Rather replace it everywhere in this file. One benefit will be that it will be possible to replace it completely in the `IOUtils.toString()` methods by using the `Charset` itself to omit `UnsupportedCharsetException` (or `UnsupportedEncodingException`, see [commons.io API](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#toString(java.io.InputStream,%20java.nio.charset.Charset))).

https://github.com/openhab/openhab1-addons/blob/b4b7a293e85c031693785831685949860829e920/bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java#L510

https://github.com/openhab/openhab1-addons/blob/b4b7a293e85c031693785831685949860829e920/bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java#L605

bruestel: Now I understand what you meant.
Unfortunately, this method isn't available in the current  commons.io API version. 
`The method toString(InputStream, String) in the type IOUtils is not applicable for the arguments (InputStream, Charset)`

cweitkamp: I see. We are using 2.2 and the expected method is available from 2.3. What a pity.

bundles/action/org.openhab.action.pushover/src/main/java/org/openhab/action/pushover/internal/Pushover.java
```diff
@@ -8,11 +8,14 @@
...
 package org.openhab.action.pushover.internal;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.nio.charset.StandardCharsets;
```
cweitkamp: As you are using a static import now this one is not needed anymore.

bruestel: removed it

schrej-zz:
Hi,
did this went into the 2.5M1? I fail to use ".withHtmlFormatting" and I get an error message, that 'withHtmlFormatting' is not a member of 'org.openhab.action.pushover.internal.Pushover'.

bruestel:
> Hi,
> did this went into the 2.5M1? I fail to use ".withHtmlFormatting" and I get an error message, that 'withHtmlFormatting' is not a member of 'org.openhab.action.pushover.internal.Pushover'.

I really don't know. I've added the addon as jar file in my installation (2.4). I can send you the jar or you can compile it yourself.

PR #5851, opened 2019-04-27
Fix CUL 2s blocking problem after sending FS20 message

MichaelKlemm:
Currently the CUL sending thread will produce a timeout of 2s if a command is sent that will not provide a response. This is the case when sending a FS20 message ("F" prefix). The major purpose of this PR is to provide a generic solution for such commands and only fix the problem for the FS20 protocol as there is no other equipment for testing available.

Additional Minor changes and refactorings:
 - Use a queue that is blocking instead of polling in a busy loop
 - Use monitors / synchronized methods to wait for a response instead of using a boolean local member


MichaelKlemm:
Hi! How to proceed with this PR?

PR #5865, opened 2019-05-31
Only 100 messages per month are allowed.

steam73:
For some time now (may 2019) only 100 messages per month are allowed.


hakan42:
LGTM.

I'll transfer your change to the OH2 binding which I am writing just now too.

9037568:
Thanks, @steam73 !

PR #5868, opened 2019-06-17
Fixed restriction on special characters in URLs

CSchlipp:
Escaping special characters in URLs allows usage of calendar-ids with special characters, e.g. private-home --> private%2Dhome


9037568:
Have you tested this and found that it actually works? Or just guessing that it should?

CSchlipp:
I've tested both, number and '-' with Nextcloud, caldav binding could sync appointment from these encoded URLs. 
Other numbers (and probably also special characters) should work as well, as long as they are encoded correctly.

Issue #5870, opened 2019-06-20
[caldav] NCDF org/threeten/bp/temporal/TemporalAdjuster


openhab-5iver:
OH S1618 (did not occur in S1615). Likely caused by #5860, which looks to be the only changes to the binding between these builds. This occurs on startup...

```
2019-06-20 16:36:27.477 [ERROR] [org.quartz.core.JobRunShell] - Job event-reloader.scottandlisa threw an unhandled Exception: 
java.lang.NoClassDefFoundError: org/threeten/bp/temporal/TemporalAdjuster
        at net.fortuna.ical4j.model.TimeZoneRegistryImpl.<init>(TimeZoneRegistryImpl.java:125) ~[?:?]
        at net.fortuna.ical4j.model.TimeZoneRegistryImpl.<init>(TimeZoneRegistryImpl.java:116) ~[?:?]
        at net.fortuna.ical4j.model.DefaultTimeZoneRegistryFactory.createRegistry(DefaultTimeZoneRegistryFactory.java:48) ~[?:?]
        at net.fortuna.ical4j.data.CalendarBuilder.<init>(CalendarBuilder.java:105) ~[?:?]
        at org.openhab.io.caldav.internal.job.EventReloaderJob.loadEvents(EventReloaderJob.java:339) ~[?:?]
        at org.openhab.io.caldav.internal.job.EventReloaderJob.loadEvents(EventReloaderJob.java:325) ~[?:?]
        at org.openhab.io.caldav.internal.job.EventReloaderJob.execute(EventReloaderJob.java:141) ~[?:?]
        at org.quartz.core.JobRunShell.run(JobRunShell.java:202) [185:org.openhab.core.scheduler:2.5.0.201906200304]
        at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573) [185:org.openhab.core.scheduler:2.5.0.201906200304]
Caused by: java.lang.ClassNotFoundException: org.threeten.bp.temporal.TemporalAdjuster cannot be found by org.openhab.io.caldav_1.14.0.201906170308
        at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:484) ~[?:?]
        at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:395) ~[?:?]
        at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:387) ~[?:?]
[...]

PR #5880, opened 2019-07-28
[caldav] Addes missing dependencies

bjoernbrings:
Signed-off-by: Björn Brings <bjoernbrings@web.de>

Should solve issue #5870. Thanks @cweitkamp for much prework.
I guess all those dependencies were required formally in the past, but as they have not been declared explicitly by ical4j (which was changed in 2.0.1) and the functions have not been used by anyone there was no error occuring. This adds lots of (not really necessary) libaries but is from my POV the best solution for now.
We have to be carefull with this library as they release bigger/breaking changes even with minor releases.


wborn:
Thanks for the PR! Can you also update the NOTICE file for these dependency changes?

bjoernbrings:
done

PR #5884, opened 2019-08-05
[fritzboxtr064] Reduced logging for TAM and call list XML-requests

cweitkamp:
- Reduced logging for TAM and call list XML-requests


9037568:
@cweitkamp what's the need for reducing these when they're already at DEBUG level?

cweitkamp:
In general it is not really necessary. More or less a personal issue. I always use TRACE for logging requests and responses. I think I was annoyed during my last debug session for the binding because my list of missed calls was too long ... I will close this PR.

Issue #5889, opened 2019-08-20
[serial] Isuess receiving serial commands glued together


marekhalmo:
If your serial device sends commands fast the serial pooler will glue the comands to a single line.
Example:
My serial input gateway sends C1 and S1 commands on serial port when a contact 1 is clicked.
The serial binding receives "C1\r\nS1\r\n" even thou the arduino code sends separate commands (Serial.println("C1"); ..code.. Serial.println("S1");

If i set the serial binding to be sensitive to ON("C1\r\n") the binding wont work.. if i set the serial binding to ON("C1\r\nS1\r\n") then the binding is triggered.

This is due to fact that the [serial device](https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/SerialDevice.java) uses "equals" instead of contains when comparing received command and actual commands.

The fix is to use contains

Issue should not be dependent on environment. Tested on Windows10 as Openhabian

marekhalmo:
Sorry for the duplicate pull request, could not get the DCO to work for me.. The second is expected to be valid

marekhalmo:
All done..
Please check the code and merge..
the build jar to check is here
Just add it to the addons folder and you are ready to go.

Here is the code to check the old/new functionality, the old version will not catch the ON/OFF events. The new version will. Please make sure that you set the proper COM port. This is a windows version
[serial.items]
Switch onOffInput "ON/OFF" { serial="COM9@115200,ON(S1),OFF(R1)" }

Arduino sketch (any USB connected Arduino will work)

// Test sketch that is simulating button clicks on serial input gateway

// State indicator
#define LED LED_BUILD_IN
#define SERIAL_RATE_BAUD 115200

void setup() {
  Serial.begin(SERIAL_RATE_BAUD);
}

void loop() {
  Serial.println("S1");
  delay(50);
  Serial.println("R1");
  Serial.println("C1");
  delay(500);

  Serial.println("S1");
  delay(50);
  Serial.println("R1");
  delay(100);
  Serial.println("S1");
  delay(50);
  Serial.println("R1");
  Serial.println("D1");

  delay(500);
  Serial.println("S1");
  delay(500);
  Serial.println("H1");
  delay(500);
  Serial.println("R1");
  Serial.println("E1");
  
  delay(500);
}

marekhalmo:
Anyone looking at this? The fix is ready to be merged...

[...]

PR #5890, opened 2019-08-20
Fix Serial binding where serial commands in rapid succession were not…

marekhalmo:
… recognized

https://github.com/openhab/openhab1-addons/issues/5889


marekhalmo:
Will try again :)

PR #5896, opened 2019-09-12
Typo lead to confusing instructions

bitDealer:
The address in the first example is 0x21, but the explenation under it statet the address as 0x20. The second example shows the right way


9037568:
Thanks, @bitDealer !
And thanks, @wborn for the review!

PR #5906, opened 2019-10-22
Updated to use MyQ v5 API

computergeek1507:


kaikreuzer:
Sorry for the late feedback!

Could you add some info on what this PR includes?
Is it from your pov safe to merge, are there any positive user reports already?

Does it include or supersede https://github.com/openhab/openhab1-addons/pull/5823?

computergeek1507:
Chamberlain shutdown the v4 API. The other pull request was to fix a bug in the V4 API. I switched my focused on the v2 binding, I will have to check if the v1 is still working with these changes.

PR #5908, opened 2019-11-04
[Fritzboxtr064] Improve README

mueller-ma:
* Translate log message
* Wifi => Wi-Fi, see https://en.wikipedia.org/wiki/Wi-Fi
* Use capital letter after bullet point

Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>


9037568:
Thanks, @mueller-ma !

PR #5910, opened 2019-11-07
[fritzboxtr064] Added uptime

cweitkamp:
- Added uptime


cweitkamp:
As we do not have a "Reuild" label here I will close and open the PR to restart Jenkins.

wborn:
We have a working "rebuild" label now here too. :-)

PR #5928, opened 2020-01-01
[HTTP binding] Update README.md

JanVerstuyft:
added a section on the dynamic urls that it is not supported when using the HTTP cache


9037568:
Please explain this new section. What does it mean and what is it referring to?

JanVerstuyft:
After reading the comment on https://community.openhab.org/t/dynamic-urls-for-http/34848 and having the same issue.  I just was trying to update the documentation.

The dynamic building of the url (including date, state, ....) is only working when you define the url on the item itself.
It is not working when you define the url on the http.conf service configuration on a cache item.

Since myself and other persons are having troubles with the documentation being not clear, I wanted to update this.

9037568:
Well, I call that a defect.

[This jar](https://drive.google.com/open?id=1ekYsT9QBJ8QlezkbQ2ubj-srGgqu_kDQ) contains a fixed version that should work properly. Care to try it out?

JanVerstuyft:
Hi,

this patch is indeed working.  Indeed, I prefer to have it as a defect ;).  

KR

Jan

9037568:
Would you like to open an actual defect to report this defect? You found it.
If not, I'll do it while working on the PR to fix it.

JanVerstuyft:
I've created the bug.  With number 5929

Thanks for the work.

PR #5935, opened 2020-01-29
[caldav-personal] Filters now allow any character, but require single quotes

openhab-5iver:
This change makes single quotes around the filters a requirement, which is what [the documentation already shows](https://www.openhab.org/addons/bindings/caldav-personal1/#filtering). Single quotes inside the filter will stop the matching, so they should be replaced with a period (.) in the Item definition. The benefit of this change is that the characters in the filters no longer have any other limitation. Previously, they were limited, [as discussed in the forum](https://community.openhab.org/t/caldav-personal-regular-expression/84751/13?u=5iver). 

PRs for the documentation are in #5936 and #5937.

Signed-off-by: Scott Rushworth <openhab@5iver.com>

### DIFF
--- a/bundles/binding/org.openhab.binding.caldav-personal/src/main/java/org/openhab/binding/caldav_personal/internal/CalDavBindingProviderImpl.java
+++ b/bundles/binding/org.openhab.binding.caldav-personal/src/main/java/org/openhab/binding/caldav_personal/internal/CalDavBindingProviderImpl.java
@@ -51,9 +51,9 @@ public class CalDavBindingProviderImpl extends AbstractGenericBindingProvider im
     private static final String REGEX_TYPE = "type:'?([A-Za-z]+)'?";
     private static final String REGEX_EVENT_NR = "eventNr:'?([0-9]+)'?";
     private static final String REGEX_VALUE = "value:'?([A-Za-z]+)'?";
-    private static final String REGEX_FILTER_NAME = "filter-name:'?([A-Za-z\\.\\*\\+\\- \\|]+)'?";
-    private static final String REGEX_FILTER_CATEGORY = "filter-category:'?([A-Za-z-_]+(, ?[A-Za-z-_]+)*)'?";
-    private static final String REGEX_FILTER_CATEGORY_ANY = "filter-category-any:'?([A-Za-z-_]+(, ?[A-Za-z-_]+)*)'?";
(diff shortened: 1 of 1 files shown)


openhab-5iver:
Any idea why [the build failed](https://ci.openhab.org/job/PR-openHAB1-Addons/2674/console)? Looks systemic, since they've been failing for a month and a half! Maybe OH1 addons are not going into the 2.5.x distro?

9037568:
>Any idea why the build failed? 

No, not a clue. Perhaps there's a wizard in the know among the @openhab/add-ons-maintainers

9037568:
FWIW, the build was successful in my local.

openhab-5iver:
I have not been successful building this. What procedure did you follow?

9037568:
I just used the command line and executed a `mvn install` from the top of the binding's folder tree.

openhab-5iver:
```
[ERROR]   Missing requirement: org.openhab.binding.caldav-personal 1.15.0.qualifier requires 'package org.openhab.io.caldav 0.0.0' but it could not be found
```

9037568:
Can anyone in @openhab/maintainers explain this build failure?

kaikreuzer:
@openhab-5iver We actually didn't plan to ship any further updates of 1.x add-ons - it isn't included in the 2.5.x release process anymore. I'd only want to revert this and spend additional efforts on it, if really urgent and necessary.

openhab-5iver:
@kaikreuzer , wow... I wish that had been made clearer in the announcements. This PR is not critical, but the filter functionality is pretty much unusable with it. The 2.x version is coming along, so hopefully this binding will be moved to legacy soon anyhow. I suggest communicating in the forum that 1.x bindings are no longer going to be updated in the distro. It makes sense to me to archive the openhab1-addons repo and put in a PR like in the ESH repo to let people know this repo is no longer maintained.

kaikreuzer:
Well, yes, sorry - I absolutely agree that this should be clearly communicated. I still hesitated because I wasn't sure whether we should really do it or not (that's why I wrote above "didn't plan to"...). For critical fixes, it might still make sense, but it will mean that we have to adapt the 2.5.x build processes and make sure that updates are taken into consideration, which might mean that we could need yet another 2.5.x distro delivery... Something I'd want to avoid.

@openhab/1-x-add-ons-maintainers  What do you think how we should go about it?

PR #5941, opened 2020-02-26
Replace "must not" by "need not"

7usr7local:
I guess it had been a translation error from german - saying "must not be ordered" means it has to be be *not* in order. That doesn't make sense to me. Guessing from the german wording "muss nicht sortiert sein" I guess the meaning should be that it may be or may be not sorted - which is "need not be ordered" in english.

### DIFF
--- a/bundles/binding/org.openhab.binding.caldav-personal/README.md
+++ b/bundles/binding/org.openhab.binding.caldav-personal/README.md
@@ -37,7 +37,7 @@ Binding file: org.openhab.binding.caldav-command-version.jar
 Used to execute commands through an event, triggered at the start or the end of an event.
 The event summary is free selectable. The event description must fullfill special syntax.
 Syntax is `<BEGIN|END>:<Item-Name>:<Command>`.
-Each item trigger must be a single line without linebreaks. Every line which starts with "BEGIN" will be executed at the begin of the event. Each line with an "END" will be executed at the end of the event. You can define multiple lines, which must not be ordered. For example:
+Each item trigger must be a single line without linebreaks. Every line which starts with "BEGIN" will be executed at the begin of the event. Each line with an "END" will be executed at the end of the event. You can define multiple lines, which need not be ordered. For example:
 
 ```
 BEGIN:Heater_Livingroom:22


7usr7local:
Hi @9037568,

propably I got something wrong - I just tried to improve the documentation public available here: https://www.openhab.org/addons/bindings/caldav-personal1/ with a text at the bottom

> "Caught a mistake or want to contribute to the documentation? Edit this page on GitHub"

containing a link I followed which lead me to create that pull request.

The only thing I can say for sure is 

- the sentence I tried to improve is confusing.
- I was trying to help.

Maybe I made the wrong improvement because I didn't get the meaning right. (Maybe it really *must* *not* be ordered - but what for? If the code does not rely on the order - the order should not matter at all, isn't it? If it must not be ordered in a special way - it would be more reasonable to mention the way it *should* be ordered.)

Maybe I didn't do the improvement at the right place, because the link pointed to a wrong location.

Maybe I missed some formalism because I tried to help without studying the projects charta.

Maybe the text at the bottom is just void, and help by non-members is not appreciated.

Maybe ... something completely different is happening.

9037568:
There's nothing wrong with what you did. But this repository is closed and not taking updates.

7usr7local:
Hello @9037568 ,

thanks to replying to my question. I am still not sure, I got it right.
Is it because openhab1 is deprecated?
So what's the way to correct errors in its documentation?
Sorry for taking so long to respond, I somehow missed your reaction.

Thanks,

Peter

ptrooms:
> Hello @9037568 ,
> 
> thanks to replying to my question. I am still not sure, I got it right.
> Is it because openhab1 is deprecated?
> So what's the way to correct errors in its documentation?
> Sorry for taking so long to respond, I somehow missed your reaction.
> 
> Thanks,
> 
> Peter
As I do understand it is more strategic. The (ability to) support of OH1 bindings by "community lead-engineers" went simply by "choice" into a (forced?) stop.
Perhaps to get rid off the old-factory (architecture) that's in the way of  making a new - independant ? -  approaches.

I can also imagine that the original way of doing things is becoming to  difficult and/or violating "developing" rights and/or while the founding "smarthome" platform is archived by Eclipse and thus no longer receives that founded company support. Consequences is that OH1 is dropped whoch will in toime also be the case for OH2 in favor  of the upcoming OH3 architecture.

openhab-5iver:
https://github.com/openhab/openhab-core/pull/1284

This also contains a link to the discussion.

PR #5945, opened 2020-03-25
Safer check in case meter_reset null

Hk-tang:
Possible null when using `"true".equalsIgnoreCase(arguments.get("meter_reset"))`.
In case it could be null, use `Boolean.parseBoolean(arguments.get("meter_reset"))`


Hk-tang:
The force pushes were because I could not get the sign-off correctly

9037568:
So you've exchanged one potential exception for another. Not very helpful.
Doesn't matter, though, because this repository is closed.

PR #5949, opened 2020-04-08
[comfoair1] add an additional example to the readme.md file

bechte:
I have set up a demo instance with my CA350 and extracted another very well working example that includes a slider to set the actual fan level to an arbitrary value between 0 and 100. The slider also sets the fan level according the the following ranges:

- 0-24 : sets fan level 1 = away
- 25-49 : sets fan level 2 = low
- 50-74 : sets fan level 3 = middle
- 75-100 : sets fan level 4 = high

The outgoing fan will be set slightly lower than the incoming fan. This is a recommendation of Zehnder service. The offset will be 2,3,4, or 5 in regard to the above ranges.

Within the rules you will also find a reset of the default values of a level that one needs to adjust to their individual needs. The idea behind this is, that you can use the slider or the setpoint to switch the fan level both, very fine or just rough by tapping through the levels.


9037568:
This repository is closed to new code.

PR #5954, opened 2020-04-29
Update documentation for insteonhub and insteonplm to indicate not supported.

robnielsen:
Also indicate that the Insteon binding has replaced the both bindings.


robnielsen:
@wborn, is it possible to get this PR merged? It only contains updated documentation to indicate insteonplm and insteonhub bindings are not supported and have been replaced by the insteon binding.

PR #5958, opened 2020-05-22
Add inactive badge

pfink:
Maybe it would be good idea to have badges showing the maintainance status of our repositories at a glance. https://www.repostatus.org/ offers such badges. If you think this is a good idea, I'd create PRs for all repos.


wborn (approved):
Personally I think it makes sense to add such a badge for repo's that are not actively maintained and haven't been archived (yet) as it helps to manage expectations for contributors. I can't speak for the other maintainers but since there have been no objections I'll proceed with merging this PR. :wink:

PR #5961, opened 2020-07-28
The xbee license is actual a GPL not LGPL

Hilbrand:
See https://github.com/andrewrapp/xbee-api/blob/master/COPYING


wborn (approved):
I probably made a copy/paste error when I added this. :-)

